PR #1, opened 2015-11-21
Dictionary: EPWING support

miurahr:
- Implement RFE#1123.
- Change IDictionary interface not to expose
  dictionary HashMap to Manager.


src/org/omegat/core/dictionaries/EBDict.java
```diff
@@ -0,0 +1,372 @@
...
+         * convert (\uFF01 - \uFF5E) to (\u0021- \u007E)
+         *
+         * @param text
+         * @return String converted
+        */
+        public String convertZen2Han(String text) {
+            int i;
+            StringBuilder result = new StringBuilder(text.length());
+            for (i = 0; i < text.length(); i++ ) {
+                char c = text.charAt(i);
```
miurahr: EPWING standard use Shift_JIS for character encoding. It is why charAt() is safe here.

amake:
This CI error is legitimate. You have a compilation error.

src/org/omegat/core/dictionaries/StarDict.java
```diff
@@ -147,7 +149,10 @@ public StarDict(File ifoFile) throws Exception {
                 mem.write(b);
             }
         }
-        return result;
+    }
+
+    public int entrySize() {
```
amake: This method name is confusing. I would expect this to mean "the size of *an* entry" but it's actually the number of entries. So I would change it to `size()` or `entryCount()` or something like that.

src/org/omegat/gui/dictionaries/IDictionariesPane.java
```diff
@@ -25,7 +25,7 @@
 
 package org.omegat.gui.dictionaries;
 
-public interface IDictionaries {
+public interface IDictionariesPane {
```
amake: Why did you rename this?

miurahr: There are `IDictionary` (from core.dictionaries) and `IDictionaries` (from gui.dictionaries)
It is easy for me to confuse it. So I'd like to make this more obvious name.

src/org/omegat/core/dictionaries/StarDict.java
miurahr: I'll take `size()`

miurahr:
Test fails because master branch has a new test for LingvoDSL.
To resolve this, I need to rebase the PR branch to current master.

miurahr:
Now we can implement as a plugin for EPWING dictionary.
I'd like to close here.

PR #2, opened 2015-11-21
dictionary: add LingvoDSL test

miurahr:
RFE#1160


src/org/omegat/core/dictionaries/LingvoDSL.java
```diff
@@ -66,7 +66,7 @@ public LingvoDSL(File file) {
                 }
                 s = RE_SKIP.matcher(s).replaceAll("");
                 if (Character.isWhitespace(s.codePointAt(0))) {
-                    trans.append(s).append('\n');
+                    trans.append(s.substring(1)).append('\n');
```
amake: You don't know how many `char`s that first codepoint is, so `substring(1)` is wrong. If your intent is to strip the first whitespace codepoint, it should be `s.substring(s.offsetByCodePoints(0, 1))`, or you could grab the first codepoint first as `int cp = s.codePointAt(0)` and then do `s.substring(Character.charCount(cp))`.

However it seems to me that if you just want to trim the whitespace you should just do `s.trim()`.

test/src/org/omegat/core/dictionaries/LingvoDSLTest.java
```diff
@@ -0,0 +1,78 @@
...
+
+    @Test
+    public void testReadArticleRussian() throws Exception {
+        LingvoDSL l = new LingvoDSL(new File("test/data/dicts/test.dsl"));
+        Map<String, Object> map = l.readHeader();
+        String word = "tool";
+        Object data = map.get(word);
+        assertNotNull(data);
+        String result = l.readArticle(word, data);
+        assertEquals("\u0441\u0442\u0430\u043d\u043e\u043a\n", result); //станок
```
amake: As a general rule, we don't allow non-ASCII characters in the source, as they will turn into mojibake in some environments.

amake:
Are you planning to address the other entry formats (with `{}`, etc.) that you mentioned in [your comment on RFE#1160](https://sourceforge.net/p/omegat/feature-requests/1160/#331c)?

src/org/omegat/core/dictionaries/LingvoDSL.java
miurahr: s.trim() seems to be best and simplest.

miurahr:
> Are you planning to address the other entry formats (with {}, etc.) that you mentioned in your comment on RFE#1160? If you prefer to make that a separate pull request, you can.

No I do not plan to address it. Because I'm not an actual user for LinvgoDSL and don't have a live dictionary data. I'd like to wait feedback from LinvgoDSL users.

amake:
OK, thanks for clarifying. We can leave that improvement for a future RFE then.

test/data/dicts/License.txt
```diff
@@ -1,3 +1,7 @@
+Liceses:
```
amake: Typo

test/src/org/omegat/core/dictionaries/LingvoDSLTest.java
```diff
@@ -0,0 +1,77 @@
...
+/**
+ * Dictionary test
+ *
+ * @author Hiroshi Miura
+ */
+public class LingvoDSLTest extends TestCore {
+
+    @Test
+    public void testReadFileDict() throws Exception {
+        LingvoDSL l = new LingvoDSL(new File("test/data/dicts/test.dsl"));
```
amake: `test/data/dicts` is currently home to a Hunspell dictionary. Since Hunspell dictionaries are often assumed to be housed in their own directories, it would be better to keep things separate. I have moved `test.dsl` to `test/data/dicts-lingvo`.

amake: You have specified the test dictionary path as a "magic" string repeated throughout this file. It would be better to make it a `static final` class member so that it's easy to change if necessary.

test/src/org/omegat/core/dictionaries/LingvoDSLTest.java
```diff
@@ -0,0 +1,77 @@
...
+import org.junit.Test;
+
+import org.omegat.core.TestCore;
+
+/**
+ * Dictionary test
+ *
+ * @author Hiroshi Miura
+ */
+public class LingvoDSLTest extends TestCore {
```
amake: `TestCore` includes a bunch of `org.omegat.core.Core` initialization stuff that should only be necessary for tests of GUI components. To maximize the number of tests that can be performed in a headless environment, it's best to extend `TestCase` whenever possible.

amake:
I moved this ticket to https://sourceforge.net/p/omegat/bugs/786/

I have merged this in trunk. I had to make some additional changes, which I described above in comments.

PR #3, opened 2015-11-21
Dictionary: Prefix search support

miurahr:
Depend on PR #1 #2 
Implement RFE#1124


miurahr:
I'd like re-submit on current master.

PR #4, opened 2015-11-21
Support Full-width and Half-width character insensitive search

miurahr:
Implement RFE#140

### DIFF
--- a/src/org/omegat/core/search/Searcher.java
+++ b/src/org/omegat/core/search/Searcher.java
@@ -174,30 +174,19 @@ public class Searcher {
         switch (expression.searchExpressionType) {
         case EXACT:
         default:
-            // escape the search string, it's not supposed to be a regular
-            // expression
-            text = StaticUtils.escapeNonRegex(text, false);
+            if (!expression.fullHalfWidthInsensitive) {
+                // escape the search string, it's not supposed to be a regular
+                // expression
+                text = StaticUtils.escapeNonRegex(text, false);
+            } else {
+                // handle half-width and full-width character insensitive match
+                // create regex expression to match both
+                text = StringUtil.createFullHalfMatchExpression(text);
+            }
 
             // space match nbsp (\u00a0)
             if (expression.spaceMatchNbsp) {
                 text = text.replaceAll(" ", "( |\u00A0)");
-            }
-
-            // handle half-width and full-width character insensitive match
-            if (expression.fullHalfWidthInsensitive) {
-                if (StringUtil.containsFullWidthExpression(text)) {
-                    // create a matcher for the full/half width search strings
-                    String full = StringUtil.getFullWidthText(text);
-                    String half = StringUtil.getHalfWidthText(text);
-                    // half can contain regex character generated from full-width character
-                    half = StaticUtils.escapeNonRegex(half, false);
-                    StringBuilder sb = new StringBuilder(half);
-                    sb.insert(0, "(");
-                    sb.append("|");
-                    sb.append(full);
-                    sb.append(")");
-                    text = sb.toString();
-                }
             }
 
             // create a matcher for the search string
--- a/src/org/omegat/core/search/Searcher.java
+++ b/src/org/omegat/core/search/Searcher.java
@@ -183,6 +183,11 @@ public class Searcher {
                 text = text.replaceAll(" ", "( |\u00A0)");
             }
 
+            // handle half-width and full-width character insensitive match
+            if (expression.fullHalfWidthInsensitive) {
+                text = StringUtil.fullHalfWidthMatchExpression(text);
+            }
+
             // create a matcher for the search string
             m_matchers.add(Pattern.compile(text, flags).matcher(""));
             break;
--- a/src/org/omegat/core/search/Searcher.java
+++ b/src/org/omegat/core/search/Searcher.java
@@ -208,9 +208,15 @@ public class Searcher {
                             : text.substring(wordStart, spacePos).trim();
 
                     if (!word.isEmpty()) {
-                        // escape the word, if it's not supposed to be a regular
-                        // expression
-                        word = StaticUtils.escapeNonRegex(word, false);
+                        if (!expression.fullHalfWidthInsensitive) {
+                            // escape the search string, it's not supposed to be a regular
+                            // expression
+                            word = StaticUtils.escapeNonRegex(word, false);
+                        } else {
+                            // handle half-width and full-width character insensitive match
+                            // create regex expression to match both
+                            word = StringUtil.createFullHalfMatchExpression(word);
+                        }
 
                         // create a matcher for the word
                         m_matchers.add(Pattern.compile(word, flags).matcher(""));
--- a/src/org/omegat/core/search/Searcher.java
+++ b/src/org/omegat/core/search/Searcher.java
@@ -185,7 +185,19 @@ public class Searcher {
 
             // handle half-width and full-width character insensitive match
             if (expression.fullHalfWidthInsensitive) {
-                text = StringUtil.fullHalfWidthMatchExpression(text);
+                if (StringUtil.containsFullWidthExpression(text)) {
+                    // create a matcher for the full/half width search strings
+                    String full = StringUtil.getFullWidthText(text);
+                    String half = StringUtil.getHalfWidthText(text);
+                    // half can contain regex character generated from full-width character
+                    half = StaticUtils.escapeNonRegex(half, false);
+                    StringBuilder sb = new StringBuilder(half);
+                    sb.insert(0, "(");
+                    sb.append("|");
+                    sb.append(full);
+                    sb.append(")");
+                    text = sb.toString();
+                }
             }
 
             // create a matcher for the search string
--- a/src/org/omegat/util/StringUtil.java
+++ b/src/org/omegat/util/StringUtil.java
@@ -31,8 +31,9 @@ package org.omegat.util;
 
 import java.text.MessageFormat;
 import java.text.Normalizer;
+import java.util.HashMap;
 import java.util.Locale;
-import java.util.regex.Pattern;
+import java.util.Map;
 
 /**
  * Utilities for string processing.
@@ -519,6 +520,7 @@ public class StringUtil {
      * Analyze text whether contains full/half width character
      * that has a codepoint in both full-width and half-width expression in Unicode
      *    ex. ASCII, full-width alphabet, full-width katakana, half-width katakana, etc.
+* 
      *
      * @param  String
      * @return String expresssion such as '(A|\uff21)(B|\uff22)(C|\uff23)'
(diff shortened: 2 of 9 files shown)


amake:
A general note for the patch as a whole: "Zen" and "Han" are Japanese-specific terms, but the issue itself is not limited to Japanese users. In English this would be "Full-width" and "Half-width".

Instead of "zenHan" in variable and method names, I'd rather see "width". For the user-visible label in the Search window, I prefer "Character width insensitive". You could add a tooltip on the checkbox to explain in more detail: "Lets half-width characters match equivalent full-width characters, and vice versa".

src/org/omegat/util/StringUtil.java
```diff
@@ -511,4 +511,31 @@ public static String format(String str, Object... arguments) {
...
+     * FIXME: handling 0x20-0x2f, 0x5c
+     *
+     * @param text
+     * @return String with regex
+     */
+    public static String replaceZenHan(String text) {
+        int i;
+        StringBuilder result = new StringBuilder(text.length());
+        for (i = 0; i < text.length(); i++ ) {
+            char c = text.charAt(i);
```
amake: Iterating by character is problematic. Please use the `codePoint`-based API:
https://docs.oracle.com/javase/6/docs/api/java/lang/String.html#codePointAt%28int%29

src/org/omegat/util/StringUtil.java
```diff
@@ -511,4 +511,31 @@ public static String format(String str, Object... arguments) {
...
+            if (0x0030 < (int) c && (int) c  < 0x007f && (int) c != 0x5c) {
+                result.append(String.format("(\\u%04x|\\u%04x)", (int) c - 0x0021 + 0xff01, (int) c ));
```
amake: Just FYI: `String.format` is very slow in Java. Since this method is going to be used in a performance-sensitive situation, I would rather see `Integer.toHexString` and multiple `appends` to `result`.

src/org/omegat/core/search/Searcher.java
```diff
@@ -183,6 +183,11 @@ public void search() throws TranslationException, PatternSyntaxException, IOExce
                 text = text.replaceAll(" ", "( |\u00A0)");
             }
 
+            // handle ZenHan insensitive match
+            if (expression.zenHanInsensitive) {
+                text = StringUtil.replaceZenHan(text);
```
amake: You should normalize the width of the search string (just once, of course) in addition to the `text.`

src/org/omegat/util/StringUtil.java
```diff
@@ -511,4 +511,31 @@ public static String format(String str, Object... arguments) {
...
+     *
+     * @param text
+     * @return String with regex
+     */
+    public static String replaceZenHan(String text) {
+        int i;
+        StringBuilder result = new StringBuilder(text.length());
+        for (i = 0; i < text.length(); i++ ) {
+            char c = text.charAt(i);
+            if (0x0030 < (int) c && (int) c  < 0x007f && (int) c != 0x5c) {
```
amake: You don't need all these casts to `int`.

src/org/omegat/util/StringUtil.java
```diff
@@ -511,4 +511,31 @@ public static String format(String str, Object... arguments) {
...
+     *
+     * replace (\uFF01 - \uFF5E) to (\u0021- \u007E)
+     * such as text = text.replaceAll("u\0021", "(\uFF01|\u0021)");
+     * except for chars that consist of regex.
+     * FIXME: handling 0x20-0x2f, 0x5c
+     *
+     * @param text
+     * @return String with regex
+     */
+    public static String replaceZenHan(String text) {
```
amake: I don't think this is a good way to go about matching character width for search purposes. Building a regex is unnecessarily complex and not reusable. I haven't benchmarked, but it smells unperformant as well. Also you are not addressing hangul at all.

Instead of this, I suggest the following algorithm:
1. Use a new util method `boolean containsFullWidth(String)` to check if the comparison text needs normalization at all. In the vast majority of cases (i.e. not Japanese text) this will return `false`.
2. If above returns `true`, normalize the comparison text with a new util method `String toHalfWidth(String)`.

I would base the implementation of both of the above on Okapi's [FullWidthConversionStep](https://bitbucket.org/okapiframework/okapi/src/8aa39fcb9977e5cf8e96b566bddbdfc89efa78b5/okapi/steps/fullwidthconversion/src/main/java/net/sf/okapi/steps/fullwidthconversion/FullWidthConversionStep.java?at=master&fileviewer=file-view-default#FullWidthConversionStep.java-104). It is, as far as I know, comprehensive in its handling of full-width/half-width characters.

miurahr:
Change issue title reflect comment;
>A general note for the patch as a whole: "Zen" and "Han" are Japanese-specific terms, but the issue itself is not limited to Japanese users. In English they would be "Full-width" and "Half-width".

amake:
Another issue I forgot to mention: The `SearchWindowController` remembers the state of all search window options, and also determines the display state of the Advanced panel if any options are not at their default settings. This new option needs to be added to both of these features. In `SearchWindowController` see:

- `loadPreferences()`
- `savePreferences()`
- `loadAdvancedOptionPreferences()`

src/org/omegat/util/StringUtil.java
miurahr: I'd like to try a following algorithm;

1. Build half-width version of search text if necessary
2. Build full-width version of search text if possible
3. If (1) and (2) is same, there is no characters to be aware for full/half width, just return original text
4. Escape special characters for regex with `'\'` ex. `'\|'`
5. Build regex `'(<half-width text>|<full-width text>)'` and return.

To implement 1 and 2, I can use a way Okapi's detection of characters incl. Hangul, katakana, etc. 

A method for 4 would be reusable so it can be independent util method, `String escapeRegexChars(String text)`

miurahr:
Thanks @amake for your suggestions

> In SearchWindowController see:
>
> loadPreferences()
> savePreferences()
> loadAdvancedOptionPreferences()

I'd like to try this.

src/org/omegat/util/StringUtil.java
amake: Why do you think steps 2–5 are necessary? We only need to normalize the texts to one form (half-width or full-width) and then do a normal comparison. There is no need to build a regex.

miurahr:
Take2

miurahr:
1 error for CI itself

amake:
Yeah I'm not sure why this is failing. Maybe it's because this PR existed before I set up CI testing. Don't worry about it.

amake:
Or it might be because you force-pushed over your initial commit. It's generally not a good idea to do that. I will squash your changes before committing to SVN anyway, so it's best to just add to the branch.

src/org/omegat/util/StringUtil.java
```diff
@@ -511,4 +511,223 @@ public static String format(String str, Object... arguments) {
...
+        // Create regex
+        StringBuilder sb = new StringBuilder("(");
+        sb.append(escapeRegexChars(sbhw.toString()));
+        sb.append("|");
+        sb.append(sbfw);
+        sb.append(")");
+        return sb.toString();
+    }
+
+    public static String escapeRegexChars(String text) {
```
amake: The standard library already has a method for this:
https://docs.oracle.com/javase/7/docs/api/java/util/regex/Pattern.html#quote%28java.lang.String%29

src/org/omegat/util/StringUtil.java
```diff
@@ -511,4 +511,223 @@ public static String format(String str, Object... arguments) {
...
+                case 0xFFED: sbfw.setCharAt(i, (char)0x25A0); break;
+                case 0xFFEE: sbfw.setCharAt(i, (char)0x25CB); break;
+            }
+        }
+        // whether we have a character to be converted.
+        if ( sbhw.toString().equals(sbfw.toString()) ) {
+            return text;
+        }
+        // Create regex
+        StringBuilder sb = new StringBuilder("(");
```
amake: Please explain why you think it's necessary to create a regex like this. As I mentioned before, we can just normalize to half-width or full-width and do a normal comparison.

miurahr:
>Or it might be because you force-pushed over your initial commit. It's generally not a good idea to do that. I will squash your changes before committing to SVN anyway, so it's best to just add to the branch.

There are two policy depending on project. 

- One is that commit should be split a part of meaning patch. When PR again, squashing it and force push. When merging it, history will be clean. OpenStack project use this way.
- Other is a way  as you mentioned. When PR again, just add diff to change somewhere. When accepting it, maintainer may squash for clean history.

For OmegaT project,  the latter is the way, I see.

amake:
If our main development was in git then the former might make sense, but since I have to rebase anyway to commit to svn (and when there are new files we need to add svn props so we *must* use svn) it's easier for me to squash.

src/org/omegat/util/StringUtil.java
miurahr: Because I think that creating full-width and half-width text in one iteration is better for performance.

Now I understand that following snippet will be better in Searcher.java
```
            // handle half-width and full-width character insensitive match
            if (expression.fullHalfWidthInsensitive) {
                if (StringUtil.containsFullWidthExpression(text)) {
                    m_matchers.add(Pattern.compile(StringUtil.getFullWidthText(text), flags).matcher(""));
                    text = StringUtil.getHalfWidthText(text);
                }
            }
            // create a matcher for the search string
            m_matchers.add(Pattern.compile(text, flags).matcher(""));
```

I can avoid creating regex using `m_matchers.add()`.
Will PR Take3 soon ...

amake: I took a closer look at the `Searcher` class and now I understand why you wanted to implement it as one big regex. That would certainly fit the existing implementation style.

In my initial suggestion what I intended was that we should normalize `expression.text` in `Searcher.search()` and then also normalize `text` in `Searcher.searchString()`. However I just realized that we will have offset problems if we do this because e.g. `プ  (U+30D7)` and `ﾌﾟ (U+FF8C U+FF9F)` are different lengths.

miurahr: m_matchers is a list for AND condition not OR condition. It is because a previous snippet is always fails.

in searchString():
```
        for (Matcher matcher : m_matchers) {
            // check the text against the current matcher
            // if one of the search strings is not found, don't
            // bother looking for the rest of the search strings
            matcher.reset(text);
            if (!matcher.find())
                return false;
```

I'd like to fallback to create a regex.

amake: Having two `Matcher`s like you suggested earlier wouldn't let us match mixed-width text anyway. I think you're right, we have to use a regex where we replace any character that has both half- and full-width forms with `(<half-width form>|<full-width form>)`.

Sorry it took me a while to see that your first approach was correct.

miurahr:
I think current version of PR is not optimal because it traverse original text 4 times in maximum.
And the search text often contains alphabet that is always considered  to check full-width alphabet.

amake:
The problem with Jenkins might be this:
https://github.com/jenkinsci/ghprb-plugin/issues/104

I've made the fix on Jenkins, so hopefully it should work for your next push.

src/org/omegat/util/StringUtil.java
```diff
@@ -511,4 +512,235 @@ public static String format(String str, Object... arguments) {
...
+
+    /**
+     * generate regex expression from two char or String parameter
+     *
+     * returns "(" + a + "|" + b + ")"
+     *
+     * @param char|String, char|String
+     * @return String  (a|b)
+     **/
+    private static final String genFullHalfMatchRegex(char a, char b) {
```
miurahr: It looks better to have a name such as 'String regOr(a,b)'.

miurahr: Also  (int cp1, int cp2) as parameter is better.

src/org/omegat/util/StringUtil.java
```diff
@@ -511,4 +512,235 @@ public static String format(String str, Object... arguments) {
...
+                // Others
+                case 0x2502: result.append(genFullHalfMatchRegex((char)cp, (char)0xFFE8)); break;
+                case 0x2190: result.append(genFullHalfMatchRegex((char)cp, (char)0xFFE9)); break;
+                case 0x2191: result.append(genFullHalfMatchRegex((char)cp, (char)0xFFEA)); break;
+                case 0x2192: result.append(genFullHalfMatchRegex((char)cp, (char)0xFFEB)); break;
+                case 0x2193: result.append(genFullHalfMatchRegex((char)cp, (char)0xFFEC)); break;
+                case 0x25A0: result.append(genFullHalfMatchRegex((char)cp, (char)0xFFED)); break;
+                case 0x25CB: result.append(genFullHalfMatchRegex((char)cp, (char)0xFFEE)); break;
+                // No match, add ch
+                default: result.append(codePoint2String(cp)); break;
```
amake: You don't need `codePoint2String`; you can use [`StringBuilder.appendCodePoint(int)`](https://docs.oracle.com/javase/7/docs/api/java/lang/StringBuilder.html#appendCodePoint%28int%29).

src/org/omegat/util/StringUtil.java
```diff
@@ -511,4 +512,235 @@ public static String format(String str, Object... arguments) {
...
+                continue;
+            }
+            // Full width Hangul
+            if (( cp > 0x3131 ) && ( cp <= 0x314E )) {
+                result.append(genFullHalfMatchRegex((char)cp, (char)(cp+0xCE70)));
+                continue;
+            }
+            switch ( cp ) {
+                // space
+                case 0x0020: result.append(genFullHalfMatchRegex((char)cp, (char)0x3000)); break;
```
amake: Instead of doing `result.append(genFullHalfMatchRegex(cp, <variant>))` in every single case, why don't you make a variable before the `switch` to hold the `<variant>` and then do the appending later.

src/org/omegat/util/StringUtil.java
```diff
@@ -511,4 +512,235 @@ public static String format(String str, Object... arguments) {
...
+                // Half width Katakana
+                case 0xFF61: result.append(genFullHalfMatchRegex((char)cp, (char)0x3002)); break;
```
amake: Your katakana handling will not work for these cases:
- `cp[i] == ｶ && cp[i+1] == ﾞ` will not match `ガ (U+30AC)`
- `cp[i] == ガ` will not match `ｶﾞ (U+FF76 U+FF9E)`
However you don't need to consider `cp[i] == カ && cp[i+1] == ゛` because we normalize to Unicode NFC when loading files.

test/src/org/omegat/util/StringUtilTest.java
```diff
@@ -237,4 +237,26 @@ public void testMatchCapitalization() {
...
+        assertEquals(text, StringUtil.createFullHalfMatchExpression(text));
+        text = "abc";
+        assertEquals("(a|\uff41)(b|\uff42)(c|\uff43)", StringUtil.createFullHalfMatchExpression(text));
+        text = "\u3000";
+        assertEquals("(\u3000|\u0020)", StringUtil.createFullHalfMatchExpression(text));
+        text = "\uff5c";
+        assertEquals("(\uff5c|\\|)", StringUtil.createFullHalfMatchExpression(text));
+        text = "\ud840\udc0b\uff41";
+        assertEquals("\ud840\udc0b(\uff41|a)", StringUtil.createFullHalfMatchExpression(text));
+    }
```
amake: A test case for katakana:
```
text = "\uff76\uff9e";
assertEquals("(\uff76\uff9e|\u30ac)", StringUtil.createFullHalfMatchExpression(text));
text = "\u30ac";
assertEquals("(\u30ac|\uff76\uff9e)", StringUtil.createFullHalfMatchExpression(text));

```

src/org/omegat/core/search/Searcher.java
```diff
@@ -174,9 +174,15 @@ public void search() throws TranslationException, PatternSyntaxException, IOExce
         switch (expression.searchExpressionType) {
         case EXACT:
```
amake: I would expect width-insensitivity to work on all modes, not just `EXACT`.

src/org/omegat/util/StringUtil.java
miurahr: Oh, I'll do `result.appendCodePoint(cp);`

src/org/omegat/util/StringUtil.java
miurahr: I want to change it to a simple table lookup from looong switch-case statement.

src/org/omegat/util/StringUtil.java
miurahr: Thank you for pointing out this. I feel some frastration for a histrical character code standard in Japan. (-.-);;

src/org/omegat/core/search/Searcher.java
miurahr: It is not good for `REGEXP`, but good for `EXACT` and `KEYWORD`. A user who use `REGEXP` may want to control search behavior by itself. 
I'd like to make option disable when selecting `REGEXP`.

src/org/omegat/util/StringUtil.java
amake: I agree. Half-width katakana is a pain. I wish people would stop using it!

src/org/omegat/util/StringUtil.java
amake: I disagree that a table lookup will be "simpler" in the end. Creating the table will take a similar number of lines, and then you will be wasting a lot of memory with persistent objects instead of inline `int`s. I agree that a long `switch` statement is ugly, but I think it's the most appropriate choice here.

src/org/omegat/core/search/Searcher.java
amake: > A user who use REGEXP may want to control search behavior by itself. 
You can make the same argument about the case-sensitivity option, but that works for all search modes. And the user can have full control by simply turning the sensitivity-related options off.

> I'd like to make option disable when selecting REGEXP.
I consider having this work in all modes to be a requirement for accepting this pull request.

miurahr: It is too hard for me to implement it. Because method should recognize regex expression and produce valid expression. it is not valid only with skipping chars that used for regexp.
```
Ex.
 "^\t[abc]+$"  become like "^t[abc\uff41\uff42\uff43]+$"
"^\t\$[abc]+$" become like "^\t\$\uff04[abc\uff41\uff42\uff43]+$"
"\p{javaWhitespace}"  keep it
```

It may become like an implementation of regex parser to support REGEXP mode and full-width and half-width insensitive search.

miurahr: I see I'd like to do turning the sensitivity-related options off when regex.

miurahr:
I think I've done responses for all comments.

amake:
I have implemented this using my initial strategy of normalizing the width of the search string and comparison strings. This allows width-insensitive search in all search modes. The major caveat is that search results can't be accurately highlighted in some cases.

PR #6, opened 2016-03-06
Dictionary: Improve test coverage

miurahr:
- Dictionary test: .idx.gz file type test
- Util: Base64 test.
  Base64 Public domain code has a test.


amake:
Thanks. I have applied the change from `.idx` to `.idx.gz`.

However with `Base64.java`, since it's not our code to begin with I don't think we need its unit tests. Really we should have been bundling it like `diff.java` as a separate JAR. But actually it turns out that since Java 1.6 the standard library contains a Base64 implementation so instead I have just removed `Base64.java` and switched to that instead.

PR #7, opened 2016-03-07
Dictionary: prefix match search for StarDict

miurahr:
-Introduce Trie4J library
-Use TRIE tree structure for internal data structure
 of dictionary index
 * Use DoubleArray structure for performance


miurahr:
Every complex part for data structure handling are inside a Trie4J library. (Thanks to great guys!)
We can just use it for application.

Main advantage to use Trie(aka. Prefix Tree, radix tree) is to add prefix search capability for keyword.
And there is a bonus  to reduce data size by cost of computing.

FYI: reference for Trie and DoubleArray Trie
https://en.wikipedia.org/wiki/Trie
http://linux.thai.net/~thep/datrie/datrie.html

lib/Licenses.txt
```diff
@@ -113,3 +113,5 @@ sources, in lib/sources/
 
 StringUtil uses code from the FullWidthConversionStep in the Okapi Framework
 (http://okapi.opentag.com/) under the GNU Public License version 2.1 or greater.
+
+trie4j-0.9.1.jar is dictributed under Apache 2.0 License 2.0
```
amake: Typo: "dictributed"

src/org/omegat/Bundle.properties
```diff
@@ -1402,7 +1402,7 @@ ABOUTDIALOG_COPYRIGHT=Copyright \u00a9 2000-2016 Keith Godfrey, Maxym Mykhalchuk
...
...s (LGPL License)\n  VLDocking Framework 3.0.5-SNAPSHOT (LGPL License)\n  Hunspell by L\u00e1szl\u00f3 N\u00e9meth and others (LGPL License)\n  JNA by Todd Fast, Timothy Wall and others (LGPL License)\n  Swing-Layout 1.0.4 (LGPL License)\n  Jmyspell 2.1.4 (LGPL License)\n  SVNKit 1.8.5 (TMate License)\n  Sequence Library (Sequence Library License)\n  ANTLR 3.4 (ANTLR 3 License)\n  SQLJet 1.1.10 (GPL License)\n  JGit (Eclipse Distribution License)\n  JSch (JSch License)\n  Base64 (public domain)\n  Diff (GPL)\n  trilead-ssh2-1.0.0-build217 (Trilead SSH license)\n  Lucene (Apache License 2.0)\n  TinySegmenter (public domain)\n  commons-*.jar (Apache License 2.0)\n  jwordsplitter-3.4.jar (Apache License 2.0)\n  LanguageTool.jar (LGPL License)\n  morfologik-*.jar (Morfologik license)\n  segment-1.4.1.jar (Segment License)\n  lucene-gosen-2.0.2-ipadic.jar (LGPL License)\n  pdfbox-app-1.8.1.jar (Apache License 2.0)\n  KoreanAnalyzer-3x-120223.jar (Apache License 2.0)\n  SuperTMXMerge-for_OmegaT.jar (LGPL License)\n  groovy-all-2.2.2.jar (Apache Licence 2.0)\n  slf4j-*.jar (MIT License)\n  juniversalchardet-1.0.3.jar (GPL)\n  DictZip from JDict (GPL)\n trie4j-0.9.1.jar (Apache License 2.0)
```
amake: Each continuing line should start with two spaces, but you only have one for `trie4j`.

amake:
trie4j requires Java 1.7. Now that trunk is moving to Java 1.8 we will be able to include this.

miurahr:
Rebased to current  trunk.

src/org/omegat/core/dictionaries/StarDict.java
```diff
@@ -199,6 +202,8 @@ public StarDictDict(File ifoFile) throws Exception {
                 mem.close();
                 idx.close();
                 bais.close();
+                MapDoubleArray<Object> result = new MapDoubleArray(loaded);
+                loaded = null;
```
amake: There is no reason to set null to this variable when it is a local variable that will go out of scope immediately afterwards.

src/org/omegat/core/dictionaries/StarDict.java
```diff
@@ -252,16 +257,16 @@ private void addIndex(final String key, final int start, final int len, final Ma
...
-            if (dictData == null) {
-                return Collections.emptyList();
-            }
             List<DictionaryEntry> result = new ArrayList<DictionaryEntry>();
-            if (dictData instanceof Entry) {
-                result.add(new DictionaryEntry(word, ((Entry) dictData).getArticle()));
-            } else if (dictData instanceof Entry[]) {
-                for (Entry entry : (Entry[]) dictData) {
-                    result.add(new DictionaryEntry(word, entry.getArticle()));
+            Iterable<String> it = data.predictiveSearch(word);
```
amake: *Always* doing a predictive search is going to be very noisy. I think we need to only do this when `Preferences.DICTIONARY_FUZZY_MATCHING` is `true`.

amake:
New libraries need to be added to `/release/WebStart-specific/OmegaT.jnlp` as well.

nbproject/build-impl.xml~
```diff
@@ -0,0 +1,1413 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
amake: This file should not have been committed.

miurahr:
Rebased against current master.
Fix indent to replace tab char to space.

amake:
I have folded this, #9 and [RFE#1245](https://sourceforge.net/p/omegat/feature-requests/1245/) into one development. A prototype is available here:
https://omegat.ci.cloudbees.com/job/omegat-prototype/42/

Let me know if there's anything I missed.

miurahr:
The prototype looks perfect to merge, but I found it lead me a heavy load for system.
```
top - 19:39:21 up 7 days, 18:41,  6 users,  load average: 2.22, 1.87, 1.45
Tasks: 402 total,   1 running, 301 sleeping,   0 stopped, 100 zombie
%Cpu(s): 55.4 us,  1.3 sy,  0.0 ni, 43.0 id,  0.2 wa,  0.0 hi,  0.0 si,  0.0 st
KiB Mem:   7836064 total,  7704888 used,   131176 free,    47224 buffers
KiB Swap:  5118972 total,  3284372 used,  1834600 free.   630360 cached Mem

  PID USER      PR  NI    VIRT    RES    SHR S  %CPU %MEM     TIME+ COMMAND                                                                                      
 1964 miurahr   20   0 5902856 1.421g  33568 S 201.9 19.0   3:21.06 java  
```

Condition: It use Eijiro for dictionary, and tested source data is OmegaT readme.txt file.

I'd like to investigate this tonight.

miurahr:
Sorry, it is strange why it consume such huge memory.  When config not to use prediction, it consume less memory less than 1GB.

It seems to load many articles than necessary.

amake:
Thanks for inviting me to your test project. I checked it out, and I also find that it's relatively slow to load the Eijiro dictionary (~10 s on my machine) and also slow to display entries.

I tested the memory usage this way:

1. Open OmegaT with `--no-team`
2. Attach with `jvisualvm` and watch the Monitor tab
3. Load the project and wait for the Dictionary pane to be populated
4. Perform a GC
5. Take a heap dump

- Current `topic/aaron/gradle` (cc59b4d52f5adbb4f4c3607a1ad3022cd3524f0b)
  - Peak heap size: 2.171 GB
  - Peak used heap: 1.137 GB
  - Post-GC used heap: 0.330 GB
  - Instances of StarDictDict.Entry: 2,189,840 (0.081 GB)
  - Instances of String: 380,066 (0.001 GB)
  - Instances of char[]: 377,024 (0.021 GB)
- Current `master` (r8835)
  - Peak heap size: 1.085 GB
  - Peak used heap: 0.531 GB
  - Post-GC used heap: 0.388 GB
  - Instances of StarDictDict.Entry: 2,189,840 (0.081 GB)
  - Instances of String: 2,428,057 (0.063 GB)
  - Instances of char[]: 2,424,976 (0.147 GB)

So we use a lot more memory up front creating the MapTrie (and MapDoubleArray?), but after that it's more efficient as we no longer store all the String keys formerly used in the Map.

amake:
Part of the problem is simply that Eijiro is a huge dictionary. A typical case probably won't be so big.

Is there a way to reduce the memory used to build the MapDoubleArray? Really the only issue here is peak memory usage; after the peak we are actually more efficient now.

miurahr:
I agree Eijiro is a huge and  a special case.
> Part of the problem is simply that Eijiro is a huge dictionary. A typical case probably won't be so big.

Even though it is special case, it show us a difference among implementations. Thank you for measurement.

> So we use a lot more memory up front creating the MapTrie (and MapDoubleArray?), but after that it's more efficient as we no longer store all the String keys formerly used in the Map.

It is cue for us to merge it to trunk, because it is more efficient. 

For future enhancement, we can cache the resulted MapDoubleArray data for next dictionary usage. It will be another RFE.

amake:
> For future enhancement, we can cache the resulted MapDoubleArray data for next dictionary usage. It will be another RFE.

Do you mean persist it to a file, or do you have something different in mind?

miurahr:
>Do you mean persist it to a file, or do you have something different in mind?

I mean persist it to a file.

miurahr:
I found  two problems with tokenizer option with this improvement.
It found so many idioms when tokenizer on.

It becomes duplicated entries, for example, test project with 'provide' looking up.
screenshot from 2016-05-15 13 07 47

In this example, we can see duplicated results.

Also, when hit idioms, it does not show whole original clause, ex.
```
provide optimum moisture without the use of oils
油分を使わずに[使用せずに]最適な水分量を与える
```

It may be necessary to consider how to combine prefix search (hit 'provided' from key 'provide') and prediction.

miurahr:
A previous duplication is a issue in dictionary, index has duplicated keys.
Here is better example for a duplication by prototype.

screenshot from 2016-05-15 16 24 21
screenshot from 2016-05-15 16 25 34

Same articles are retrieved  from 'provide' and 'provid'.

amake:
I know the cause. I'll fix it as soon as I can.

amake:
Fixed in https://omegat.ci.cloudbees.com/job/omegat-prototype/44/

miurahr:
An issue for title is fixed.
screenshot from 2016-05-16 00 37 48

But duplication issue still continues. It show same entries twice.

miurahr:
IMHO, an ideal way to look-up dictionary is as follows:

omegat_dictionary_architecture_ideal

If we want to realize this, we need to tokenize a result of lookups and count matches.
Prefix search may help us to predict words, we can use just stemming for token, and then we should classify or  filter a result, because it  easily become the false-positive.

amake:
I agree that the current implementation gives too much noise. Currently I am thinking about only doing predictive lookup for stems, like this (Python pseudocode):
```
for token in tokenize(source_text):
  stem = get_stem(token)
  if stem != token:
    return dict_lookup_predictive(stem)
  else:
    return dict_lookup(token)
```

amake:
I tried loading the Eijiro project on an instance where the JVM was started with `-Xmx1024M` (the current default for Windows) and it actually managed to load the dictionary without issue. Of course there was some GC activity, but it didn't feel any slower than with unbounded max heap, which is ~4GB(?) on OS X.

So unless there are any additional issues to cover in this RFE, I will commit to trunk soon.

miurahr:
Can you add filter to reduce duplication?

Here is a super-dirty patch what I want to do;

```
diff --git a/src/org/omegat/core/dictionaries/DictionaryData.java b/src/org/omegat/core/dictionaries/DictionaryData.java
index 733894c..e965e6e 100644
--- a/src/org/omegat/core/dictionaries/DictionaryData.java
+++ b/src/org/omegat/core/dictionaries/DictionaryData.java
@@ -173,7 +173,8 @@ public class DictionaryData<T> {
         }
         List<Entry<String, T>> result = new ArrayList<>();
         if (predictive) {
-            data.predictiveSearch(word).forEach(w -> get(w, data.get(w), result));
+            List<String> keys = new ArrayList<>();
+            data.predictiveSearch(word).forEach(w -> get(w, keys, data.get(w), result));
         } else {
             get(word, data.get(word), result);
         }
@@ -181,6 +182,17 @@ public class DictionaryData<T> {
         return result;
     }
 
+    private <U> void get(U key, Collection<U> keys, Object value, Collection<Entry<U, T>> into) {
+        if (value == null) {
+            return;
+        }
+        if (keys.contains(key)) {
+            return;
+        }
+        keys.add(key);
+        get(key, value, into);
+    }
+
     /**
      * Unpack the given stored object (singular, or array) into the given
      * collection.
```

amake:
As far as I can tell, your patch doesn't change anything. The keys you're trying to filter are already unique. A quick test with a debug print line after the `if (keys.contains(key))` line turned up nothing.

What exactly are you trying to do?

miurahr:
In my environment, It changes behavior, that make duplication fixed. (the issue reported https://github.com/omegat-org/omegat/pull/7#issuecomment-219292931)
WIthout this modification, topic/aaron/dict branch HEAD does produce this issue.

amake:
1. I tried again with your patch plus debug print line, and could see no effect whatsoever. It doesn't even make sense that it would change anything: the output of `data.predictiveSearch(word)` will not have duplicates, so filtering the output as your patch does will accomplish nothing.

2. I cannot reproduce the duplication you show in comment 7. I am using 5b273eaf9edf581458ece3db83b301b1c2560df3.

miurahr:
Thank you for testing again. I used b881795df1ed1bef32dd1796a8b0477a9d82f810 on Linux Mint 17.3 64bit with JDK 8u91-b14-0 ubuntu4~14.04. I run omegat with 'gradle run'.  I'd like to test again from clean environment.

I agree that the output of data.predictiveSearch(word) will not have duplicates in API document and I wondering  why  it happened. It may be my problem in environment, but I'd like to investigate it.

amake:
Your git revision is OK. I wonder if it's possible that you have two copies of the Eijiro dictionary somewhere in your project's dictionary path? I think that would explain the duplication you're seeing.

miurahr:
Sorry, the duplication is not produced when I  tested again from cleaning repository. It would bw a issue in my operation. Thanks for great job!

It is Ok to merge it.

Now I find another behavior that now it don't look up when exact keyword match some articles.
In previous example, following article is no more hit against original provide' term.
```provide $__ a month tocover medical cost``` 

it would be a side effect of https://github.com/omegat-org/omegat/commit/bcda35a62dc3688f1fd06c920f84b93ae1fa970f and it should be addressed as another RFE.

amake:
Yes, that's intentional. To prevent noise, I think it's best to only fall back to predictive search if a regular search doesn't find anything.

We could loosen that restriction if we make some improvements to the presentation layer like you suggested earlier.

amake:
Committed to trunk.

PR #9, opened 2016-04-29
StarDict: Update dictzip library to fledged version.

miurahr:
RFE#1242

- Replace dictzip.jar to dictzip-0.8.1 that is now  tested and generalized library. 
 It changes an interface for seek() than original one.
 It has also capability to write a dictzip file.
 https://bintray.com/miurahr/maven/dictzip-lib/view

- Replace to try (resource) { block } style for input streams.


amake:
Thanks for this. We're likely going to move to building with Gradle soon so we will be able to include your library by pulling from jcenter rather than with a JAR. (This also affects #7.)

Also, I don't think we need the sources JAR or even the sources ZIP anymore:

- The sources JAR will be available through Gradle's classpath settings (if desired).
- The sources ZIP was only required before because you had made a special modification of jdictd specifically for OmegaT. Since you are now publishing and maintaining this as a standalone library, we would only need to include your source for licensing reasons. However since it is GPLv2+, that isn't necessary.

miurahr:
Is it better to move on Gradle feature branch for this work?
If so, I'd like to close here and propose again.

amake:
No, please wait for the move to Gradle to hit trunk.

amake:
This will be incorporated into the patch for #7.

PR #11, opened 2016-07-16
LanguageTool network bridge

chelobaka:
This is an implementation of a new LanguageTool network API with JSON replies. Tests don't use external path because one have to setup it first to work. One way to do it: use official server [https://languagetool.org/api/v2/check](https://languagetool.org/api/v2/check) and change LanguageToolWrapper constructor parameter to false. Also the document should be small because this server rejects large number of requests coming one right after another which happens during document opening in OmegaT.


amake:
I have a number of issues:

1. Using `Boolean`s as a three-state flag is really bad practice. If you really need three states you should use an enum; otherwise it should be separate `boolean` flags.
2. Having a boolean flag that branches the class's main functionality is a strong smell indicating that we should be splitting this into separate classes: one for the built-in LT and one for querying an external LT.
3. I'd like to see some unit tests for the new functionality. OmegaT already ships with almost all of LT, so it should be possible to launch an LT server either as-is or with just one(?) more LT package, which could be a test-runtime-only dependency.

src/org/omegat/languagetools/LanguageToolWrapper.java
```diff
@@ -118,13 +167,18 @@ protected JLanguageTool getLanguageToolInstance(Language ltLang) {
     @Override
     public synchronized List<Mark> getMarksForEntry(SourceTextEntry ste, String sourceText, String translationText,
             boolean isActive) throws Exception {
+
         if (translationText == null || !isEnabled()) {
```
amake: Sorry for giving you bad information in the dev ML. `translationText` is indeed required, so the multiple null checks were redundant. Thanks for catching that.

chelobaka:
I almost finished refactoring this code and I have some questions left.

1. During initialization of network bridge we can do sanity checks on local path, free local socket, remote server availability. Shall the code fallback to native bridge in case of a failure or it's better to create a dummy bridge and switch to it so a user can see it's configuration is bad?
2. I have a working implementation of configuration dialog for LT. Can I register the wrapper into the core so when a user changes LT settings the wrapper will reread configuration and apply changes to its bridge? It's the way spellchecking dialog works.
3. Right now I have tests for network flavor of LT which talk to official languagetool.org. I agree that keeping everything in one place is the best solution but staring and closing external process safely  would require extra logic. Do the tests run on machine without Internet access? If not then my solution could be a good compromise.

P.S. I see some changes landed in current LT code. I'll merge them as soon I finish current task.

amake:
1. I would log the errors and fall back to the built-in implementation.
2. If you want to be notified of changes in the values or if you must store the values as local members then I would suggest [registering a `PropertyChangeListener`](https://omegat.ci.cloudbees.com/job/omegat-trunk/javadoc/org/omegat/util/Preferences.html#addPropertyChangeListener-java.beans.PropertyChangeListener-). Otherwise you can just get the values directly from `Preferences` at the point that you need them (it's just a `HashMap` lookup, so there's no performance benefit to caching the values locally).
3. I would not want to have our tests depend on network access or be hitting a third party server. Your current patch has code for starting and stopping a server; why can't this be used? Is that not precisely what it does?

chelobaka:
I've updated and refactored the code. Test related code is ready but not included because I can't figure out how to set `test` dependency. Adding _testCompile 'org.languagetool:languagetool-server:3.4'_ to build.gradle doesn't seem to do anything. Adding  it as normal dependency works fine.

chelobaka:
Weird stuff with Jenkins error. My local branch is surely synchronized with github.

chelobaka:
OK, the code is rebased against master. A little piece of test related code slipped into one of the previous commits so I pushed small helper function so the code would compile.

amake:
There were a lot of changes I wanted to recommend, but it would have been hard to explain by just commenting, so I have sent a pull request on your pull request (gitception!).
https://github.com/chelobaka/omegat/pull/1

chelobaka:
Thank you very much for spending time on my mumbo-jumbo!
You dropped support for applying different punctuation rule. Shall I do it some other way or it's already decided that OmegaT doesn't need it?
Right now there is only a configuration dialog left outside of the tree. Shall I fix it to work with new preferences and push?

amake:
> You dropped support for applying different punctuation rule. Should I do it some other way or it's already decided that OmegaT doesn't need it?

I dropped it because it would be better to implement a proper UI for enabling/disabling rules instead of including a stopgap secret preference that we would have to support indefinitely (at least in theory).

We will be releasing 4.0 likely in late August; I think we should target the next version with this PR. That gives us plenty of time to polish the config dialog, if you're up for it.

Off the top of my head, these are my thoughts on the UI:
- The user should be able to select the LT bridge type: maybe radio buttons representing the `BridgeType` enum options
- Values for `LANGUAGETOOL_REMOTE_URL`, `LANGUAGETOOL_LOCAL` values for should persist even if not active
- The user should be able to pick from a list of rules, but the list should not be too detailed (e.g. `FRENCH_FOO_RULE`, `GERMAN_BAR_RULE`). I believe LT defines rule "categories"; that's probably what we should expose.
- We will only be able to expose rules we know about, i.e. that are included in the bundled LT version. However newer server instances might have new rules we don't know about, which the user won't be able to configure. Thus we should probably offer a toggle between "blacklist" and "whitelist" mode, which will allow a bit better control.
- The list of selected rules should probably be persisted as a comma-delimited string: `FOO_RULE,BAR_RULE,BAZ_RULE`.

More thoughts:
- The`LANGUAGETOOL_LOCAL_PORT` pref seems redundant. Anyone needing to change the port should just launch the server themselves.
- The `LanguageToolNetworkBridge` constructor for the local installation should probably take the path to `languagetool-server.jar` instead of the path to the installation folder.

chelobaka:
I still don't have a full picture of UI because some questions need to be answered first.
-  Each LanguageTool rule has a category,  issue type and id. Lists of categories and issue types can be extracted from LanguageTool without referring to a specific language, but list of rules can be extracted only after creating JLanguageTool instance which requires passing language(s) to a constructor.
- Blacklisting some categories is handy in some situations. For example _TYPOS_ category contains spellchecking errors which OmegaT handles by itself. But if we want for example to enable/disable different punctuation rule it can be done only by its id.
- Should we keep two lists (categories and rule ids) in preferences dialog, so a user would be able to select in both? Rule id list can populated after a project load.
- I'm not sure about whitelist idea. What's the point of using newer LanguageTool and avoiding all new rules altogether? Maybe we could enable `debug` output so a user could track new annoying rule and add it to the blacklist? Current output of locally spawned server should be dealt with.

Another question I have is about communication between LanguageToolWrapper and configuration dialog. Earlier you suggested using _PropertyChangeListener_ for this and it worked because the dialog had top level boolean property. Change events always fire for boolean properties and having a statefull wrapper I could determine if the bridge needs to be recreated. Now there only String properties left and it's impossible to find out what should be done having information only on one variable change. It looks like that _restart_ logic should reside inside the dialog and the dialog has to be able to access wrapper instance. If so how it's better to do: register our wrapper in the core like spellchecker or make the wrapper a singleton?

src/org/omegat/languagetools/LanguageToolNetworkBridge.java
```diff
@@ -0,0 +1,272 @@
...
+import java.util.Map;
+
+import org.omegat.core.Core;
+import org.omegat.core.data.SourceTextEntry;
+import org.omegat.gui.editor.mark.Mark;
+import org.omegat.util.Language;
+import org.omegat.util.Log;
+import org.omegat.util.OStrings;
+
+import net.arnx.jsonic.JSON;
```
amake: You can now parse JSON natively in Java 8 without dependencies:
http://www.adam-bien.com/roller/abien/entry/converting_json_to_map_with

Did you evaluate this approach?

src/org/omegat/languagetools/LanguageToolNetworkBridge.java
```diff
@@ -113,10 +115,25 @@ public LanguageToolNetworkBridge(String path, int port) throws Exception {
         // Run the server
         ProcessBuilder pb = new ProcessBuilder("java", "-cp", serverJar.getAbsolutePath(), SERVER_CLASS_NAME, "--port",
                 Integer.toString(port));
-        pb.inheritIO();
-
+        pb.redirectErrorStream(true);
         server = pb.start();
 
+        // Create a thread to consume server output
+        new Thread(new Runnable() {
```
amake: You can make this much more concise:

```
new Thread(() -> {
    try (InputStream is = server.getInputStream()) {
        int b;
        while ((b = is.read()) != -1) {
            // Discard
        }
    } catch (IOException e) {
        // Do nothing
    }
}).start();
```

amake:
Yes, I checked the API and how the LanguageTool GUI works, and it looks like we really do need to provide language-specific, per-rule configuration (rather than just by category).

I think taking a page from the LanguageTool GUI and having basically a `JTree` of categories with rules as leaves should suffice; there's no need to separate them in the UI. For persistence, though, we probably need to store the selected categories and rules separately, since they are configured differently.

> I'm not sure about whitelist idea. What's the point of using newer LanguageTool and avoiding all new rules altogether?

Existing rules may have desirable bug fixes, while new rules may be undesirable for whatever reason (too many false positives, etc.). A whitelist is the only way to salvage that use case. But we could leave out the whitelist option and take the stance that the user should just configure the server as desired in that case.

> Another question I have is about communication between LanguageToolWrapper and configuration dialog

I don't think there should be any communication; that would be overly tight coupling.

> Now there only String properties left and it's impossible to find out what should be done having information only on one variable change

Recreate the bridge when any of the properties change. It's not that expensive, and it's not that common a situation (it will never happen in a loop).

amake:
I should have mentioned that since the LanguageTool configuration UI depends on the user's active language, we should simply make the dialog openable only when a project is loaded, and use the project's language. The dialog can instantiate its own instance using `LanguageToolWrapper`'s static methods.

chelobaka:
On nashorn JSON parsing:
I reproduced JMH benchmark given in the comments of the blog post. At first sight nashorn solution looked very slow with very short JSON example. I decided to extend the tests with JSONIC and JSON.simple and put real LanguageTool response in tests. The picture changed dramatically: 
>Benchmark                                 Mode  Cnt      Score             Error          Units
JSONBenchmark.jackson         thrpt    20       63044,363 ± 2727,630  ops/s
JSONBenchmark.json_simple  thrpt    20       16733,050 ±  850,395   ops/s
JSONBenchmark.jsonic             thrpt    20       21407,726 ± 1000,077  ops/s
JSONBenchmark.nashorn         thrpt    20       76886,642 ± 2756,419  ops/s 

Test source is [here](https://gist.github.com/chelobaka/b0be93cc2e0c37c8c83879c7a0468c1a). The  downside of using nashorn would be poor code readability. No idea about its memory footprint though.

chelobaka:
Latest commit implements JSON parsing using Nashorn. The code is ugly: a lot of casting and redundant loop. Can we extend Mark class so it includes rule and category identifiers. This way filtering can be moved inside the wrapper.

amake:
The casting is to be expected since we are processing JSON. Either we do it ourselves or a library does it for us; I'm not really concerned. As far as the redundant loop goes, you can stream the `matches` variable, map directly to `Mark`s, and collect to a list instead of creating an intermediate list with an intermediate data class (the `LanguageToolMatch` class is not necessary as currently implemented).

> Can we create an extended Mark class for LT case so it includes rule and category identifiers.

It sounds like you want to query LT for all rules and then filter the output? I'd prefer not to query it for unnecessary rules in the first place, rather than discard the results. Some rules may be costly to evaluate.

src/org/omegat/languagetools/LanguageToolNetworkBridge.java
```diff
@@ -227,29 +222,18 @@ public void destroy() {
         }
 
         List<Map<String,Object>> matches = (List) response.get("matches");
-        ArrayList<LanguageToolMatch> ltMatches = new ArrayList<>();
+        List<Mark> r = new ArrayList<>();
 
         matches.stream().forEach((match) -> {
```
amake: Rather than create a new `ArrayList` and repeatedly add to it in a `forEach` I would recommend mapping and collecting directly to a list, like:

```
return matches.stream().map(match -> /* create and return new Mark */).collect(Collections.toList())
```

chelobaka:
Yes, it's better to filter out rules as soon as possible. Cleaned up JSON parser and some other parts of `LanguageToolNetworkBridge`.

chelobaka:
Thanks for the tip  :+1:

chelobaka:
Right now the dialog only communicates with preferences. Communication with the wrapper and filtering inside bridges should follow later.
Some classes and methods were taken from LanguageTool code. Importing LanguageTool package
didn't make sense because most of the logic was modified to fit our needs.

chelobaka:
According to my testing everything works as expected now.

chelobaka:
Aaron, can you please give a feedback on commit c5dc32bf9b043d93184e7557c11c17f7022f0876? I'm not quite sure how should I modify current code to meet new multithreading concept. Also ThreadLocal memebers are non-static there which puzzles me too. Is there any reason for this?

src/org/omegat/languagetools/LanguageToolWrapper.java
```diff
@@ -136,52 +166,21 @@ public synchronized void onProjectChanged(PROJECT_CHANGE_TYPE eventType) {
...
-        } catch (Exception ex) {
-            // bitext rules can be not defined
-            return null;
+    @Override
+    public void propertyChange(PropertyChangeEvent evt) {
+        if (!Core.getProject().isProjectLoaded()) {
+            return;
+        }
+        // This property is changed in the end of configuration dialog saving,
+        // so at this point every other related properties are already changed.
```
amake: It's actually an accident that this works: `PropertyChangeEvent`s aren't supposed to be sent when the new value is equal to the old value, but due to how the `Preferences` backing store is implemented, they will be sent every time in the case of non-string values.

I understand the appeal of what you did here (you only check one property; you only rebuild once, after all properties have finished changing) but I'd rather not rely on this always being the case. If you want to keep things basically the same but be safer, I'd recommend defining another preference like `LANGUAGETOOL_PREFS_CHANGED_AT` where the value is a string representing the date and time of the last change to the prefs.

chelobaka:
Thank you for your work, Aaron. I added new property as you suggested.

After your changes to the dialog form OK/Cancel buttons are located right near to dialog border. It doesn't look very clean to me. Was it intended?

Also I can translate new strings into Russian. Shall I do it here or make another pull request after this one is merged?

amake:
> I added new property as you suggested.

Great. One thought: Right now we are only intending to use the property to detect changes to the settings, so as long as it differs over time we are OK. However, in case there is ever a reason we want to actually know the date/time of the last change, it's best to use a format that doesn't depend on time zone or user locale. For instance XML often uses this format (from `TMXWriter2`):

```
DateFormat format = new SimpleDateFormat("yyyyMMdd'T'HHmmss'Z'", Locale.ENGLISH);
format.setTimeZone(TimeZone.getTimeZone("UTC"));
```

> OK/Cancel buttons are located right near to dialog border

I develop on OS X, which gives reasonable padding by default. I added a bit in the latest PR; you can adjust if necessary.

amake:
> Also I can translate new strings into Russian. Shall I do it here or make another pull request after this one is merged? 

There may be a Russian translation team, in which case you should coordinate with them (they may have a TM where your strings should go, so they can generate the translated docs instead of you editing and committing by hand). Check with @suzume for that.

chelobaka:
I wondered if we need to store disabled categories per language as we do with rules. It doesn't cost anything and may be useful for some users.

amake:
Yes, ideally we would store the target-language rules namespaced to the language, and the bilingual rules namespaced to both languages.

amake:
If there's nothing more you'd like to add, I think this development is about ready to be merged.

There is one more thing I would suggest, though it's not strictly necessary: I've been working on an "Issues" window to replace the Tag Validation window. The Issues window will aggregate tag problems, spelling mistakes, and LanguageTool issues into one view. Each issue type, when possible, has a context menu with relevant actions, e.g. ignore word or add to dictionary for spelling issues. Prototype build here:
https://omegat.ci.cloudbees.com/job/omegat-prototype/57/

For LanguageTool issues I thought it would be good to offer a context menu action to disable the rule that produced the issue. To that end, rather than duplicating the code that manages storing the disabled rules/categories list, I thought it would be good to aggregate those into a reusable form.

My thought is to make a `LanguageToolPrefs` class with static methods like `disableRule(String rule)` that adds the provided rule name to the relevant preference. I did something similar for managing disabled issue types: https://github.com/omegat-org/omegat/blob/topic/aaron/issues3/src/org/omegat/gui/issues/IssueProviders.java#L66

If we add something like this then it feels like there's not much point in maintaining a strict separation between the LanguageTool config dialog and the LanguageToolWrapper, so I guess we could get rid of the indirection of using `LANGUAGETOOL_PREFS_CHANGED_AT` and just have the dialog call `setBridgeFromCurrentProject()` directly. This feels like tight coupling to me, but maybe it's not a problem. Do you have any thoughts on a better way to do this?

One interesting issue that arises from combining the Issues window with your work is that it becomes possible to spam a remote LanguageTool server with repeated requests, one for each translated segment in the project, as fast as one's connection will allow. I'm thinking of perhaps simply disallowing use of a remote LT bridge with the Issues window. Do you have any thoughts on this?

chelobaka:
Sorry for being away for a long time. I agree on your previous comment that we should keep rules and categories separated by both languages. It looks like it's better to change dialog so it would work with rule instances instead of their IDs so we can easily see the class of a rule (normal, bitext, external) when storing settings. Please provide your opinion on that matter.

Moving settings related code to a separate class seems OK. After you reply on my previous question I'll do it and push here for a review.

I don't see a problem with a tight coupling here. Over-engineering usually does more harm than good.

I stumbled upon a spam issue when tested network bridge with official LT server. It's configured to reject fast consecutive queries and opening a document in a project resulted only a couple of first segments receive their LT hints. Running a personal LT server (somewhere on the LAN for example) shouldn't cause any problems with spam queries. Maybe we can add a hint in the configuration dialog so a user would be notified that remote server can reject queries in some cases?

amake:
I did a little testing and it isn't really necessary to namespace the bilingual rules by both languages. The worst that can happen with the current implementation is that a bilingual rule for a different set of languages shows up as a manually defined rule.

That said, if you want to go to the trouble I think it would be OK to change the view model to use Rule instance directly.

> How are you going to retrieve rule ID for a disable command? Shall we set it somewhere within the Mark instance?

The current implementation uses the list of `RuleMatch`es so it can get the rule ID from there. When we merge your work, I will refactor the bridges to allow obtaining a list of some intermediate representation that abstracts both `RuleMatch` and the JSON parsing result from the network bridge.

> I stumbled upon a spam issue when tested network bridge with official LT server. It's configured to reject fast consecutive queries

Interesting. I'm a bit torn, since ideally we wouldn't allow the user to spam the server even if there are mitigations in place on the server side. I'm not sure how to do that without damaging the privately-run-server scenario. Maybe we can just blacklist the official LanguageTool server from use when checking the whole project.

chelobaka:
Can you have a look at LanguageToolPrefs prototype please? https://gist.github.com/chelobaka/d25aef100c4330eecb15ec0873fce589

I just want to be sure I understood you correctly.

amake:
Yes, that's what I was talking about. Thanks!

amake:
Sorry, I conflicted you again on trunk. We're no longer storing dependency binaries in the repo (yay!) so you can remove the new LanguageTool JARs.

chelobaka:
Pushed `LanguageToolPrefs` and merged master. I hope I didn't mess `build.gradle` (fingers crossed).

The only open question now is spamming issue. The simplest solution would be adding a special member to `LanguageToolNetworkBridge` which will hold the time of last check and if next request comes sooner than a defined threshold (50 ms for example) after that time then it would be discarded and the timer updated.

chelobaka:
Latest changes in master broke this pull request once again. I decided not to touch it myself since IssueProvider interacts with LangauageToolWrapper and I can break things easily without full understanding of how they work.

amake:
Yes, sorry, I'm actually working on a squash of this PR against master right now. I'll let you know how to proceed soon.

amake:
I've squashed this against master in #12. Let's continue discussion over there.

PR #12, opened 2016-09-22
LanguageTool network bridge squash

amake:
This is a squash of #11 in preparation for commit to svn trunk.

Changes vs #11:

- The check on the remote server only looked at the response headers, but I found that it's easy to supply a URL for the old XML API, in which case communication will fail when we try to parse the JSON response. The check now looks to make sure the server is not sending XML.
- Enabling/disabling rules on the native bridge was only affecting the LT instance in the thread that made the change, so issue checking (which happens on a special thread pool) did not see the rule changes at all. I fixed this but my solution is very unsatisfying.

I haven't really tested the network bridge beyond the unit tests, so I will do that before committing. That should give me a better idea of how important spam prevention will be.


src/org/omegat/languagetools/LanguageToolIssueProvider.java
```diff
@@ -124,7 +118,7 @@ public int getSegmentNumber() {
 
         @Override
         public String getDescription() {
-            return "<html>" + match.getMessage().replace("<suggestion>", "<i>").replace("</suggestion>", "</i>")
+            return "<html>" + result.message.replace("<suggestion>", "<i>").replace("</suggestion>", "</i>")
                     + "<html>";
```
chelobaka: Shouldn't the closing tag look like `</html>`?

amake: Yes, thanks, I'll fix that.

amake:
I changed the dialog layout to put the radio buttons next to the options they correspond to. Otherwise the only hint about what goes with what is the enabled/disabled graying, which might not be noticeable for some users.

I also changed the wording to what I thought would be clearer. If you don't like any of the changes or have additional suggestions please let me know.

(Feel free to open pull requests against this branch, too.)

chelobaka:
I like new dialog layout. It's cleaner now. Probably can't help with improving multithreading model because of lack of time now.

amake:
I did some testing and found that using `MultiThreadedJLanguageTool` instead of `JLanguageTool` + `ThreadLocal` doesn't speed anything up. In fact it seemed a bit slower, but that could have been my imagination. It reduces garbage a bit, but on the other hand I can reliably get [concurrency exceptions](languagetool-org/languagetool/issues/555) with it so it seems a poor choice.

src/org/omegat/Bundle.properties
```diff
@@ -2036,10 +2036,16 @@ GUI_LANGUAGETOOL_URL=URL:
 
 GUI_LANGUAGETOOL_RULES=Rules
 
-GUI_LANGUAGETOOL_BAD_RULE_ID=Rule ID can only contain latin letters and underscores.
+GUI_LANGUAGETOOL_BAD_RULE_ID=Rule ID can only contain Latin letters and underscores.
```
chelobaka: During debugging network bridge I found out that there are rules with dots in names, so I changed regexp to `"^[A-Za-z_\\.]+$"` Not sure if LT devs just overlooked those rule names in their code.

chelobaka:
Thank you for your hard work Aaron! Current state is a huge improvement over my code.

src/org/omegat/Bundle.properties
amake: OK, we can clarify the wording of this string. Can you point me to an example of a rule with a `.` in the ID?

chelobaka: There are two rules (rulegroups actually, but we see them as rules from outside) in Russian `grammar.xml`:
`<rulegroup id="V.А" name="Обозначение физических величин «В⋅А»">`
`<rulegroup id="KWT.H" name="Обозначение физических величин «кВт⋅ч»">`

amake: Wow. I would never have found those.

amake:
This is now committed to trunk, so I will close this PR.

Thanks for all your hard work.

If you have any more changes to make please open a PR any time.

amake:
@chelobaka I wrote up this PR as an RFE:
https://sourceforge.net/p/omegat/feature-requests/1273/

If you have a SourceForge account I can add you to the OmegaT project.

chelobaka:
Thank you Aaron! I don't have a SourceForge account. In case I'll have something to contribute github pull request should be enough. The only thing I'm interested now is a kramdown filter but I can't say when I could be able to come with a complete solution. Nested HTML makes it a bit un-trivial.

PR #13, opened 2016-10-12
Fix parsing of resource bundle cli parameter

chelobaka:
I had to apply this patch to be able to pass custom resource bundle to the program during its translation.


amake:
This is fixed in trunk.

Note that while that code was incorrect, it is only there for backwards compatibility. The preferred way of specifying that is `--resource-bundle=<path>`, which should have worked without your fix.

chelobaka:
Thank you Aaron. Localization documentation suggests passing the parameter the old way http://www.omegat.org/en/howtos/localizing_omegat.php. It needs to be updated.

PR #14, opened 2016-10-16
Update dictzip library

miurahr:
Dictzip library have updated with some minor fix.
The project home moves to https://github.com/dictzip/dictzip-java


amake:
Thanks. This is committed to trunk (r9340).

PR #16, opened 2016-10-25
Added 'RELAX NG schema for XML (*.rng)'.

tgraham-antenna:
This adds an XML filter for RELAX NG, which is a schema language for XML.  See http://relaxng.org/

RELAX NG has an XML syntax and a compact, non-XML syntax.  This handles the XML syntax only.

Tests are included.

### DIFF
--- a/Plugins.properties
+++ b/Plugins.properties
@@ -15,6 +15,7 @@ filter=org.omegat.filters3.xml.android.AndroidFilter \
     org.omegat.filters3.xml.helpandmanual.HelpAndManualFilter \
     org.omegat.filters3.xml.properties.PropertiesFilter \
     org.omegat.filters3.xml.schematron.SchematronFilter \
+    org.omegat.filters3.xml.relaxng.RelaxNGFilter \
     org.omegat.filters3.xml.camtasiawindows.CamtasiaWindowsFilter \
     org.omegat.filters2.html2.HTMLFilter2 \
     org.omegat.filters2.hhc.HHCFilter2 \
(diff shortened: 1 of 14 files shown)


amake:
I'm just curious: Why did you remove `isFileSupported()`?

tgraham-antenna:
Because neither it nor the DocBook version that it was based on would reject files where I'd introduced errors that the method was supposed to catch.

amake:
I would much prefer to figure out what's wrong with the method, if anything, rather than simply remove it.

I've pushed a new unit test for `DocBookFilter.isFileSupported()` in e961ef85afb9a27d78ac4cf754294ba467cebc8b. It seems to work fine. If you can't get yours to work then feel free to push it anyway and I will take a look.

tgraham-antenna:
On 25/10/2016 14:21, Aaron Madlon-Kay wrote:
...
>  1. The package name does not match the actual path for
>     |RelaxNGDialect.java| and |RelaxNGFilter.java|.

Sorry.  Had copied them from original in a previous SVN checkout.

>  2. |RelaxNGDialect.java| has unused imports

Fixed.

>  3. None of the following in |RelaxNGDialect| are actually used. Please
>     remove them if they are not needed.
>       * |RELAXNG_ROOT_TAG|
>       * |RELAXNG_XMLNS|
>       * |checkIgnoreTags()|

Removed.

>  4. There is commented-out code in |RelaxNGDialect| and |RelaxNGFilter|.
>     If it is not needed please remove it.

Removed.

>  5. Please update the copyright notice on |RelaxNGFilterTest.java|.

Done.

amake:
This is now committed to trunk. Thanks for your contribution.

PR #17, opened 2016-12-21
XML 'DOCTYPE' with systemId but not publicId requires 'SYSTEM'

tgraham-antenna:
See https://www.w3.org/TR/xml/#NT-ExternalID

### DIFF
--- a/src/org/omegat/filters3/xml/DTD.java
+++ b/src/org/omegat/filters3/xml/DTD.java
@@ -83,6 +83,10 @@ public class DTD extends XMLPseudoTag {
             res.append("\"" + publicId + "\"");
         }
         if (systemId!=null) {
+	    if (publicId==null) {
+		res.append(" ");
+		res.append("SYSTEM");
+	    }
             res.append(" ");
             res.append("\"" + systemId + "\"");
         }


amake:
From the reference link you provided it looks like there should only be *either* `PUBLIC <value> <value>` *or* `SYSTEM <value>`. Why is adding `SYSTEM` conditional in your patch? Also it looks like it's possible to write out both `PUBLIC` and `SYSTEM` content; we should really be ensuring that only one is written. In fact we shouldn't even allow creation of a `DTD` object with both.

tgraham-antenna:
Adding `SYSTEM` is conditional to avoid `PUBLIC <value> SYSTEM <value>`.

In XML terms, the DOCTYPE declaration is a special form of an 'external entity'.  From https://www.w3.org/TR/xml/#dt-extent, an XML external entity has to have a system identifier and may also have a public identifier. (You can ignore the `NDATA` notation identifier as not applying to DOCTYPE declarations and as not being used in XML at all these days.  It's a carryover from SGML that people thought would be used, even though it was barely used in SGML.)

SGML allowed/allows `PUBLIC <value>`, i.e., with no system identifier.  There is a very remote chance that someone may use OmegaT on SGML documents, but the real reason why I didn't fiddle with the `if (systemId!=null)` is that I wanted to minimise my changes.

amake:
> an XML external entity has to have a system identifier and may also have a public identifier

Where do you see that it can also have a public identifier? The grammar at your link shows only that it can be one or the other:
```
doctypedecl ::= '<!DOCTYPE' S Name (S ExternalID)? S? ('[' intSubset ']' S?)? '>'
ExternalID ::= 'SYSTEM' S SystemLiteral | 'PUBLIC' S PubidLiteral S SystemLiteral
```
If my reading of the spec is not wrong, perhaps files "in the wild" have both?
> I wanted to minimise my changes.
I'm more concerned with making *correct* changes than minimizing them.

tgraham-antenna:
Maybe it's a problem with my terminology.  The 'system identifier' is the `SystemLiteral` in the production, and the 'public identifier' is the `PubidLiteral`.

From https://www.w3.org/TR/xml/#dt-extent, "The SystemLiteral is called the entity's system identifier." and " In addition to a system identifier, an external identifier may include a public identifier."

The relevant examples from that section (which use `ENTITY` instead of `DOCTYPE` because they're examples of external entities) are:

````
<!ENTITY open-hatch
         SYSTEM "http://www.textuality.com/boilerplate/OpenHatch.xml">
<!ENTITY open-hatch
         PUBLIC "-//Textuality//TEXT Standard open-hatch boilerplate//EN"
         "http://www.textuality.com/boilerplate/OpenHatch.xml">
````

DocBook documents tend to use both public and system identifiers, which may be why this hasn't been a problem for OmegaT before.

amake:
OK, thanks for clarifying. I was under the (apparently mistaken) assumption that the parser was passing `SystemLiteral` for the OmegaT `systemId` member and `PubidLiteral SystemLiteral` for the `publicId` member, in which case we should only output one of them. But you are saying it's `systemId = SystemLiteral` and `publicId = PubidLiteral`, which makes a lot more sense. In that case your patch is correct.

amake:
This is now in trunk, r9417.

PR #18, opened 2017-01-15
Update Bundle_zh_TW.properties

hiunnhue:


suzume:
I've applied your changes to svn r9555. I'm closing the issue here. Thank you for your work.

PR #20, opened 2017-07-28
Fix remote LanguageTool version check

chelobaka:
LanguageTool 3.8 with OmegaT master returns its API version as integer.


amake:
Thanks for the report.

PR #21, opened 2017-09-30
Suggestion

OrionRandD:
I am not sure if this is the place to ask a new feature for OmegaT. But, I really would like the devs to add the possibility to add Vim keybindings to this tool, since it would increase the translation workflow a lot... Thx anyway...

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -17,7 +17,7 @@ applicationName = 'OmegaT'
 mainClassName = 'org.omegat.Main'
 
 ext {
-    javaVersion = '1.8'
+    javaVersion = '1.9'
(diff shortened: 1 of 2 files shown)


amake:
Please make feature requests here:
https://sourceforge.net/p/omegat/feature-requests/new/

mich2016:
I'd also like this feature.  Vim or Vi.  Use it most of the time. 

Word count on the frequency of the text would be great too.  Here's how I do it with a shell script:

$cat /Users/mich/bin/speech.txt | tr -cs '[A-Z][a-z]' '\012' | sort | uniq -c | sort -nr |more

Mich

amake:
This is not the place to discuss anything other than pull requests.

PR #22, opened 2018-02-04
Use variables in regex replacements

amake:
This PR is for code-review purposes.


src/org/omegat/core/search/Searcher.java
```diff
@@ -628,7 +630,24 @@ public boolean searchString(String origText, boolean collapseResults) {
...
             while (true) {
                 int start = matcher.start();
-                foundMatches.add(new SearchMatch(start, matcher.end()));
+                if (m_searchExpression.mode == SearchMode.REPLACE && m_searchExpression.searchExpressionType == SearchExpression.SearchExpressionType.REGEXP) {
+                    String repl = m_searchExpression.replacement;
+                    Matcher replaceMatcher = Pattern.compile("(?<!\\\\)\\$(\\d+)").matcher(repl);
+                    while (replaceMatcher.find()) {
+                        int varId = Integer.parseInt(replaceMatcher.group(1));
+                        repl = repl.substring(0, replaceMatcher.start())
+                            + matcher.group(varId) // yes, from source matcher!
```
amake: I'm sure you did not consider this to be within the scope of your patch, but for completeness: when the user specifies a group that doesn't exist (e.g. search: `.` replace: `$1`) the error message has some problems:

screen shot 2018-02-04 at 10 58 33

1. The message is confusing: Technically, yes, we are searching a directory tree, but this is a detail the user doesn't really know about.
2. The part about groups may be understandable to a power user using regex replacements, but it could be improved
3. It appears parented to the main window, when it would be better on the search window
4. It appears even if we merely searched and did not initiate a replacement

At the very least we should catch the OOB exception and provide a better error message.

src/org/omegat/core/search/Searcher.java
```diff
@@ -662,6 +681,55 @@ public boolean searchString(String origText, boolean collapseResults) {
         return true;
     }
 
+    // Interpret case changes in regex. Must be called _AFTER_ variables replacement!
+    private static String replaceCase(String txt) {
```
amake: Since this doesn't have anything specifically to do with search, I'd like to see it moved to a util class like `StringUtil` and have some unit tests added.

src/org/omegat/core/search/ReplaceMatch.java
```diff
@@ -0,0 +1,46 @@
...
+
+package org.omegat.core.search;
+
+/**
+ * Extends search match for regular expression replacement:
+ * we must store the text for replacement because it will differ from one target string to another
+ *
+ * @author Thomas Cordonnier (t_cordonnier@yahoo.fr)
+ */
+public class ReplaceMatch extends SearchMatch {
```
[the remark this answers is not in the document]
t-cordonnier: In your other comments you don't say why. I wrote like this because we are in an object-oriented language, where it seems logical to consider that a replace match is a special case of search match which contains also text to replace with.
I prefer that rather than big containers like search expression, where you have to look for some fields to decide whenever another one should be filled or not (for example rootDir should be filled only if mode == search)
In case you wrote something about that in the coding guidelines, can you give a link?

An alternative could be the following:
class SearchMatch {
            ...
           class ReplaceMatch {
                    public String replacement;
                    public SearchMatch getSearch() { return SearchMatch.this; }
           }
}

src/org/omegat/core/search/Searcher.java
```diff
@@ -662,6 +681,55 @@ public boolean searchString(String origText, boolean collapseResults) {
...
+            int idx = txt.indexOf("\\");
+            if (idx == -1) {
+                return txt;
+            } else {
+                return txt.substring(0, idx) + replaceCase(txt.substring(idx));
+            }
+        }
+        // Double symbols are longer, so they must be treated first
+        if (txt.startsWith("\\u\\L")) {
+            return txt.substring(4, 5).toUpperCase() + replaceCase("\\L" + txt.substring(5));
```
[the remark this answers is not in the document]
t-cordonnier: Correct (I was alerted by Didier just after I sent the patch)
One point which may cause problems, however: 
logically, the target language should be used;
but in OmegaT, the replace screen can be used to fill untranslated segments, in which case it "replaces" based on source language. In other terms, there is a use case where source language should be used instead of target one.
What do you think should be done?

src/org/omegat/gui/editor/filter/ReplaceFilter.java
```diff
@@ -107,13 +110,23 @@ public void replaceAll() {
...
             List<SearchMatch> found = getReplacementsForEntry(trans);
             if (found != null) {
                 int off = 0;
                 StringBuilder o = new StringBuilder(trans);
                 for (SearchMatch m : found) {
-                    o.replace(m.getStart() + off, m.getEnd() + off, replacement);
-                    off += replacement.length() - m.getLength();
+                    String repl;
+                    try {
+                        repl = ((ReplaceMatch) m).getReplacement();
```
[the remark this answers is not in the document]
t-cordonnier: Not only to impact less code. Another reason is to avoid to repeat an identical string in all replace match instances: the only case where it can differ is when you use regex replacements.
That is also one other reason for using a special ReplaceMatch class: this class is susceptible to have thousands of instances for a single search, while only a few of them really need to have a string distinct from the original one.

src/org/omegat/core/search/Searcher.java
t-cordonnier: Yes, I simply forgot to check that case. No problem to correct this.

src/org/omegat/core/search/ReplaceMatch.java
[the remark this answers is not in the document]
amake: Yes, I meant because of the cast. Sorry my comments were so scattered; I'll gather my thoughts on this below, not mixed in with the code.

src/org/omegat/core/search/Searcher.java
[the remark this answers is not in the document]
amake: Unfortunately it seems even worse than that: `replaceCase` is used in `searchString`, and we call this against lots of different text that could be in any language, such as notes and external files.

Ideally we can handle this by supplying the language/locale as a parameter to `searchString`. I think it's fair to assume that notes and external files should be in the source language.

src/org/omegat/core/search/Searcher.java
```diff
@@ -628,7 +630,24 @@ public boolean searchString(String origText, boolean collapseResults) {
...
+                    Matcher replaceMatcher = Pattern.compile("(?<!\\\\)\\$(\\d+)").matcher(repl);
+                    while (replaceMatcher.find()) {
+                        int varId = Integer.parseInt(replaceMatcher.group(1));
+                        repl = repl.substring(0, replaceMatcher.start())
+                            + matcher.group(varId) // yes, from source matcher!
+                            + repl.substring(replaceMatcher.end());
+                        replaceMatcher.reset(repl);
+                    }
+                    if (!repl.equals(m_searchExpression.replacement)) {
+                        foundMatches.add(new ReplaceMatch(start, matcher.end(), replaceCase(repl)));
```
amake: `replaceCase` is only used here, when the replacement string is match-specific (i.e. when there was a replacement). I suppose it doesn't make much sense for a user to want use case transformations on a constant replacement string, but it feels inconsistent to me that the `\u`, etc., operators would behave as literals when there's no backreference.

How is this handled in other tools? If it's standard behavior to become literals in the absence of backreferences then I guess we should do so as well.

src/org/omegat/gui/editor/filter/ReplaceFilter.java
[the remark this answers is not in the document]
amake: Let's move this discussion down to a separate thread.

amake:
Regarding the `SearchResult` / `ReplaceResult` / casting issue:

Casting to me is a strong smell that something is wrong with the overall "shape" of the implementation. My first instinct was simply to combine the result classes, but really we should probably be separating searching and searching+replacing entirely, with core search functionality in a parameterized type (`Searcher<T extends SearchMatch>`) and searching+replacing being a specialization of that (`SearchReplacer extends Searcher<ReplaceMatch>`).

In that case I don't mind having `ReplaceMatch` extend `SearchMatch`. However I think trying to get away with using a `SearchMatch` when the replacement is identical is a premature optimization. By introducing backreference substitution we are now entering a world where, conceptually, a given match's replacement may be unique so the most logical thing is for the match object to carry its replacement around with it. Signalling that the replacement is not unique by means of the type system feels like a conceptual abuse, and (without having benchmarked it) I don't think it buys us very much: it's a single field per object, at worst all pointing to the same string. If there is some low-RAM target environment where search-and-replace performs poorly, I'd like to see some benchmarks before we start optimizing.

amake:
Now that you're a member of the organization, you have write access to this repo. Feel free to push to this branch. When we're finished, I will squash and commit to svn.

src/org/omegat/core/search/Searcher.java
t-cordonnier: In Perl, or in Bash, there is the clear distinction between interpolated strings (in double quotes) and non-interpolated ones (in simple quotes). When a string is interpolated, case operators _are_ applied even if there was finally no variable in it (note that I do not speak about regular expressions here, only about interpolation of variables, $1 or $x for example)
Sorry to make reference to our fork, but in DGT-OmegaT specific replace screen, this is one of the reasons why it makes sense to have a checkbox to activate or not the interpolation: when inactivated, $1 remains $1 and \L remains \L, as is. Only users which are really familiar with regex replacements should activate it.

src/org/omegat/core/search/Searcher.java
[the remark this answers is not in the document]
t-cordonnier: That should be a good reason why to split this big "Searcher" class into specific ones, linked by inheritance: then "searchString" would be overridden and only the version specific to replacements would make calls to replaceCase.

t-cordonnier:
Sorry to open a new thread but I do not see any "reply" in the previous one (while I have in your other comments)

"Casting to me is a strong smell that something is wrong with the overall "shape" of the implementation. "
Probably because it is an exception, right? If I had written
```String repl = m.getReplacement(); if (repl == null) repl = this.replacement; ```
and implemented like this:
```
class SearchMatch {
       ...
       public String getReplacement() { return null; }
}
class ReplaceMatch extends SearchMatch {
       ...
       @Override public String getReplacement() { return this.replacement; }
}
```
would you consider that this is more correct?

File: src/org/omegat/core/search/Searcher.java [hunk 7]
```diff
@@ -628,7 +630,24 @@ public boolean searchString(String origText, boolean collapseResults) {
             }
             while (true) {
                 int start = matcher.start();
-                foundMatches.add(new SearchMatch(start, matcher.end()));
+                if (m_searchExpression.mode == SearchMode.REPLACE && m_searchExpression.searchExpressionType == SearchExpression.SearchExpressionType.REGEXP) {
+                    String repl = m_searchExpression.replacement;
+                    Matcher replaceMatcher = Pattern.compile("(?<!\\\\)\\$(\\d+)").matcher(repl);
```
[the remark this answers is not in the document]
t-cordonnier: Done.

src/org/omegat/core/search/Searcher.java
t-cordonnier: Done.
For the moment I committed in github only small things, in order to check that I had correct write access. More will follow during next days.

src/org/omegat/core/search/SearchExpression.java
```diff
@@ -60,6 +62,7 @@
     public SearchMode mode;
     public String text;
     public String rootDir;
+    public String replacement;  // used for regular expressions
```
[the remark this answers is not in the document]
t-cordonnier: Well sorry this phrase was ambiguous: indeed the variable was always filled but the only case where it was read was when it is copied before the loop. Not relevant anymore.

src/org/omegat/core/search/Searcher.java
```diff
@@ -628,7 +630,24 @@ public boolean searchString(String origText, boolean collapseResults) {
             }
             while (true) {
                 int start = matcher.start();
-                foundMatches.add(new SearchMatch(start, matcher.end()));
+                if (m_searchExpression.mode == SearchMode.REPLACE && m_searchExpression.searchExpressionType == SearchExpression.SearchExpressionType.REGEXP) {
```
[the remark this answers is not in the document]
t-cordonnier: See new code, this is not relevant anymore

src/org/omegat/core/search/Searcher.java
t-cordonnier: OK sorry I understood later why you wrote that. Yes, you are true, \L should always be interpreted when we are in regex mode.
With the new code the rule is: always interpreted in regex mode, never interpreted in "exact search" mode.
However you should consider the opportunity to add a checkbox for optional support of regex variables: only users who really know what they are doing should use it.

src/org/omegat/core/search/Searcher.java
t-cordonnier: OK I could move the error message inside SearchThread and find better text (points 1 & 2)
But now it is modal, and difficult to say whenever it refers to search window or main window

amake:
It looks like your commits disappeared for a bit due to overzealous syncing from SourceForge's git to GitHub. I believe I have restored everything now.

> Sorry to open a new thread but I do not see any "reply" in the previous one (while I have in your other comments)

I think that is the correct thing to do. The "main" thread doesn't group comments together like review threads.

> Probably because it is an exception, right? If I had written
> `String repl = m.getReplacement(); if (repl == null) repl = this.replacement;`
> and implemented like this:

It's actually a couple of things.

1. Yes, using exceptions like that for control flow is a bad smell, but also:
2. The fact that there is "hidden" data in the result type that requires secret knowledge (cast to some other class) to access is a very bad smell
3. Inheritance here is probably wrong. Per _Effective Java_ item 10: "There is no way to extend an instantiable class and add a value component while preserving the equals contract, unless you’re willing to forgo the benefits of object-oriented abstraction." At the moment we aren't overriding `equals` so it's not an immediate concern, but I think it's the wrong direction. If we're going to have separate classes, we should probably use composition: `ReplaceMatch` would have a `SearchMatch` field, which it would return from an `asSearchMatch()` accessor. `Searcher` should be parameterized on the match type (`Searcher<SearchMatch>` or `Searcher<ReplaceMatch>` so casts aren't necessary.

amake:
Here's what's happening with the deletions and closures:

1. Initially I pushed the `topic/aaron/regex-replace` branch to SourceForge git by accident.
2. From there it was synced by our CI to GitHub.
3. Thomas pushed some commits to GitHub.
4. Commits were made to SourceForge svn, which syncs to SourceForge git, which triggered a sync to GitHub. This reset the branch to the last commit present on SourceForge git.
5. I restored Thomas's commits and pushed them to the GitHub branch.
6. In the hopes of preventing another reset, I deleted the branch on SourceForge git.
7. The deletion was synced to GitHub, thus closing the PR.

amake:
From Thomas via email:

> I finally removed the ReplaceMatch and put everything in SearchMatch, because the only way to avoid the cast would have been to split also the Searcher class: even if I see lot of reasons why it should be done (see in the comments!) that would be another topic which should not slow down the review for "variables in regex replacements".

I think this is prudent. I think `Searcher` is full of spaghetti code that needs refactoring, but it would definitely slow us down, and I appreciate that your time is valuable. If you're interested in helping to improve this some other time, we would be very grateful.

src/org/omegat/core/search/SearchMatch.java
```diff
@@ -32,10 +32,12 @@
  */
 public class SearchMatch implements Comparable<SearchMatch> {
     private int start, end;
+    private String replacement;
 
-    public SearchMatch(int start, int end) {
+    public SearchMatch(int start, int end, String replacement) {
```
amake: I think we could have two constructors here, rather than forcing non-replacement instances to specify `null` for `replacement`.

test/src/org/omegat/util/StringUtilTest.java
```diff
@@ -364,4 +366,17 @@ public void testRstrip() {
             // Should fail when stripping null string.
         }
     }
+
+    @Test
+    public void testCaseConversion() {
```
amake: Thank you! This is probably the first time someone has supplied a unit test when I asked for one :)

src/org/omegat/core/search/Searcher.java
```diff
@@ -628,7 +631,24 @@ public boolean searchString(String origText, boolean collapseResults) {
...
+                        String repl = m_searchExpression.replacement;
+                        Matcher replaceMatcher = PatternConsts.REGEX_VARIABLE.matcher(repl);
+                        while (replaceMatcher.find()) {
+                            int varId = Integer.parseInt(replaceMatcher.group(1));
+                            repl = repl.substring(0, replaceMatcher.start())
+                                + matcher.group(varId) // yes, from source matcher!
+                                + repl.substring(replaceMatcher.end());
+                            replaceMatcher.reset(repl);
+                        }
+                        foundMatches.add(new SearchMatch(start, matcher.end(), StringUtil.replaceCase(repl, Core.getProject().getProjectProperties().getTargetLanguage().getLocale())));
```
amake: Please humor my nitpicking here:

In general in the OmegaT codebase there is too much referencing of `Core`, especially for project properties. It's basically a big ball of global variables, and it overly couples things to the OmegaT concept of a project. One far-off goal of mine is to decouple all of this so that we can modularize OmegaT and e.g. offer just an API dependency for plugins, etc.

In the case of the `Searcher`, there's no reason a `Searcher` should have to know or care about what a project is. So instead of accessing the locale through `Core`, I think the `Searcher` should receive the locale at the point of instantiation via its constructor and keep it in a private field.

Granted, `Searcher` already looks to `Core` for a few other things, but I would like to keep those things from growing.

src/org/omegat/core/threads/SearchThread.java
```diff
@@ -82,6 +84,9 @@ public void run() {
                 // bad regexp input
                 // alert user to badness
                 window.displayErrorRB(e, "ST_REGEXP_ERROR");
+            } catch (IndexOutOfBoundsException e) {
+                // replacement with group which does not exist
+                window.displayErrorRB(null, "ST_REGEXP_REPLACEGROUP_ERROR", e.getMessage().substring(e.getMessage().lastIndexOf(" ") + 1));
```
amake: I think this munging of the exception message is a bit dangerous. I imagine you did it this way because there's no other way to retrieve the problematic group from the exception object.

Instead of this, I'd rather detect the issue ahead of time and throw a custom exception before attempting to access the group. E.g. in `Searcher`:

```
int varId = Integer.parseInt(replaceMatcher.group(1));
if (varId > matcher.groupCount()) {
    throw new InvalidRegexGroupException(OStrings.getString("ST_REGEXP_REPLACEGROUP_ERROR", varId));
}

t-cordonnier: One potential inconvenient of this method is that you change the interface of a public method (searchString), because it throws a new exception. Being a public method, are you sure it will not have an impact on plugins?

src/org/omegat/core/search/Searcher.java
t-cordonnier: In the meantime I found that the project is already in the private fields. So, we can replace Core.getProject() by m_project.

src/org/omegat/core/threads/SearchThread.java
amake: If the custom exception extends `RuntimeException` then it is not a checked exception and doesn't affect the method signature.

src/org/omegat/core/search/Searcher.java
amake: Ah yes, that's perfect. I missed that. Thanks.

src/org/omegat/core/threads/SearchThread.java
t-cordonnier: OK fine.
Since this exception will probably later be used in other locations, do you have any particular preference about where to put it (i.e. somewhere in org.omegat.util for example?)

amake: I don't know that we will necessarily use the exception elsewhere. If we do, we would probably be abstracting the entire search/replace mechanism, in which case I think an appropriate location for it would become apparent.

For now I think we can make it a nested static class within `Searcher`.

t-cordonnier: Finally this was not necessary: since IndexOutOfBoundsException is public I could throw my own instance in the searcher, with a customized message.

amake: Yes, that's better than my idea. Thanks.

amake:
Everything looks good. I will squash and commit to svn.

amake:
SourceForge svn is out of service due to a data center move. I will commit when they're back up.

PR #23, opened 2018-05-10
Side by side

x24george:
Hi Aaron,  

Here is the side-by-side modifications I've mentioned.  I appreciate you taking the time to consider these changes.   If there is anything I can do to make these changes a part of a future release please let me know! 

The main changes are the creation of the SideBySideEditorController and SourceOnlySegmentBuilder.  The SourceOnlySegmentBuilder (and it's parent AbstractSegmentBuilder) is a copy of the original SegmentBuilder but modified to only display source elements.  I started to refactor the SegmentBuilder/SourceOnlySegmentBuilder into a common class (AbstractSegmentBuilder) but stopped since I was uncertain about the future of my changes.  I can continue that work plus make some optimizations if there is still a possibility this optional mode can make it to a future release.


amake:
As discussed on the dev mailing list, this PR doesn't meet our criteria for this feature, so I'm going to close it.

PR #24, opened 2018-06-26
Can read SRX format instead of CONF

t-cordonnier:
SRX format support

After discussion:
If CONF file is present and SRX absent, convert it. Next time, OmegaT 4 will use the SRX (while CONF is kept for older versions, but not updated anymore)
Rule is valid for global config directory as well as for project-specific segmentation rules. 

Last patch adds support for cascade attribute. This was not needed in the past, since you only used SRX for default rules. But now, if a user copies as segmentation.srx a file coming from another tool, this is required for full compatibility.

### DIFF
--- a/src/org/omegat/core/segmentation/SRX.java
+++ b/src/org/omegat/core/segmentation/SRX.java
@@ -37,6 +37,7 @@ import java.io.IOException;
 import java.io.Serializable;
 import java.net.URL;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
 
 import javax.xml.bind.JAXBContext;
@@ -242,16 +243,10 @@ public class SRX implements Serializable {
             List<MapRule> newMap = new ArrayList<MapRule>();
             Srx data = (Srx) SRX_JAXB_CONTEXT.createUnmarshaller().unmarshal(rulesUrl);
 
-            for (Languagerule rules : data.getBody().getLanguagerules().getLanguagerule()) {
+            // Correction: in SRX, the same "languagerulename" can be used more than once
+            HashMap<String,List<Rule>> mapping = new HashMap<>();
 
-                String lang = rules.getLanguagerulename();
-                String pattern = DEFAULT_RULES_PATTERN;
-                for (Languagemap lm : data.getBody().getMaprules().getLanguagemap()) {
-                    if (lm.getLanguagerulename().equals(rules.getLanguagerulename())) {
-                        pattern = lm.getLanguagepattern();
-                        break;
-                    }
-                }
+            for (Languagerule rules : data.getBody().getLanguagerules().getLanguagerule()) {
                 List<Rule> rulesList = new ArrayList<Rule>(rules.getRule().size());
                 for (gen.core.segmentation.Rule r : rules.getRule()) {
                     boolean isBreak = "yes".equalsIgnoreCase(r.getBreak());
@@ -259,8 +254,13 @@ public class SRX implements Serializable {
                             .getContent()));
                 }
 
-                newMap.add(new MapRule(lang, pattern, rulesList));
+                mapping.put(rules.getLanguagerulename(), rulesList);
             }
+
+            for (Languagemap lm : data.getBody().getMaprules().getLanguagemap()) {
+                newMap.add(new MapRule(lm.getLanguagerulename(), lm.getLanguagepattern(), mapping.get(lm.getLanguagerulename())));
+            }
+
             Log.log("using segmentation rules from " + rulesUrl);
             // set rules only if no errors
             SRX res = new SRX();
@@ -269,7 +269,7 @@ public class SRX implements Serializable {
             res.setSegmentSubflows((data.getHeader() != null) && ("yes".equals(data.getHeader().getSegmentsubflows())));	// not really used
             return res;
         } catch (Exception ex) {
-            ex.printStackTrace();
+            Log.log(ex);
             return null;
         }
     }
--- a/src/org/omegat/core/segmentation/SRX.java
+++ b/src/org/omegat/core/segmentation/SRX.java
@@ -139,10 +139,14 @@ public class SRX implements Serializable {
                 gen.core.segmentation.Rule jaxbRule = factory.createRule();
                 lr.getRule().add(jaxbRule);
                 jaxbRule.setBreak(rule.isBreakRule() ? "yes" : "no");
-                jaxbRule.setBeforebreak(factory.createBeforebreak());
-                jaxbRule.getBeforebreak().setContent(rule.getBeforebreak());
-                jaxbRule.setAfterbreak(factory.createAfterbreak());
-                jaxbRule.getAfterbreak().setContent(rule.getAfterbreak());
+                if (rule.getBeforebreak() != null) {
+                    jaxbRule.setBeforebreak(factory.createBeforebreak());
+                    jaxbRule.getBeforebreak().setContent(rule.getBeforebreak());                
+                }
+                if (rule.getAfterbreak() != null) {
+                    jaxbRule.setAfterbreak(factory.createAfterbreak());                
+                    jaxbRule.getAfterbreak().setContent(rule.getAfterbreak());
+                }
             }
         }
 
--- a/src/org/omegat/core/segmentation/SRX.java
+++ b/src/org/omegat/core/segmentation/SRX.java
@@ -106,9 +106,9 @@ public class SRX implements Serializable {
     /**
      * Saves segmentation rules into specified directory.
      * @param srx OmegaT object to be written; if null, means that we want to delete the file
-     * @param outDir where to put the file. The file name is forced to {@link #SRX_SENTSEG}
+     * @param outDir where to put the file. The file name is forced to {@link #SRX_SENTSEG} and will be in standard SRX format.
      */
-    public static void saveTo(SRX srx, File outDir) throws IOException {
+    public static void saveToSrx(SRX srx, File outDir) throws IOException {
         File outFile = new File (outDir, SRX_SENTSEG);
 
         if (srx == null) {
@@ -117,10 +117,6 @@ public class SRX implements Serializable {
             return;
         }
 
-        saveToSrx(srx, outFile);
-    }
-
-    public static void saveToSrx(SRX srx, File outFile) throws IOException {
         ObjectFactory factory = new ObjectFactory();
         Srx jaxbObject = factory.createSrx();
         jaxbObject.setVersion("2.0");
@@ -182,7 +178,7 @@ public class SRX implements Serializable {
         if (inFile.exists()) {
             SRX srx = loadConfFile(inFile);
             try {
-                saveToSrx(srx, new File(configDir, SRX_SENTSEG));
+                saveToSrx(srx, configDir);
             } catch (Exception o3) {
                 Log.log(o3); // detail why conversion failed, but continue
             }
--- a/src/org/omegat/core/segmentation/SRX.java
+++ b/src/org/omegat/core/segmentation/SRX.java
@@ -29,7 +29,6 @@ package org.omegat.core.segmentation;
 
 import java.beans.ExceptionListener;
 import java.beans.XMLDecoder;
-import java.beans.XMLEncoder;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.FileNotFoundException;
@@ -111,14 +110,6 @@ public class SRX implements Serializable {
      */
     public static void saveTo(SRX srx, File outDir) throws IOException {
         File outFile = new File (outDir, SRX_SENTSEG);
-        if (! outFile.exists()) {
-            // Unless SRX file exists, check for CONF file. If present, overwrite it. Else, fallback to SRX
-            outFile = new File (outDir, CONF_SENTSEG);
-            if (! outFile.exists()) {
-                outFile = new File (outDir, SRX_SENTSEG);
-            }
-        }
-        // else: SRX file exists, overwrite or delete it
 
         if (srx == null) {
             outFile.delete();
@@ -126,24 +117,7 @@ public class SRX implements Serializable {
             return;
         }
 
-        if (outFile.getName().endsWith(".conf")) {
-            saveToConf(srx, outFile);
-        } else if (outFile.getName().endsWith(".srx")) {
-            saveToSrx (srx, outFile);
-        }
-    }
-
-    public static void saveToConf(SRX srx, File outFile) throws IOException {        
-        try {
-            srx.setVersion(CURRENT_VERSION);
-            XMLEncoder xmlenc = new XMLEncoder(new FileOutputStream(outFile));
-            xmlenc.writeObject(srx);
-            xmlenc.close();
-        } catch (IOException ioe) {
-            Log.logErrorRB("CORE_SRX_ERROR_SAVING_SEGMENTATION_CONFIG");
-            Log.log(ioe);
-            throw ioe;
-        }
+        saveToSrx(srx, outFile);
     }
 
     public static void saveToSrx(SRX srx, File outFile) throws IOException {
@@ -204,7 +178,13 @@ public class SRX implements Serializable {
         // If file was not present or not readable
         inFile = new File(configDir, CONF_SENTSEG);
         if (inFile.exists()) {
-            return loadConfFile(inFile);
+            SRX srx = loadConfFile(inFile);
+            try {
+                saveToSrx(srx, new File(configDir, SRX_SENTSEG));
+            } catch (Exception o3) {
+                Log.log(o3); // detail why conversion failed, but continue
+            }
+            return srx;
         }
 
         // If none of the files (conf and srx) are present,
--- a/src/org/omegat/core/segmentation/SRX.java
+++ b/src/org/omegat/core/segmentation/SRX.java
@@ -5,6 +5,7 @@
 
  Copyright (C) 2000-2006 Keith Godfrey and Maxym Mykhalchuk
                2008 Alex Buloichik
+               2018 Thomas Cordonnier
                Home page: http://www.omegat.org/
                Support center: https://omegat.org/support
 
@@ -41,6 +42,7 @@ import java.util.List;
 
 import javax.xml.bind.JAXBContext;
 import javax.xml.bind.JAXBException;
+import javax.xml.bind.Marshaller;
 
 import org.omegat.util.Language;
 import org.omegat.util.Log;
@@ -48,6 +50,7 @@ import org.omegat.util.OStrings;
 
 import gen.core.segmentation.Languagemap;
 import gen.core.segmentation.Languagerule;
+import gen.core.segmentation.ObjectFactory;
 import gen.core.segmentation.Srx;
 
 /**
@@ -55,12 +58,14 @@ import gen.core.segmentation.Srx;
  * It loads and saves its data from/to SRX file.
  *
  * @author Maxym Mykhalchuk
+ * @author Thomas Cordonnier
  */
 public class SRX implements Serializable {
 
     private static final long serialVersionUID = 2182125877925944613L;
 
     public static final String CONF_SENTSEG = "segmentation.conf";
+    public static final String SRX_SENTSEG = "segmentation.srx";
 
     /** Context for JAXB rules processing. */
     protected static final JAXBContext SRX_JAXB_CONTEXT;
@@ -83,16 +88,6 @@ public class SRX implements Serializable {
     }
 
     /**
-     * Initializes SRX rules to defaults.
-     */
-    private void init() {
-        this.mappingRules = new ArrayList<MapRule>();
-        this.includeEndingTags = true;
-        this.segmentSubflows = true;
-        initDefaults();
-    }
-
-    /**
      * Creates an empty SRX, without any rules.
      * <p>
      * Please do not call directly unless you know what you are doing.
@@ -110,13 +105,35 @@ public class SRX implements Serializable {
     }
 
     /**
-     * Saves segmentation rules into specified file.
+     * Saves segmentation rules into specified directory.
+     * @param srx OmegaT object to be written; if null, means that we want to delete the file
+     * @param outDir where to put the file. The file name is forced to {@link #SRX_SENTSEG}
      */
-    public static void saveTo(SRX srx, File outFile) throws IOException {
+    public static void saveTo(SRX srx, File outDir) throws IOException {
+        File outFile = new File (outDir, SRX_SENTSEG);
+        if (! outFile.exists()) {
+            // Unless SRX file exists, check for CONF file. If present, overwrite it. Else, fallback to SRX
+            outFile = new File (outDir, CONF_SENTSEG);
+            if (! outFile.exists()) {
+                outFile = new File (outDir, SRX_SENTSEG);
+            }
+        }
+        // else: SRX file exists, overwrite or delete it
+
         if (srx == null) {
             outFile.delete();
+            new File(outDir, CONF_SENTSEG).delete();
             return;
         }
+
+        if (outFile.getName().endsWith(".conf")) {
+            saveToConf(srx, outFile);
+        } else if (outFile.getName().endsWith(".srx")) {
+            saveToSrx (srx, outFile);
+        }
+    }
+
+    public static void saveToConf(SRX srx, File outFile) throws IOException {        
         try {
             srx.setVersion(CURRENT_VERSION);
             XMLEncoder xmlenc = new XMLEncoder(new FileOutputStream(outFile));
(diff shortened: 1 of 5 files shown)


amake:
I rebased your branch onto the latest master to re-trigger the Jenkins check (your parent on master had an error in it, plus the Jenkins job was broken). You'll want to reset your local branch to the remote one. Sorry for the inconvenience.

src/org/omegat/core/segmentation/SRX.java
```diff
@@ -110,40 +105,105 @@ public SRX copy() {
...
+        }
+        
+        // Note: as in previous version (4.1) return a new SRX with defaults
+        return SRX.getDefault();
+    }
+    
     /**
      * Loads segmentation rules from an XML file. If there's an error loading a
-     * file, it calls <code>initDefaults</code>.
+     * file, it calls <code>getDefaultFromJar</code>.
```
amake: There is no `getDefaultFromJar`. Did you mean `getDefault`?

src/org/omegat/core/segmentation/SRX.java
```diff
@@ -157,32 +217,65 @@ public static SRX loadSRX(File configFile) {
                     sb.append("\n");
                 }
                 Log.logErrorRB("CORE_SRX_EXC_LOADING_SEG_RULES", sb.toString());
-                res = new SRX();
-                res.initDefaults();
-                return res;
+                return SRX.getDefault();
```
amake: The result of `SRX.getDefault()` and `res = new SRX(); res.initDefaults()` is not exactly the same: the former sets `includeEndingTags` and `segmentSubflows` to true, while the latter does not. Is it safe to replace the former with the latter in each situation?

src/org/omegat/core/segmentation/SRX.java
```diff
@@ -110,40 +105,105 @@ public SRX copy() {
...
         }
+        
+        saveToSrx (srx, outFile);
+    }
+    
+    public static void saveToSrx(SRX srx, File outFile) throws IOException {        
+        ObjectFactory factory = new ObjectFactory();
+        Srx jaxbObject = factory.createSrx();
+        jaxbObject.setVersion("2.0");
+        jaxbObject.setHeader (factory.createHeader());
```
amake: Please clean up the spaces before the argument lists in this line and the following ones.

src/org/omegat/core/segmentation/SRX.java
```diff
@@ -379,11 +438,38 @@ public static SRX getDefault() {
             MapRule maprule = getMappingRules().get(i);
             if (maprule.getCompiledPattern().matcher(srclang.getLanguage()).matches()) {
                 rules.addAll(maprule.getRules());
+                if (! this.cascade) {
```
amake: `!` should not have trailing whitespace.

src/org/omegat/util/Preferences.java
```diff
@@ -670,9 +670,9 @@ public static void setSRX(SRX newSrx) {
         SRX oldValue = srx;
         srx = newSrx;
 
-        File srxFile = new File(StaticUtils.getConfigDir() + SRX.CONF_SENTSEG);
+        File srxDir = new File(StaticUtils.getConfigDir());
         try {
-            SRX.saveTo(srx, srxFile);
+            SRX.saveTo(srx, srxDir);	// defaults to SRX format, but if conf file exists, overwrite it
```
amake: I don't think this comment is accurate w.r.t. overwriting the conf.

src/org/omegat/core/segmentation/SRX.java
```diff
@@ -157,32 +217,65 @@ public static SRX loadSRX(File configFile) {
...
+                    rulesList.add(new Rule(isBreak, r.getBeforebreak().getContent(), r.getAfterbreak()
+                            .getContent()));
+                }
+
+                newMap.add(new MapRule(lang, pattern, rulesList));
+            }
+            Log.log("using segmentation rules from " + rulesUrl);
+            // set rules only if no errors
+            SRX res = new SRX(); res.setMappingRules(newMap);
+            res.setCascade(! ((data.getHeader() != null) && ("no".equals(data.getHeader().getCascade())))); 	// in OmegaT, defaults to true
```
amake: Should the string comparisons be case-insensitive?

src/org/omegat/core/segmentation/SRX.java
```diff
@@ -379,11 +438,38 @@ public static SRX getDefault() {
...
+     * Holds value of property cascade: true, unless we read an SRX where it was set to false.
+     */
+    private boolean cascade = true;
+
+    /**
+     * Getter for property cascade.
+     * 
+     * @return Value of property cascade.
+     */
+    public boolean isSCascade() {
```
amake: I believe the `S` is a typo.

src/org/omegat/core/segmentation/SRX.java
amake: Is this sufficient to implement support for cascading? If so, why have a getter for the `cascade` property? Also I'd really like to see the expected behavior nailed down in some unit tests.

src/org/omegat/util/Preferences.java
```diff
@@ -762,8 +762,8 @@ public static synchronized void initSegmentation() {
         }
         didInitSegmentation = true;
 
-        File srxFile = new File(StaticUtils.getConfigDir(), SRX.CONF_SENTSEG);
-        SRX s = SRX.loadSRX(srxFile);
+        File srxDir = new File(StaticUtils.getConfigDir());
+        SRX s = SRX.loadSRX(srxDir); // may read SRX or CONF
```
amake: I personally would err on the side of *not* documenting the behavior of `loadSRX` outside of the method itself, but if you feel like it's surprising and needs to be mentioned then that's OK.

src/org/omegat/core/segmentation/SRX.java
t-cordonnier: Yes, the method getDefaultFromJar existed in the very first patch, and was deleted when I "merged" the behaviour of initDefaults() and getDefault()

src/org/omegat/core/segmentation/SRX.java
t-cordonnier: Actually you don't implement support for segment subflows nor includeEndingTags, so it has no consequence. But if one day you really implement segment subflows, you will have to ask yourself why the default constructor does not set default values as well.

src/org/omegat/core/segmentation/SRX.java
t-cordonnier: > Is this sufficient to implement support for cascading?

If we look to the definition in SRX specification : " If the cascade attribute is set to "no" then aggregation of the <rule> elements will terminate. ". That is what I do here.

>  If so, why have a getter for the cascade property? 

The good question is : why do we have getters in this class, after all? Why do we have getters for segment subflows, which are not supported?
Don't forget one thing: I did not write this class, I just modified it. I added this getter because other parameters, even if not really used, did have one, so I tried to remain coherent.

src/org/omegat/core/segmentation/SRX.java
t-cordonnier: Corrected

t-cordonnier:
Sorry for the delay, I was notified only today about your comments.
For the moment I corrected only things about style. The rest I will comment tomorrow.

src/org/omegat/util/Preferences.java
t-cordonnier: I would say that the problem is the name of the method.
Previous version loaded only conf format, this was not in the documentation but since the previous line explicitly mentioned a CONF file, it could aspect clear enough.
Now that the previous line only mentions a dir, either we document it here, or we may search a better name for the method, what do you think?

src/org/omegat/core/segmentation/SRX.java
t-cordonnier: Strictly speaking, the SRX specification does not say that values are case insensitive (if I good remember, it also contains a sample showing the contrary). I know that in the case of maprules you decided to accept them as case-insensitive, this can sound logical if we consider that ISO-639 and BCP 47 are case-insensitive, but be conscient that SRX specification does not require case-insensitivity here, meaning that you can break compatibility with SRX files coming from other tools.
Please decide.

src/org/omegat/util/Preferences.java
t-cordonnier: True, this comment was coherent with the first version of my patches, where I tried to find a way to maintain both formats. But during our discussions it became clear that you want the old format to disappear, so I removed any possibility to save to old conf format. 
Now it saves always and only to SRX.

src/org/omegat/core/segmentation/SRX.java
```diff
@@ -110,40 +105,105 @@ public SRX copy() {
     }
 
     /**
-     * Saves segmentation rules into specified file.
+     * Saves segmentation rules into specified directory.
+     * Possibilities:
+     * -- if file does not exist : SRX format
+     * -- if conf file or srx file exists: owerwrite it
```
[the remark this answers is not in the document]
t-cordonnier: Point 1: see the answer I posted in Preferences.java --> yes, the comment is obsolete, now it can only save to SRX.

Point 3: as the lines saying "if (srx == null) outFile.delete()" are not written by me, I would like that we check together if the situation can really happen, or if it is just an old piece of code which is not valid anymore. Then we can check whenever we should comment this case or not.

src/org/omegat/core/segmentation/SRX.java
```diff
@@ -110,40 +105,105 @@ public SRX copy() {
...
      */
-    public static void saveTo(SRX srx, File outFile) throws IOException {
+    public static void saveTo(SRX srx, File outDir) throws IOException {
+        File outFile = new File (outDir, SRX_SENTSEG);       
+        
         if (srx == null) {
```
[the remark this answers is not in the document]
t-cordonnier: Another possibility is to remove the distinction between saveTo and saveToSrx : this was useful when two formats were supported, but now that I removed saveToConf, we could have only one method.

src/org/omegat/core/segmentation/SRX.java
```diff
@@ -110,40 +105,105 @@ public SRX copy() {
...
-    public static SRX loadSRX(File configFile) {
-        if (!configFile.exists()) {
-            return null;
-        }
+    private static SRX loadConfFile (File configFile) {
         SRX res;
         try {
-
             MyExceptionListener myel = new MyExceptionListener();
             XMLDecoder xmldec = new XMLDecoder(new FileInputStream(configFile), null, myel);
```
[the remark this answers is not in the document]
t-cordonnier: This part was not written by me (Git blame confirmed it to me). I can do the change, but this should not be a reason why to reject the patches.

src/org/omegat/core/segmentation/SRX.java
```diff
@@ -110,40 +105,105 @@ public SRX copy() {
...
+        
         try {
-            srx.setVersion(CURRENT_VERSION);
-            XMLEncoder xmlenc = new XMLEncoder(new FileOutputStream(outFile));
-            xmlenc.writeObject(srx);
-            xmlenc.close();
-        } catch (IOException ioe) {
+            Marshaller m = SRX_JAXB_CONTEXT.createMarshaller();
+            m.setProperty(Marshaller.JAXB_FORMATTED_OUTPUT, Boolean.TRUE);
+            m.marshal(jaxbObject, new FileOutputStream(outFile));
```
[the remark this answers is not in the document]
t-cordonnier: OK no problem (I just did it like that to stay in coherence with other parts of the code not written by me - see comment about FileInputStream some lines after)

src/org/omegat/core/segmentation/SRX.java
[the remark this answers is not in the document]
t-cordonnier: Finally did the change but with a small difference: since XMLDecoder implements AutoClosable and its method close(), according to the doc, "closes the input stream associated with this stream", it is better to use try with XMLDecoder.

src/org/omegat/core/segmentation/SRX.java
```diff
@@ -110,40 +105,105 @@ public SRX copy() {
...
+     **/
+    public static SRX loadSRX(File configDir) {
+        File inFile = null;
+        try { 
+            inFile = new File (configDir, SRX_SENTSEG); 
+            if (inFile.exists()) {
+                return loadSrxFile(inFile.toURI().toURL()); 
+            }
+        } catch (Exception o2) {
+
```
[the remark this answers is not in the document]
t-cordonnier: I agree, but before to do the change, small question: 
when SRX file is absent, we try to find CONF file instead, that's logical;
but if SRX is present but corrupted, should we fallback to CONF file, or to default? 
In case the file is present but corrupted, are you sure it is enough to put the info in the log? (or should we send a message through the GUI?)

src/org/omegat/core/segmentation/SRX.java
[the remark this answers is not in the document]
t-cordonnier: Corrected

src/org/omegat/core/segmentation/SRX.java
t-cordonnier: > Also I'd really like to see the expected behavior nailed down in some unit tests.

That's not so easy.
Writing a unit test will only prove that the segmenter works as I wrote it. But a correct test would be to have a normative sample, with expected behaviour, and check that in both cases (with cascade = true then with cascade = false) the segmenter gives the correct result.
Do you have such an example?

src/org/omegat/core/segmentation/SRX.java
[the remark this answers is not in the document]
amake: Yes, we might as well combine the two.

src/org/omegat/core/segmentation/SRX.java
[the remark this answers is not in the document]
amake: > In case the file is present but corrupted, are you sure it is enough to put the info in the log? (or should we send a message through the GUI?)

You're right, this is probably important enough to have a GUI message. But I definitely think we should not be reaching out to the GUI from this class: I have long wanted to modularize OmegaT, meaning splitting the GUI and internals into separate modules. The internals should not know about the GUI; this is definitely an internal class.

Instead the `loadSRX` method should throw an exception if the file is corrupted (or if it's missing, for that matter) and the caller should retry loading the defaults.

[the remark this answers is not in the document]
amake: > but if SRX is present but corrupted, should we fallback to CONF file, or to default?

To clarify on this point, I think we should not fall back to CONF; the road from CONF to SRX should be one-way only.

src/org/omegat/core/segmentation/SRX.java
amake: It appears you're right; the members are not actually used for anything substantial. Your change is fine, then.

src/org/omegat/core/segmentation/SRX.java
amake: My instinct was that there isn't much danger in being generous in interpreting booleans, but I am satisfied by the logic put forth [here](https://stackoverflow.com/a/10511947/448068) that one ought to follow the spec, which is the XSD, and [the XSD says](https://www.gala-global.org/sites/default/files/uploads/pdfs/srx-20.xsd) `cascade` is an enumerated type with values `yes` and `no`. So the current code is fine.

src/org/omegat/util/Preferences.java
amake: I do think the method could be renamed something like `loadSRXFromDir`. But I don't think it's relevant to the Preferences class whether the result came from a CONF or an SRX file.

src/org/omegat/core/segmentation/SRX.java
amake: > The good question is : why do we have getters in this class, after all? Why do we have getters for segment subflows, which are not supported?
> Don't forget one thing: I did not write this class, I just modified it. I added this getter because other parameters, even if not really used, did have one, so I tried to remain coherent.

I haven't forgotten that you didn't write the class, and I am not blaming you for code written by those who came before you. I simply wanted to understand why you wrote it the way you did; it appears the answer is "to remain coherent". That's a perfectly reasonable answer.

Now that I understand your reason, we can think about whether or not coherency is a good thing in this instance, given that the current code has some obvious problems. I would be OK with you not adding this new getter, but I don't feel that strongly about it.

amake: >  Writing a unit test will only prove that the segmenter works as I wrote it.

This, in and of itself, has at least some value.

> But a correct test would be to have a normative sample, with expected behaviour, and check that in both cases (with cascade = true then with cascade = false) the segmenter gives the correct result.
> Do you have such an example?

No, but as the author of this feature I figured you would.

amake:
You certainly don't need to apologize for any delays when I am taking forever to get back to you myself.

I have learned a valuable lesson: I shouldn't have left so many comments about formatting. Since you fixed the formatting most of our conversation is buried in "outdated" collapsed chunks. Next time I will mention formatting just once.

It seems like you may have missed a couple of my comments. I am happy to take care of them myself, but I can't push to your branch. Can you check the "Allow edits from maintainers" checkbox on the right?

t-cordonnier:
> Can you check the "Allow edits from maintainers" checkbox on the right?

Actually it seems to be ckecked. But is it related to commits or comments on Github?

src/org/omegat/util/Preferences.java
t-cordonnier: Better loadSegmentationFromDir, no? The idea is to avoid to use the name "SRX" for a method which _may_ load from another format.

amake:
> Actually it seems to be ckecked.

Sorry! I must have done something wrong on my end, then. I'll take another look.

> But is it related to commits or comments on Github?

The "learn more" link says:

> If checked, users with write access to *<branch>* can add new commits to your peco branch. You can always change this setting later.

src/org/omegat/util/Preferences.java
amake: The class `SRX` predates any support for the actual SRX XML serialization format, so I think the best way to understand it is as a synonym for "segmentation".

Since `loadSRX` is a method on the class `SRX` and it returns an instance of `SRX`, I think having "SRX" in the method name is fine. The caller need not know or care about what the underlying serialization format was. However it's also a bit redundant; a good compromise would be to simply name it `loadFromDir` since the call will be `SRX.loadFromDir()` and thus obvious what we are loading.

I think the `SRX` class should probably be renamed to `Segmentation` or `SegmentationRules` or something like that.

amake:
> Sorry! I must have done something wrong on my end, then. I'll take another look.

This was my own mistake. I managed to push the commits.

src/org/omegat/core/segmentation/SRX.java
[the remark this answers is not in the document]
t-cordonnier: OK finally I kept only saveToSrx, so that the output format appears in the method name.

src/org/omegat/core/segmentation/SRX.java
```diff
@@ -110,40 +105,105 @@ public SRX copy() {
...
+        
         if (srx == null) {
             outFile.delete();
             return;
         }
+        
+        saveToSrx (srx, outFile);
+    }
+    
+    public static void saveToSrx(SRX srx, File outFile) throws IOException {
```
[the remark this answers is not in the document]
t-cordonnier: See previous answer

src/org/omegat/util/Preferences.java
t-cordonnier: >  The class SRX predates any support for the actual SRX XML serialization format, so I think the best way to understand it is as a synonym for "segmentation".
> I think the SRX class should probably be renamed to Segmentation or SegmentationRules or something like that.

For me this is a bad idea, because SRX is not the only format/algorithm existing for segmentation.
Using this name while supporting only your old serializer (conf) format was not a problem because this format is semantically equivalent to SRX (i.e. you can convert from/to SRX without loosing any kind of information)
In another project I have created own segmentation formats, which give some possibilities better than SRX:
http://www.silvestris-lab.org/node/34
(description and samples are in the downloadable zip)
In the future I plan to implement them in DGT-OmegaT, simply because we have in DGT some rules which are absolutely impossible to express in SRX (for example "never cut inside a parenthesis"). I say in the future because the specification is not complete for now.
I did not begin this work but I am pretty sure that when I will do it, I will probably keep all other classes (MapRule, LanguageRule, Rule, and even Segmenter) unchanged, but simply replace the class SRX by a new one dedicated to my formats, and maybe add some classes for specific features.
So, for me SRX is the correct class name, because it means: in-memory representation of SRX file (also of a CONF file because the format is semantically equivalent) by opposition to any other segmentation format.

> a good compromise would be to simply name it loadFromDir

Yes, considering the fact that the method is static, it will always be preceeded by the class name, so no need to repeat it.
However, note the "from" in the name. Calling the method loadSRX, as actually, is correct in the sense that "you load SRX data". But if you have named it "loadFromSRX", then it would have been an error, because the method can load something else. 
Definitively I prefer loadFromDir. This is also coherent with the eventuality that, in the future, we load more than one file (for example, to have the list of abbreviations in a separate file)

src/org/omegat/core/segmentation/SRX.java
t-cordonnier: > Now that I understand your reason, we can think about whether or not coherency is a good thing in this instance, given that the current code has some obvious problems. I would be OK with you not adding this new getter, but I don't feel that strongly about it.

In the meantime I found a location where these accessors may/should be used.
Actually, your segmentation customizer (GUI) does not allow to modify these parameters. However, you have a class org.omegat.core.segmentation.datamodels.SRXOptionsModel which seems written for that but actually unused (I checked that it is called nowhere)
This is the only class which uses getters and setters for segment subflows (but actually it does not support cascade)

Now the questions are:
1. do you plan at any time to implement a window or a panel to edit these fields? if not, shouldn't we delete SRXOptionsModel?
2. Now that cascade is really supported by OmegaT, shouldn't we add a checkbox in the already existing GUI panel?
3. More complicated. What to do when the user checks/unchecks the box?
Two possibilities.

a. The simplest, of course, is to simply set the boolean value.
But if you know what cascade means, you certainly understand that all language rules are affected when you uncheck the box : all rules coming from next language maps are not used anymore.
So, second possibility:
b. In my project, for which I just sent you the link in previous message, I also implemented the possibility to "uncascade" an SRX file: not only it sets cascade="no", but it copies all the rules which would have applied to a given language in the correct location, ensuring that the resulting file is equivalent to the previously cascaded file.

My proposal would be the following: add a checkbox in the GUI, but when the user unchecks it, display a dialog "do you want to uncascade current rules?", with 3 possibilities: 1) YES (apply the algorithm and refreshes the editor), 2) NO (uncheck the box but do not touch the rules), 3) CANCEL (leave all as is, keep the box checked)

What do you think?

t-cordonnier:
> It seems like you may have missed a couple of my comments

I corrected most of the whitespace issues, simply did not give a comment for each of them.
But reading your corrections, I understand that I also missed the extra whitespaces on empty lines. Thanks for the correction.

src/org/omegat/core/segmentation/SRX.java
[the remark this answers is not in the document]
t-cordonnier: > You're right, this is probably important enough to have a GUI message.
> Instead the loadSRX method should throw an exception

OK, I will study this possibility. This is important because now that we use a standard format, we are susceptible to load segmentation rules copied by the user from another CAT tool, and what is correctly read by one CAT tool may not work in another one...

I will write again once it is done.

src/org/omegat/util/Preferences.java
t-cordonnier: OK, renaming to loadFromDir is done.

t-cordonnier: > I did not begin this work but I am pretty sure that when I will do it, I will probably keep all other classes (MapRule, LanguageRule, Rule, and even Segmenter) unchanged, but simply replace the class SRX by a new one dedicated to my formats, and maybe add some classes for specific features. Or, if I want to keep compatibility with all formats, I could have an abstract class SegmentationRules for which SRX is one implementation.

Finally I did it: implemented a (currently limited) support for CSCX format in DGT-OmegaT. And this works exactly as I had described before: an interface ISegmentationData which contains list of methods any format should support, then two implementations - SRX.java, which supports old CONF format because it is semantically equivalent; and CSC.java, which implements the specific part. Segmenter, LanguageRule and other classes are currently quite not affected.

I do not say that you should support such a format in OmegaT, that is not the question. But it confrms what I said about the name of SRX class.

amake:
Sorry for the long delay on this. Since this is a pretty major change, I'd like to wait until OmegaT 4 has been promoted to "standard" and we can put this in OmegaT 5.

brandelune:
All this is way beyond my understanding, but reading @amake's last comment, it seems that we have a window of opportunity now that 4 is stable is 5 is heading toward stability with the work on the manual undergoing.

brandelune:
@t-cordonnier I understand that this merge is not a priority at the moment, but it would be nice to set a milestone, even a "floating" one. Just to not let it rot. So I'm setting it for 6.1, with the possibility to push it further if needed. Thank you for your work.

miurahr:
A feature is integrated into #239

PR #25, opened 2018-07-01
Support offline mode for team projects

chelobaka:
This pull request (re)implements offline mode for team projects. Right now when a team project is opened or saved OmegaT throws an error to a user instead of handling the situation gracefully.

I didn't remove exception logging because I don't know about project policies on that matter.

### DIFF
--- a/src/org/omegat/core/data/RealProject.java
+++ b/src/org/omegat/core/data/RealProject.java
@@ -71,6 +71,7 @@ import org.omegat.core.segmentation.Segmenter;
 import org.omegat.core.statistics.CalcStandardStatistics;
 import org.omegat.core.statistics.Statistics;
 import org.omegat.core.statistics.StatisticsInfo;
+import org.omegat.core.team2.IRemoteRepository2;
 import org.omegat.core.team2.RebaseAndCommit;
 import org.omegat.core.team2.RemoteRepositoryProvider;
 import org.omegat.core.threads.CommandMonitor;
@@ -218,8 +219,6 @@ public class RealProject implements IProject {
      *
      * @param props
      *            project properties
-     * @param isNewProject
-     *            true if project need to be created
      */
     public RealProject(final ProjectProperties props) {
         config = props;
@@ -336,30 +335,34 @@ public class RealProject implements IProject {
             Core.getMainWindow().showStatusMessageRB("CT_LOADING_PROJECT");
 
             if (remoteRepositoryProvider != null) {
-                tmxPrepared = null;
-                glossaryPrepared = null;
-                
-                remoteRepositoryProvider.switchAllToLatest();
+                try {
+                    tmxPrepared = null;
+                    glossaryPrepared = null;
 
-                remoteRepositoryProvider.copyFilesFromRepoToProject("", '/' + RemoteRepositoryProvider.REPO_SUBDIR,
-                        '/' + RemoteRepositoryProvider.REPO_GIT_SUBDIR, '/' + RemoteRepositoryProvider.REPO_SVN_SUBDIR,
-                        '/' + OConsts.FILE_PROJECT,
-                        '/' + config.getProjectInternalRelative() + OConsts.STATUS_EXTENSION,
-                        '/' + config.getWritableGlossaryFile().getUnderRoot(),
-                        '/' + config.getTargetDir().getUnderRoot());
+                    remoteRepositoryProvider.switchAllToLatest();
 
-                // After adding filters.xml and segmentation.conf, we must reload them again
-                config.loadProjectFilters();
-                config.loadProjectSRX();
-                
-                loadFilterSettings();
-                
-                loadTranslations();
+		            remoteRepositoryProvider.copyFilesFromRepoToProject("", '/' + RemoteRepositoryProvider.REPO_SUBDIR,
+		                    '/' + RemoteRepositoryProvider.REPO_GIT_SUBDIR, '/' + RemoteRepositoryProvider.REPO_SVN_SUBDIR,
+		                    '/' + OConsts.FILE_PROJECT,
+		                    '/' + config.getProjectInternalRelative() + OConsts.STATUS_EXTENSION,
+		                    '/' + config.getWritableGlossaryFile().getUnderRoot(),
+		                    '/' + config.getTargetDir().getUnderRoot());
+
+		            // After adding filters.xml and segmentation.conf, we must reload them again
+		            config.loadProjectFilters();
+		            config.loadProjectSRX();
+                } catch (IRemoteRepository2.NetworkException e) {
+                    Log.logErrorRB("TEAM_NETWORK_ERROR", e.getCause());
+                    setOfflineMode();
+                }
+            }
+
+            loadFilterSettings();
+            loadTranslations();
+
+            if (remoteRepositoryProvider != null && isOnlineMode) {
                 Core.getMainWindow().showStatusMessageRB("TEAM_REBASE_AND_COMMIT");
                 rebaseAndCommitProject(true);
-            } else {
-                loadFilterSettings();
-                loadTranslations();
             }
 
             // Set project specific segmentation rules if they exist, or defaults otherwise.
@@ -647,7 +650,7 @@ public class RealProject implements IProject {
                 numberOfCompiled++;
             }
         }
-        if (remoteRepositoryProvider != null && config.getTargetDir().isUnderRoot() && commitTargetFiles) {
+        if (remoteRepositoryProvider != null && config.getTargetDir().isUnderRoot() && commitTargetFiles && isOnlineMode) {
             tmxPrepared = null;
             glossaryPrepared = null;
             // commit translations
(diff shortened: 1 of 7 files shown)


src/org/omegat/core/data/RealProject.java
```diff
@@ -336,39 +336,47 @@ public synchronized void loadProject(boolean onlineMode) {
...
+                    for (String file : new String[]{
+                            '/' + config.getProjectInternalRelative() + FilterMaster.FILE_FILTERS,
+                            '/' + config.getProjectInternalRelative() + SRX.CONF_SENTSEG}) {
+                        remoteRepositoryProvider.copyFilesFromRepoToProject(file);
+                    }
+                    // After adding filters.xml and segmentation.conf, we must reload them again
+                    config.loadProjectFilters();
+                    config.loadProjectSRX();
+
+                } catch (Exception e) {
```
amake: It doesn't seem right that *any* exception in this block leads to being considered offline. We should be defining our own `NetworkUnavailableException` or similar to wrap each remote provider's specific exception, then handle only that one.

chelobaka: Should this wrapping be done inside RemoteRepositoryProvider? I see no reason to do it in RealProject.

amake: Definitely not in `RealProject`. Whether `RemoteRepositoryProvider` is the best place or not, I don't recall off the top of my head. There might be a lower layer that is more appropriate, depending on the specific provider's implementation.

chelobaka:
It turned out we already had special exception for network errors but it wasn't used everywhere it should be. I scanned all repostory provider implementations and wrapped network related exceptions with common network error exception. Tested only with GIT and SVN repositories, though HTTP should work as well as it's pretty straightforward.

Some commits are just code cleanups.

RFE is here https://sourceforge.net/p/omegat/bugs/849/

ordtrogen:
This may not be relevant for the issue at hand, but while you're at it, in RealProject.java 

>for (String file : new String[]{
>    '/' + config.getProjectInternalRelative() + FilterMaster.FILE_FILTERS,
>    '/' + config.getProjectInternalRelative() + SRX.CONF_SENTSEG}) {

Wouldn't that look better as " ... String files : new String[] ..."  as it's an array (i.e plural)?

amake:
No, because the array is not being assigned to `files`; `files` is the name of the variable used to iterate the array, thus it is only ever refers to a single file.

ordtrogen:
Oh, you're right of course.  Well, at least I got confused by the use of a for-loop. Does that have an advantage over just two simple calls?  (feel free to ignore my OT comments :-)  Not terribly important)

remoteRepositoryProvider.copyFilesFromRepoToProject('/' + config.getProjectInternalRelative() + FilterMaster.FILE_FILTERS);
remoteRepositoryProvider.copyFilesFromRepoToProject('/' + config.getProjectInternalRelative() + SRX.CONF_SENTSEG);

amake:
The advantage is that you don't duplicate code: you write what you want to do just once, and you do it to all the items in the list. For a list of length two one can argue that it doesn't save very much, but when you consider that the list may be expanded in the future, I think it's fine.

amake:
I rebased your branch onto the latest trunk, only to find out that the Jenkins tests were failing because GitHub dropped support for TLS1.1 and Jenkins apparently has a problem with this. I re-pushed the same commit a few times just to test the Jenkins job; please disregard them.

The Jenkins failure is now meaningful: you added a method to `IProject` but didn't implement the method in all of the instances. Please address this.

(Since I mucked with your branch, you will want to reset your local branch to the remote one.)

src/org/omegat/core/data/RealProject.java
```diff
@@ -336,39 +335,47 @@ public synchronized void loadProject(boolean onlineMode) {
...
+                            '/' + config.getProjectInternalRelative() + FilterMaster.FILE_FILTERS,
+                            '/' + config.getProjectInternalRelative() + SRX.CONF_SENTSEG}) {
+                        remoteRepositoryProvider.copyFilesFromRepoToProject(file);
+                    }
+                    // After adding filters.xml and segmentation.conf, we must reload them again
+                    config.loadProjectFilters();
+                    config.loadProjectSRX();
+
+                } catch (IRemoteRepository2.NetworkException e) {
+                    Log.logErrorRB("TEAM_NETWORK_ERROR");
```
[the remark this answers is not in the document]
chelobaka: Would exception name (without stack trace) suffice here?

src/org/omegat/gui/main/MainWindowMenu.java
```diff
@@ -696,13 +696,13 @@ private void onProjectStatusChanged(final boolean isProjectOpened) {
         if (Core.getParams().containsKey(CLIParameters.NO_TEAM)) {
             projectTeamNewMenuItem.setEnabled(false);
         }
-        if (isProjectOpened && Core.getProject().isRemoteProject()
+        if (isProjectOpened && Core.getProject().isRemoteProject() && Core.getProject().isOnline()
```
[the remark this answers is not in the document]
chelobaka: Agreed. Simple error dialog should be enough here.

src/org/omegat/core/team2/impl/HTTPRemoteRepository.java
```diff
@@ -184,7 +184,7 @@ protected void retrieve(Properties etags, String file, String url, File out) thr
                 Log.logDebug(LOGGER, "Retrieve " + url + ": not modified");
                 return;
             default:
-                throw new RuntimeException("HTTP response code: " + conn.getResponseCode());
+                throw new NetworkException("HTTP response code: " + conn.getResponseCode());
```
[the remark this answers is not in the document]
chelobaka: Bad mistake. My tests showed that network failures for IP-based addresses or failures after the host was found (easy to do in debug mode) throw a `SocketException`. Pushed the fix commit.

src/org/omegat/gui/main/ProjectUICommands.java
```diff
@@ -488,11 +489,11 @@ protected Void doInBackground() throws Exception {
...
                                     needToSaveProperties = true;
                                 }
+                            } catch (IRemoteRepository2.NetworkException e) {
+                                // Do nothing. Network errors are handled in RealProject.
                             } catch (Exception e) {
                                 Log.logErrorRB(e, "TF_PROJECT_PROPERTIES_ERROR");
-                                    Log.log(e);
-                            throw new IOException(OStrings.getString("TF_PROJECT_PROPERTIES_ERROR") + "\n"
-                                            + e.getMessage(), e);
+                                throw e;
```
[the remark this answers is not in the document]
chelobaka: I found no reason for this wrapping. Enclosing scope still catches `Exception`.

chelobaka:
I made requested changes but can't follow Alex' advice. Here's what I wrote in SF ticket:
_There are three methods in RealProject which check for remote remoteRepositoryProvider, and these methods don't seem to be used anythere. They are compileProjectAndCommit, teamSyncPrepare and teamSync. I don't really know what to do with them._

amake:
I think your issue was resolved in the [the SF ticket](https://sourceforge.net/p/omegat/bugs/849/). If you are still blocked on something please let us know.

chelobaka:
> Need to check online mode on the every check RealProject.remoteRepositoryProvider != null

Methods checking condition:
* loadProject (Done)
* compileProjectAndCommit (Done)
* saveProject (Done)
* teamSyncPrepare (Done)
* teamSync (Not required, depends on `teamSyncPrepare` change of `preparedStatus` which can happen only online)
* isRemoteProject (Not required)

> change online mode only inside executeExclusively

Online mode is changed only inside `saveProject` and `loadProject` methods. The former fulfils this condition and the latter doesn't seem to require it.

> set RealProject.preparedStatus to NONE on every change online mode

Changes happen only in methods `setOnlineMode` and `setOfflineMode`. Both methods set `preparedStatus` to NONE.

src/org/omegat/core/data/RealProject.java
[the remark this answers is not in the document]
amake: I think we'd need the exception's message, but you're right we don't need the trace.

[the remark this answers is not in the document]
amake: Oops I missed what you did with this. It looks good now.

src/org/omegat/core/team2/impl/SVNRemoteRepository2.java
```diff
@@ -120,8 +119,9 @@ public void switchToVersion(String version) throws Exception {
             ourClientManager.getUpdateClient().doCheckout(url, baseDirectory, SVNRevision.HEAD,
                     toRev, SVNDepth.INFINITY, false);
             Log.logInfoRB("SVN_FINISH", "checkout");
-        } catch (Exception ex) {
+        } catch (SVNException ex) {
```
amake: Now we don't handle any non-`SVNException`s. Can you add a clause for `Exception` that just logs and re-throws it, like in `commit()`?

src/org/omegat/core/data/RealProject.java
```diff
@@ -1851,18 +1856,20 @@ public boolean existEntryInProject(EntryKey key) {
 
     void setOnlineMode() {
         if (!isOnlineMode) {
+            isOnlineMode = true;
+            preparedStatus = PreparedStatus.NONE;
             Log.logInfoRB("VCS_ONLINE");
             Core.getMainWindow().displayWarningRB("VCS_ONLINE", "VCS_OFFLINE");
         }
-        isOnlineMode = true;
```
amake: I would actually feel better if `isOnlineMode` and `preparedStatus` were set unconditionally here and in `setOfflineMode()`.

src/org/omegat/core/team2/impl/SVNRemoteRepository2.java
chelobaka: This try block throws only `SVNException`, so I just narrowed exception scope. I can revert this change but can't see why.

amake: The only *declared* exception is `SVNException`. Anything else that might happen due to e.g. a bug in SVNKit will not be caught.

Also, you might take a look at the SVNKit code and say "they are wrapping all exceptions with SVNException, so it's OK", and you might be right for this version of SVNKit. But you don't know about future versions.

So the safe play is to catch everything.

chelobaka: Thanks for explanation. I just reverted the change because extra catch won't bring any functionality here.

amake: That's good, too. Thanks.

amake:
Committed to trunk, r10462. Thanks for your hard work, and sorry for the slow response time.

PR #26, opened 2018-07-03
Show/log human-readable error on version check failure

chelobaka:
Version checking without network connection makes OmegaT dump exception stack trace to log. Since this is a common sutuation it's better to log human-readable error and don't pollute the log. The same error is shown in a dialog on manual check failure.

Version check classes have some hardcoded strings for log messages, but since these classes use custom loggers (without RB notation) I'm sure if it's a good idea to localize those messages.

### DIFF
--- a/src/org/omegat/gui/dialogs/VersionCheckDialog.java
+++ b/src/org/omegat/gui/dialogs/VersionCheckDialog.java
@@ -61,9 +61,9 @@ public class VersionCheckDialog {
                         new VersionCheckDialog(VersionChecker.getInstance().getRemoteVersion()).show(parent);
                     }
                 } catch (Exception e) {
-                    Log.logWarningRB("VERSION_CHECK_FAILED", e.toString());
+                    Log.logWarningRB("VERSION_CHECK_FAILED", e);
                     JOptionPane.showMessageDialog(parent,
-                            OStrings.getString("VERSION_CHECK_FAILED", e.getCause().toString()),
+                            OStrings.getString("VERSION_CHECK_FAILED", e.getCause()),
                             OStrings.getString("ERROR_TITLE"),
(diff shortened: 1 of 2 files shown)


amake:
I amended and force-pushed your commit just to re-trigger the Jenkins check.

src/org/omegat/gui/dialogs/VersionCheckDialog.java
```diff
@@ -61,8 +61,11 @@ protected void done() {
                         new VersionCheckDialog(VersionChecker.getInstance().getRemoteVersion()).show(parent);
                     }
                 } catch (Exception e) {
-                    Log.log(e);
-                    JOptionPane.showMessageDialog(parent, e.getLocalizedMessage(), OStrings.getString("ERROR_TITLE"),
+                    Log.logWarningRB("VERSION_CHECK_FAILED");
+                    Log.log(e.toString());
```
amake: You can combine these two lines as well if you use a placeholder in the properties string.

src/org/omegat/gui/dialogs/VersionCheckDialog.java
```diff
@@ -61,8 +61,11 @@ protected void done() {
...
+                    JOptionPane.showMessageDialog(parent,
+                            OStrings.getString("VERSION_CHECK_FAILED"),
```
amake: If you want to leave out the details of the error here, I guess we need a separate string. However since this dialog is only shown when the user explicitly requests a version check, not explaining the reason seems inconsiderate to me. But I don't feel that strongly about it either way.

chelobaka:
Requested changes applied.

amake:
I have squashed and applied this on trunk. Thanks for your contribution.

PR #28, opened 2018-10-24
Implementation of the .php filter for Moodle

keko:
Implementation for the RFE: https://sourceforge.net/p/omegat/feature-requests/1407/
With the colaboration of the "Oficina de Software Libre do CIXUG".

The code of the filter is adapted from the .dtd filter. It changes the pattern and some conditions in the «if» into the processFile function.


src/org/omegat/filters2/moodlephp/MoodlePHPFilter.java
```diff
@@ -0,0 +1,218 @@
...
+
+    protected static final Pattern RE_ENTITY = Pattern.compile("\\$string\\['(.+)'] (=) '(.+)';",
+            Pattern.DOTALL);
+
+    protected Map<String, String> align;
+
+    /**
+     * If true, will remove non-translated segments in the target files
+     */
+    public static boolean removeStringsUntranslated = false;
```
amake: I know you based this on the Mozilla DTD filter, and this part is the same as in that one, but it doesn't make any sense for this to be `static`. Can you please make it non-static?

src/org/omegat/filters2/moodlephp/MoodlePHPFilter.java
```diff
@@ -0,0 +1,218 @@
...
+        // If the value is null the default is false
+        if ((removeStringsUntranslatedStr != null) && (removeStringsUntranslatedStr.equalsIgnoreCase("true"))) {
+            removeStringsUntranslated = true;
+        } else {
+            removeStringsUntranslated = false;
+        }
+
+        StringBuilder block = new StringBuilder();
+        boolean isInBlock = false;
+        boolean foundQuotes = false;
```
amake: `foundQuotes` isn't actually used for anything. If that's correct, please remove it.

keko: Remove it.

src/org/omegat/filters2/moodlephp/MoodlePHPFilter.java
keko: Changed it.

amake:
Thank you very much for your contribution. I have squashed and committed to svn trunk.

PR #29, opened 2018-11-22
Iadaatpa

e-gomez:


amake:
Thanks for your contribution.

First, can you give us some information about this MT service? How can we sign up and test it?

Second, I see you are adding a new JSON parser dependency. I would prefer to avoid adding new dependencies when we already have e.g. Jackson and JSONIC pulled in by other dependencies (see the output of `gradlew dependencies`). If you have a good reason why we should add this one as well, please let us know.

Third, I see lots of problems with naming conventions and spacing. Please follow existing conventions within the code (methods should be CamelCase, not snake_case, etc.).

Finally, have you looked into publishing this as a plugin? There are many MT plugins already available that you can use for reference:
https://sourceforge.net/p/omegat/wiki/Plugins/

amake:
Also please address the testing issue reported by Jenkins:
https://omegat.ci.cloudbees.com/job/github-pr-test/139/findbugsResult/new/

e-gomez:
The IADAATPA Consortium will supply infrastructure services to the EU in the field of Digital Service Infrastructures (DSIs), which will deliver a highly secure, scalable Machine Translation platform, which will make automated translation ubiquitous across European Public Administrations.

The consortium includes Pangeanic, KantanMT, Prompsit, Tilde, Everis and DCU, with support from public administrations in Spain (Segittur and SESIAD). The IADAATPA promises to accelerate Trans-European connectivity, improve access to digital services, and facilitate interoperability between digital services and link national infrastructure to the existing core MT@EC automated translation platform.

The IADAATPA platform will integrate seamlessly with existing European eDelivery Service, ensuring that other technology players can leverage its capabilities, scalability and services. This will ensure that the IADAATPA platform can be deployed across other digital services, by other MT providers and technology developers broadening the range of multilingual services across the European Union.

amake:
Thank you, that's useful, but how does one go about getting access? What are the conditions of use? I'm trying to figure out if it makes sense to include this in OmegaT itself vs asking you to publish it as a plugin instead.

PR #36, opened 2019-09-17
Glossaries change layout

t-cordonnier:
3 patches, conform to last message in the user list:
Patch 1: add colors for source, translation and comment. These colors are configurable in the colors panel
Patch 2: can disable merging of entries (seems that some people don't like it)
Patch 3: change layout for merged entries, in order to put each comment just after the translation it is associated to (see discussion in users list)

### DIFF
--- a/src/org/omegat/gui/glossary/DefaultGlossaryRenderer.java
+++ b/src/org/omegat/gui/glossary/DefaultGlossaryRenderer.java
@@ -113,13 +113,8 @@ public class DefaultGlossaryRenderer implements IGlossaryRenderer {
 
     @Override
     public void render(GlossaryEntry entry, StyledDocument doc) {
-        int start = doc.getLength();
         DocTarget trg = new DocTarget(doc);
         render(entry, trg);
-        int end = doc.getLength();
-        SimpleAttributeSet attrs = new SimpleAttributeSet();
-        attrs.addAttribute(TooltipAttribute.ATTRIBUTE_KEY, new TooltipAttribute(renderToHtml(entry)));
-        doc.setCharacterAttributes(start, end - start, attrs, false);
         trg.append("\n\n");
     }
 
@@ -130,6 +125,7 @@ public class DefaultGlossaryRenderer implements IGlossaryRenderer {
         String[] targets = entry.getLocTerms(false);
         String[] comments = entry.getComments();
         boolean[] priorities = entry.getPriorities();
+        String[] origins = entry.getOrigins(false);
         StringBuilder commentsBuf = new StringBuilder();
         for (int i = 0, commentIndex = 0; i < targets.length; i++) {
             if (i > 0 && targets[i].equals(targets[i - 1])) {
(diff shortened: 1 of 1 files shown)


src/org/omegat/gui/glossary/DefaultGlossaryRenderer.java
```diff
@@ -145,7 +141,9 @@ protected void render(GlossaryEntry entry, IRenderTarget<?> trg) {
                 trg.append(", ");
             }
 
-            trg.append(bracketEntry(targets[i]), priorities[i] ? PRIORITY_ATTRIBUTES : null);
+            SimpleAttributeSet attrs = new SimpleAttributeSet (priorities[i] ? PRIORITY_ATTRIBUTES : NO_ATTRIBUTES);
```
amake: There is an unneeded space after `SimpleAttributeSet`.

Just FYI, we have a Gradle task to auto-format files changed on the current branch:

```
./gradle spotlessChangedApply
```

This is based on our format config available here: https://github.com/omegat-org/omegat/blob/master/docs_devel/eclipse-formatting.xml

t-cordonnier: After your comments in the other pull request, I now use gradle task "check" and read the report in HTML about the code I modified. Sorry that I did not see this one.  This is now corrected.

Thanks for the tip about global task spotlessChangedApply. That said, if it automatically changes all classes, including those I did not modify, then I prefer to apply the change in a dedicated commit, rather than including the change in a functional commit. In my work I often use gitk to check when and where I implemented a dedicated feature (that's what I do when you ask me for a feature implemented in DGT) and to do that, I prefer to have atomic commits, instead of seeing in a big commit changes which are not related to the feature described in commit's comment. 
I accept to correct the style in my code before pushing in GitHub, so I should not add new style issues. But correcting already existing issues is another topic, which should be in own commit, whenever you do it automatically or not.

amake: > That said, if it automatically changes all classes, including those I did not modify

Unfortunately it can't be applied to classes, only files. It's only supposed to touch files you changed, but there are probably situations where it gets it wrong.

> But correcting already existing issues is another topic, which should be in own commit, whenever you do it automatically or not.

Yes. I want to simply apply it to the whole codebase, but I imagine that would make accepting patches from DGT very hard, so I have not done it yet. Maybe some time in the future.

t-cordonnier:
Le mercredi 23 octobre 2019 à 10:22:50 UTC+2, Aaron Madlon-Kay <notifications@github.com> a écrit :  
> Unfortunately it can't be applied to classes, only files. It's only supposed to touch files you changed, but there are probably situations where it gets it wrong.
Yes, in reality I meant files, not classes. 
I did not know that it affects only files which I changed. In such a case, yes it makes sense that I use it. But I still do not want to see in a commit changes which are not related to the feature, simply because they are in the same file. And of course you are right, if we apply the style changes before, it can make conflict with patch taken from DGT. I will have a look how I can merge the two constraints, I will tell you later.
RegardsThomas

PR #37, opened 2019-10-09
1466 Allow users to choose Apertium server

MarcRiera:
https://sourceforge.net/p/omegat/feature-requests/1466/


src/org/omegat/core/machinetranslators/ApertiumTranslate.java
```diff
@@ -83,18 +94,23 @@ private String apertiumCode(Language language) {
 
     @Override
     protected String translate(Language sLang, Language tLang, String text) throws Exception {
-        String prev = getFromCache(sLang, tLang, text);
-        if (prev != null) {
-            return prev;
-        }
-
```
amake: Did you mean to delete the caching mechanism? That doesn't seem desirable to me.

src/org/omegat/core/machinetranslators/ApertiumTranslate.java
```diff
@@ -43,9 +51,12 @@
...
  */
 public class ApertiumTranslate extends BaseTranslate {
-    protected static final String GT_URL = "https://www.apertium.org/apy/translate?q=";
+    protected static final String PROPERTY_APERTIUM_CUSTOM = "apertium.server.custom";
+    protected static final String PROPERTY_APERTIUM_URL = "/translate?q=";
+    protected static final String PROPERTY_APERTIUM_URL2 = "&markUnknown=no&langpair=#sourceLang#|#targetLang#&key=";
+    protected static final String PROPERTY_APERTIUM_SERVER = "https://www.apertium.org/apy";
     // Specific OmegaT key
-    protected static final String GT_URL2 = "&markUnknown=no&langpair=#sourceLang#|#targetLang#&key=bwuxb5jS+VwSJ8mLz1qMfmMrDGA";
+    protected static final String PROPERTY_APERTIUM_API_KEY = "bwuxb5jS+VwSJ8mLz1qMfmMrDGA";
```
amake: The value of a `PROPERTY_*` constant is meant to be a human-readable key used for storing the value in preferences and for supplying the value on the command line. See e.g. the [Microsoft MT connector](https://github.com/omegat-org/omegat/blob/3dff23f3af974d1a1b8015375255b60bb91e6879/src/org/omegat/core/machinetranslators/MicrosoftTranslatorAzure.java#L61)

Here they are doing double duty, serving as a (human-*un*readable) keys *and* the actual values when no other value has been supplied.

Please make separate definitions for keys and default values.

src/org/omegat/core/machinetranslators/ApertiumTranslate.java
```diff
@@ -83,18 +94,23 @@ private String apertiumCode(Language language) {
...
+
+        if (!useCustomServer()) {
+            server = PROPERTY_APERTIUM_SERVER;
+        }
+        if (!useCustomServer()) {
+            apiKey = PROPERTY_APERTIUM_API_KEY;
+        }
+
+        String url2 = PROPERTY_APERTIUM_URL2.replace("#sourceLang#", sourceLang).replace("#targetLang#", targetLang);
+        String url = server + PROPERTY_APERTIUM_URL + URLEncoder.encode(trText, "UTF-8") + url2 + apiKey;
```
amake: Building the URL by concatenation seems dangerous to me. Instead I would suggest a format string, like:

```java
protected static final String APERTIUM_URL_FORMAT = "https://%s/apy/translate?q=%s&markUnknown=no&langpair=%s|%s&key=%s";

...

String url = APERTIUM_URL_FORMAT.format(server, trText, sourceLang, targetLang, apiKey);
```

src/org/omegat/core/machinetranslators/ApertiumTranslate.java
```diff
@@ -138,4 +154,67 @@ protected String getJsonResults(String json) {
...
+        dialog.panel.valueField1.setText(getServerUrl());
+        dialog.panel.valueField1.setColumns(20);
+        dialog.panel.valueLabel2.setText(OStrings.getString("APERTIUM_CUSTOM_SERVER_KEY_LABEL"));
+        dialog.panel.valueField2.setText(getCredential(PROPERTY_APERTIUM_API_KEY));
+
+        dialog.panel.valueLabel1.setEnabled(apiCheckBox.isSelected());
+        dialog.panel.valueLabel2.setEnabled(apiCheckBox.isSelected());
+        dialog.panel.valueField1.setEnabled(apiCheckBox.isSelected());
+        dialog.panel.valueField2.setEnabled(apiCheckBox.isSelected());
+        dialog.panel.temporaryCheckBox.setEnabled(apiCheckBox.isSelected());
```
amake: You could obviate the need for these lines by storing the `ItemListener` set above as a local variable and calling its `itemStateChanged` with `null`.

src/org/omegat/core/machinetranslators/ApertiumTranslate.java
MarcRiera: Sorry, I removed it for testing and forgot to add it back. It is now in code again.

src/org/omegat/core/machinetranslators/ApertiumTranslate.java
MarcRiera: Thanks, I agree that the approach was messy. Fixed in latest commit.

src/org/omegat/core/machinetranslators/ApertiumTranslate.java
MarcRiera: You are absolutely right, something similar was already in place but it is better to fix it now to prevent future issues. Fixed in latest commit.

src/org/omegat/core/machinetranslators/ApertiumTranslate.java
MarcRiera: Thanks, I managed to simplify this part with your suggestion.

MarcRiera:
I have noticed two aspects which could be improved:

1. If the custom URL is incorrect and the plugin is unable to retrieve a translation, the plugin simply shows the request URL in the MT panel. It would be better to handle this kind of error more nicely to guide the user to check the setting, with messages like "404 error. Please check the Apertium server URL is correct".

2. MT caching poses a problem if the user changes the URL halfway during a translation or if the server is updated and the results change (rare situation, but more likely if connecting to localhost). I have not found a way to flush the MT cache (which would fix the first situation), so the only solution I can think of is to add an additional checkbox in the plugin's settings to enable/disable caching.

What do you think?

src/org/omegat/core/machinetranslators/ApertiumTranslate.java
```diff
@@ -138,4 +159,63 @@ protected String getJsonResults(String json) {
...
+        };
+
+        apiCheckBox.addItemListener(toggleInterface);
+
+        dialog.panel.itemsPanel.add(apiCheckBox,1);
+        dialog.panel.valueLabel1.setText(OStrings.getString("APERTIUM_CUSTOM_SERVER_URL_LABEL"));
+        dialog.panel.valueField1.setText(getCredential(PROPERTY_APERTIUM_SERVER_URL));
+        dialog.panel.valueField1.setColumns(20);
+        dialog.panel.valueLabel2.setText(OStrings.getString("APERTIUM_CUSTOM_SERVER_KEY_LABEL"));
+        dialog.panel.valueField2.setText(getCredential(PROPERTY_APERTIUM_SERVER_KEY));
```
amake: The `temporaryCheckBox` needs to be properly initialized. See `isCredentialStoredTemporarily` in `BaseTranslate`.

src/org/omegat/core/machinetranslators/ApertiumTranslate.java
```diff
@@ -93,8 +105,17 @@ protected String translate(Language sLang, Language tLang, String text) throws E
...
-        String url = GT_URL + URLEncoder.encode(trText, "UTF-8") + url2;
+        String server = getCredential(PROPERTY_APERTIUM_SERVER_URL);
+        String apiKey = getCredential(PROPERTY_APERTIUM_SERVER_KEY);
+
+        if (!useCustomServer()) {
+            server = APERTIUM_SERVER_URL;
+        }
+        if (!useCustomServer()) {
+            apiKey = APERTIUM_SERVER_KEY;
+        }
```
amake: These two if statements can be combined.

src/org/omegat/core/machinetranslators/ApertiumTranslate.java
```diff
@@ -93,8 +105,17 @@ protected String translate(Language sLang, Language tLang, String text) throws E
         String sourceLang = apertiumCode(sLang);
         String targetLang = apertiumCode(tLang);
 
-        String url2 = GT_URL2.replace("#sourceLang#", sourceLang).replace("#targetLang#", targetLang);
-        String url = GT_URL + URLEncoder.encode(trText, "UTF-8") + url2;
+        String server = getCredential(PROPERTY_APERTIUM_SERVER_URL);
```
amake: Do you consider the server URL to be a "credential" (requiring secure storage)? I don't think it should be, in which case it should be stored as a normal preference and not with the credentials API.

src/org/omegat/core/machinetranslators/ApertiumTranslate.java
```diff
@@ -43,9 +51,13 @@
  * @author Didier Briel
  */
 public class ApertiumTranslate extends BaseTranslate {
-    protected static final String GT_URL = "https://www.apertium.org/apy/translate?q=";
+    protected static final String PROPERTY_APERTIUM_SERVER_CUSTOM = "apertium.server.custom";
+    protected static final String PROPERTY_APERTIUM_SERVER_URL = "apertium.server.url";
+    protected static final String PROPERTY_APERTIUM_SERVER_KEY = "apertium.server.key";
+    protected static final String APERTIUM_SERVER_URL = "https://www.apertium.org/apy";
```
amake: I would add `_DEFAULT` or something to `APERTIUM_SERVER_URL` and `APERTIUM_SERVER_KEY` to make it clearer what they are for.

amake:
>     1. If the custom URL is incorrect and the plugin is unable to retrieve a translation, the plugin simply shows the request URL in the MT panel. It would be better to handle this kind of error more nicely to guide the user to check the setting, with messages like "404 error. Please check the Apertium server URL is correct".

Yes, that sounds good. I believe other MT connectors show error messages like this.

>     2. MT caching poses a problem if the user changes the URL halfway during a translation or if the server is updated and the results change (rare situation, but more likely if connecting to localhost). I have not found a way to flush the MT cache (which would fix the first situation), so the only solution I can think of is to add an additional checkbox in the plugin's settings to enable/disable caching.

Currently there is no way to clear the cache without restarting OmegaT.

It seems a bit odd to me to add an option to enable/disable the cache just for this connector, but I'm having a hard time imagining a better solution.

What about clearing the cache if the URL changed?

MarcRiera:
> What about clearing the cache if the URL changed?

This is what I had in mind, but it is not possible to do it currently (MT plugins can only add and retrieve cached entries). However, I have investigated how the cache is handled in ```BaseTranslate``` and it should be easy to add a function there to clear the cache that can be called from the plugin if the URL has changed.

Thinking again about the localhost situation, given how rare it is to work on a language pair while translating (for several reasons, including Apertium's compilation times), I think we can simply ignore it for the moment.

src/org/omegat/core/machinetranslators/ApertiumTranslate.java
MarcRiera: Agree, fixed in latest commit.

src/org/omegat/core/machinetranslators/ApertiumTranslate.java
MarcRiera: Fixed in latest commit, thanks for pointing this out.

src/org/omegat/core/machinetranslators/ApertiumTranslate.java
MarcRiera: Combined in latest commit, no reason to keep them separated.

src/org/omegat/core/machinetranslators/ApertiumTranslate.java
MarcRiera: It is not, so I have changed it to a normal preference. Thanks!

src/org/omegat/core/machinetranslators/BaseTranslate.java
```diff
@@ -145,6 +145,14 @@ protected String putToCache(Language sLang, Language tLang, String text, String
         return cache.put(sLang + "/" + tLang + "/" + text, result);
     }
 
+    /**
+     * Clear the machine translation cache.
+     */
+    protected void clearCache() {
+        cache.clear();
+        return;
```
amake: This `return` is not necessary.

src/org/omegat/core/machinetranslators/ApertiumTranslate.java
```diff
@@ -123,6 +143,9 @@ protected String getJsonResults(String json) {
         if (rootNode.containsKey("responseStatus")) {
             code = (Integer) rootNode.get("responseStatus");
         }
+        else {
```
amake: Our coding style would put this `else` on the previous line with the closing bracket. To avoid having to nitpick these things, I made a Gradle task to auto-format files changed on the current branch:

```
./gradle spotlessChangedApply
```

Could you run it and see if it works for you?

src/org/omegat/core/machinetranslators/ApertiumTranslate.java
```diff
@@ -138,4 +161,79 @@ protected String getJsonResults(String json) {
...
+                String server = panel.valueField1.getText().trim();
+                String apiKey = panel.valueField2.getText().trim();
+
+                if (!getCustomServerUrl().equals(server)) {
+                    clearCache();
+                }
+
+                System.setProperty(PROPERTY_APERTIUM_SERVER_URL, server);
+                Preferences.setPreference(PROPERTY_APERTIUM_SERVER_URL, server);
+                setCredential(PROPERTY_APERTIUM_SERVER_KEY, apiKey, temporary);
```
amake: With this code we can have a situation where `PROPERTY_APERTIUM_SERVER_CUSTOM` is `true` but there is no value for `PROPERTY_APERTIUM_SERVER_URL` or `PROPERTY_APERTIUM_SERVER_KEY`.

There are a lot of different levels of hand-holding you can do to solve this. Right now I guess this is handled by the code path that shows `APERTIUM_CUSTOM_SERVER_NOTFOUND`. If that's good enough then you can leave it.

If you wanted to be nicer, I would add `DocumentListener`s to the text fields that enable/disable the OK button according to whether the required fields are filled in:

```java
Runnable updateOk = () -> {
    boolean needsFields = apiCheckBox.isSelected();
    boolean hasFields = !dialog.panel.valueField1.getText().trim().isEmpty()
            && !dialog.panel.valueField2.getText().trim().isEmpty();
    boolean canConfirm = !needsFields || hasFields;
    dialog.panel.okButton.setEnabled(canConfirm);
};
// Call updateOk.run() in listeners
```

It should probably also be called in the `ItemListener` below.

src/org/omegat/core/machinetranslators/ApertiumTranslate.java
MarcRiera: Thanks, I did not know about this and it has detected other things in addition to this. Fixed in the latest commit.

src/org/omegat/core/machinetranslators/ApertiumTranslate.java
MarcRiera: Thanks, I have added this check to prevent users from clicking the OK button with a blank URL field (the API key is optional, so it is not checked). If the URL is not valid the ```APERTIUM_CUSTOM_SERVER_NOTFOUND``` message is shown to prompt the user to check it.

amake: If the API key is not supplied then the URL will have `&key=` with a blank value. I take it that's not a problem?

MarcRiera: Yes, it is fine, the API itself handles this nicely.

amake:
I've squashed and committed to master over at SourceForge. Thanks very much for your contribution.

PR #38, opened 2019-11-07
Bug in the implementation of the filter for the php files of Moodle.

keko:
The filter truncates some strings. If the string contain the sequence '; in the inside, it cuts the string in that site. Example:

```php
$string['auth_ldap_attrcreators'] = 'List of groups or contexts whose members are allowed to create attributes. Separate multiple groups with \';\'. Usually something like \'cn=teachers,ou=staff,o=myorg\'';
```
In the interface of Omegat shows the next:

> List of groups or contexts whose members are allowed to create attributes. Separate multiple groups with \

### DIFF
--- a/src/org/omegat/filters2/moodlephp/MoodlePHPFilter.java
+++ b/src/org/omegat/filters2/moodlephp/MoodlePHPFilter.java
@@ -61,7 +61,7 @@ public class MoodlePHPFilter extends AbstractFilter {
 
     public static final String OPTION_REMOVE_STRINGS_UNTRANSLATED = "unremoveStringsUntranslated";
 
-    protected static final Pattern RE_ENTITY = Pattern.compile("\\$string\\['(.+)'] (=) '(.+)';",
+    protected static final Pattern RE_ENTITY = Pattern.compile("\\$string\\['(.+)'\\] (=) '(.+)(';)$",
             Pattern.DOTALL);
 
     protected Map<String, String> align;
@@ -135,9 +135,11 @@ public class MoodlePHPFilter extends AbstractFilter {
                 processBlock(block.toString(), outFile);
                 block.setLength(0);
             }
-            if (!Character.isWhitespace(c)) { // In the regexp, there could be whitespace between " and >
-                previousChar = c;
+            if (c == quotes && previousChar == 92) {
+                previousChar = 0;
             }
+            else
+                previousChar = c;
         }
     }
 
--- a/src/org/omegat/filters2/moodlephp/MoodlePHPFilter.java
+++ b/src/org/omegat/filters2/moodlephp/MoodlePHPFilter.java
@@ -117,7 +117,7 @@ public class MoodlePHPFilter extends AbstractFilter {
 
         StringBuilder block = new StringBuilder();
         boolean isInBlock = false;
-        int quotes = 39;
+        final char quotes = '\'';
         int previousChar = 0;
         int c;
         while ((c = inFile.read()) != -1) {
@@ -135,7 +135,7 @@ public class MoodlePHPFilter extends AbstractFilter {
                 processBlock(block.toString(), outFile);
                 block.setLength(0);
             }
-            if (c == quotes && previousChar == 92) {
+            if (c == quotes && previousChar == '\\') {
                 previousChar = 0;
             }
             else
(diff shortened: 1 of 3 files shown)


src/org/omegat/filters2/moodlephp/MoodlePHPFilter.java
```diff
@@ -61,7 +61,7 @@
 
     public static final String OPTION_REMOVE_STRINGS_UNTRANSLATED = "unremoveStringsUntranslated";
 
-    protected static final Pattern RE_ENTITY = Pattern.compile("\\$string\\['(.+)'] (=) '(.+)';",
+    protected static final Pattern RE_ENTITY = Pattern.compile("\\$string\\['(.+)'\\] (=) '(.+)(';)$",
```
amake: Is this regex really sufficient? It looks like a [Moodle strings file](https://docs.moodle.org/dev/String_API#Adding_language_file_to_plugin) can be arbitrary PHP, so it could have whitespace or comments on the end. There's also some syntax involving HTML-like tags.

If you want to improve this, there is a common regex pattern for [matching only unescaped characters](https://stackoverflow.com/a/11819111/448068). This would let you match only the "real" end of the string literal and ignore anything afterwards.

src/org/omegat/filters2/moodlephp/MoodlePHPFilter.java
```diff
@@ -135,9 +135,11 @@ protected void processFile(BufferedReader inFile, BufferedWriter outFile, Filter
                 processBlock(block.toString(), outFile);
                 block.setLength(0);
             }
-            if (!Character.isWhitespace(c)) { // In the regexp, there could be whitespace between " and >
-                previousChar = c;
+            if (c == quotes && previousChar == 92) {
```
amake: Representing the chars as ints is quite cryptic; I would suggest changing them to char literals: `'\\'` here, and `'\''` above. In fact there doesn't seem to be a point in having them be `int` at all; you might as well make them `char`.

While you're at it, you might make the `quotes` variable a static final constant, or at least make the local variable `final` to indicate that it's not meant to be modified by the loop.

test/src/org/omegat/filters/MoodlePHPFilterTest.java
```diff
@@ -38,10 +38,10 @@
     public void testParse() throws Exception {
         parse(new MoodlePHPFilter(), "test/data/filters/MoodlePHP/file.php");
     }
-
+
```
amake: A lot of garbage whitespace was added to this file. Could you remove them please?

src/org/omegat/filters2/moodlephp/MoodlePHPFilter.java
```diff
@@ -135,9 +135,11 @@ protected void processFile(BufferedReader inFile, BufferedWriter outFile, Filter
...
+                previousChar = 0;
             }
+            else
```
amake: Our coding conventions require brackets for the `else`.

You can run `./gradlew spotlessChangedApply` to run an autoformatter on the files changed in your branch.

src/org/omegat/filters2/moodlephp/MoodlePHPFilter.java
keko: I do not develop very well. Sometimes I am able to adapt code to meet my needs (I rarely get it).

I don't know what php files are like. I needed Omegat to work natively with the Moodle php files. I made a filter that worked with Moodle files. But a translator using OmegaT found a failure in the filter.

Both pattern, the new and the previous, they work ok. The string is filtered before that the pattern executes. The error was in the while loop.

I will be very grateful if you can improve the filter.

src/org/omegat/filters2/moodlephp/MoodlePHPFilter.java
keko: I did something, but I am not able to do everything you suggest.

test/src/org/omegat/filters/MoodlePHPFilterTest.java
keko: Solved.

src/org/omegat/filters2/moodlephp/MoodlePHPFilter.java
keko: Done.

src/org/omegat/filters2/moodlephp/MoodlePHPFilter.java
amake: Sure, no problem. Thanks for your contribution.

PR #41, opened 2020-04-26
add reference to the 3 tmx scores

brandelune:
based on the descriptions in Preferences.

### DIFF
--- a/doc_src/en/OmegaT4_EditorsPanes.xml
+++ b/doc_src/en/OmegaT4_EditorsPanes.xml
@@ -214,9 +214,12 @@
     <guimenuitem>Ctrl+R</guimenuitem> to replace the target field with the match or use
     <guimenuitem>Ctrl+I </guimenuitem> to insert it at the cursor position.</para>
 
-    <para>The matching percentage is roughly equivalent to taking the number
-    of common words in the matched and the matching segment and dividing by
-    the number of words in the longer of the two. The selected fuzzy match is
+    <para>three matching percentages are in order: the percentage calculated with
+    stemming and ignoring tags and numbers (generally the highest), the percentage
+    calculated without stemming and still ignoring tags and numbers (generally slightly
+    lower) and the percentage calculated with the full text, including tags and numbers
+    (generally the lowest).</para>
+    <para>The selected fuzzy match is
     highlighted in bold, words that are missing in the segment you are
     translating are colored blue and words adjacent to the missing parts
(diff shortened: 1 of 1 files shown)


doc_src/en/OmegaT4_EditorsPanes.xml
```diff
@@ -214,9 +214,12 @@
     <guimenuitem>Ctrl+R</guimenuitem> to replace the target field with the match or use
     <guimenuitem>Ctrl+I </guimenuitem> to insert it at the cursor position.</para>
 
-    <para>The matching percentage is roughly equivalent to taking the number
-    of common words in the matched and the matching segment and dividing by
-    the number of words in the longer of the two. The selected fuzzy match is
+    <para>three matching percentages are in order: the percentage calculated with
```
amake, line 217:
```suggestion
    <para>The three matching percentages are in order: the percentage calculated with
```

amake: This is definitely an improvement. Things that could be improved more:

1. Formatting: This would probably be more readable as a list of three items
2. Jargon: Is stemming explained anywhere in the manual? I can't remember. It's not explained in the UI, but that's acceptable due to limited space. Here, we could even just link to e.g. https://en.wikipedia.org/wiki/Stemming
3. Implication: It would be nice to clarify that while you can change the sort order of the five fuzzy matches, you can't change which percentage is used to select the five fuzzy matches.

Bonus: You could reflow the second paragraph. I do this in Emacs with `M-q`.

brandelune: Ok, I'll rewrite and resubmit. Thank you for the remarks. Re. emacs, the reflowing did not use the same width as the xml so I gave it up. I'll do that too.

amake: You can make changes as additional commits and just push them to this branch. No need to close the PR.

I think I might have found that Emacs's reflow didn't match the original, but I don't consider that a problem. To keep the diff from getting crazy, though, I would reflow only the paragraphs you actually touch.

brandelune:
The github interface always confuses me. Sorry. Ok for emacs. It's just a matter of changing the default width, not a big deal.

amake:
If you could specify the default width in a `.dir-locals.el` that would be great, since it would prevent people from creating diff noise when reflowing in the future.

brandelune:
Thank you for the suggestion. I was not aware of this solution.
https://www.gnu.org/software/emacs/manual/html_node/emacs/Directory-Variables.html
I'll check that later!

PR #42, opened 2020-04-27
add reference to the 3 fuzzy types

brandelune:
also add a directory settings file for emacs and reformat the subsequent paragraph


amake:
I've squashed and committed to master on SourceForge.

Thanks!

PR #43, opened 2020-05-25
Add Yandex.Cloud MT engine

chelobaka:
This PR implements Yandex.Cloud machine translation engine. The engine API also allows to specify glossary pairs, but I got weird translation results with them and prefer to have base functionality code reviewed and accepted first.

SF ticket: https://sourceforge.net/p/omegat/feature-requests/1494/


src/org/omegat/util/WikiGet.java
```diff
@@ -421,11 +421,15 @@ private static String getStringContent(HttpURLConnection conn, String defaultCha
...
         public final int code;
         public final String message;
+        public final String body;
 
         public ResponseError(HttpURLConnection conn) throws IOException {
             super(conn.getResponseCode() + ": " + conn.getResponseMessage());
             code = conn.getResponseCode();
             message = conn.getResponseMessage();
+            try (InputStream in = conn.getErrorStream()) {
+                body = IOUtils.toString(in, "utf-8");
```
lc-nyovchev, line 431: It's a good practice on JDK 7+ to use `StandardCharsets.UTF-8.name()` rather than hardcoded "utf-8". 
https://stackoverflow.com/questions/6698354/where-to-get-utf-8-string-literal-in-java

src/org/omegat/core/machinetranslators/YandexCloudTranslate.java
```diff
@@ -0,0 +1,245 @@
...
+public class YandexCloudTranslate extends BaseTranslate {
+
+    protected static final String PROPERTY_OATH_TOKEN = "yandex.cloud.oauth-token";
+    protected static final String PROPERTY_FOLDER_ID = "yandex.cloud.folder-id";
+
+    protected static final int MAX_TEXT_LENGTH = 10000;
+    protected static final int AIM_TOKEN_TTL_SECONDS = 3600; // Recommended value
+
+    protected static final String IAM_TOKEN_URL = "https://iam.api.cloud.yandex.net/iam/v1/tokens";
+    protected static final String TRANSLATE_URL = "https://translate.api.cloud.yandex.net/translate/v2/translate";
```
lc-nyovchev, line 67: What's the benefit of these instance variables being protected? I'd imagine they are not accessed outside of the scope of YandexCloudTranslate, so maybe we can consider making them private?

src/org/omegat/core/machinetranslators/YandexCloudTranslate.java
```diff
@@ -0,0 +1,245 @@
...
+
+    private String AIMErrorMessage = null;
+    private String cachedAIMToken = null;
+    private long lastAIMTokenTime = 0;
```
lc-nyovchev, line 71: nitpick: it's probably slightly better to denote this long as a long: https://stackoverflow.com/questions/769963/javas-l-number-long-specification

e.g change:
`private long lastAIMTokenTime = 0;`
to 
`private long lastAIMTokenTime = 0L;`

src/org/omegat/core/machinetranslators/YandexCloudTranslate.java
```diff
@@ -0,0 +1,245 @@
...
+            Map<String, String> translationNode = (Map<String, String>) translationsList.get(0);
+            return translationNode.get("text");
+        } catch (Exception e) {
+            Log.logErrorRB(e, "MT_JSON_ERROR");
+            return OStrings.getString("MT_ENGINE_YANDEX_CLOUD_BAD_TRANSLATE_RESPONSE");
+        }
+    }
+
+    @Override
+    public boolean isConfigurable() {
```
lc-nyovchev, line 215: nitpick:

Overridden public methods should appear in the class body before private methods. 

Basically something like this:

```
public class MyClass extends SomeBaseClass {

   public static String PUBLIC_CONSTANT = "this should be first";
   private static String PRIVATE_CONSTANT = "then private constants";
   private String someInstanceVariable = "then non static intance variables";
   
   // then constructors
   public MyClass() {
        super();
   }
   
   @Override
   public void overriddenMethod() {
       // then public overridden methods
   } 

   private String helperMethod() {
       // then private methods
   }
}
```

### Source:

- the Java Language specification: https://docs.oracle.com/javase/specs/jls/se8/html/jls-8.html
- some checkstyle plugins: https://checkstyle.sourceforge.io/apidocs/com/puppycrawl/tools/checkstyle/checks/coding/DeclarationOrderCheck.html, https://checkstyle.org/styleguides/sun-code-conventions-19990420/CodeConventions.doc2.html#a1852
- SO: https://stackoverflow.com/questions/4668218/are-there-any-java-method-ordering-conventions

src/org/omegat/core/machinetranslators/YandexCloudTranslate.java
```diff
@@ -0,0 +1,245 @@
...
+            return OStrings.getString("MT_ENGINE_YANDEX_CLOUD_FOLDER_ID_NOT_FOUND");
+        }
+
+        String AIMToken = getAIMToken(oAuthToken);
+        if (AIMToken == null) {
+            return AIMErrorMessage;
+        }
+
+        StringBuilder requestBuilder = new StringBuilder();
+        requestBuilder.append("{");
```
amake, line 142: You can combine a lot of these calls in a "fluent" fashion. It might help readability to group related pieces together.

src/org/omegat/core/machinetranslators/YandexCloudTranslate.java
```diff
@@ -0,0 +1,245 @@
...
+    protected static final String PROPERTY_OATH_TOKEN = "yandex.cloud.oauth-token";
+    protected static final String PROPERTY_FOLDER_ID = "yandex.cloud.folder-id";
+
+    protected static final int MAX_TEXT_LENGTH = 10000;
+    protected static final int AIM_TOKEN_TTL_SECONDS = 3600; // Recommended value
+
+    protected static final String IAM_TOKEN_URL = "https://iam.api.cloud.yandex.net/iam/v1/tokens";
+    protected static final String TRANSLATE_URL = "https://translate.api.cloud.yandex.net/translate/v2/translate";
+
+    private String AIMErrorMessage = null;
```
amake, line 69: Dumb question: Is AIM different from IAM? Because I see both. And I see that `iam` occurs in strings that must be correct in order for this connector to function, but `AIM` only occurs in identifiers, so I am curious if it's not a typo.

chelobaka: The question is not dumb at all and it's really a typo.

src/org/omegat/core/machinetranslators/YandexCloudTranslate.java
chelobaka: I made this class using `IBMWatsonTranslate` as a base and had the same doubts. Changing to private.

chelobaka:
Thank you for your feedback! I fixed all issues you pointed to, found another typo and a blind copy/paste bug.

amake:
Thanks! I've committed this to SourceForge.

PR #44, opened 2020-06-12
Implement glossary support for Yandex.Cloud MT

chelobaka:
This feature required some internal changes which were discussed in relevant ticket https://sourceforge.net/p/omegat/feature-requests/1494/


src/org/omegat/core/machinetranslators/YandexCloudTranslate.java
```diff
@@ -162,7 +196,7 @@ protected String translate(Language sLang, Language tLang, String text) {
         String tr = extractTranslation(response);
 
         if (tr == null) {
-            return "";
+            return EMPTY_STRING;
```
amake, line 199: I don’t think we really need a constant for the empty string. 

Or at least give it some semantics: “no translation”, “default glossary payload” etc.

src/org/omegat/core/machinetranslators/YandexCloudTranslate.java
```diff
@@ -230,4 +264,39 @@ private String getIAMToken(final String oAuthToken) {
...
         return cachedIAMToken;
     }
+
+    private String getGlossaryConfigPart(SourceTextEntry ste) {
+
+        if (!Preferences.isPreference(PROPERTY_USE_GLOSSARY)) {
+            return EMPTY_STRING;
+        }
+
+        List<GlossaryEntry> glossaryHits = Core.getGlossaryManager().searchSourceMatches(ste);
```
amake, line 274: I don’t really like MT providers being coupled to the core. Avoiding this is why I suggested the entry point should just accept a list of glossary terms instead of the `SourceTextEntry`.

src/org/omegat/core/machinetranslators/YandexCloudTranslate.java
```diff
@@ -230,4 +264,39 @@ private String getIAMToken(final String oAuthToken) {
...
+
+        if (glossaryHits.size() == 0) {
```
amake, line 276: Please use `isEmpty()`.

src/org/omegat/gui/exttrans/IMachineTranslation.java
```diff
@@ -65,7 +66,7 @@ default void setEnabled(boolean enabled) {
      *            text for translation
      * @return translated text, or null if translation impossible
      */
-    String getTranslation(Language sLang, Language tLang, String text) throws Exception;
```
amake, line 68: We need to keep the old method in the interface as well, since plugins might not be using `BaseTranslate`.

src/org/omegat/core/machinetranslators/YandexCloudTranslate.java
chelobaka: Wouldn't passing a `GlossaryManager` be a compromise choice here? Passing glossary terms would take a search on each MT call and most of MT engines don't use those terms.

chelobaka:
I moved alternative method into the interface and extended the interface with another method to avoid invoking glossary search for every MT provider.

src/org/omegat/gui/exttrans/MachineTranslateTextArea.java
```diff
@@ -193,7 +195,12 @@ private String getTranslation(Language source, Language target) throws Exception
                     }
                 }
             }
-            return translator.getTranslation(source, target, src);
+            if (translator.isUsingGlossary()) {
+                List<GlossaryEntry> glossaryTerms = Core.getGlossaryManager().searchSourceMatches(currentlyProcessedEntry);
```
amake, line 199: Thanks, this is better wrt coupling.

This will recompute glossary terms for every provider that wants them. (Right now there’s only one, but that may change.) Instead it should only be computed once. 

But is it expensive? If not then I don’t see a need to muddy the provider API with `isUsingGlossary`: just compute it once and provide it to all; they can ignore if so desired.

src/org/omegat/core/machinetranslators/BaseTranslate.java
```diff
@@ -28,13 +28,16 @@
...
+import java.util.List;
 import java.util.Map;
 import java.util.regex.Matcher;
 
 import javax.swing.JCheckBoxMenuItem;
 
 import org.omegat.core.Core;
+import org.omegat.core.data.SourceTextEntry;
 import org.omegat.gui.exttrans.IMachineTranslation;
+import org.omegat.gui.glossary.GlossaryEntry;
```
amake, line 40: Once the API changes have settled down, please clean the imports here.

chelobaka:
I hope I did it right this time.

src/org/omegat/gui/exttrans/MachineTranslateTextArea.java
```diff
@@ -127,9 +129,10 @@ private void startSearchThread(SourceTextEntry newEntry, boolean force) {
         UIThreadsUtil.mustBeSwingThread();
 
         clear();
+        List<GlossaryEntry> glossaryTerms = Core.getGlossaryManager().searchSourceMatches(newEntry);
```
amake, line 132: Thanks for your patience on this. This is definitely what I asked for... but now I'm really curious about how expensive this call is, since it's happening on the UI thread and I don't see a nice way to move it off without duplicating work or doing some complex synchronization.

If you have time, could you time it e.g. with `System.currentTimeMillis()` and make sure it's not crazy?

chelobaka: Tested on a project with 1386 glossary terms. First run was 92 ms, all subsequent runs were 1-3 ms. CPU is Core i5-6200U@2.3GHz.

Also I noticed that `startSearchThread` is called twice for each segment: first time when I switch to a new segment, second time when I press Ctrl+M (automatic MT translation is disabled).

chelobaka:
Fixed useless glossary searches on segment switching.

amake:
Actually I think we're trying to avoid the wrong thing here. It's my fault because I just saw the glossary lookup happening in the loop and wanted it out of there without looking at the rest of the search logic.

Here's the important stuff to my mind now:

1. The glossary lookup is fast enough that it's not a big deal to repeat
2. There are lots of scenarios where we don't even need to do it
3. It's probably better to prioritize getting it off the UI thread rather than making sure it happens just once

I think we should move it into `FindThread.getTranslation`, at the bottom of the `if` when we know we need it.

amake:
Also I should add that only supplying glossary entries when `force` is true is just wrong: if "Automatically fetch translations" is on and "Untranslated segments only" is off, then we would do the query and cache the result without glossary entries. The user wouldn't know to force-reload it.

chelobaka:
Agreed.

amake:
Thanks for your patience. I've squashed and merged to master.

PR #46, opened 2020-06-24
Queue scripts executed during events

briacp:
This PR should fix [bug 860](https://sourceforge.net/p/omegat/bugs/860/).
When executing scripts during events, the scripts are sorted by filename, the corresponding SwingWorkers are created and queued.

The elements of the queue are then executed and removed from the queue when done. The state of the SwingWorker is monitored with the PropertyChangeListener to check the `state` property is DONE.

As a side-effect, this fix also removes the unwanted `canceled (0ms)` that were appearing when several scripts were launched in rapid succession.

### DIFF
--- a/src/org/omegat/gui/scripting/ScriptingWindow.java
+++ b/src/org/omegat/gui/scripting/ScriptingWindow.java
@@ -40,13 +40,18 @@ import java.awt.event.MouseEvent;
 import java.awt.event.MouseMotionAdapter;
 import java.awt.event.WindowAdapter;
 import java.awt.event.WindowEvent;
+import java.beans.PropertyChangeEvent;
+import java.beans.PropertyChangeListener;
 import java.io.File;
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.LinkedList;
 import java.util.Locale;
 import java.util.Map;
+import java.util.Queue;
 import java.util.concurrent.CancellationException;
 import java.util.concurrent.ExecutionException;
 import java.util.logging.Level;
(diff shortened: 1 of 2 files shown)


briacp:
implemented in [092f4f1]

src/org/omegat/gui/scripting/ScriptingWindow.java
```diff
@@ -567,17 +572,23 @@ protected void done() {
...
     }
 
     public void executeScriptFile(ScriptItem scriptItem, Map<String, Object> additionalBindings) {
+    	cancelScriptQueue();
         try {
             String scriptString = scriptItem.getText();
-            executeScript(scriptString, scriptItem, additionalBindings);
+            ScriptWorker worker = createScriptWorker(scriptString, scriptItem, additionalBindings);
+            currentScriptWorker = worker;
+            worker.execute();
```
amake, line 591: Why not continue to call `executeScript(String, ScriptItem)` here instead of duplicating its contents?

src/org/omegat/gui/scripting/ScriptingWindow.java
```diff
@@ -586,8 +597,56 @@ public void executeScriptFile(ScriptItem scriptItem, Map<String, Object> additio
...
+                // TODO: Do we really want to handle the exception here, like this?
+                // This method can be called in instances when the Scripting Window
+                // is not visible, so it might make more sense to let the caller
+                // handle the exception.
+                logResult(StringUtil.format(OStrings.getString("SCW_SCRIPT_LOAD_ERROR"), scriptItem.getFile()), e);
+            };
+        }
+
+        executeScriptWorkers();
+    }
```
amake, line 618: This looks like basically the same thing as `executeScriptFile(ScriptItem, Map)`; why not have that wrap its argument in a list and delegate to this `executeScriptFiles`?

amake:
Sorry I took so long to look at this PR.

In addition to the comments above, I would have pointed out several formatting issues (which I have already fixed in https://github.com/omegat-org/omegat/commit/e6e503df3096aa938373d52202f64b2778b5114b).

src/org/omegat/gui/scripting/ScriptingWindow.java
amake: Also `ArrayList<ScriptItem>` is needlessly specific; it can be `List<ScriptItem>`. Then for my previous suggestion you can wrap with `Arrays.asList`.

PR #47, opened 2020-08-15
XLIFF: set state attr of target

miurahr:
XLIFF filter modify `<target state="needs-translation">`  to`<target state="translated">` if  translation of sentence is done.

Here is a proposed patch on https://sourceforge.net/p/omegat/feature-requests/1506/


src/org/omegat/filters3/xml/DefaultXMLDialect.java
```diff
@@ -110,7 +110,9 @@ public void defineIntactTags(String[] tags) {
 
     /** The set of defined paragraph tags. */
     private MultiMap<String, String> translatableTagAttributes = new MultiMap<String, String>();
-
+    /** The set of defined translation status tags */
+    private MultiMap<String, String> translationStatusAttributes = new MultiMap<>();
```
amake, line 114: This doesn't appear to ever be populated? If it's not intended for `DefaultXMLDialect` to have a real implementation of this, then I don't really like it having these dummy maps.

src/org/omegat/filters3/xml/Handler.java
```diff
@@ -637,6 +637,15 @@ private void translateButDontFlash() throws TranslationException {
...
                 translation = src;
+            } else if (lead instanceof XMLTag) {
+                String tag = ((XMLTag) lead).getTag();
+                for (int i = 0; i < ((XMLTag) lead).getAttributes().size(); i++) {
+                    Attribute attr = ((XMLTag) lead).getAttributes().get(i);
+                    if (dialect.getTranslationStatusAttributes().containsPair(tag, attr.getName())
+                            && dialect.getTranslationStatusTransitions().containsKey(attr.getValue())) {
+                        attr.setValue(dialect.getTranslationStatusTransitions().get(attr.getValue()));
+                    }
+                }
```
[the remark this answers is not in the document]
amake: On second thought this entire thing could be moved into the Dialect as a method like `handleXMLTag(XMLTag tag)`:

- `DefaultXMLDialect` would just have an empty implementation (and no dummy maps)
- `XLIFFDialect` would have an implementation that does what you're doing here

miurahr:
> This kind of looks like it marks everything as "translated" even if it's not really translated according to OmegaT's concept of (un)translated-ness. If that's the case, I think we need to change that.

Yes and no. 

- Yes; view from OmegaT

- No;  view from receiver of XLIFF file who may be a manager or customer, and use "state"  to handle workflow.
 "state" field may be an important place to see.

I'd like to see the specification. There are several pre-defined values.

Value | Description
-- | --
final | Indicates the terminating state.
needs-adaptation | Indicates only non-textual information needs adaptation.
needs-l10n | Indicates both text and non-textual information needs adaptation.
needs-review-adaptation | Indicates only non-textual information needs review.
needs-review-l10n | Indicates both text and non-textual information needs review.
needs-review-translation | Indicates that only the text of the item needs to be reviewed.
needs-translation | Indicates that the item needs to be translated.
new | Indicates that the item is new. For example, translation units that were not in a previous version of the document.
signed-off | Indicates that changes are reviewed and approved.
translated | Indicates that the item has been translated.

In my stupid understand, we can transit "needs-translation" to "translated" or "needs-review-translation".

translation" and untranslated(translation string is same as source) fields to be "translated", It is not able to separate for OmegaT.

We can add an option to select a behavior.

1.  transit "state" even when untranslated items.
2. select value whether "translated", "needs-review-translation"  from "needs-translation"

miurahr:
I'd like to take a way of 

> On second thought this entire thing could be moved into the Dialect as a method like `handleXMLTag(XMLTag tag)`

as a signature of `void handleXMLTag(XMLTag tag, boolean translated)`

Concrete method in XLIFFDialect can handle tag, and attributes according to boolean translated value.
Conceptually XLIFFDialect or other filters can revert state from "translated" to "needs-translation", or set another custom value such as "x-no-need-translation",  when boolean translated is false.

Because I don't know actual work-flows other than Crowdin importer behavior observation, it is just an idea.

miurahr:
I've added options to change behaviors. All the "needs-translation" state become "translated" in default.
User can select to keep it when untranslated, and change to "needs-review-translation" instead of "translated".

src/org/omegat/filters3/xml/xliff/EditXLIFFOptionsDialog.java
```diff
@@ -95,112 +99,128 @@ public int getReturnStatus() {
...
-        jPanel1.add(ignoreTypeForBptTagsCB);
-
-        getContentPane().add(jPanel1, java.awt.BorderLayout.CENTER);
-
-        pack();
-    }// </editor-fold>//GEN-END:initComponents
+        // <editor-fold defaultstate="collapsed" desc="Generated Code">//GEN-BEGIN:initComponents
+        private void initComponents() {
+
+                buttonGroup1 = new javax.swing.ButtonGroup();
```
amake, line 105: Do you have any idea why the indent changed so much here?

src/org/omegat/Bundle.properties
```diff
@@ -1410,6 +1410,9 @@ XLIFF_OPTIONS_36_ONLY=The following options do not apply if "Compatibility with
 XLIFF_OPTIONS_FORCE2F=&Force shortcut to "f" for <it pos="end"> tags
 XLIFF_OPTIONS_IGNORE4PH=&Ignore type and ctype attributes when building <ph> tag shortcuts
 XLIFF_OPTIONS_IGNORE4BPT=Ignore type and ctype attributes when building <&bpt> and <ept> tag shortcuts
+XLIFF_OPTIONS_STATE=Specify state attribute handling behavior
+XLIFF_OPTIONS_STATE_KEEP_UNTRANSLATED=Keep "meeds-translation" when untranslated
```
amake, line 1414:
```suggestion
XLIFF_OPTIONS_STATE_KEEP_UNTRANSLATED=Keep "needs-translation" when untranslated
```

src/org/omegat/filters3/xml/xliff/EditXLIFFOptionsDialog.java
```diff
@@ -214,7 +234,7 @@ private void okButtonActionPerformed(java.awt.event.ActionEvent evt) {// GEN-FIR
 
     private void cancelButtonActionPerformed(java.awt.event.ActionEvent evt) {// GEN-FIRST:event_cancelButtonActionPerformed
         doClose(RET_CANCEL);
-    }//GEN-LAST:event_cancelButtonActionPerformed
+    }
```
amake, line 237: This change looks wrong.

src/org/omegat/filters3/xml/xliff/XLIFFDialect.java
```diff
@@ -160,6 +166,47 @@ public Boolean validateContentBasedTag(String tag, Attributes atts) {
         return "mrk".equals(tag) && atts != null && "protected".equals(atts.getValueByName("mtype"));
     }
 
+    /**
+     * When translation is different from src, transit <target state="needs-translated"> attr
+     * to <target state="translated"> value.
+     * https://sourceforge.net/p/omegat/feature-requests/1506/
```
amake, line 172: I think for Javadoc purposes we need

- To escape `<` (and `>`?) as HTML entities: `&lt;` (and `&gt;`?)
- Put the link in a `@see <a href="...">...</a>`

src/org/omegat/filters3/xml/xliff/XLIFFDialect.java
```diff
@@ -160,6 +166,47 @@ public Boolean validateContentBasedTag(String tag, Attributes atts) {
...
+     * @param tag XML tag to be processed.
+     * @param translated is a value considered as translated?
```
amake, line 174:
```suggestion
     * @param translated is the value considered translated?
```

src/org/omegat/filters3/xml/xliff/XLIFFOptions.java
```diff
@@ -55,6 +55,8 @@
     private static final String OPTION_FORCE_SHORTCUT_2_F = "forceshortcut2f";
     private static final String OPTION_IGNORE_TYPE_4_PH_TAGS = "ignoretype4phtags";
     private static final String OPTION_IGNORE_TYPE_4_BPT_TAGS = "ignoretype4bpttags";
+    private static final String OPTION_STATE_KEEP_NEEDS = "statekeepneeds";
```
amake, line 58: I find `KEEP_NEEDS` to be a bit cryptic. Could you make it `KEEP_NEEDS_TRANSLATION`?

If that's too long for your taste, it could be `KEEP_NEEDS_T9N`

Please expand the method names to match.

src/org/omegat/filters3/xml/xliff/EditXLIFFOptionsDialog.java
```diff
@@ -7,6 +7,7 @@
                2007-2012 Didier Briel
                2013 Didier Briel, Piotr Kulik
                2014 Didier Briel, Aaron Madlon-Kay, Piotr Kulik
+	       2020 Hiroshi Miura
```
amake, line 10: Please fix this indent.

src/org/omegat/filters3/xml/xliff/EditXLIFFOptionsDialog.java
amake: If it's something to do with tabs vs spaces, could you set your editor to 4 spaces and try to reformat this file?

(You probably would want to use the NetBeans visual editor to make some edit, not do "reformat code" or whatever.)

amake (changes requested):
Thanks. It looks much nicer, and thanks for the bit from the XLIFF spec.

I'm a bit confused about one thing: Why would you *not* want to keep `needs-translation` when the segment is not translated? That seems like a useless option to me.

src/org/omegat/filters3/xml/xliff/EditXLIFFOptionsDialog.java
miurahr: I've used NetBeans 12.0  visual editor to edit the form.
 It may caused by miss configuration (no configuration) of NetBeans.
I'd like to change or flush netbeans configuration then open-save.

src/org/omegat/filters3/xml/xliff/EditXLIFFOptionsDialog.java
miurahr: I'm not sure why this removed because here is a part of auto generated code block....
I'd like to hand write here to revert.

src/org/omegat/Bundle.properties
```diff
@@ -1410,6 +1410,9 @@ XLIFF_OPTIONS_36_ONLY=The following options do not apply if "Compatibility with
 XLIFF_OPTIONS_FORCE2F=&Force shortcut to "f" for <it pos="end"> tags
 XLIFF_OPTIONS_IGNORE4PH=&Ignore type and ctype attributes when building <ph> tag shortcuts
 XLIFF_OPTIONS_IGNORE4BPT=Ignore type and ctype attributes when building <&bpt> and <ept> tag shortcuts
+XLIFF_OPTIONS_STATE=Specify state attribute handling behavior
+XLIFF_OPTIONS_STATE_KEEP_UNTRANSLATED=Keep "needs-translation" when untranslated
```
miurahr, line 1414: It is better to be consistent with a variable name of option class.

src/org/omegat/filters3/xml/Handler.java
```diff
@@ -637,8 +638,12 @@ private void translateButDontFlash() throws TranslationException {
             // untranslated is written out uncompressed
             if (compressed.equals(translation)) {
                 translation = src;
+                isTranslated = false;
             }
```
amake, line 642: OmegaT has a concept of "translated" (vs "untranslated") that is independent of whether `compressed.equals(translation)`: you can set "allow translation to be equal to source", in which case the segment can be "translated" even if `compressed.equals(translation) == true`.

But it looks like there is no way to determine the true translation status of the segment at this point, is that right? (Sorry, it's been a long time since I looked at this code.)

If that's the case then I guess this will have to do.

src/org/omegat/filters3/xml/xliff/EditXLIFFOptionsDialog.java
```diff
@@ -214,7 +233,7 @@ private void okButtonActionPerformed(java.awt.event.ActionEvent evt) {// GEN-FIR
 
     private void cancelButtonActionPerformed(java.awt.event.ActionEvent evt) {// GEN-FIRST:event_cancelButtonActionPerformed
         doClose(RET_CANCEL);
-    }//GEN-LAST:event_cancelButtonActionPerformed
+    }// GEN-LAST:event_cancelButtonActionPerformed
```
amake, line 236: NetBeans might have removed this comment initially because its opening sibling was accidentally formatted to have a space after the `//` like this.

I think if the spaces following the `//` are removed, NetBeans will correctly recognize this block as generated code.

src/org/omegat/filters3/xml/xliff/XLIFFDialect.java
```diff
@@ -160,6 +166,48 @@ public Boolean validateContentBasedTag(String tag, Attributes atts) {
...
+            if (tag.getTag().equals("target") && attr.getName().equals("state")) {
+                String current = attr.getValue();
+                if ("needs-translation".equals(current)) {
+                    if (!keepStateNeedsTranslation || translated) {
+                        if (changeStateToNeedsReviewTranslation) {
+                            attr.setValue("needs-review-translation");
+                        } else {
+                            attr.setValue("translated");
+                        }
+                    } else {
```
amake, line 190: I kind of liked before how you had a transition table defined as a Map. Was there a reason you switch to this style?

amake (commented):
Could you comment on this, please?

> I'm a bit confused about one thing: Why would you _not_ want to keep `needs-translation` when the segment is not translated? That seems like a useless option to me.

miurahr:
> I'm a bit confused about one thing: Why would you not want to keep needs-translation when the segment is not translated? That seems like a useless option to me.

It is why I don't have confidence whether all workflow scenario require to make all "needs-translation" segment "translated".

I'm confused in a case in my short experience with Crowdin as follows:

1. Download XLIFF export from crowdin.
2. Translate it with OmegaT and change all the segment state into "translated"
3. Upload XLIFF to crowdin.
4. Download XLIFF again from crowdin, to translate some segment which is missed to work(forget).
5. When open it with OmegaT, all the segment marked as "translated", then I need to search segment, not using "Next untranslated segment " key.

Keeping "needs-translation" is not so harm in many case except translator decide that some segment should keep as source.
Then I add an option for one who works partial translation such as casual volunteers for OSS and Wiki. 

Ok, I'd like to drop the option. If anyone report the issue in future, it will be a time to change.

src/org/omegat/filters3/xml/Handler.java
miurahr: I got it.
We should not check  `compressed.equals(translation)` because of useless, and there is no way to detect actual "untranslated" status on OmegaT.

amake: > We should not check `compressed.equals(translation)` because of useless

Since the filter is already treating `compressed.equals(translation)` as if it represented the actual status, I think it's OK to use it here as you are doing.

Making the actual status available at this level is an improvement that could be made in the future.

In other words I think this part is OK as it is; no need to change it.

src/org/omegat/filters3/xml/xliff/XLIFFDialect.java
miurahr: I've rewrite it with Map.

src/org/omegat/Bundle.properties
```diff
@@ -1410,6 +1410,9 @@ XLIFF_OPTIONS_36_ONLY=The following options do not apply if "Compatibility with
 XLIFF_OPTIONS_FORCE2F=&Force shortcut to "f" for <it pos="end"> tags
 XLIFF_OPTIONS_IGNORE4PH=&Ignore type and ctype attributes when building <ph> tag shortcuts
 XLIFF_OPTIONS_IGNORE4BPT=Ignore type and ctype attributes when building <&bpt> and <ept> tag shortcuts
+XLIFF_OPTIONS_STATE=Specify state attribute handling behavior
+XLIFF_OPTION_KEEP_TARGET_STATE_NEEDS_T9N=Keep "needs-translation" when untranslated
```
miurahr, line 1414: No need this now.

src/org/omegat/filters3/xml/xliff/XLIFFDialect.java
```diff
@@ -160,6 +177,25 @@ public Boolean validateContentBasedTag(String tag, Attributes atts) {
         return "mrk".equals(tag) && atts != null && "protected".equals(atts.getValueByName("mtype"));
     }
 
+    /**
+     * Handle &lt;target state="needs-translated"&gt; attribute according to
+     * option settings. In default, change to &lt;target state="translated"&gt;.
+     * It also handle it when source has state="new" to "translated" or "needs-translation".
```
miurahr, line 183: Comment is outdated.

miurahr (commented):
It needs more changes.

miurahr:
I believe I've adjusted all the review comments.

amake (approved):

amake:
Thanks for your patience on this.

Unfortunately I led you down the wrong path with some of my comments. I apologize.

- I think we *should* use the (un)translated flag, even though it's flawed
- The map looked cleaner, but in the end I was able to simplify the logic to something that looks similar

Please check my changes; if they look OK to you then I will squash and merge.

miurahr:
**It looks good and ok**!  Thanks.

amake:
Committed in https://github.com/omegat-org/omegat/commit/307fb82b4e111fef8b2e7b43c7201694183b10a0

Thank you for your contribution!

PR #48, opened 2020-08-16
Refactoring statistics functionality for better expansibility

bvfalcon:
Refactoring in statistics: calculation separated from presentation.
Calculation statistics method returns Bean with results. Bean contains
methods for presentation. This approach is very helpful for
expansibility with plugins.

Moved from SourceForge merge request: https://sourceforge.net/p/omegat/code/merge-requests/11/


amake:
Thanks for moving the discussion to GitHub.

From discussion on SourceForge:

>> 2. I see you have introduced a new result object, `StatsResult`. However I feel that its role is basically the same as `StatisticsInfo`. Is there a reason you chose to add a new class instead of expanding `StatisticsInfo`? If there is not a strong reason (or if the reason was "to make the changes to existing code as minimal as possible") then I think we should consider consolidating into `StatisticsInfo`.
> 2. Two reasons
>     1) StatisticsInfo is now used as field iof RealProject instance. StatsResult instance is more heavy than StatisticsInfo instance.
>     2) RealProject changed property numberOfTranslatedSegments of StatisticsInfo (line 1503). StatsResult has only read possibilities (getters)

OK, but then I'd like to keep the old role of `StatisticsInfo`: it shouldn't be recreated, but should be updated in an "outvar" style.

I've pushed this change.

> 3. Yes, in aspect of usage in `CalcStandardStatistics.buildProjectStats` this is absolutely equivalent (see  and )

OK I think you're right. I added an assertion just in case.

> 4. Yes, I use LF line endings. Also I do in last commit spaces instead of tabs.

`StatsResult.java` still had CRs. I removed them.

amake:
If my changes look OK to you then I will squash and apply to master on SourceForge.

bvfalcon:
Thank you Aaron. Yes, all right.

amake:
Applied as https://github.com/omegat-org/omegat/commit/d92abb9d1a240cc660209e7dd59ec936a8b80009

Thanks for your contribution!

PR #51, opened 2020-08-23
ProjectFilesList dialog respect LAF for color selection

miurahr:
use LAF's UIManager for colors selection  of project files list dialog.

https://sourceforge.net/p/omegat/bugs/1016/


miurahr:
> Unfortunately this is a regression on Mac: the alternating highlights are missing.

It is because AquaLAF defines both 'TextArea.inactiveBackground' and  'TextArea.background' as 'white'. which is as same as WindowsLAF.

https://github.com/openjdk/jdk/blob/efba976cfbb97655846e1f04ee70fc4e33910454/src/java.desktop/macosx/classes/com/apple/laf/AquaLookAndFeel.java#L376-L377

IMHO, we can implement our own custom LAFs , or override LAF value , instead of fixed color values.

amake:
I'm not advocating for sticking with a fixed color value. I'm OK with whatever solution fixes the problem *without causing regressions*.

However, making our own LAF sounds like a major departure from our current situation; do I understand correctly that

- Currently we use the OS's default LAF, so OmegaT (mostly) looks like a native app on each platform
- Making our own LAF would abandon looking like a native app, and would instead create a cross-platform-consistent look (that is *not* like the OS default appearance)

If so, then that is a very big departure from how OmegaT has looked until now. I think we should be very cautious about pursuing such a solution.

miurahr:
Could you see PoC PR https://github.com/omegat-org/omegat/pull/53 what we can do.

IMO, it is worth doing

- Change  constant value into properties
- Override a part of standard LAF values

We can continue discussion around introduction of new LAF. PoC illustrate it can do and also keep backword UI compatibility.

miurahr:
Here is a difficulty when override a color of System LAF.
When Mac and Windows, System LAF in Java always light theme. So[ overriding a same value with original source ](https://github.com/omegat-org/omegat/pull/53/commits/8a3f0bf1f5d55f0805f574fa7962301d2c21e5b6)works fine.

But on GTK LAF on dark mode on Mint Linux,  project files selection talble become light (because of override light values).

We may need to determine whether system LAF is light mode or dark mode, to decide which color should be used.
Currently I don't know how to do it.

If we add an **option** to select another LAF(such as darcula) by configuration, users who use OmegaT with GTK LAF can change it to dark theme of OmegaT instead of system LAF, then problem may solved.
Users on Windows and Mac still can use System LAF as default.

miurahr:
I've added [dark mode detection](https://github.com/omegat-org/omegat/pull/53/commits/8650ce64e2cc0f0f428a24b152207d07b4e23909) on PoC code and switch application colors according to system theme which is dark or not.
It seems work fine.

miurahr:
I've pushed a hack for  regression on Mac and Windows.

miurahr:
I found it is not worked well. I'd like to rethink a way.

miurahr:
Could you please review changes, @amake ?

 There are several points which we should see;

- Borrows isDarkTheme() static method from NetBeans(Apache-2.0 license), which is quite simple, and is difficult to be implemented differently.
- Manage immediate color number and Color.Constants of table in UIManager hash table to provide application specific colors and manage LAF difference in one place, LookAndFeelHelper class.
- Use Color.darker() and Color.brighter() standard methos, which use a factor 0.7. It is different from current factor 0.96.(0xffffff vs. 0xf5f5f5)

src/org/omegat/util/gui/LookAndFeelHelper.java
```diff
@@ -0,0 +1,113 @@
...
+     * Heuristic detection of dark theme.
+     * <p>
+     *     isDarkTheme method derived from NetBeans licensed by Apache-2.0
+     *     and re-licensed into GPL3.
+     * @return true when dark theme, otherwise false.
+     */
+    public static boolean isDarkTheme() {
+        if(isDark != null) {
+            return isDark;
+        }
```
amake, line 51: I don't see a need for this to be a public API.

Then, since it's private and we only call it once, there's no need to cache the result.

```suggestion
    private static boolean isDarkTheme() {
```

src/org/omegat/util/gui/LookAndFeelHelper.java
```diff
@@ -0,0 +1,113 @@
...
+public class LookAndFeelHelper {
+
+    /** is LaF theme dark? */
+    private static Boolean isDark = null;
+
+    /**
+     * Heuristic detection of dark theme.
+     * <p>
+     *     isDarkTheme method derived from NetBeans licensed by Apache-2.0
+     *     and re-licensed into GPL3.
```
amake, line 45: I don't think we can just relicense Apache-2.0 code. It's enough to say that this particular method is under Apache-2.0. If you can link to the original source that would be nice.

```suggestion
```

src/org/omegat/util/gui/LookAndFeelHelper.java
```diff
@@ -0,0 +1,113 @@
...
+import org.omegat.util.Log;
+
+/**
+ * Helper to initialize LaF.
+ */
+public class LookAndFeelHelper {
+
+    /** is LaF theme dark? */
+    private static Boolean isDark = null;
+
```
amake, line 40: Not needed per other comment.

```suggestion
```

src/org/omegat/util/gui/LookAndFeelHelper.java
```diff
@@ -0,0 +1,113 @@
...
+                foreground.getGreen(),
+                foreground.getBlue(),
+                null)[2];
+        float background_brightness = Color.RGBtoHSB(
+                background.getRed(),
+                background.getGreen(),
+                background.getBlue(),
+                null)[2];
+        isDark = background_brightness < foreground_brightness;
+        return isDark;
```
amake, line 85: Per comments above.

```suggestion
        return background_brightness < foreground_brightness;
```

src/org/omegat/util/gui/LookAndFeelHelper.java
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    /**
+     * Set LookAndFeel default and load application colors
+     */
+    public static void setDefaultLaf() {
+        // Workaround for JDK bug 6389282 (OmegaT bug bug 1555809)
+        // it should be called before setLookAndFeel() for GTK LookandFeel
+        // Contributed by Masaki Katakai (SF: katakai)
+        UIManager.getInstalledLookAndFeels();
```
amake, line 95: I think we don't need the bug workaround anymore: The JDK bug has been fixed since Java 7.
https://bugs.java.com/bugdatabase/view_bug.do?bug_id=6389282

So we can remove the try/catch and the get/set.

Incidentally the original OmegaT bug was https://sourceforge.net/p/omegat/bugs/215/

src/org/omegat/util/gui/LookAndFeelHelper.java
```diff
@@ -0,0 +1,113 @@
...
+                        UIManager.getColor("TextArea.background").brighter());  // NOI18N
+            } else {
+                UIManager.put("OmegaT.AlternatingHilite",
+                        UIManager.getColor("TextArea.background").darker());  // NOI18N
+            }
+            UIManager.put("OmegaT.SpecialForeground", Color.BLACK);
+            UIManager.put("OmegaT.SpecialBackground", new Color(0xC8DDF2));
+        } catch (Exception e) {
+            // do nothing
+            Log.logErrorRB("MAIN_ERROR_CANT_INIT_OSLF");
```
amake, line 110: We won't need the string for this in `Bundle.properties` anymore either.

amake (changes requested):
The overall approach looks good. Thanks for doing the legwork on this.

Are you aware of [DockingUI.java](https://github.com/omegat-org/omegat/blob/master/src/org/omegat/util/gui/DockingUI.java)? It seems like some sort of consolidation is in orderーit would be nice to have all custom LAF definitions in one place.

src/org/omegat/util/gui/LookAndFeelHelper.java
miurahr: I see, I'd like to drop it.

src/org/omegat/util/gui/LookAndFeelHelper.java
miurahr: removed

miurahr:
I've tried to consolidate into DockingUI about isDarkTheme() and setDefaultColors().
Both methods are not only for DockingUI but for overall application LaF. So class name DockingUI might be changed in future, when more and more generic LaF methods are added.

miurahr:
I'd like change DockingUI class as same as #54

amake:
Code-wise, this LGTM.

Unfortunately the new Mac appearance is not acceptable:

Before
Screen Shot 2020-09-01 at 21 17 10

After
Screen Shot 2020-09-01 at 21 17 58

amake:
I've pushed a tweak to give results much closer to the previous color.

Left: current master
Right: this PR

# macOS 10.15.6
Screen Shot 2020-09-01 at 21 17 10 Screen Shot 2020-09-01 at 21 51 38
# Windows 10
Screen Shot 2020-09-01 at 21 44 45 Screen Shot 2020-09-01 at 21 40 12

# Ubuntu 20.04 Light
Screen Shot 2020-09-01 at 21 45 45 Screen Shot 2020-09-01 at 21 37 01

# Ubuntu 20.04 Dark
Screen Shot 2020-09-01 at 21 46 24 Screen Shot 2020-09-01 at 21 34 12

amake:
There is a weird regression in Ubuntu where the highlight (orange) doesn't get applied to the Filename column. Do you have any idea why that is?

If my changes look OK to you then I will squash and apply this to master.

amake (approved):

miurahr:
> There is a weird regression in Ubuntu where the highlight (orange) doesn't get applied to the Filename column. Do you have any idea why that is?

This is not a regression,  but it is also able to be reproduced without the change.

Mint Linux 20 in light mode (v5.3.0beta, 2ee9bd0d3)
Mint-Linux-20-light-filelist2ee9bd0d3

When using 3rd party LAF library such as darcula, this problem is gone. It may be a GTKLookAndFeel or GTK provisioning in Ubuntu/Mint issue.

amake:
> If my changes look OK to you then I will squash and apply this to master.

Am I OK to merge this?

miurahr:
> If my changes look OK to you then I will squash and apply this to master.

Yes, it is OK for merge! Thanks.

amake:
I've squashed and committed in https://github.com/omegat-org/omegat/commit/12c087ce9bf054acee05e806a1bd273a1695bf8d. Thanks for your contribution.

PR #52, opened 2020-08-23
Added concurrency in statistics calculation

bvfalcon:


bvfalcon:
Where are some bugs appears with this commit. Cancelled

PR #53, opened 2020-08-24
WIP: PoC: switch system LAF, and custom  one according to property, override system value with properties.

miurahr:
This is PoC topic branch to indicate what can do. 
Things doing here are

- Override system LAF value using UIManager and properties
- Add patches related to LAF and colors, #1013 and #1016
- Move fixed color constant to properties.
- Have a mechanism to switch properties group according to LAF configuration.

Things we should consider but not doing yet are

- Determine LAF name in the code and switch App colors group, e.g.  Aqua, Windows and GTK
- Determine whether LAF mode is dark or light, and switch App color.


miurahr:
PoC has been end of its role.

PR #54, opened 2020-08-24
GUI: Externalize Color schemes to properties

miurahr:
- All immediate color values into properties.
- Introduce light and dark color schemes. When OS is dark theme, omegat uses dark color set.
- Implement https://sourceforge.net/p/omegat/feature-requests/1514/ feature.


src/org/omegat/util/gui/ResourcesUtil.java
```diff
@@ -64,11 +66,26 @@ public static Image getImage(final String resourceName) {
...
     }
+
+    /**
+     * Load default color scheme from classpath.
+     */
+    public static Properties getBundleColorProperties(final String style) {
+        String resourcePath = "/org/omegat/" + "ColorScheme_" + style + ".properties";
+        Properties properties = new Properties();
+        try {
+            InputStream is = ResourcesUtil.class.getResourceAsStream(resourcePath);
```
amake, line 84: This PR will need some reconciliation with some of the other work you're doing, but I just wanted to point out this for now:

This should use [try-with-resources](https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html).

src/org/omegat/Main.java
```diff
@@ -251,6 +254,24 @@ private static void applyConfigFile(String path) {
...
+            int colorCode = Integer.parseInt(v.toString(), 16);
+            UIManager.put(k.toString(), new Color(colorCode));
+        });
+    }
+
+    /**
+     * Huristic detection of dark theme.
+     * @return true is dark theme, otehrwise it seems light theme.
+     */
+    private static boolean isDarkTheme() {
```
miurahr, line 270: This should be as same as #51

src/org/omegat/util/gui/Styles.java
```diff
@@ -59,55 +59,55 @@ private Styles() {
...
+        COLOR_MARK_COMES_FROM_TM(UIManager.getColor("OmegaT.MarkComesFromTM")),
+        COLOR_MARK_COMES_FROM_TM_XICE(UIManager.getColor("OmegaT.MarkComesFromTMXice")),
+        COLOR_MARK_COMES_FROM_TM_X100PC(UIManager.getColor("OmegaT.MarkComesFromTMX100PC")),
+        COLOR_MARK_COMES_FROM_TM_XAUTO(UIManager.getColor("OmegaT.MarkComesFromTMXAuto")),
+        COLOR_MARK_COMES_FROM_TM_XENFORCED(UIManager.getColor("OmegaT.MarkComesFromTMXenForced")),
+        COLOR_REPLACE(UIManager.getColor("OmegaT.Replace")),
+        COLOR_LANGUAGE_TOOLS(UIManager.getColor("OmegaT.LanguageTools")),
+        COLOR_TRANSTIPS(UIManager.getColor("OmegaT.TransTips")),
+        COLOR_SPELLCHECK(UIManager.getColor("OmegaT.SpellCheck")),
         COLOR_TERMINOLOGY(Color.ORANGE),
```
miurahr, line 95: This also be able to be externalized.

miurahr:
Can I place an initialization function which load colors scheme properties onto `org.omegat.core.Core` class?
It seems natural to load colors when calling `Core.initializeGUI` in `Main.runGUI()`.

```
        System.setProperty("swing.aatext", "true");

        try {
            Core.initializeGUI(PARAMS);
        } catch (Throwable ex) {
            Log.log(ex);
            showError(ex);
            return 1;
        }
```

How about add the call here.

```
     public static void initializeGUI(final Map<String, String> params) throws Exception {
         cmdLineParams = params;
 
         // 1. Initialize project
         currentProject = new NotLoadedProject();
 
+        // 2. Load default color scheme
+        DockingUI.loadDefaultColors();

-        // 2. Initialize application frame
+        // 3. Initialize application frame
         MainWindow me = new MainWindow();
         mainWindow = me;
```

amake:
Why does the call need to go there? (I'm not arguing against it; I just want to understand.) Is it because you need to initialize the colors before creating the `MainWindow`?

My only real problem is that it looks strange to call `DockingUI` there, just because the name doesn't fit. Do you want to think about the refactoring of `DockingUI` you described [here](https://github.com/omegat-org/omegat/pull/51#issuecomment-683252906) for this PR?

miurahr:
> Why does the call need to go there? (I'm not arguing against it; I just want to understand.) Is it because you need to initialize the colors before creating the MainWindow?

In my understanding,  procedures in Main.runGUI before calling initilizeGUI are all  tweaks for platform specific issues such as OSX integration, X11 application name, and (I've dropped)  JDK bug against LAF.

It is because loading Colors definition seems a part of initialization of GUI,  not tweaks, that I propose it.

I'm also feel strange a class name DockingUI when place this function. OK, I'd like to consider better class name.

miurahr:
I've changes name org.omegat.util.gui.DockingUI to org.omegat.util.gui.UIDesignManager.
I put loading of default colors in initialize() , and it is called in MainWindow constructor.

src/org/omegat/ColorScheme_dark.properties
```diff
@@ -0,0 +1,38 @@
+# OmegaT application default colors scheme
+# This is a table of default colors definitions when no customization.
+# Here is a definition for dark theme, such as GTK LAF(dark), OSX Mojava and 3rd party dark LAFs.
+OmegaT.ActiveSource=28a028
```
amake, line 4: Could you normalize the color names to be `CamelCase.lowerCamelCase` please?

src/org/omegat/util/gui/ResourcesUtil.java
```diff
@@ -41,6 +43,7 @@
     private ResourcesUtil() {
     }
 
+    private static final String APPROOT = "/org/omegat/";
```
amake, line 46: Could this be `APP_ROOT` please?

src/org/omegat/util/gui/ResourcesUtil.java
```diff
@@ -64,11 +67,28 @@ public static Image getImage(final String resourceName) {
...
+     * @return Properties object loaded when succeeded, otherwise null
+     */
+    public static Properties getBundleColorProperties(final String style) {
+        String resourcePath = APPROOT + "ColorScheme_" + style + ".properties";
+        Properties properties = new Properties();
+        try (InputStream is = ResourcesUtil.class.getResourceAsStream(resourcePath)) {
+            properties.load(is);
+        } catch (Exception e) {
+            return null;
+        }
```
amake, line 91: The consumer of this (`UIDesignManager.loadDefaultColors`) blows up via `Objects.requireNonNull` anyway, so I would just make this method throw whatever exceptions it wants and handle them at a higher layer.

amake (changes requested):

src/org/omegat/ColorScheme_light.properties
```diff
@@ -0,0 +1,45 @@
...
+OmegaT.markComesFromTMXenForced=ffccff
+OmegaT.replace=0000ff
+OmegaT.languageTools=0000ff
+OmegaT.transTips=0000ff
+OmegaT.spellCheck=ff0000
+OmegaT.terminology=ff6600
+OmegaT.matchesChanged=0000ff
+OmegaT.matchesUnchanged=00ff00
+OmegaT.matchesInsActive=0000ff
+OmegaT.MatchesInsInactive=808080
```
miurahr, line 35: Need to fix.

miurahr (commented):
There is a backlog.
It can be externalize too.
```
UIManager.put("OmegaT.specialForeground", Color.BLACK);    
UIManager.put("OmegaT.specialBackground", new Color(0xC8DDF2));
```

src/org/omegat/ColorScheme_dark.properties
miurahr: done changing names.

miurahr:
I believe all the adjustments done.

src/org/omegat/ColorScheme_dark.properties
```diff
@@ -0,0 +1,40 @@
...
+OmegaT.removeTextTarget=ff0000
+OmegaT.NBSP=c8c8c8
+OmegaT.whiteSpace=808080
+OmegaT.bidiMarkers=c80000
+OmegaT.paragraphStart=aeaeae
+OmegaT.markComesFromTM=fa8072
+OmegaT.markComesFromTMXice=af76df
+OmegaT.markComesFromTMX100PC=ff9408
+OmegaT.markComesFromTMXAuto=ffd596
+OmegaT.markComesFromTMXenForced=ffccff
```
amake, line 20: These are a bit confusing because it looks like they are about TMX. Could we change them to:

```suggestion
OmegaT.markComesFromTmXice=af76df
OmegaT.markComesFromTmX100pc=ff9408
OmegaT.markComesFromTmXauto=ffd596
OmegaT.markComesFromTmXenforced=ffccff
```

Then for consistency `OmegaT.NBSP` should become `OmegaT.nbsp`.

src/org/omegat/ColorScheme_dark.properties
```diff
@@ -0,0 +1,40 @@
+# OmegaT application default colors scheme
+# This is a table of default colors definitions when no customization.
+# Here is a definition for dark theme, such as GTK LAF(dark), OSX Mojava and 3rd party dark LAFs.
```
amake, line 3: I think people understand the concept of a dark color scheme, so I would simplify the description and instead document the format of the expected values:

```suggestion
# OmegaT default dark color scheme
#
# Format is 8-bit RGB in hexadecimal (like standard HTML colors without the
# leading '#')
```

src/org/omegat/util/gui/UIDesignManager.java
```diff
@@ -470,20 +476,39 @@ private static boolean isDarkTheme() {
...
+                    int a = (int)(val & 0xFF);
+                    int b = (int)(val >> 8 & 0xFF);
+                    int g = (int)(val >> 16 & 0xFF);
+                    int r = (int)(val >> 24 & 0xFF);
+                    UIManager.put(k.toString(), new Color(r, g, b, a));  // hasAlpha
+                }
+            });
+        } catch (IOException e) {
+            LOGGER.warning("Cannot load color style defaults.");
+        }
```
amake, line 495: If we hit this catch because the file can't be found, then that's a programming/packaging error and we probably want to just crash.

src/org/omegat/util/gui/UIDesignManager.java
```diff
@@ -470,20 +476,39 @@ private static boolean isDarkTheme() {
...
+            ResourcesUtil.getBundleColorProperties(scheme).forEach((k, v) -> {
+                if (v.toString().length() <= 6) {
+                    UIManager.put(k.toString(), new Color(Integer.parseInt(v.toString(), 16)));  // int(rgb)
+                } else {
+                    long val = Long.parseLong(v.toString(), 16);
+                    int a = (int)(val & 0xFF);
+                    int b = (int)(val >> 8 & 0xFF);
+                    int g = (int)(val >> 16 & 0xFF);
+                    int r = (int)(val >> 24 & 0xFF);
+                    UIManager.put(k.toString(), new Color(r, g, b, a));  // hasAlpha
```
amake, line 490: I didn't see any colors with alpha in the .properties files you added. And the color picker UI in the Preferences window doesn't seem to let you set colors with alpha.

I would prefer to remove this code if we don't have a use case for it.

amake (changes requested):

src/org/omegat/util/gui/UIDesignManager.java
miurahr: Here is a case to use it, sorry I've missed it.
```
# Shadow and Stroke has alpha
OmegaT.shadow=80808080
OmegaT.stroke=EED20080
```

amake:
I made some tweaks. Please take a look; if it looks OK then I will squash and merge.

Things worth mentioning:

1. I think you accidentally swapped the light and dark colors for the Preferences search field. I fixed them and also removed one of the custom colors in favor of a default LAF color.
2. I added a leading `#` to the color properties. I think this will help people understand the format since it looks like colors in HTML and CSS. Also there are nice editor features that can show you the colors when they are formatted this way; here's my Emacs with [rainbow-mode](https://elpa.gnu.org/packages/rainbow-mode.html):
    image
    (Unfortunately it doesn't handle RGBA correctly)

miurahr:
Thanks and it looks fine!
rainbow-mode is very cool and helpful. I hope someone who is good color design help us making good defaults :+1:

PR #56, opened 2020-09-05
Fix background of dictionary on linux dark mode

miurahr:
Workaround and hack for GTKLookAndFeel bug that TextPane.background become always white even when dark mode.

ref: https://sourceforge.net/p/omegat/bugs/1013/

omegat-linux-darkmode-hack
Linux Mint 20 Dark theme

omegat-hack-background-light
Linux Mint 20 Light theme

### DIFF
--- a/src/org/omegat/util/gui/UIDesignManager.java
+++ b/src/org/omegat/util/gui/UIDesignManager.java
@@ -477,6 +477,12 @@ public final class UIDesignManager {
         Color hilite;
         if (isDarkTheme()) {
             hilite = UIManager.getColor("TextArea.background").brighter();  // NOI18N
+            // Hack for JDK GTKLookAndFeel bug.
+            // TextPane.background is always white but should be a text_background of GTK.
+            // List.background is as same color as text_background.
+            if (Platform.isLinux()) {
+                UIManager.put("TextPane.background", UIManager.getColor("List.background"));
+            }
         } else {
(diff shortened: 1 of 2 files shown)


src/org/omegat/util/gui/UIDesignManager.java
```diff
@@ -477,6 +477,12 @@ private static void setDefaultColors() {
         Color hilite;
         if (isDarkTheme()) {
             hilite = UIManager.getColor("TextArea.background").brighter();  // NOI18N
+            // Hack for JDK GTKLookAndFeel bug.
+            // TextPane.background is always white but should be a text_background of GTK.
+            // List.background is as same color as text_background.
+            if (Platform.isLinux()) {
```
amake, line 483: I'm basically OK with this, but I wonder: what about other LAFs? Would this cause a regression in a LAF that sets `TextPane.background` more appropriately?

Instead of checking `isLinux()`, what about a check to see if `TextPane.background` is white?

miurahr: > Instead of checking isLinux(), what about a check to see if TextPane.background is white?

Oh, it is better way to check.

amake:
Committed in https://github.com/omegat-org/omegat/commit/266f7ebaaa19800a85b8e37f60dae4c47b481ca6.

Thank you for your contribution!

PR #58, opened 2020-11-25
Remove Yandex.Translate MT engine

chelobaka:
The service was terminated in favor of Yandex.Cloud.


amake:
Thanks. I note this was requested in https://sourceforge.net/p/omegat/feature-requests/1522/

I reverted your changes to the localized properties files because the deletion from the base file should propagate naturally as the localizations are updated. Doing it now just risks extra churn.

I also found and removed some additional references to the old service.

PR #59, opened 2020-11-29
Support dsl.dz dictionary format

miurahr:
Implemnt [RFE#1528](https://sourceforge.net/p/omegat/feature-requests/1528/)

You can download wiktionary dictionaries which format is .dsl.dz
https://github.com/open-dsl-dict/wiktionary-dict

This enhancement support a .dsl.dz format.


amake:
I think you neglected to commit the new `test.dsl.dz` file.

src/org/omegat/core/dictionaries/LingvoDSL.java
```diff
@@ -61,48 +67,56 @@ public IDictionary loadDict(File file) throws Exception {
...
-            data.add(word.toString(), trans.toString());
-        }
-        data.done();
-        return data;
+        return new LingvoDSLDict(file, language);
     }
 
     static class LingvoDSLDict implements IDictionary {
         protected final DictionaryData<String> data;
+        private final String dataFile;
```
amake, line 75: This isn't used outside of the constructor; should we get rid of it?

src/org/omegat/core/dictionaries/LingvoDSL.java
```diff
@@ -61,48 +67,56 @@ public IDictionary loadDict(File file) throws Exception {
...
+
+        LingvoDSLDict(File file, Language language) throws Exception {
+            dataFile = file.getPath();
+            data = new DictionaryData<>(language);
+            if (dataFile.endsWith(".dz")) {
+                try (InputStreamReader reader = new InputStreamReader(new GZIPInputStream(new FileInputStream(file),
+                        8192), StandardCharsets.UTF_16)) {
+                    loadData((new BufferedReader(reader)).lines());
```
amake, line 83: `BufferedReader` could be included in the `try (...)` statement so that it gets auto-closed.

(The implementation on Java 8 just closes the reader that it wraps, so semantically there's no difference, but it feels nicer to close it.)

In case you put `BufferedReader` outside the `try (...)` because the line was getting too long, I should point out that you can declare multiple variables inside the `try` and they will all get closed ([details](https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html)):

```suggestion
                try (FileInputStream fis = new FileInputStream(file);
                        GZIPInputStream gis = new GZIPInputStream(fis, 8192);
                        InputStreamReader isr = new InputStreamReader(gis,  StandardCharsets.UTF_16);
                        BufferedReader reader = new BufferedReader(isr)) {
                    loadData(reader.lines());
```

src/org/omegat/core/dictionaries/LingvoDSL.java
```diff
@@ -61,48 +67,56 @@ public IDictionary loadDict(File file) throws Exception {
...
+                }
+            } else {
+                loadData(Files.lines(file.toPath(), StandardCharsets.UTF_16));
+            }
+        }
 
-        LingvoDSLDict(DictionaryData<String> data) throws Exception {
-            this.data = data;
+        private void loadData(Stream<String>stream) {
```
amake, line 90:
```suggestion
        private void loadData(Stream<String> stream) {
```

amake (commented):
Thanks, this looks good. Just some small comments.

amake:
Thanks! I squashed and merged to master.

PR #60, opened 2020-12-04
StarDict performance improvement by RandomAccessFile

oho-sugu:
Performance improvement for StarDict uncompressed file type (.dict file type).

How?

- Use RandomAccessFile
- Call constructor (and create file hundle) of RandomAccessFile only once.

Why?

- skip method of FileInputStream is slow, because it read all file data before start point.
- Call constructor each time is slow, because Making file hundle is heavy operation and maybe can't use OS's file cache.

### DIFF
--- a/src/org/omegat/core/dictionaries/StarDict.java
+++ b/src/org/omegat/core/dictionaries/StarDict.java
@@ -159,6 +159,14 @@ public class StarDict implements IDictionaryFactory {
                 throw new FileNotFoundException("No .dict.dz or .dict files were found for " + dictName);
             }
 
+            if (dictType == DictType.DICTFILE) {
+                try {
+                    dictFile = new RandomAccessFile(new File(dataFile), "r");
+                } catch (FileNotFoundException e) {
+                    throw new FileNotFoundException("No .dict files were found for " + dictName);
+                }
+            }
+
             try {
                 data = loadData(getFile(".idx.gz", ".idx").get());
             } catch (NoSuchElementException ex) {
@@ -199,6 +207,17 @@ public class StarDict implements IDictionaryFactory {
             is.close();
             newData.done();
             return newData;
+        }
+
+        @Override
+        public void dispose() {
+            if (dictFile != null) {
+                try {
+                    dictFile.close();
+                } catch (IOException e) {
+                    System.err.println(e.getMessage());
+                }
+            }
         }
 
         @Override
@@ -250,13 +269,6 @@ public class StarDict implements IDictionaryFactory {
          */
         private String readDictArticleText(int start, int len) {
             String result = null;
-            if(dictFile == null){
-                try {
-                    dictFile = new RandomAccessFile(new File(dataFile), "r");
-                } catch (FileNotFoundException e){
-                    System.err.println(e.getMessage());
-                }
-            }
             try {
                 byte[] data = new byte[len];
                 dictFile.seek(start);
--- a/src/org/omegat/core/dictionaries/StarDict.java
+++ b/src/org/omegat/core/dictionaries/StarDict.java
@@ -35,6 +35,7 @@ import java.io.FileInputStream;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.RandomAccessFile;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.util.List;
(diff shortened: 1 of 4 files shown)


src/org/omegat/core/dictionaries/StarDict.java
```diff
@@ -247,16 +250,17 @@ private synchronized String readArticle(int start, int len) {
...
-                long moved = in.skip(start);
-                if (moved < start) {
-                    long moved2 = in.skip(start - moved);
-                    if (moved2 < start - moved) {
-                        throw new IOException("Cannot seek dictionary.");
-                    }
+            if(dictFile == null){
+                try {
+                    dictFile = new RandomAccessFile(new File(dataFile), "r");
+                } catch (FileNotFoundException e){
```
amake, line 256: Please follow project whitespace conventions.

```suggestion
            if (dictFile == null) {
                try {
                    dictFile = new RandomAccessFile(new File(dataFile), "r");
                } catch (FileNotFoundException e) {
```

amake: However I am not sure about instantiating the file here. If a user has loaded a dictionary, I think it's OK to assume they want to use it. So I would unconditionally instantiate the `RandomAccessFile` elsewhere, like in the constructor.

amake (changes requested):
Thank you for your contribution. This does look like it would be a nice speedup on large files.

But I am a bit concerned about how we're supposed to clean up the `RandomAccessFile`. In [DictionariesManager#fileChanged](https://github.com/omegat-org/omegat/blob/431fe8f019ee4df1c27126a0f9764e7b7314fd9e/src/org/omegat/core/dictionaries/DictionariesManager.java#L114) you can see that it's possible to unload or reload a dictionary. We can't just leave the `RandomAccessFile` open or we will leak resources.

So I think if we're going to do this then we need a `dispose` method on `IDictionary` and to call it when removing the dictionary.

src/org/omegat/core/dictionaries/StarDict.java
oho-sugu: I think you're right.
I will move instantiate to constructor.
Thank you.

oho-sugu:
Thank you for response.

I'm sorry but I was afraid that any side-effect for other code, I'm not familiar on OmegaT whole source code.
Adding dispose method is nice idea, I think so too.
I have modified and recommited my change.

Please re-review my code.
Thanks.

src/org/omegat/core/dictionaries/StarDict.java
amake: Thanks

amake:
I realized that we actually use a `RandomAccessFile` in both the `DICTZIP` and `DICTFILE` code paths, so I made some adjustments.

I don't have StarDict files of both types to test with at the moment; could you double check that my changes are OK?

oho-sugu:
Some error occurs with zipped dict file (DICTZIP). 
I think this is because wrong treat of DictZipInputStream.
OK, I'll try to correct.

oho-sugu:
In constructor of DictZipInputStream, they read Zip Header for get some information.
Each word searching make constructing DictZipInputStream with same RandomAccessFile and lead some error ("Not in GZIP format").

After this modification (3cb26fd...), on my computer, Zipped and Notzipped stardict file works fine.

amake:
Thanks for testing and for your fix.

I was not happy with `StarDictDict` having both `RandomAccessFile dataFile` and `DictZipInputStream din` members when only one should actually be used. I think it revealed some big design flaws in the entire class, so I did a big refactoring. Please see if this works for you.

(Sorry for so many rounds of back-and-forth.)

oho-sugu:
I have tested and works fine.
Nice refactoring!

> (Sorry for so many rounds of back-and-forth.)
Please do not worry.
I was looking for CAT tool for my private translation project (EN->JP) and just found OmegaT.
I felt OmegaT is super powerfull tool for my goal, but when I use dictionary, it frequently freezed.
So I found source code in github and PRed.
If my contribution will be useful and help project, I'm happy.
And I learned many from this conversation too.
Thank you.

amake:
Thank you for your contribution.

I have squashed and merged to our main repository on SourceForge. The change will be synced to GitHub shortly.

For documentation purposes I have made a ticket on SourceForge: https://sourceforge.net/p/omegat/feature-requests/1532/

If you have a SourceForge account I will be happy to make you owner of that ticket. If not, no problem.

oho-sugu:
I have no SourceForge account, and I don't mind that.
Thank you.

PR #63, opened 2020-12-21
javadoc: fix argument desc of TeamTool constructor

miurahr:


amake:
Merged to master on SourceForge.

Thanks!

PR #65, opened 2021-01-02
Prefer xdg-open on Linux instead of java.awt.Desktop browse/open

chelobaka:
This PR aims to alleviate java.awt.Desktop shortcomings on Linux platform. OmegaT uses two methods of this class: `browse` and `open` so the wrapper implements only them. While `browse` is not supported at all in some DEs like KDE Plasma, the `open` method also has its problems. On my PC when I open a directory from OmegaT main menu (for example project root) Java starts VSCodium to open the directory. I wasn't frustrated enough to find where it's configured, but `xdg-open` handles directory paths correctly launching default file manager.

The code was tested only on Linux though it's simple enough to assume other platform should behave nicely.

### DIFF
--- a/src/org/omegat/util/gui/DesktopWrapper.java
+++ b/src/org/omegat/util/gui/DesktopWrapper.java
@@ -36,9 +36,8 @@ import java.net.URI;
 /**
  * Wrapper class for java.awt.Desktop with `xdg-open` path on Linux platform.
  */
-public class DesktopWrapper {
+public final class DesktopWrapper {
 
-    private final static Desktop awtDesktop = Desktop.getDesktop();
     private final static boolean useXDGOpen;
 
     static {
@@ -57,7 +56,7 @@ public class DesktopWrapper {
         if (useXDGOpen) {
             xdgOpen(uri.toString());
         } else {
-            awtDesktop.browse(uri);
+            Desktop.getDesktop().browse(uri);
         }
     }
 
@@ -65,7 +64,7 @@ public class DesktopWrapper {
         if (useXDGOpen) {
             xdgOpen(file.getPath());
         } else {
-            awtDesktop.open(file);
+            Desktop.getDesktop().open(file);
         }
     }
 
--- a/src/org/omegat/util/gui/DesktopWrapper.java
+++ b/src/org/omegat/util/gui/DesktopWrapper.java
@@ -1,0 +1,88 @@
+/**************************************************************************
+ OmegaT - Computer Assisted Translation (CAT) tool
+ with fuzzy matching, translation memory, keyword search,
+ glossaries, and translation leveraging into updated projects.
+
+ Copyright (C) 2021 Lev Abashkin
+ Home page: http://www.omegat.org/
+ Support center: https://omegat.org/support
+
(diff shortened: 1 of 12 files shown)


src/org/omegat/util/gui/DesktopWrapper.java
```diff
@@ -0,0 +1,88 @@
...
+
+import java.awt.Desktop;
+import java.io.File;
+import java.io.IOException;
+import java.net.URI;
+
+/**
+ * Wrapper class for java.awt.Desktop with `xdg-open` path on Linux platform.
+ */
+public class DesktopWrapper {
```
amake, line 39: To make this really unmodifiable it also has to be `final` or else someone can subclass it.

src/org/omegat/util/gui/DesktopWrapper.java
```diff
@@ -0,0 +1,88 @@
...
+
+    private final static Desktop awtDesktop = Desktop.getDesktop();
```
amake, line 41: I'm not so sure about caching this. Are we so sure that the `Desktop` object can't ever change, and that it never will in the future?

None of these methods should ever be called in a loop, so I would be fine calling `Desktop.getDesktop()` every time, until there is evidence that doing so is somehow causing a problem.

src/org/omegat/util/gui/DesktopWrapper.java
```diff
@@ -0,0 +1,88 @@
...
+        ProcessBuilder pb = new ProcessBuilder("xdg-open", s);
+        pb.redirectOutput(devNull);
+        pb.redirectError(devNull);
+        Process p = pb.start();
+
+        try {
+            return p.waitFor() == 0;
+        } catch (InterruptedException ex) {
+            return false;
+        }
```
amake, line 83: Why are we catching this exception? It seems to me that it makes more sense to let it throw.

amake (commented):
Thanks for this. Just some small comments.

src/org/omegat/util/gui/DesktopWrapper.java
chelobaka: Thanks for the tip.

src/org/omegat/util/gui/DesktopWrapper.java
chelobaka: I hesitated about this one, because it's unlikely to change during program run, but I agree that this is an over-optimization.

src/org/omegat/util/gui/DesktopWrapper.java
chelobaka: Is there any useful information we can extract from this exception? I suppose that running time of the thread should always be very short and even if it gets interrupted we don't really care about it. Propagating it up the stack will require changing other code without any perceivable benefit.

amake: I was thinking that letting it bubble up one level would make this method nicer, but I forgot that aside from the static init block we aren't try-catching, so never mind. You're right it's good as it is.

amake:
I've squashed to `master`. Thanks for your contribution!

PR #66, opened 2021-01-19
key/value filter: strip quote of value when exist

miurahr:
Implement [RFE#1328](https://sourceforge.net/p/omegat/feature-requests/1328/) that request 
to support joomla message catalog, which has double-quote for value.


amake:
Thanks for your contribution.

I have pushed to `master` with a small change: `String.format` is quite costly, so I split it into three separate calls to `write`.

PR #67, opened 2021-01-21
Add Aaron's semi-public announcement about scripts

brandelune:
https://sourceforge.net/p/omegat/feature-requests/1547/#a5ee

"I don't remember if I've ever said this publicly, so I should do so now:
OmegaT scripts are hard to maintain: they are basically untestable, and actually currently have no tests. We generally don't do all that great at maintaining API compatibility as a platform, but for the Java world you at least have the compiler to yell at you; for scripts it's incredibly easy to break them without noticing.
I do not consider the scripts core to OmegaT; they are there for users' convenience, but I do not commit to maintaining them.
If a script breaks then I expect the author to fix it. If the author is unable or unwilling to, I will remove it.
So I hope Kos is OK with including his script, given the above."


amake:
Thanks. I applied to master.

PR #68, opened 2021-01-23
Add plain text translation option to Yandex.Cloud

chelobaka:
Yandex.Cloud MT gives better results when translating in plain text mode. We can let our users to choose what's more important to them: tags or quality.


amake:
Is it not worthwhile to try to remove any tags of the option is off? What does the service do to them in that case?

chelobaka:
If the option is off then the service assumes format string is `PLAIN_TEXT` and removes all tags itself. There is no need to do it on our side.

amake (approved):
Looks good. Thanks as always.

Is there an RFE on SourceForge for this?

chelobaka:
I received this feature request personally, so probably there's no RFE.

amake:
OK. It's very helpful to me if you can make an RFE to go along with a PR.

I made it this time: https://sourceforge.net/p/omegat/feature-requests/1550/

I've committed to `master`. Thanks for your contribution.

chelobaka:
Thank you!

PR #70, opened 2021-01-27
add missing spaces in description

Presskopp:


amake:
These files are generated from translation resources so I think you need to talk to @kosivantsov  to get this fixed the right way.

kosivantsov:
Even if this PR is accepted, next time the docs are generated, your work will be lost. You need to edit doc sources, and the accepted way to do it described here: https://tinyurl.com/y46a2usy
You may also ask @Ragnar-F to fix it, they are actively working on DE localisation right now

PR #71, opened 2021-01-28
Fix bug #1033

chelobaka:
Fixes bug https://sourceforge.net/p/omegat/bugs/1033/

It was reported by Kos and he tested this fix.


amake:
Thanks! I've applied to `master`.

PR #75, opened 2021-02-20
INI-filter: allow continuous multiline

miurahr:
- Accept continuous lines as individual value.
- implement [RFE#1554](https://sourceforge.net/p/omegat/feature-requests/1554/)

### DIFF
--- a/src/org/omegat/filters2/text/ini/INIFilter.java
+++ b/src/org/omegat/filters2/text/ini/INIFilter.java
@@ -6,6 +6,7 @@
  Copyright (C) 2000-2006 Keith Godfrey and Maxym Mykhalchuk
                2009-2010 Alex Buloichik
                2011 Alex Buloichik, Didier Briel
+               2021 Hiroshi Miura
                Home page: http://www.omegat.org/
                Support center: https://omegat.org/support
 
@@ -51,6 +52,7 @@ import org.omegat.util.StringUtil;
  * @author Didier Briel
  */
 public class INIFilter extends AbstractFilter {
+    private static String unnamed_format = "[$$UnNamee_%d_]"; // general key cannot be [..]
     protected Map<String, String> align;
 
     public String getFileFormatName() {
@@ -94,8 +96,11 @@ public class INIFilter extends AbstractFilter {
         LinebreakPreservingReader lbpr = new LinebreakPreservingReader(reader); // fix for bug 1462566
         String str;
         String group = null;
+        int unnamed_counter = 0;
 
         while ((str = lbpr.readLine()) != null) {
+            String key;
+            String value;
             String trimmed = str.trim();
             boolean hasQuote = false;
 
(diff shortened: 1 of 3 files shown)


src/org/omegat/filters2/text/ini/INIFilter.java
```diff
@@ -111,34 +116,41 @@ public void processFile(BufferedReader reader, BufferedWriter outfile, FilterCon
...
-            int afterEqualsPos = str.offsetByCodePoints(equalsPos, 1);
-
-            // writing out everything before = (and = itself)
-            outfile.write(str.substring(0, afterEqualsPos));
-
-            String key = (group != null ? group + '/' : "") + str.substring(0, equalsPos).trim();
-            String value = str.substring(afterEqualsPos);
-
             value = leftTrim(value);
+
```
miurahr, line 153: unnecessary empty line

src/org/omegat/filters2/text/ini/INIFilter.java
```diff
@@ -111,34 +116,41 @@ public void processFile(BufferedReader reader, BufferedWriter outfile, FilterCon
...
-                equalsPos = str.offsetByCodePoints(str.length(), -1);
-            }
-
-            // advance if there're spaces after =
-            while (str.codePointCount(equalsPos, str.length()) > 1) {
-                int nextOffset = str.offsetByCodePoints(equalsPos, 1);
-                if (str.codePointAt(nextOffset) != ' ') {
-                    break;
+                int equalsPos = str.offsetByCodePoints(str.length(), -1);
+                int afterEqualsPos = str.offsetByCodePoints(equalsPos, 1);
```
miurahr, line 120: would be better to declare at top of the method code.

src/org/omegat/filters2/text/ini/INIFilter.java
```diff
@@ -107,37 +114,47 @@ public void processFile(BufferedReader reader, BufferedWriter outfile, FilterCon
...
-                equalsPos = str.offsetByCodePoints(str.length(), -1);
-            }
-
-            // advance if there're spaces after =
-            while (str.codePointCount(equalsPos, str.length()) > 1) {
-                int nextOffset = str.offsetByCodePoints(equalsPos, 1);
-                if (str.codePointAt(nextOffset) != ' ') {
-                    break;
+                // produce virtual key
+                key = (group != null ? group + '/' : "") + String.format(unnamed_format, unnamed_counter);
```
amake, line 133: Three things:

1. Sorry for the bikeshedding, but can we use camelCase for `unnamed_format` and `unnamed_counter`?

2. `String#format` is surprisingly expensive; since we only have one use of `unnamed_format`, I would suggest instead making a builder method like

    ```java
    private String buildKey(String group, int counter) {
        StringBuilder sb = new StringBuilder();
        if (group != null) {
            sb.append(group).append('/');
        }
        sb.append("[$$UnNamee_").append(counter).append("_]");
        return sb.toString();
    }
    ```
3. Is `[$$UnNamee_%d_]` really the intended key? Not UnName*****d*****? Is there some significance of this particular string, or is it something arbitrary you decided?

amake (commented):
Thank you! LGTM except for some very small things.

miurahr:
> Is [$$UnNamee_%d_] really the intended key? Not UnNamed? Is there some significance of this particular string, or is it something arbitrary you decided?

It is arbitrary which just form of  `[..]` which should not a key. It can be ` [%d]`.

src/org/omegat/filters2/text/ini/INIFilter.java
miurahr: Fixed comments No.1 and No.2

miurahr:
> It is arbitrary which just form of [..] which should not a key. It can be [%d].

Umm, `[ID6]` can be a key, which added as test case, so virtual key should be '#%d#'.
I've updated a patch.

src/org/omegat/filters2/text/ini/INIFilter.java
```diff
@@ -107,37 +122,47 @@ public void processFile(BufferedReader reader, BufferedWriter outfile, FilterCon
...
+                // advance if there're spaces after =
+                while (str.codePointCount(equalsPos, str.length()) > 1) {
+                    int nextOffset = str.offsetByCodePoints(equalsPos, 1);
+                    if (str.codePointAt(nextOffset) != ' ') {
+                        break;
+                    }
+                    equalsPos = nextOffset;
                 }
-                equalsPos = nextOffset;
+                key = (group != null ? group + '/' : "") + str.substring(0, equalsPos).trim();
```
amake, line 160: I think this would also be clearer as a `build` method, but it's OK if you don't want to bother.

(I know this is from the existing code.)

amake (approved):
LGTM. I was OK with an arbitrary string for the virtual keys; I just thought `UnNamee` looked like a typo.

src/org/omegat/filters2/text/ini/INIFilter.java
miurahr: Use buildKey() for here now

test/src/org/omegat/filters/INIFilterTest.java
```diff
@@ -50,6 +50,12 @@ public void testLoad() throws Exception {
         checkMulti("Value", "Section/ID", null, null, null, null);
         checkMulti("Value2", "Section/ID2", null, null, null, null);
         checkMulti("Value3", "Section/ID3", null, null, null, null);
+        checkMulti("Value4", "Section/ID4", null, null, null, null);
+        checkMulti("Continued Value4", "Section/#0#", null, null, null, null);
+        checkMulti("Value5", "Section/ID5", null, null, null, null);
+        checkMulti("Continued Value5 start with space", "Section/#1#", null, null, null, null);
```
amake, line 56: Sorry, I don't know why it took me so long to realize this, but:

1. Would it not make more sense for the value for `ID5` to be `Value5\n Continued Value5 start with space`? Are multiline values always really intended to be **collections of multiple values**? Or are they **single values spanning multiple lines**? If the latter, then we should extract as a single string and rely on segmentation to split them up.
2. If the answer to (1) is "no" (they are **collections** of values) then I think the keys for a collection should probably go like
  - `Section/ID5/0`
  - `Section/ID5/1`
  - `Section/ID5/2`

(How many levels of nesting can INI have? I see `[section]` and then the value ID. If that is the limit of nesting, then I think a `three/level/key` is sufficient to identify "virtual" values.)

miurahr: There is no concrete specification for multi-line value.
cf. https://en.wikipedia.org/wiki/INI_file

I don't know which is better, but (2) seems to be safe.

amake: I did (2) above, to the degree it's possible with the current implementation of the filter: https://github.com/omegat-org/omegat/pull/75/commits/f00ec0924a5d6484fc574c95c6e6000fd2acd3fd

Unfortunately at the time when we would build the key `Section/ID5/0` we don't yet know if there are continuation lines, so I output it as `Section/ID5` and then the first continuation line is `Section/ID5/1`.

It's also not feasible to implement (1) with the current implementation; we would have to build up values until we know there are no more continuation lines.

I would like to see an example of a real-world INI file that this improvement is targeting, so I can get a better feel for how important (1) is.

miurahr: One of targets of translation projects is a game translation. Some game project which use old version of Unreal Engine takes INI style file. (new version has a l10n feature which use po files)

eg. https://docs.unrealengine.com/udk/Three/LocalizedTextFiles.html

There seems no standard specification especially for multi line messages.

Here is a small example;

```
[System]

Hiscore = HighScore
Yes = Yes
No = no
Finished_All_Notification = Wow! It seems you have finished all the stages
You are awesome!
There will be more of them!
Congratulations = Congratulations

[UI]

EnableTranslation = You can edit the translation any of the available languages.
All you need to do is:
• browse to the localization folder
• Copy the localization file, of the specific language, you want to change
• Paste it to the folder


[Language]

English = English
Chinese = Chinese
Russian = Russian
German = German
French = French
Spanish = Spanish


```

amake: Thanks, that's helpful.

It seems pretty clear to me that the correct strategy is to parse as

- Key: `Finished_All_Notification`
- Value: `Wow! It seems you have finished all the stages\nYou are awesome!\nThere will be more of them!`

But I am not interested in rewriting the filter to support this.

What we have now is a compromise that at least is an improvement over the status quo. So I will squash and merge unless you are interested in working on it further.

miurahr: It is OK to merge it. Thank you.

amake:
Thanks for your contribution. I've squashed to master.

PR #77, opened 2021-02-26
add rudimentary support for dir attribute modification

brandelune:
I am just adding "dir" as a translatable attribute in both HTML and XHTML filters.


amake:
Thanks. I've merged to `master` on SourceForge.

PR #80, opened 2021-03-19
.tsv extension support for glossaries

brandelune:
It appears that the legit .tsv extension is not recognized for glossaries:
https://twitter.com/_niten/status/1372776521056755716

I've just added the extension as a possibility. I tested it on my side. Let me know if I missed something.


brandelune:
I'm cleaning the changes.txt and will PR again.

PR #81, opened 2021-03-19
Topic/jc/tsv

brandelune:
I'm not sure how to do formal tests, or if it is necessary.
Also, as far as documentation is concerned, it looks like .tsv files will be read in the default os encoding.

### DIFF
--- a/release/changes.txt
+++ b/release/changes.txt
@@ -1,7 +1,7 @@
 ----------------------------------------------------------------------
  OmegaT 5.5.0
 ----------------------------------------------------------------------
-  9 Enhancements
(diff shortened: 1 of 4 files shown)


amake:
> I'm not sure how to do formal tests, or if it is necessary.

Unfortunately the `GlossaryManager` doesn't have any unit tests right now, so it would be a big pain. The change is trivial enough that I think it's OK.

> Also, as far as documentation is concerned, it looks like .tsv files will be read in the default os encoding.

The encoding is detected with [juniversalchardet](https://github.com/albfernandez/juniversalchardet).

PR #82, opened 2021-03-20
Gradle: bump dictzip@0.9.5

miurahr:
dictzip moves to mavenCentral


amake:
Thank you! I've committed to `master`.

I removed the `mavenCentral()` line because it's included by default.

PR #83, opened 2021-03-21
glossary annex rewrite with updated info

brandelune:
See https://sourceforge.net/p/omegat/documentation/367/


amake:

PR #84, opened 2021-03-22
Consolidate file downloading utilities

miurahr:
Consolidate http connection utility methods into org.omegat.util.net.HttpConnectionUtils class.

There are several similar functions in WikiGet, StaticUtils and DictionaryManager.

This is also a preparation for plugin file download feature for #69.

### DIFF
--- a/src/org/omegat/util/StaticUtils.java
+++ b/src/org/omegat/util/StaticUtils.java
@@ -39,6 +39,7 @@ import java.io.InputStream;
 import java.io.PushbackInputStream;
 import java.net.URI;
 import java.net.URISyntaxException;
+import java.net.URL;
 import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.List;
@@ -50,6 +51,7 @@ import java.util.zip.ZipInputStream;
 
 import org.apache.commons.io.FileUtils;
 import org.omegat.util.Platform.OsType;
+import org.omegat.util.net.HttpConnectionUtils;
 
 /**
  * Static functions taken from CommandThread to reduce file size.
@@ -438,6 +440,17 @@ public final class StaticUtils {
             haystack.replace(globIndex, globIndex + 1, replacement);
             current = globIndex + replacement.length();
         }
+    }
+
+    /**
+     * Download a file to memory.
+     * @Deprecated
+     * This method is replaced to HttpConnectionUtils.getURL(url)
+     * No longer accept timeout.
+     */
+    @Deprecated
+    public static String downloadFileToString(URL url, int timeout) throws IOException {
+        return HttpConnectionUtils.getURL(url);
     }
 
     /**
--- a/src/org/omegat/util/StaticUtils.java
+++ b/src/org/omegat/util/StaticUtils.java
@@ -39,9 +39,6 @@ import java.io.InputStream;
 import java.io.PushbackInputStream;
 import java.net.URI;
 import java.net.URISyntaxException;
-import java.net.URL;
-import java.net.URLConnection;
-import java.nio.charset.StandardCharsets;
 import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.List;
@@ -52,7 +49,6 @@ import java.util.zip.ZipEntry;
 import java.util.zip.ZipInputStream;
 
 import org.apache.commons.io.FileUtils;
-import org.apache.commons.io.IOUtils;
 import org.omegat.util.Platform.OsType;
 
 /**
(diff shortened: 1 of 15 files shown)


amake:
Thanks, I think this is a great improvement.

However I'm concerned about one thing: I think third-party plugins may use some of the old public methods, so rather than removing them I think we should deprecated them and keep them for at least one release.

miurahr:
> However I'm concerned about one thing: I think third-party plugins may use some of the old public methods,

I'm also afraid that. There seems to be no clean definitions to open API which one is usable and stable for plugins.
IMHO, now is a good chance to define a clean API for plugins and encourage plugins developer update their products.

Can we define a java package such as `org.omegat.openapi` that is usable for plugins? 

> so rather than removing them I think we should deprecate them and keep them for at least one release.

Ok, I'd like to do it.

src/org/omegat/util/net/HttpConnectionUtils.java
```diff
@@ -0,0 +1,398 @@
...
+     */
+    private static final int TIMEOUT_MS = 10_000;
+
+    /**
+     * Don't instanciate util class.
+     */
+    private HttpConnectionUtils() {}
+
+    /**
+     * Download a file to memory
```
miurahr, line 82: javadoc: it should end with period(.)

src/org/omegat/util/net/HttpConnectionUtils.java
```diff
@@ -0,0 +1,398 @@
...
+    public static String getURL(URL url) throws IOException {
+        URLConnection urlConn = url.openConnection();
+        urlConn.setConnectTimeout(TIMEOUT_MS);
+        urlConn.setReadTimeout(TIMEOUT_MS);
+        try (InputStream in = urlConn.getInputStream()) {
+            return IOUtils.toString(in, StandardCharsets.UTF_8);
+        }
+    }
+
+    public static void downloadZipFileAndExtract(URL url, File dir, List<String> expectedFiles) throws IOException {
```
miurahr, line 93: need javadoc comment

src/org/omegat/util/net/HttpConnectionUtils.java
```diff
@@ -0,0 +1,398 @@
...
+                throw new FileNotFoundException("Could not extract expected files from zip; expected: "
+                        + expectedFiles + ", extracted: " + extracted);
+            }
+        } catch (IllegalArgumentException ex) {
+            throw new FlakyDownloadException(ex);
+        }
+    }
+
+    /**
+     * Downloads a binary file from a URL
```
miurahr, line 109: javadoc: need period

src/org/omegat/util/net/HttpConnectionUtils.java
```diff
@@ -0,0 +1,398 @@
...
+        }
+        try (InputStream in = conn.getInputStream()) {
+            return IOUtils.toByteArray(in);
+        } finally {
+            conn.disconnect();
+        }
+    }
+
+    /**
+     * Method call without additional headers for possible calls from plugins.
```
miurahr, line 168: javadoc: need @param here

miurahr (commented):
There are several place I can improve javadoc comments.

src/org/omegat/util/StaticUtils.java
```diff
@@ -445,15 +443,14 @@ private static void replaceGlobs(StringBuilder haystack, String needle, String r
     }
 
     /**
-     * Download a file to memory
+     * Download a file to memory.
+     * @Deprecated
+     * This method is replaced to HttpConnectionUtils.getURL(url)
+     * No longer accept timeout.
```
amake, line 449: Was there a reason to get rid of the timeout parameter?

src/org/omegat/util/net/HttpConnectionUtils.java
```diff
@@ -0,0 +1,398 @@
...
+        }
+    }
+
+    /**
+     * Downloads a binary file from a URL
+     * @param jarFileURL HTTP URL of the file to be downloaded
+     * @param saveFilePath path of the file
+     * @throws IOException
+     */
+    public static boolean downloadBinaryFile(final URL jarFileURL, final Map<String, String> headers,
```
amake, line 114: Where did this method come from? It isn't used anywhere.

Is this something you need for #69?

amake (commented):
Sorry for the delay on this. I just had some minor questions.

Also I wonder if the package `org.omegat.util.net` isn't too much. What about just `org.omegat.util`? Or even `org.omegat.net`?

src/org/omegat/util/StaticUtils.java
miurahr: Functions which use the function set a same timeout value so I remove it. Now there is a compat function so I'd like to revert to support timeout.

src/org/omegat/util/net/HttpConnectionUtils.java
miurahr: > Is this something you need for #69?

Yes, I plan to use it to download Jar file from repository.

src/org/omegat/util/StaticUtils.java
miurahr: Now accept timeout.

amake:
Sorry for the delay on this.

I will commit this after the release of 5.5.0 shortly. Can I get your opinion on this last issue please:

> Also I wonder if the package `org.omegat.util.net` isn't too much. What about just `org.omegat.util`? Or even `org.omegat.net`?

miurahr:
> Sorry for the delay on this.
> 
> I will commit this after the release of 5.5.0 shortly. Can I get your opinion on this last issue please:
> 
> > Also I wonder if the package `org.omegat.util.net` isn't too much. What about just `org.omegat.util`? Or even `org.omegat.net`?

OK, I'd like to move it to `org.omegat.util`.

amake:
Thank you and sorry for the delay. I've squashed and committed to `master`.

PR #85, opened 2021-03-24
Update to use DeepLv2 API

jatunda:
DeepL v1 api was deprecated in 2018.

### DIFF
--- a/src/org/omegat/core/machinetranslators/DeepLTranslate.java
+++ b/src/org/omegat/core/machinetranslators/DeepLTranslate.java
@@ -57,7 +57,7 @@ import org.omegat.util.WikiGet;
  */
 public class DeepLTranslate extends BaseTranslate {
     protected static final String PROPERTY_API_KEY = "deepl.api.key";
-    protected static final String DEEPL_URL = "https://api.deepl.com/v1/translate";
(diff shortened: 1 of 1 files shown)


amake:
Is v2 really compatible with v1 such that no other changes are necessary?

jatunda:
"Currently, the two DeepL API versions differ only in the URL used for access." < from their API docs.

amake:
I have committed this to `release/5.4.2`. It will be in the v5.4.2 release, which I will put out shortly.

amake:
I'm seeing a report that this change has broken functionality:

https://sourceforge.net/p/omegat/mailman/message/37249135/

I don't have a DeepL account so I can't test this myself. 

Can you comment on this? Is there anything that needs to be done to use v2, like regenerate a token/key or something?

amake:
This document says that v2 is not available for use in CAT tools:

https://www.deepl.com/docs-api/accessing-the-api/api-versions/

This information agrees with that:

https://community.sdl.com/product-groups/translationproductivity/f/openexchange_applications/31027/studio-2021-deepl---error/102536#102536

Why in the world did you make this PR?

jatunda:
I'm stupid. I was using a developer DeepL account, and my quick read of the docs led me to think that v1 was deprecated. 
Revert this change.

PR #86, opened 2021-03-24
Support markUnknown feature of Apertium

unhammer:
Untested! I'd like to add support for markUnknown, but I get this error when building:
```
$ /snap/bin/gradle assembleDist

Welcome to Gradle 6.8.3!

Here are the highlights of this release:
 - Faster Kotlin DSL script compilation
 - Vendor selection for Java toolchains
 - Convenient execution of tasks in composite builds
 - Consistent dependency resolution

For more details see https://docs.gradle.org/6.8.3/release-notes.html

Starting a Gradle Daemon (subsequent builds will be faster)

FAILURE: Build failed with an exception.

* Where:
Build file '/home/me/src/omegat/build.gradle' line: 10

* What went wrong:
Plugin [id: 'com.github.spotbugs', version: '4.7.0'] was not found in any of the following sources:

- Gradle Core Plugins (plugin is not in 'org.gradle' namespace)
- Plugin Repositories (could not resolve plugin artifact 'com.github.spotbugs:com.github.spotbugs.gradle.plugin:4.7.0')
  Searched in the following repositories:
    Gradle Central Plugin Repository

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 10s
```

@MarcRiera I based it off your PR https://github.com/omegat-org/omegat/pull/37 you may be interested in this too

### DIFF
--- a/src/org/omegat/Bundle.properties
+++ b/src/org/omegat/Bundle.properties
@@ -2179,6 +2179,7 @@ APERTIUM_CUSTOM_SERVER_URL_LABEL=Custom server URL
 APERTIUM_CUSTOM_SERVER_KEY_LABEL=API Key (optional)
 APERTIUM_CUSTOM_SERVER_NOTFOUND=Unable to connect to server. Ensure the server URL is correct.
 APERTIUM_CUSTOM_SERVER_INVALID=The server is not a valid Apertium server. Ensure the URL is correct.
+APERTIUM_MARKUNKNOWN_LABEL=Mark unknown words
 
 MT_ENGINE_GOOGLE2_API_KEY_LABEL=API Key:
 MT_ENGINE_GOOGLE2_PREMIUM_LABEL=Premium Edition
--- a/src/org/omegat/core/machinetranslators/ApertiumTranslate.java
+++ b/src/org/omegat/core/machinetranslators/ApertiumTranslate.java
@@ -5,6 +5,7 @@
 
  Copyright (C) 2010 Alex Buloichik, Ibai Lakunza Velasco, Didier Briel
                2019 Marc Riera Irigoyen
+               2021 Kevin Brubeck Unhammer
                Home page: http://www.omegat.org/
                Support center: https://omegat.org/support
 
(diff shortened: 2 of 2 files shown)


amake:
> * What went wrong:
> Plugin [id: 'com.github.spotbugs', version: '4.7.0'] was not found in any of the following sources:

Weird. My machine and [CI](https://dev.azure.com/omegat-org/OmegaT/_build/results?buildId=2396&view=results) have no problem resolving `com.github.spotbugs` 4.7.0.

The only thing I can see is

> `$ /snap/bin/gradle assembleDist`

You're using some system Gradle that is different from our wrapper. I had no problems with my system Gradle which is also 6.8.3, but just to remove one variable please try invoking `./gradlew` instead.

unhammer:
Ooh `./` – I thought `gradlew` was a command :see_no_evil:  maybe https://omegat.sourceforge.io/manual-latest/en/chapter.installing.and.running.html#building.OmegaT.from.source should include the `./` before `gradlew`.

Now I'm able to test and it seems to work for me :) The only issue is that translations seem to be cached (so if I turn on the switch after translating something, I the unknown words aren't retranslated with marking when I click the segment again), but I suppose it's not a big problem since people will not tend to turn it on and off in the same session.

amake:
> (maybe https://omegat.sourceforge.io/manual-latest/en/chapter.installing.and.running.html#building.OmegaT.from.source should include the `./` before `gradlew`)

It doesn't work on Windows then, so I guess we just need to be super verbose and list both patterns.

MarcRiera:
Tested, works as expected!

@unhammer Regarding the caching issue: there's a method that could be used to fix it (clearCache, see here: https://omegat.sourceforge.io/javadoc-latest/org/omegat/core/machinetranslators/BaseTranslate.html#clearCache--). As far as I know, the caching is used to prevent duplicate requests, which is important for MT plugins for services with word/request limits. Given that it is now possible to connect to third-party or local Apertium API services, clearing the cache without the user knowing about it could lead to unexpected behaviour if someone ever builds a limited service based on the Apertium API.

As you have already pointed out, turning this on and off in the middle of a translation session should be rare, so I wouldn't change anything else. Maybe a generic note about caching could be added in the MT section of OmegaT's user guide (because this is not limited to Apertium).

unhammer:
@amake Does this seem mergable? Anything more you need from me?

amake:
Sorry, I haven't had time to review it yet. I will let you know.

amake:
Thanks for your contribution! I have committed to `master` with minor changes.

Documented here: https://sourceforge.net/p/omegat/feature-requests/1563/

Upstream will be synced to GitHub shortly.

unhammer:
Thank you! :-)

PR #87, opened 2021-03-27
unify DTD URI to local URI

brandelune:
The /doc_src/Readme.md file says:

> For efficiency reason, the path to the DTD in the DocBook files (e.g.,
> `AboutOmegaT.xml`) has been changed to a local path
> (`../../../docbook-xml-4.5/docbookx.dtd` instead of an `http://`
> reference).

But there was a discrepancy in the files, some had the local URI some had the absolute URI.

This patch unifies the URI to the local path given in the Readme.md file.


amake:
Is the result of generating the docs still the same after this change?

brandelune:
I'm still in the process of installing the system so as to do it "the hard way" so I supposed that would work. Ignore until I report back.

brandelune:
I managed to create the HTML manual, and the FO file works too (I checked the internal XML).

 The problem I have, is that the PDF has # instead of letters except for some parts of the manual (looks like the contents of <code>...</code> that properly display their contents.) Considering the error message I get, this seems to be related to this issue: https://bz.apache.org/bugzilla/show_bug.cgi?id=47279

amake:
I also tried regenerating the manual after this change and saw no difference.

Thanks, I've committed to `master`.

PR #89, opened 2021-04-16
change author to Vincent Bideau

brandelune:


amake:
Did you discuss this with Vincent?

Is he not part of the "OmegaT team"?

amake:
To clarify, I'm OK with this change but I just wanted to make sure Vincent is OK with it too.

brandelune:
My understanding was that the two of you, or maybe Vincent and Didier had discussed that in the past and for license purposes it was better to have a clearly identified author. But maybe "author" is not the right term here, "Editor" or "Coordinator" or something like that could be better. In any case, just leave this PR on the side until that's settled.

amake:
I don't recall discussing that but it sounds fine to me. If DocBook has some markup that lets us credit both OmegaT contributors in general while also giving specific credit to Vincent for his work then that seems ideal.

brandelune:
Ok, I'll check docbook, and see with him what's the best way to credit his work. Let's close this PR for now. I'll open another one when everything is settled.

PR #90, opened 2021-04-16
topic/jc/bundle

brandelune:
Add reference to drag & drop in the "your project is empty" UI message.

### DIFF
--- a/src/org/omegat/Bundle.properties
+++ b/src/org/omegat/Bundle.properties
@@ -230,14 +230,14 @@ ENCODING_AUTO=<auto>
 
 # MainWindow
 TF_INTRO_EMPTYPROJECT=Your project does not contain any files in a supported file format.\n\
-    (In the user's manual, see 'Files to translate' > 'File formats' \
-    for a list of supported file formats.)\n\
-    To add files to the project, click on the Copy Files to Source Folder... button \
-    in the Project Files window or call this action from the Project menu and \
-    select the source files you want to copy. You may also copy the files \
-    to your project's 'source' folder using your favorite file manager and \
-    reload the project by calling Reload from the Project menu.
-
+    See Options > File filters for a list of supported file formats.\n\
+    To add files to the project you can\n\
+    - drag and drop files on this Editor pane\n\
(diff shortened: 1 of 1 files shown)


src/org/omegat/Bundle.properties
```diff
@@ -230,14 +230,14 @@ ENCODING_AUTO=<auto>
...
-    select the source files you want to copy. You may also copy the files \
-    to your project's 'source' folder using your favorite file manager and \
-    reload the project by calling Reload from the Project menu.
-
+    See Options > File filters for a list of supported file formats.\n\
+    To add files to the project you can\n\
+    - drag and drop files on this Editor pane\n\
+    - use Project > Copy Files to Source Folder...\n\
+    - click on Copy Files to Source Folder... in the Project Files window\n\
+    - copy files to the project\'s /source/ folder with your file manager, \
```
amake, line 238: A very pedantic question if you don't mind:

What is your intent in changing `'source'` to `/source/`?

I don't have a strong opinion on this but I was thinking:

- `/foo/` in some contexts indicates italics
- In another context `/foo/` could be a path, but it would be an absolute path with a trailing slash

I feel like this is maybe harder for users to understand as a path, because it looks absolute but is actually relative, and at the same time it's confusable as some sort of other emphasis or styling.

brandelune: Not a pedantic question at all.

I checked the doc and there are instances of /folder, /folder/, ’folder’, "folder", folder/ and folder. And inconsistencies in identical segments:
The resulting TMX is stored in the <f1>/omegat/</f1> folder under the name <f2>align.tmx</f2>.
The resulting TMX is stored in the <f3>omegat</f3> folder under the name <f4>align.tmx</f4>.

So it's a bit messy but I did not think about the /.../ italics meaning.

I guess we'll have to unify the documentation at one point, so if you have a preference I'm a taker. I personally would go for /folder in places where we don't have visual cues (the doc can use <filename> to clarify the meaning).

amake:
I went with apostrophes for now. Thanks.

PR #92, opened 2021-04-27
Improve plugin information management

miurahr:
- Manage plugins status whether is bundled or installed
- Better version and author detection
    - Set "OmegaT team" for bundled plugins
    - Set version as same as OmegaT for bundled plugins
- Support extended metadata plugin-link  [RFE#1544](https://sourceforge.net/p/omegat/feature-requests/1544/)
- Plugin category name from plugin-category or old OmegaT-Plugin metadata.

- PluginInformation class is now POJO data class, and introduce its builder

### DIFF
--- a/src/org/omegat/core/data/PluginInformation.java
+++ b/src/org/omegat/core/data/PluginInformation.java
@@ -4,6 +4,7 @@
           glossaries, and translation leveraging into updated projects.
 
  Copyright (C) 2020 Briac Pilpre
+               2021,2022 Hiroshi Miura
                Home page: http://www.omegat.org/
                Support center: https://omegat.org/support
 
@@ -25,102 +26,115 @@
 
 package org.omegat.core.data;
 
+import java.net.URL;
 import java.util.Properties;
 import java.util.jar.Attributes;
 import java.util.jar.Manifest;
 
-public class PluginInformation {
-    private static final String PLUGIN_NAME = "Plugin-Name";
-    private static final String PLUGIN_VERSION = "Plugin-Version";
-    private static final String PLUGIN_AUTHOR = "Plugin-Author";
-    private static final String PLUGIN_DESCRIPTION = "Plugin-Description";
+import org.omegat.filters2.master.PluginUtils;
 
-    private static final String IMPLEMENTATION_VENDOR = "Implementation-Vendor";
-    private static final String IMPLEMENTATION_TITLE = "Implementation-Title";
-    private static final String IMPLEMENTATION_VERSION = "Implementation-Version";
-    private static final String BUNDLE_VERSION = "Bundle-Version";
-    private static final String BUNDLE_NAME = "Bundle-Name";
-    private static final String BUILT_BY = "Built-By";
+/**
+ * Plugin information POJO data class.
+ * @author Briac Pilpre
+ * @author Hiroshi Miura
+ */
+public class PluginInformation {
+
+    public enum Status {
+        INSTALLED,
+        BUNDLED,
+    }
 
     private final String className;
     private final String name;
     private final String version;
     private final String author;
     private final String description;
+    private final PluginUtils.PluginType category;
+    private final String link;
+    private final URL url;
+    private final Status status;
 
-    public PluginInformation(String className, Manifest manifest) {
+    /* The class is recommend to build from builder. */
+    private PluginInformation(String className, String name, String version, String author, String description,
+                             PluginUtils.PluginType category, String link, URL url, Status status) {
         this.className = className;
-        Attributes attrs = manifest.getMainAttributes();
-        name = findName(manifest);
-        version = findVersion(manifest);
-        author = findAuthor(manifest);
-        description = attrs.getValue(PLUGIN_DESCRIPTION);
+        this.name = name;
+        this.version = version;
+        this.author = author;
+        this.description = description;
+        this.category = category;
+        this.link = link;
+        this.url = url;
+        this.status = status;
     }
 
-    public PluginInformation(String className, Properties props) {
-        this.className = className;
-        name = null;
-        version = null;
-        author = null;
-        description = null;
-    }
-
-    private String findName(Manifest m) {
-        Attributes attrs = m.getMainAttributes();
-        if (attrs.getValue(PLUGIN_NAME) != null) {
-            return attrs.getValue(PLUGIN_NAME);
-        } else if (attrs.getValue(BUNDLE_NAME) != null) {
-            return attrs.getValue(BUNDLE_NAME);
-        } else if (attrs.getValue(IMPLEMENTATION_TITLE) != null) {
-            return attrs.getValue(IMPLEMENTATION_TITLE);
-        }
-        return null;
-    }
-
-    private String findVersion(Manifest m) {
-        Attributes attrs = m.getMainAttributes();
-        if (attrs.getValue(PLUGIN_VERSION) != null) {
-            return attrs.getValue(PLUGIN_VERSION);
-        } else if (attrs.getValue(BUNDLE_VERSION) != null) {
-            return attrs.getValue(BUNDLE_VERSION);
-        } else if (attrs.getValue(IMPLEMENTATION_VERSION) != null) {
-            return attrs.getValue(IMPLEMENTATION_VERSION);
-        }
-        return null;
-    }
-
-    private String findAuthor(Manifest m) {
-        Attributes attrs = m.getMainAttributes();
-        if (attrs.getValue(PLUGIN_AUTHOR) != null) {
-            return attrs.getValue(PLUGIN_AUTHOR);
-        } else if (attrs.getValue(IMPLEMENTATION_VENDOR) != null) {
-            return attrs.getValue(IMPLEMENTATION_VENDOR);
-        } else if (attrs.getValue(BUILT_BY) != null) {
-            return attrs.getValue(BUILT_BY);
-        }
-        return null;
-    }
-
+    /**
+     * @return className of plugin entry point
+     */
     public String getClassName() {
         return className;
     }
 
+    /**
+     * @return name of plugin
+     */
     public String getName() {
         return name;
     }
 
+    /**
+     * @return version of plugin
+     */
     public String getVersion() {
         return version;
     }
 
+    /**
+     * @return description of plugin features
+     */
     public String getDescription() {
         return description;
     }
 
+    /**
+     * @return author(s) of plugin
+     */
     public String getAuthor() {
         return author;
     }
 
+    /**
+     * @return category type of plugin as PluginType enum
+     */
+    public final PluginUtils.PluginType getCategory() {
+        return category;
+    }
+
+    /**
+     * @return link URL of plugin homepage
+     */
+    public final String getLink() {
+        return link;
+    }
+
+    /**
+     * @return manifest URL of plugin jar
+     */
+    public URL getUrl() {
+        return url;
+    }
+
+    /**
+     * @return true if plugin is bundled with OmegaT distribution, otherwise false when 3rd party plugin
+     */
+    public final boolean isBundled() {
+        return status == Status.BUNDLED;
+    }
+
+    /**
+     * @return string expression of PluginInformation class.
+     */
     @Override
     public String toString() {
         StringBuilder builder = new StringBuilder();
@@ -130,6 +144,10 @@ public class PluginInformation {
         return builder.toString();
     }
 
+    /**
+     * It is identical if status is differed.
+     * @return hashCode of plugin
+     */
     @Override
     public int hashCode() {
         final int prime = 31;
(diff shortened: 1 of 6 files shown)


src/org/omegat/core/data/PluginInformation.java
```diff
@@ -173,4 +267,55 @@ public boolean equals(Object obj) {
         return true;
     }
 
+    @Override
+    public final int compareTo(PluginInformation pluginInformation) {
```
miurahr, line 271: Here is a point how to sort plugins, that affect to order of preferences-plugins table in default.

src/org/omegat/filters2/master/PluginUtils.java
```diff
@@ -269,8 +292,12 @@ protected static void loadFromManifest(final Manifest m, final ClassLoader class
                     continue;
                 }
                 if (loadClass(clazz, classLoader)) {
-                    PLUGIN_INFORMATIONS.add(new PluginInformation(clazz, m));
-                }
+                    if (mu == null) {
+                        PLUGIN_INFORMATIONS.add(new PluginInformation(clazz, m, null, PluginInformation.STATUS.BUNDLED));
```
miurahr, line 296: When `bundled` plugin is loaded, `mu` become null, because L.137 call here.
If there is better way to determine bundled plugins, it can be improved.

src/org/omegat/core/data/PluginInformation.java
```diff
@@ -42,56 +53,107 @@
...
         if (attrs.getValue(PLUGIN_VERSION) != null) {
             return attrs.getValue(PLUGIN_VERSION);
         } else if (attrs.getValue(BUNDLE_VERSION) != null) {
             return attrs.getValue(BUNDLE_VERSION);
         } else if (attrs.getValue(IMPLEMENTATION_VERSION) != null) {
             return attrs.getValue(IMPLEMENTATION_VERSION);
+        } else if (mainAttrs.getValue(PLUGIN_VERSION) != null) {
+            return mainAttrs.getValue(PLUGIN_VERSION);
+        } else if (mainAttrs.getValue(BUNDLE_VERSION) != null) {
+            return mainAttrs.getValue(BUNDLE_VERSION);
```
miurahr, line 146: Duplicated conditions should be removed.

miurahr: It is not duplicated. 
A part here is for bundled plugin or Okapi framework, that has Plugin-Version: in Main attribution and each filter has no version.

src/org/omegat/core/data/PluginInformation.java
```diff
@@ -173,4 +267,55 @@ public boolean equals(Object obj) {
         return true;
     }
 
+    @Override
+    public final int compareTo(PluginInformation pluginInformation) {
+       int score;
```
amake, line 272: The indents are not aligned in this section.

src/org/omegat/core/data/PluginInformation.java
```diff
@@ -42,56 +53,107 @@
     private static final String BUNDLE_NAME = "Bundle-Name";
     private static final String BUILT_BY = "Built-By";
 
+    public enum STATUS {
+        INSTALLED,
+        BUNDLED,
+        UPGRADABLE,
+        UNINSTALLED
```
amake, line 60: `UPGRADABLE` and `UNINSTALLED` are not used (sorry I haven't seen if they're used in your other related PR).

I can imagine how `UPGRADABLE` will be used. But what about `UNINSTALLED`? How does that work?

amake (commented):
I didn't get a chance to look at everything, but here are some comments from my first pass.

src/org/omegat/core/data/PluginInformation.java
```diff
@@ -121,6 +187,34 @@ public String getAuthor() {
...
+    public final String getCategory() {
+        return category;
+    }
+
+    public final String getLink() {
+     return link;
+    }
+
+    public final File getJarFile() {
+        return new File(url.getPath().substring(5, url.getPath().indexOf("!")));
```
amake, line 199: This string munging makes me very nervous. What is the magic `5`? Is there really no better way to extract the file path?

src/org/omegat/core/data/PluginInformation.java
miurahr: Yes, `UNINSTALLED` is not used. I'd like to remove it.

src/org/omegat/core/data/PluginInformation.java
miurahr: This can be safely delete

5 = 'file:`.length

src/org/omegat/core/data/PluginInformation.java
```diff
@@ -42,56 +53,107 @@
...
+    }
+
+    private String categoryName(final String key1, final String key2) {
+        String key = key1 != null ? key1 : key2;
+        Optional<PluginUtils.PluginType> type = Arrays.stream(PluginUtils.PluginType.values()).filter(v ->
+                v.getTypeValue().equals(key)).findFirst();
+        if (type.isPresent()) {
+            return type.get().getTypeValue();
+        }
+        return PluginUtils.PluginType.UNKNOWN.getTypeValue();
```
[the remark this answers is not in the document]
miurahr: I've got it. Rewriting it now.

src/org/omegat/core/data/PluginInformation.java
miurahr: Now we don't have plugin removal feature on Preference - Plugins.
When we got the feature, preference dialog may show "UNINSTALLED" status before restart OmegaT.

src/org/omegat/core/data/PluginInformation.java
```diff
@@ -42,56 +53,107 @@
     private static final String BUNDLE_NAME = "Bundle-Name";
     private static final String BUILT_BY = "Built-By";
 
+    public enum STATUS {
```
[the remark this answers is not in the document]
miurahr: done.

miurahr:
> private static final String PLUGIN_ID = "Plugin-Id";

It is not good to add. We can use  class-name as an identifier.
When we need some artifact id or similar for plugin installer, it can be handled out of manifest in plugin itself.

miurahr:
It is better to add test for class PluginInformation

amake:
Does this make sense to review next?

src/org/omegat/core/data/PluginInformation.java
miurahr: I've replaced with getUrl() here.

miurahr:
> Does this make sense to review next?

Yes, could you please review after resolving conflict.

src/org/omegat/core/data/PluginInformation.java
```diff
@@ -42,56 +50,91 @@
...
-        version = findVersion(manifest);
-        author = findAuthor(manifest);
+        Attributes mainAttrs = manifest.getMainAttributes();
+        Attributes attrs = manifest.getEntries().get(className);
+        if (attrs == null) {
+              attrs = manifest.getMainAttributes();
+        }
+        name = findName(attrs);
+        version = findVersion(attrs, mainAttrs);
+        author = findAuthor(mainAttrs);
```
amake, line 78: I'm confused about the difference between `mainAttrs` and `attrs`.

If `attrs` is null (when would that be?) then we use `mainAttrs` for `attrs`.

Some information is looked up in `attrs` but some only in `mainAttrs`.

Why are some things in one but not the other? Can we not make the lookup logic simpler, e.g. always look in one, and then the other?

Also would it not be better to keep `PluginInformation` a POJO (keep this lookup logic out of the constructor)? There could be a factory method e.g. `buildFromManifest`.

src/org/omegat/core/data/PluginInformation.java
```diff
@@ -166,11 +235,59 @@ public boolean equals(Object obj) {
...
-                return false;
-        } else if (!version.equals(other.version))
-            return false;
-        return true;
+            return other.version == null;
+        } else return version.equals(other.version);
     }
 
+    @Override
+    public final int compareTo(PluginInformation pluginInformation) {
```
amake, line 243: I think this method can be simplified greatly by use of [Comparator](https://docs.oracle.com/javase/8/docs/api/java/util/Comparator.html).

src/org/omegat/filters2/master/PluginUtils.java
```diff
@@ -72,8 +72,29 @@
...
+        TOKENIZER("tokenizer"),
+        MARKER("marker"),
+        MACHINETRANSLATOR("machinetranslator"),
+        BASE("base"),
+        GLOSSARY("glossary"),
+        DICTIONARY("dictionary"),
+        MISCELLANEOUS("miscellaneous"),
+        UNKNOWN("Undefined");
+
+        private final String typeValue;
```
amake, line 89: What is the need for `typeValue`? It looks like it's supposed to be a UI string, but that's not ideal because we can't translate them.

Instead of having UI strings as members of the enum class, the UI string should be looked up like `OStrings.getString("PLUGIN_TYPE_" + PluginType.FOO)`.

src/org/omegat/core/data/PluginInformation.java
miurahr: >  the difference between `mainAttrs` and `attrs`.

There is because `PluginInformation` is created from old API manifest,  and  new API manifest.

The old manifest looks like

```
Manifest-Version: 1.0
Bundle-ManifestVersion: 2
Bundle-Name: Okapi Filters for OmegaT
OmegaT-Plugin: true

Name: net.sf.okapi.lib.omegat.TSFilter
OmegaT-Plugin: filter

Name: net.sf.okapi.lib.omegat.TTXFilter
OmegaT-Plugin: filter
```

Here `Bundle Name` field is in `mainAttr`, and `Name: <class>` field is in `attr`.
When loading plugin like okapi filter, several `PluginInformation` objects are created with same 'mainAttr` and each `attr` 


When new one

```
Manifest-Version: 1.0
Plugin-Name: Okapi Filters for OmegaT
OmegaT-Plugins: net.sf.okapi.lib.omegat.OmegatOkapiFilter
```

All field is in `mainAttr` (or `attr`?) and another is null.

src/org/omegat/filters2/master/PluginUtils.java
miurahr: I use `typeValue` to compare with `Plugin-Category` value in MANIFEST, and get corresponding enum object.

src/org/omegat/core/data/PluginInformation.java
miurahr: > Also would it not be better to keep PluginInformation a POJO (keep this lookup logic out of the constructor)? There could be a factory method e.g. buildFromManifest.

I'd like to try it.

src/org/omegat/filters2/master/PluginUtils.java
miurahr: I've adjusted relative functions to synthesize the policy.

> Instead of having UI strings as members of the enum class, the UI string should be looked up like OStrings.getString("PLUGIN_TYPE_" + PluginType.FOO).

I'd like to do this for future UI change.

src/org/omegat/core/data/PluginInformation.java
miurahr: I've introduced a `PluginInformationFactory` class to make things simple.

src/org/omegat/filters2/master/PluginUtils.java
```diff
@@ -72,8 +73,41 @@
...
+        public String getTypeValue() {
+            return typeValue;
+        }
+
+        public static PluginType getTypeByValue(String str) {
+            String sType = str.toLowerCase(Locale.ENGLISH);
+            for(PluginType v : values())
+            {
+                if(v.getTypeValue().equals(sType))
+                {
```
miurahr, line 105: Here can improve coding styles.

miurahr:
I believe it is ready to merge.

miurahr:
Any further comments and requests?

amake:
> Any further comments and requests?

Sorry for the delay. I've been very busy lately. I'll try to get to this soon.

amake:
I'm really sorry, I'm moving to a new house this weekend so it will take me a while to get back to reviewing. I hope to resolve some of the backlog by the end of the month.

src/org/omegat/core/data/PluginInformationFactory.java
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    private static String lookupAttribute(Attributes attrs, String[] candicates) {
+        for (String key: candicates) {
+            if (attrs.getValue(key) != null) {
+                return attrs.getValue(key);
+            }
+        }
+        return null;
+    }
```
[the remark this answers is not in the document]
miurahr: Thanks. Merged.

miurahr:
I've reorganized  a commit history for review.

miurahr:
I believe the PR become a status of enough quality, which has a clean commit history, and ready for merge.
All non-core changes are  divided into  commits. A last commit is the core of proposed change, extension of fields of PluginInforrmation.

@amake please take a look at commits?

miurahr:
Simplify and re-based on master

miurahr:
merged into `master`.

PR #94, opened 2021-04-28
Plugin: drop old tokenizer loader

miurahr:
This part of source was added 10 years ago and have a comment 

```
// TODO remove after release new tokenizers
```

This patch remove it.


amake:
I'm not positive that it's safe to remove this. I've asked on the omegat-development mailing list if anyone knows of plugins that use this code.

If there don't seem to be any plugins using it, we can remove it.

amake:
The only reported tokenizer plugins don't use this code path, so I've committed to `master`. Thanks!

PR #97, opened 2021-05-09
Update OmegaT_webhelp.css: distinguish h2 and h3 more clearly

msoutopico:
Increased the font-size property of h2 (and also of h1, to keep the proportion) and added some soft underlining to h2, I think the result makes it a bit clearer where each section starts and what is inside the section.


amake:
Thanks. Vincent committed equivalent changes in https://github.com/omegat-org/omegat/commit/38823c907415be0130f06ff7b8e5337e1dc75d74

PR #100, opened 2021-05-27
Allow configuring Theme(LaF) and adding LaF by plugins

miurahr:
This PR is intend to introduce a theme plugin to change Look-and-Feel.
Theme plugin that specified by preference is activated before starting GUI.

Restart required when change theme.

Related  [RFE#1540](https://sourceforge.net/p/omegat/feature-requests/1540/) and [RFE#1567](https://sourceforge.net/p/omegat/feature-requests/1567/)


src/org/omegat/util/gui/UIDesignManager.java
```diff
@@ -72,9 +75,34 @@
...
+    }
+
+    public static void registerThemeloader(IThemeLoader loader) {
+        themeLoaders.add(loader);
+    }
+
+    public static boolean launchThemeLoader(String name) {
+        Optional<IThemeLoader> loader = themeLoaders.stream().filter(t -> name.equals(t.getName())).findFirst();
+        if (loader.isPresent()) {
+            loader.get().onStartup();
```
miurahr, line 99: We need to add plugin's jar file to system class path where UIManager can see LaF library.

Loading external class to ancestor of class loader in dynamic manner is hack for Java8. After Java9, we need a custom class loader and specify it when launching JVM.

miurahr:
Custom loader can load LaF library in dynamic.
It resulted in NPE now.

```
java.lang.Error: no ComponentUI class for: org.omegat.gui.editor.EditorTextArea3[,0,0,0x0,invalid,alignmentX=0.0,alignmentY=0.0,border=,flags=0,maximumSize=,minimumSize=,preferredSize=,caretColor=,disabledTextColor=,editable=true,margin=,selectedTextColor=,selectionColor=,kit=,typeHandlers=]
	at java.desktop/javax.swing.UIDefaults.getUIError(UIDefaults.java:760)
	at java.desktop/javax.swing.MultiUIDefaults.getUIError(MultiUIDefaults.java:131)
	at java.desktop/javax.swing.UIDefaults.getUI(UIDefaults.java:790)
	at java.desktop/javax.swing.UIManager.getUI(UIManager.java:1066)
	at java.desktop/javax.swing.text.JTextComponent.updateUI(JTextComponent.java:352)
	at java.desktop/javax.swing.text.JTextComponent.<init>(JTextComponent.java:326)
	at java.desktop/javax.swing.JEditorPane.<init>(JEditorPane.java:198)
	at org.omegat.gui.editor.EditorTextArea3.<init>(EditorTextArea3.java:149)
	at org.omegat.gui.editor.EditorController.<init>(EditorController.java:223)
	at org.omegat.core.Core.initializeGUI(Core.java:257)
	at org.omegat.Main.runGUI(Main.java:300)
	at org.omegat.Main.main(Main.java:190)
12195: エラー: java.lang.NullPointerException 
12195: エラー: 	at org.omegat.gui.editor.EditorTextArea3.<init>(EditorTextArea3.java:170) 
12195: エラー: 	at org.omegat.gui.editor.EditorController.<init>(EditorController.java:223) 
12195: エラー: 	at org.omegat.core.Core.initializeGUI(Core.java:257) 
12195: エラー: 	at org.omegat.Main.runGUI(Main.java:300) 
12195: エラー: 	at org.omegat.Main.main(Main.java:190) 
```

When launching OmegaT with `java -cp OmegaT.jar:plugin.jar org.omegat.Main` it works well.

src/org/omegat/Main.java
```diff
@@ -118,6 +118,14 @@ public static void main(String[] args) {
             System.exit(0);
         }
 
+        ClassLoader loader = ClassLoader.getSystemClassLoader();
+        MainClassLoader mainClassLoader = (MainClassLoader) loader;
+        try {
+            mainClassLoader.addJarToClasspath("/opt/OmegaT-default/plugins/omegat-laf-0.0.1.jar");
```
miurahr, line 124: Here should be generic, but now it is for PoC.

miurahr:
> Custom loader can load LaF library in dynamic.
> It resulted in NPE now.
> NPE is reported on GUI dialog, which LaF is customized.

It may be happened on ClassLoader hierarchy issue. 
Does anyone suggest root cause or work around?

https://github.com/openjdk/jdk/blob/master/src/java.desktop/share/classes/javax/swing/UIDefaults.java#L783-L792

```

 public ComponentUI getUI(JComponent target) {
        Object cl = get("ClassLoader");
        ClassLoader uiClassLoader =  (cl != null) ? (ClassLoader)cl : target.getClass().getClassLoader();
        Class<? extends ComponentUI> uiClass = getUIClass(target.getUIClassID(), uiClassLoader);
        Object uiObject = null;

        if (uiClass == null) {
            getUIError("no ComponentUI class for: " + target)
        }
```

miurahr:
I've tested with  https://github.com/miurahr/omegat-laf

miurahr:
`getUIClass` is a function that load LaF 
> Returns the L&F class that renders this component.

miurahr:
> It may be happened on ClassLoader hierarchy issue.
> Does anyone suggest root cause or work around?

Fixed by commit https://github.com/omegat-org/omegat/pull/100/commits/f0e9e41b426d6417f59994d8b956ddde47a5a932

release/linux-specific/OmegaT
```diff
@@ -7,4 +7,4 @@ JAVA="java"
 BUNDLED_JAVA="${REALOMEGATPATH}/jre/bin/java"
 [ -f "${BUNDLED_JAVA}" ] && JAVA="${BUNDLED_JAVA}"
 
-"${JAVA}" -jar -Xmx1024M "${REALOMEGATPATH}/@JAR_SUBST@" "$@"
+"${JAVA}" -Djava.system.class.loader=org.omegat.MainClassLoader -jar -Xmx1024M "${REALOMEGATPATH}/@JAR_SUBST@" "$@"
```
miurahr, line 10: Question.

Where is a place to do  a same thing in macOS ?

miurahr:
This modification make sub-effect to allow users to select other system laf such as 'Nimbus', 'Metal' and more.
After selecting LaF,  restart is required to take effect.

src/org/omegat/gui/preferences/view/AppearanceController.java
```diff
@@ -55,21 +65,29 @@ public String toString() {
...
         // TODO: Properly abstract the restore function
         panel.restoreWindowButton
                 .addActionListener(e -> MainWindowUI.resetDesktopLayout((MainWindow) Core.getMainWindow()));
     }
-
+    
     @Override
     protected void initFromPrefs() {
+        String currentTheme = Preferences.getPreferenceDefault(Preferences.THEME_SELECTED_NAME, "Default");
+        // Todo: check theme exist or not, when not exist fallback to default
```
miurahr, line 79: Todo

src/org/omegat/Main.java
```diff
@@ -255,6 +256,15 @@ private static void applyConfigFile(String path) {
      * Execute standard GUI.
      */
     protected static int runGUI() {
+        ClassLoader loader = ClassLoader.getSystemClassLoader();
+        MainClassLoader mainClassLoader = (MainClassLoader) loader;
+        try {
+            mainClassLoader.addJarToClasspath("/opt/OmegaT-default/plugins/omegat-laf-0.0.1.jar");
```
miurahr, line 262: Should generalize, now just for PoC.

miurahr:
https://github.com/omegat-org/omegat/pull/100/commits/8beb09767bb2e8a5bbbb6161f86fb818615cac04 Works well without setting  plugin.jar into search path.

src/org/omegat/Main.java
```diff
@@ -45,11 +45,11 @@
 import java.util.Map;
 import java.util.PropertyResourceBundle;
 import java.util.TreeMap;
-
 import javax.swing.JOptionPane;
 import javax.swing.SwingUtilities;
 import javax.swing.UIManager;
 
+import com.vlsolutions.swing.docking.DockingDesktop;
```
amake, line 52: Your IDE seems to prefer different import-ordering than we tend to use in OmegaT.

I don't really like this unnecessary churn.

Can you tell me what IDE you're using? I believe we have presets for Eclipse; maybe we can figure out something for your IDE as well.

src/org/omegat/Main.java
```diff
@@ -277,14 +281,26 @@ protected static int runGUI() {
...
+                try {
+                    UIManager.setLookAndFeel(UIDesignManager.getThemeClassName(theme));
+                } catch (Exception e) {
+                    Log.log(e);
+                    // fallback to default theme
+                    UIManager.setLookAndFeel(UIManager.getSystemLookAndFeelClassName());
+                }
+            }
+        } catch (Exception ex) {
+            // Something become wrong!!
```
amake, line 298:
```suggestion
            // Something went wrong!!
```

src/org/omegat/Main.java
```diff
@@ -277,14 +281,26 @@ protected static int runGUI() {
             // do nothing
         }
 
+        String theme = Preferences.getPreferenceDefault(Preferences.THEME_SELECTED_NAME, "Default");
```
amake, line 284: I'm a bit wary of expecting `Preferences` to be initialized this early. But if it's working then I guess it's OK.

src/org/omegat/Main.java
```diff
@@ -255,6 +254,11 @@ private static void applyConfigFile(String path) {
      * Execute standard GUI.
      */
     protected static int runGUI() {
+        ClassLoader cl = ClassLoader.getSystemClassLoader();
+        MainClassLoader mainClassLoader = (cl instanceof MainClassLoader)? (MainClassLoader) cl: new MainClassLoader(cl);
```
amake, line 258: Project whitespace conventions

```suggestion
        MainClassLoader mainClassLoader = (cl instanceof MainClassLoader) ? (MainClassLoader) cl : new MainClassLoader(cl);
```

src/org/omegat/MainClassLoader.java
```diff
@@ -0,0 +1,103 @@
+/**************************************************************************
+ OmegaT - Computer Assisted Translation (CAT) tool
+          with fuzzy matching, translation memory, keyword search,
+          glossaries, and translation leveraging into updated projects.
+
+ Copyright (C) 2021 Hiroshi Miura
+               2018 Mordechai Meisels(Original licensed Apache-2.0)
```
amake, line 7:
```suggestion
               2018 Mordechai Meisels (Original licensed Apache-2.0)
```

amake: Where exactly did this code come from?

src/org/omegat/gui/preferences/view/AppearanceController.java
```diff
@@ -55,21 +67,32 @@ public String toString() {
...
 
     private void initGui() {
+        UIDesignManager.getThemes().forEach(themeSelectionModel::addElement);
         panel = new AppearancePreferencesPanel();
         // TODO: Properly abstract the restore function
         panel.restoreWindowButton
                 .addActionListener(e -> MainWindowUI.resetDesktopLayout((MainWindow) Core.getMainWindow()));
     }
-
+
```
amake, line 76: Whitespace cleanup

src/org/omegat/gui/preferences/view/AppearanceController.java
```diff
@@ -55,21 +67,32 @@ public String toString() {
...
         panel = new AppearancePreferencesPanel();
         // TODO: Properly abstract the restore function
         panel.restoreWindowButton
                 .addActionListener(e -> MainWindowUI.resetDesktopLayout((MainWindow) Core.getMainWindow()));
     }
-
+    
     @Override
     protected void initFromPrefs() {
+        String currentTheme = Preferences.getPreferenceDefault(Preferences.THEME_SELECTED_NAME, "Default");
```
amake, line 79: We use the magic string `Default` in a lot of places. I would prefer a constant in `Preferences.java`.

src/org/omegat/Main.java
```diff
@@ -277,14 +281,26 @@ protected static int runGUI() {
             // do nothing
         }
 
+        String theme = Preferences.getPreferenceDefault(Preferences.THEME_SELECTED_NAME, "Default");
         try {
-            UIManager.setLookAndFeel(UIManager.getSystemLookAndFeelClassName());
-        } catch (Exception e) {
-            // do nothing
+            if (theme.equals("Default")) {
```
amake, line 286: This deep nesting kind of makes me nervous.

Maybe we could move the LAF setting to a separate method.

src/org/omegat/filters2/master/PluginUtils.java
```diff
@@ -100,6 +100,7 @@ public static void loadPlugins(final Map<String, String> params) {
             URL[] urls = new URL[fs.size()];
             for (int i = 0; i < urls.length; i++) {
                 urls[i] = fs.get(i).toURI().toURL();
+                THEME_PLUGIN_JARS.add(urls[i]); // FIXME: only required files should be added
```
amake, line 103: What is your plan for fixing this?

src/org/omegat/filters2/master/PluginUtils.java
```diff
@@ -114,6 +115,9 @@ public static void loadPlugins(final Map<String, String> params) {
                         // found main manifest - not in development mode
                         foundMain = true;
                     }
+                    if ("theme".equals(m.getMainAttributes().getValue("Plugin-Category"))) {
+                        // FIXME: add jar file to list
+                    }
```
amake, line 120: What is this supposed to be doing?

src/org/omegat/util/gui/UIDesignManager.java
```diff
@@ -69,12 +71,47 @@
...
+ * @author Hiroshi Miura
  */
 public final class UIDesignManager {
 
+    private static final Set<String> BLACKLIST = new HashSet<>();
+
+    static {
+        BLACKLIST.add("CDE/Motif");
+        BLACKLIST.add("Metal");
+        BLACKLIST.add("GTK+");
```
amake, line 83: Why these items? Are we sure no one wants to use them?

I think some people set the LAF on the command line (see https://docs.oracle.com/javase/tutorial/uiswing/lookandfeel/plaf.html) like

```shell
java -Dswing.defaultlaf=com.sun.java.swing.plaf.gtk.GTKLookAndFeel MyApp
```

Will that no longer work?

src/org/omegat/util/gui/UIDesignManager.java
```diff
@@ -69,12 +71,47 @@
...
+        for (UIManager.LookAndFeelInfo info: lafInfo) {
+            if (!BLACKLIST.contains(info.getName())) {
+                names.add(info.getName());
+            }
+        }
+        return names;
+    }
+
+    public static void registerTheme(String name, String className) {
+		UIManager.installLookAndFeel(new UIManager.LookAndFeelInfo( name, className));
```
amake, line 102: Indent looks wrong

src/org/omegat/gui/preferences/view/AppearanceController.java
```diff
@@ -40,6 +45,13 @@
 
     private AppearancePreferencesPanel panel;
 
+    private static DefaultComboBoxModel<String> themeSelectionModel = new DefaultComboBoxModel<>();
```
amake, line 48: It's mostly a convention, but the controller should not retain the model like this. Please see other preference pane examples like `GlossaryPreferencesController.java` where the model is set in `initGui`.

src/org/omegat/gui/preferences/view/AppearancePreferencesPanel.form
```diff
@@ -37,5 +37,32 @@
...
+          <Properties>
+            <Property name="text" type="java.lang.String" editor="org.netbeans.modules.i18n.form.FormI18nStringEditor">
+              <ResourceString bundle="org/omegat/Bundle.properties" key="MW_OPTIONMENU_APPEARANCE_THEME_LABEL" replaceFormat="java.util.ResourceBundle.getBundle(&quot;{bundleNameSlashes}&quot;).getString(&quot;{key}&quot;)"/>
+            </Property>
+          </Properties>
+        </Component>
+        <Component class="javax.swing.JComboBox" name="themeSelectCB">
+          <Properties>
+            <Property name="model" type="javax.swing.ComboBoxModel" editor="org.netbeans.modules.form.RADConnectionPropertyEditor">
+              <Connection code="AppearanceController.getComboBoxModel()" type="code"/>
```
amake, line 54: Avoiding this kind of hidden code is why I prefer to set the model in `initGui` per the other comment.

This code is completely invisible unless you happen to know where in NetBeans to look, so someone can very easily break the GUI without knowing it.

release/linux-specific/OmegaT
amake: Probably here:

https://github.com/omegat-org/omegat/blob/6aab48baa3b1c036fb66e75c1eba80ce41ff40e2/release/mac-specific/OmegaT.app/Contents/Info.plist#L81-L83

If you can supply a JAR of your LAF plugin I can test it.

src/org/omegat/util/gui/UIDesignManager.java
```diff
@@ -69,12 +71,47 @@
...
     private UIDesignManager() {
     }
 
+    public static List<String> getThemes() {
+        List<String> names = new ArrayList<>();
+        names.add("Default");
+        UIManager.LookAndFeelInfo[] lafInfo = UIManager.getInstalledLookAndFeels();
+        for (UIManager.LookAndFeelInfo info: lafInfo) {
+            if (!BLACKLIST.contains(info.getName())) {
+                names.add(info.getName());
```
amake, line 95: I see that we are using the theme "name" as the preferences value.

Is that safe? Are you sure the name is not localized? That it never changes across Java versions?

Would it not be safer to use the classname as the preference value, and show the display name for the combo box via a renderer? (See `GlossaryPreferencesController#initGui` for an example.)

src/org/omegat/Main.java
miurahr: I use IntelliJ IDEA.

src/org/omegat/Main.java
miurahr: Preferences is initialized in Main:172 so it's working.

miurahr:
Here is a flatlaf plugin jar for test.
Source is at https://github.com/miurahr/omegat-flatlaf

[plugin.zip](https://github.com/omegat-org/omegat/files/6564820/plugin.zip)

src/org/omegat/util/gui/UIDesignManager.java
miurahr: OK, I'd like drop black listing default LaFs.
Users always decide by themselves whether use  it or not.

src/org/omegat/gui/preferences/view/AppearanceController.java
miurahr: Change to set from initGUI().

src/org/omegat/util/gui/UIDesignManager.java
miurahr: >Would it not be safer to use the classname as the preference value, and show the display name for the combo box via a renderer? 

getName() returns String for menu
https://docs.oracle.com/javase/8/docs/api/javax/swing/UIManager.LookAndFeelInfo.html#getName--

>Returns the name of the look and feel in a form suitable for a menu or other presentation

In Java 16, it is the same.
https://docs.oracle.com/en/java/javase/16/docs/api/java.desktop/javax/swing/UIManager.LookAndFeelInfo.html#getName()

miurahr: I found  toString() is safer to save preference as identifier 
https://docs.oracle.com/en/java/javase/16/docs/api/java.desktop/javax/swing/UIManager.LookAndFeelInfo.html#toString()

miurahr:
When using  `LookAndFeelInfo.toString()` as a configuration key,`omegat.prefs` become like
```
    <theme>javax.swing.UIManager$LookAndFeelInfo[Metal javax.swing.plaf.metal.MetalLookAndFeel]</theme>
```

When using `LookAndFeelInfo.getName()` , it become like `<theme>CDE/Motif</theme>`

src/org/omegat/util/gui/UIDesignManager.java
miurahr: Now use `LookAndFeelInfo.toString()` as a key.

src/org/omegat/gui/preferences/view/AppearancePreferencesPanel.form
miurahr: Now explicitly configure it from Controller.

src/org/omegat/Main.java
miurahr: Fixed import order.

src/org/omegat/MainClassLoader.java
miurahr: Come from here
https://github.com/update4j/update4j/blob/master/src/main/java/org/update4j/DynamicClassLoader.java
and

src/org/omegat/filters2/master/PluginUtils.java
miurahr: I have not had clear idea yet.
I tried do this in L.118 at first, but `URL mu` is an URL of `MENIFEST.MF` not jar file.
So here is too late to do thing.
```
                    if ("theme".equals(m.getMainAttributes().getValue("Plugin-Category"))) {
                        // FIXME: add jar file to list
                     }
```

When I found a way to detect plugins which need to be in the System Class-Path, it can be fixed.

src/org/omegat/gui/preferences/view/AppearanceController.java
miurahr: done.

miurahr:
## Important design decision.

Current API provide  a way to register `Theme(LaF) name`  and  `LaF class`.
There is no way to customize LaF. It is also no way to do things that `UIDesignManager.installFlatDesign()` doing.

How we can improve API?

### Reference

- Customizing FlatLaf UI: https://www.formdev.com/flatlaf/customizing/

miurahr:
## Call for help - Theme icons for preference selection

Current PR has several icons with poor design. Any better icons are welcome.

src/org/omegat/MainClassLoader.java
miurahr: You can see OpenJDK AppClassLoader (what getSystemClassLoader() returns) if want.
https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/jdk/internal/loader/ClassLoaders.java#L163

That is not child of `URLClassLoader` in Java9+, but behave like a `URLClassLoader`

src/org/omegat/util/gui/UIDesignManager.java
miurahr: > Will that no longer work?

A configuration `java -Dswing.defaultlaf=com.sun.java.swing.plaf.gtk.GTKLookAndFeel MyApp` override a default `SystemClassLoader` so it working when user set theme as Default.

miurahr:
> ## Important design decision.
> Current API provide a way to register `Theme(LaF) name` and `LaF class`.
> There is no way to customize LaF. It is also no way to do things that `UIDesignManager.installFlatDesign()` doing.
> 
> How we can improve API?
> 
> ### Reference
> * Customizing FlatLaf UI: https://www.formdev.com/flatlaf/customizing/

src/org/omegat/util/gui/UIDesignManager.java
```diff
@@ -69,16 +72,84 @@
...
+
+    public static Set<String> getThemeKeySet() {
+        return themeImages.keySet();
+    }
+
+    public static URL getThemeImage(String key) {
+        return themeImages.getOrDefault(key, ResourcesUtil.getResourceURL(THEME_UNKNOWN_ICON));
+    }
+
+    public static String getThemeClassName(String key) {
```
amake, line 139: I don't understand why using `UIManager.LookAndFeelInfo#toString()` as the key is better than simply using the classname as the key. If we used the classname as the key then we wouldn't even need this particular lookup.

src/org/omegat/util/gui/UIDesignManager.java
```diff
@@ -69,16 +72,84 @@
...
  * @author Aaron Madlon-Kay
+ * @author Hiroshi Miura
  */
 public final class UIDesignManager {
 
+    public static final String THEME_DEFAULT_ICON = "thumb_default.png";
+    public static final String THEME_UNKNOWN_ICON = "thumb_unknown.png";
+
+    private static final Map<String, URL> themeImages = new HashMap<>();
+    private static final Map<String, IThemeDesigner> themeInitilizer = new HashMap<>();
```
amake, line 83: These are `static` and `final` so they should be UPPER_CASE.

Also there is a typo: `themeInitilizer` → `THEME_INITIALIZER`

src/org/omegat/gui/theme/IThemeDesigner.java
```diff
@@ -0,0 +1,38 @@
...
+ GNU General Public License for more details.
+
+ You should have received a copy of the GNU General Public License
+ along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ **************************************************************************/
+package org.omegat.gui.theme;
+
+import java.net.URL;
+
+public interface IThemeDesigner {
```
amake, line 29: I'm not sure "designer" makes sense. What about `IThemeInitializer`?

src/org/omegat/filters2/master/PluginUtils.java
```diff
@@ -114,6 +115,11 @@ public static void loadPlugins(final Map<String, String> params) {
                         // found main manifest - not in development mode
                         foundMain = true;
                     }
+                    if ("theme".equals(m.getMainAttributes().getValue("Plugin-Category"))) {
+                        String target = mu.toString();
+                        Arrays.stream(urls).filter(url -> target.contains(url.toString()))
+                                .forEach(THEME_PLUGIN_JARS::add);
```
amake, line 121: I appreciate the aesthetics of this style, but I should point out that I think this is much less performant than a simple `for` loop with an `if` statement.

src/org/omegat/MainClassLoader.java
```diff
@@ -0,0 +1,103 @@
...
+import java.io.IOException;
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.net.URLClassLoader;
+import java.nio.file.Paths;
+
+/**
+ * @author Hiroshi Miura
+ */
+public final class MainClassLoader extends URLClassLoader {
```
amake, line 39: Sorry, one more question about this class: Can you explain why we need it?

What does this class loader do that the default class loaders don't?

src/org/omegat/Main.java
```diff
@@ -251,10 +252,36 @@ private static void applyConfigFile(String path) {
...
+        try {
+            UIManager.setLookAndFeel(UIDesignManager.getThemeClassName(theme));
+            return theme;
+        } catch (Exception e) {
+            Log.log(e);
+            setDefaultTheme();
+            return Preferences.THEME_DEFAULT;
+        }
+    }
+
```
amake, line 275: This seems like maybe it should move into `UIDesignManager`.

src/org/omegat/MainClassLoader.java
amake: I should say that I don't know very much about how class loaders work. Sorry!

src/org/omegat/util/gui/UIDesignManager.java
amake: >  OK, I'd like drop black listing default LaFs.

Great, thanks.

> so it working when user set theme as Default

That's great. I'd like to avoid breaking existing setups if possible.

src/org/omegat/MainClassLoader.java
amake: Thanks.

release/linux-specific/OmegaT
amake: I tested your sample plugin and it worked: when launching the OmegaT.app, the theme combobox showed FlatLaf (light and dark) and changing the theme took effect after relaunching the app.

src/org/omegat/gui/preferences/view/AppearanceController.java
```diff
@@ -56,20 +69,90 @@ public String toString() {
...
+                ThemeLabel data,
+                int i,
+                boolean isSelected,
+                boolean b) {
+            setText(data.getText());
+            setIcon(data.getIcon());
+            return this;
+        }
+    }
+    public static class ThemeLabel {
```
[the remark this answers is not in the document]
miurahr: OK, I see.
I'm also not good at design. I'd like to drop it.

src/org/omegat/util/gui/UIDesignManager.java
miurahr: It is because LaF class can have multiple design. So classname may not be unique against theme.
toString include a name and classname that is unique.

src/org/omegat/MainClassLoader.java
miurahr: Because swing cannot find an external LaF of plugins. It causes ClassNotFound error (or NPE in Swing component) without it.

It is because Swing has a limitation, that use SystemClassLoader to find UI component even when child class loader load LaF and register the loader to UIManager. (It seems feature bug but not documented the feature).

src/org/omegat/util/gui/UIDesignManager.java
amake: Aha. Well that is a good reason, thanks.

amake: Actually I'm still confused though: to apply the theme we call `UIManager.setLookAndFeel(classname)`. If the classname is not enough to uniquely identify the theme, then how do different "designs" of the same theme take effect? Are there any examples of such themes in the real world?

I see in your sample plugin

https://github.com/miurahr/omegat-flatlaf/blob/7922c2cd57cc6315bf5b96b453f37ece6e7607f8/src/main/java/tokyo/northside/omegat/laf/CustomLookAndFeel.java#L35-L39

every theme has a different implementing class.

Even if it's hypothetically true that a single LAF class could implement multiple different "themes", unless such usage is common in the real world, I would prefer to require that each theme be a separate class. That would simplify a lot of our API design.

src/org/omegat/gui/preferences/view/AppearanceController.java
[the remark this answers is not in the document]
miurahr: Droped.

src/org/omegat/gui/theme/IThemeDesigner.java
miurahr: Changed.

src/org/omegat/Main.java
miurahr: done.

src/org/omegat/filters2/master/PluginUtils.java
miurahr: done

src/org/omegat/util/gui/UIDesignManager.java
miurahr: I've pushed a case that a theme use same LaF class with other, but provide different theme design (round corner of buttons, and triangle arrow)
https://github.com/miurahr/omegat-flatlaf/blob/master/src/main/java/tokyo/northside/omegat/laf/CustomLookAndFeel.java#L42

src/org/omegat/util/gui/ResourcesUtil.java
```diff
@@ -65,6 +65,16 @@ public static Image getImage(final String resourceName) {
         return Toolkit.getDefaultToolkit().getImage(resourceURL);
     }
 
+    /**
+     * Get Resource URL of image.
+     *
+     * @param imageName icon file name
+     * @return icon URL when found, otherwise null
+     */
+    public static URL getResourceURL(final String imageName) {
```
miurahr, line 74: Do we still need this method?

src/org/omegat/util/gui/UIDesignManager.java
```diff
@@ -69,16 +74,80 @@
...
+        return null;
+    }
+
+    public static void setDefaultTheme() {
+        try {
+            UIManager.setLookAndFeel(UIManager.getSystemLookAndFeelClassName());
+        } catch (Exception ex) {
+            // Something went wrong!!
+            Log.log(ex);
+            Main.showError(ex);
```
miurahr, line 132: Are there better way instead of calling Main class method?

src/org/omegat/util/gui/UIDesignManager.java
```diff
@@ -122,15 +191,17 @@ public static void initialize() throws IOException {
...
-        // Classic design overridden by flat design
-        //installClassicDesign();
-
-        installFlatDesign();
-
         // Panel notification (blinking tabs/headers) settings
         UIManager.put("DockingDesktop.notificationBlinkCount", 2);
         UIManager.put("DockingDesktop.notificationColor", Styles.EditorColor.COLOR_NOTIFICATION_MAX.getColor());
 
+        IThemeInitializer initializer = THEME_INITIALIZER.get(theme);
```
miurahr, line 198: It should be in line 193.

src/org/omegat/Main.java
```diff
@@ -255,6 +256,11 @@ private static void applyConfigFile(String path) {
      * Execute standard GUI.
      */
     protected static int runGUI() {
+        ClassLoader cl = ClassLoader.getSystemClassLoader();
+        MainClassLoader mainClassLoader = (cl instanceof MainClassLoader) ? (MainClassLoader) cl : new MainClassLoader(cl);
+        PluginUtils.getThemePluginJars().forEach(mainClassLoader::add);
+        UIManager.put("ClassLoader", mainClassLoader);
```
miurahr, line 262: Here is a hack to tell UIComponents to use custom class loader.
https://github.com/openjdk/jdk/blob/6765f902505fbdd02f25b599f942437cd805cad1/src/java.desktop/share/classes/javax/swing/UIDefaults.java#L784-L790

```
        Object cl = get("ClassLoader");
        ClassLoader uiClassLoader =
            (cl != null) ? (ClassLoader)cl : target.getClass().getClassLoader();
        Class<? extends ComponentUI> uiClass = getUIClass(target.getUIClassID(), uiClassLoader);
```

src/org/omegat/MainClassLoader.java
miurahr: Now I found a way without specify custom system loader for JVM argument.

src/org/omegat/util/gui/UIDesignManager.java
miurahr: Here is a plugin jar that has a "Flat Round Dark" and "FlatLaf Dark" theme which both use "FlatDarkLaf" class.

[omegat-flatlaf-0.0.1.jar.gz](https://github.com/omegat-org/omegat/files/6592018/omegat-flatlaf-0.0.1.jar.gz)

src/org/omegat/util/gui/UIDesignManager.java
miurahr: It is not a problem.

src/org/omegat/MainClassLoader.java
miurahr: From another point of view. System's default class loader don't allow adding  new URL(plugin) to search path, no method to do it. URLClassLoader class has a addURL(URL url) method that is **protected**.

https://github.com/openjdk/jdk/blob/6765f902505fbdd02f25b599f942437cd805cad1/src/java.base/share/classes/java/net/URLClassLoader.java#L389

MainClassLoader has methods, `add(URL url)` and `addJarToClassPath(String jarName)` and its scope is **default** (package local).
Only a Main class can manipulate a class search path in same security reason.

MainClassLoader#add method give a chance to loading plugins without restart in future.

src/org/omegat/gui/theme/IThemeInitializer.java
```diff
@@ -0,0 +1,34 @@
...
+ but WITHOUT ANY WARRANTY; without even the implied warranty of
+ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ GNU General Public License for more details.
+
+ You should have received a copy of the GNU General Public License
+ along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ **************************************************************************/
+package org.omegat.gui.theme;
+
+public interface IThemeInitializer {
```
miurahr, line 27: Current interface is to pass a argument for LookAndFeelInfo class constructor and setup() method.
It is natural both class has same methods, getName() and getClassName().

Why not asking theme plugins to provide an object that extends LookAndFeelInfo?
We can provide a class `public class ThemeInitializer extends LookAndFeelInfo` that have `void setup()` as a extension point.

miurahr: We can change `registerTheme` just accept `LookAndFeelInfo` class.

```
public static void registerTheme(UIManager.LookAndFeelInfo info) {
         UIManager.installLookAndFeel(info);
         THEME_KEY_SET.add(info.toString());
```

Then Default theme becomes like

```
    public static void loadPlugins() {
        UIDesignManager.registerTheme(new DefaultFlatThemeLaf());
    }

    public static class DefaultFlatThemeLaf extends UIManager.LookAndFeelInfo {

        public DefaultFlatThemeLaf() {
            super(Preferences.THEME_DEFAULT, defaultLaf);
            installFlatDesign();
        }
```

This brings simplicity for API. And also we can guarantee that a LaF class has a single design.
When developer want to customize LaF library,  simply inherit it and extend it in constructor as same as default theme.

Then we can use class name as a configuration key. 

Is it make sense?

miurahr: After some experiment, I want to forget the idea here 🧠

src/org/omegat/util/gui/UIDesignManager.java
miurahr: Here is another plugin examples.
https://github.com/miurahr/omegat-themes
[omegat-themes-0.0.1.jar.gz](https://github.com/omegat-org/omegat/files/6598610/omegat-themes-0.0.1.jar.gz)

src/org/omegat/gui/theme/IThemeInitializer.java
amake: > This brings simplicity for API. And also we can guarantee that a LaF class has a single design.
> When developer want to customize LaF library, simply inherit it and extend it in constructor as same as default theme.

I think this is a great idea. What was the problem with it?

I don't think it makes a lot of sense to have separate themes that require configuration outside of the class itself.

src/org/omegat/util/gui/UIDesignManager.java
amake: Thanks.

However I it looks like the "multiple themes from a single LAF class" is simply a convention that you invented, not something that is widely practiced in the real world, or that is somehow required by external concerns (e.g. the LAF API encourages it somehow).

I was hoping for something like your idea here: https://github.com/omegat-org/omegat/pull/100#discussion_r645269366

I would much prefer that each theme is required to be encapsulated in its own LAF class, and not require additional configuration.

Is there a reason that can't be done? Let's continue the discussion either in this thread or in https://github.com/omegat-org/omegat/pull/100#discussion_r645269366.

src/org/omegat/gui/theme/IThemeInitializer.java
miurahr:
```
    public static class DefaultFlatThemeLaf extends UIManager.LookAndFeelInfo {

        public DefaultFlatThemeLaf() {
            super(Preferences.THEME_DEFAULT, defaultLaf);
            installFlatDesign();
        }
```
A class inherit from  **LookAndFeelInfo** not **LookAndFeel** class, but I try to run `installFlatDesign()` in constructor.
It does not work.

`UIManager.put9)` should called after LaF class is registered with `UIManager.setLookAndFeel()` called.

Above code does it in constructor of  child of `LookAndFeelInfo`. It is wrong.

src/org/omegat/MainClassLoader.java
amake: > Now I found a way **without** specifying a custom system loader for JVM argument, but we still need custom MainClassLoader that has a capability to add URLs of plugins after class loader created.

I tried running without the JVM argument and I got these errors:

```
34572: Error: java.lang.NullPointerException 
34572: Error:   at java.util.concurrent.ConcurrentHashMap.putVal(ConcurrentHashMap.java:1011) 
34572: Error:   at java.util.concurrent.ConcurrentHashMap.putIfAbsent(ConcurrentHashMap.java:1535) 
34572: Error:   at java.lang.ClassLoader.getClassLoadingLock(ClassLoader.java:457) 
34572: Error:   at java.lang.ClassLoader.loadClass(ClassLoader.java:398) 
34572: Error:   at java.lang.ClassLoader.loadClass(ClassLoader.java:351) 
34572: Error:   at org.omegat.util.gui.UIDesignManager.setLookAndFeel(UIDesignManager.java:138) 
34572: Error:   at org.omegat.util.gui.UIDesignManager.setDefaultTheme(UIDesignManager.java:118) 
34572: Error:   at org.omegat.Main.runGUI(Main.java:288) 
34572: Error:   at org.omegat.Main.main(Main.java:184) 
34572: Error: java.lang.NullPointerException 
34572: Error:   at org.omegat.util.gui.UIDesignManager.initialize(UIDesignManager.java:189) 
34572: Error:   at org.omegat.core.Core.initializeGUI(Core.java:237) 
34572: Error:   at org.omegat.Main.runGUI(Main.java:295) 
34572: Error:   at org.omegat.Main.main(Main.java:184) 
```

The last error was unhandled so it crashed the process.

I think this will cause a lot of support issues if the JVM argument really becomes required. I think a lot of people run OmegaT by double-clicking OmegaT.jar, at least during troubleshooting.

amake: Actually I get those errors even if I do use the JVM args.

I invoked as

```sh
java -Djava.system.class.loader=org.omegat.MainClassLoader -jar build/install/OmegaT/OmegaT.jar
```

src/org/omegat/gui/theme/IThemeInitializer.java
amake: Thanks.

# Terminology

First, I think we should clarify some things by defining terms:

- A "LAF" is an implementation of [`LookAndFeel`](https://docs.oracle.com/javase/7/docs/api/javax/swing/LookAndFeel.html)
- A "configuration" is some values set to `UIManager` such as in `DefaultFlatTheme#installFlatDesign()` 
- A "theme" is a combination of 1) LAF and 2) configuration

# LAF and configuration

I note that `LookAndFeel` has a method `getDefaults` that returns an instance of `UIDefaults`. I believe that this is how a LAF is supposed to provide its "configuration".

# Default theme

The default theme doesn't tie itself to a particular LAF; it uses the default LAF, which depends on the system.

The default theme also has a lot of associated configuration, which we install as a secondary step (`DefaultFlatTheme#installFlatDesign()`).

The secondary step is only necessary because the LAF is not known until runtime; otherwise we could create a `LookAndFeel` class that inherits from some existing LAF and provide our configuration via `getDefaults`.

(Actually I have an idea for how to do this: create a `LookAndFeel` class that merely wraps the system LAF. I made this locally but I couldn't test it due to this error: https://github.com/omegat-org/omegat/pull/100#discussion_r646132685)

# Default configuration

When I tried your test plugin, I found the result quite ugly on macOS because `installFlatDesign` was not called. I think we might need to install at least some of this configuration by default, regardless of the theme. Or we can modularize it and make it easy for themes to adapt; see below.

# Theme plugins

I imagine that theme plugins will want to supply a LAF and configuration. I have a hard time imagining a theme plugin that would supply *only* a configuration.

If the plugin supplies a LAF, per above I believe they can supply their configuration as well from `getDefaults`.

If the plugin *only* supplies configuration, I believe it can already do so with an `IApplicationEventListener` that implements `onApplicationStartup()` (I haven't tested this; of course there are severe limitations such as the theme will always be activated if the plugin is loaded.)

# Improving the configuration

I think we could package our configuration (what happens in `installFlatDesign`) in an instance of `UIDefaults`, or make it take action on an instance of `UIDefaults` instead of directly on `UIManager`.

This would let theme plugins easily reuse (with modifications if desired) the default configuration.

What do you think?

amake: Sorry, I should be clear about action items:

1. Please give me your opinion on my points above
2. Please see if you can reproduce (and fix?) the error I mentioned in https://github.com/omegat-org/omegat/pull/100#discussion_r646132685
3. Once the error is resolved, I will test my implementation of the default theme entirely encapsulated in a LAF. If this works, we can discard the entire concept of "theme initializers"

src/org/omegat/MainClassLoader.java
miurahr: When calling `UIManager.setLookAndFeel(String name)`, it always use **SystemClassLoader**. When we want to load class that was loaded by our own class loader, it went NPE or ClassNotFound.
So I try always to use `UIManager.setLookAndFeel(LookAndFeel object)` form.
```
        Class<?> clazz = classLoader.loadClass(getThemeClassName(laf));
        UIManager.setLookAndFeel((LookAndFeel) clazz.getConstructor().newInstance());
```

When investigation of  call stack, it was not  handled  by our custom `MainClassLoader`, but `SystemClassLoader`.

```
34572: Error:   at java.lang.ClassLoader.loadClass(ClassLoader.java:351) 
34572: Error:   at org.omegat.util.gui.UIDesignManager.setLookAndFeel(UIDesignManager.java:138) 
```

miurahr: I'd like to reproduce the issue. I've tested with OpenJDK 11 on Linux, so I'd like to test with sevral JDK versions.

miurahr: I can reproduce the issue.
It happens when no theme configuration in omegat.prefs.
Current theme key is something like 
`   <theme>javax.swing.UIManager$LookAndFeelInfo[Default com.sun.java.swing.plaf.gtk.GTKLookAndFeel]</theme>`

but loading default is looking `Default` then it was not found in set of Laf information.
 
The fix is to load the default class and override configuration with proper value.

src/org/omegat/gui/theme/IThemeInitializer.java
miurahr: 1. It is better to realize your idea above.  I just failed the last trial because of my misunderstandings.
2. I've fixed the issue [No.2](https://github.com/omegat-org/omegat/pull/100#discussion_r646132685)
3. I don't have an instant idea how to realize an "encapsulating" class. I'd like to wait @amake for No.3.

miurahr: @amake I believe I've realized a LaF class implementation of the default theme entirely encapsulated in a LaF now.
Sample plugin project is also implemented as same way.
https://github.com/miurahr/omegat-flatlaf/tree/75ec61c4c8699ead4b89c48bf2d2c15ea6971c3f

amake: Thanks, and sorry I haven't been able to work on this for the last few days. I hope to get back to it this weekend.

Your implementation is mostly in line with what I was working on; just a few details are different. I'll give it a closer look soon.

amake (commented):
I've pushed my changes and added commentary below.

The main additional change needed is that, per the comment below, I think `UIDesignManager.registerTheme` and the `THEME_KEY_SET` mechanism is no longer needed and should be removed.

I am planning to do another round of cleanup and testing before next week, but please feel free to make more changes in the meantime.

amake:
I've made the changes I was planning to make.

Can you see if this lets you implement your theme plugin?

src/org/omegat/gui/theme/DefaultFlatTheme.java
```diff
@@ -0,0 +1,272 @@
...
+        // #EEEEEE -> #575757; Darkest border
+        Color statusAreaColor = adjustRGB(standardBgColor, 0x57 - 0xEE);
+
+        // General highlight & shadow used in a lot of places
+        //
+        // FIXME: VLDocking values have to be set to the "developer defaults"
+        // not the "LAF defaults" because that's where
+        // DockingUISettings#installUI puts them
+        UIManager.put("VLDocking.highlight", activeTitleBgColor);
+        UIManager.put("VLDocking.shadow", statusAreaColor);
```
miurahr, line 116: question: Are there any reason `VLDocking*` parameters should set through `UIManager.put`, not set `defaults`?

src/org/omegat/gui/theme/DefaultFlatTheme.java
```diff
@@ -44,113 +43,71 @@
...
+     * Apply theme default values to the supplied {@link UIDefaults} object.
+     * Suitable for third-party themes that want to modify the default theme.
+     *
+     * @param defaults the {@link UIDefaults} object such as from
+     *                 {@link LookAndFeel#getDefaults()}
+     * @param lafId    the ID of the LAF that will make use of the defaults (from
+     *                 {@link LookAndFeel#getID()})
+     * @return the modified {@link UIDefaults} object
+     */
+    public static UIDefaults setDefaults(UIDefaults defaults, String lafId) {
```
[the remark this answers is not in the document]
miurahr: I'd like to try using `setDefaults(...)` from sample 3rd-party plugin, and give a feedback here.

src/org/omegat/gui/theme/DefaultFlatTheme.java
amake: Yes. I mentioned it in the comment right above here, but we should probably document it better elsewhere too.

`UIManager` manages three "layers" of "defaults" (= sets of key-value settings). From [the Javadoc](https://docs.oracle.com/javase/7/docs/api/javax/swing/UIManager.html):

> 1. Developer defaults. With few exceptions Swing does not alter the developer defaults; these are intended to be modified and used by the developer.
> 2. Look and feel defaults. The look and feel defaults are supplied by the look and feel at the time it is installed as the current look and feel (`setLookAndFeel()` is invoked). The look and feel defaults can be obtained using the `getLookAndFeelDefaults()` method.
> 3. Sytem defaults. The system defaults are provided by Swing.

In our `setDefaults` implementation we can only set values on layer 2 (LAF defaults). However VLDocking's [`installUI`](https://github.com/omegat-org/vldocking/blob/1a786b686b4e94bc8a63e477a592ea4e7ee2c71b/src/main/java/com/vlsolutions/swing/docking/ui/DockingUISettings.java#L626) puts all of its settings into layer 1.

We can't override the VLDocking defaults from layer 2, so we have no choice but to put them in layer 1 via `UIManager#put*`.

I think the correct solution is to change VLDocking: it should let you install its defaults into layer 2. But in the meantime we can work around it by using `UIManager#put*` in our `setDefaults` method.

src/org/omegat/gui/theme/DefaultFlatTheme.java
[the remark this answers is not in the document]
miurahr: There are 3 feedbacks from 3rd-party plugin implementation,

1. UIDesignManager#initialize should call `loadColors()` after `setTheme` because it fails to detect dark custom theme, then it load light color sheme even when dark theme applied.
2. Second argument `lafId` of `setDefaults` should be changed. It is used to detect OS(Windows) but  custom LaF name does not work for it.
3. 3rd-party LaF also need to use `UIManager.put` for VLDocking related customize. 
`        UIManager.put("DockViewTitleBar.border", new MatteBorder(1, 1, 1, 1, borderColor));`


image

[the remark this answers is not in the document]
miurahr: There is future improvement space in `loadColors`. Current implementation only have 2 fixed defaults, light and dark.
Because it depends on theme color scheme,  we should allow theme provider to customize colors.

When we or plugin developer provide a high-contrast theme for weak-eyed users, it is going to be a big problem.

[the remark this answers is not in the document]
miurahr: When `setDefaults` param `lafId` intend to accept `systemLookAndFeel`  in any case, eg. even in 3rd-paty plugins, it is not necessary to get it as argument.

src/org/omegat/gui/theme/DefaultFlatTheme.java
miurahr: Got it.

 I'd like to move a sample plugin to be a `theme-skeleton` and encourage other users to propose better themes.
That is because, I've put the description on the plugin project's README.

src/org/omegat/gui/theme/DefaultFlatTheme.java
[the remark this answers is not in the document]
amake: > There are 3 feedbacks from 3rd-party plugin implementation,

Thanks for trying it out.

>     1. UIDesignManager#initialize should call `loadColors()` after `setTheme` because it fails to detect dark custom theme, then it load light color sheme even when dark theme applied.

Thanks for making this change.

>     2. Second argument `lafId` of `setDefaults` should be changed. It is used to detect OS(Windows) but  custom LaF name does not work for it.

I'm not sure what we should do with this. It's not detecting Windows the OS; it's detecting a specific Windows LAF. For other LAFs I think it's OK to pass the LAF ID and it simply won't recognize it.

>     3. 3rd-party LaF also need to use `UIManager.put` for VLDocking related customize.
>        `        UIManager.put("DockViewTitleBar.border", new MatteBorder(1, 1, 1, 1, borderColor));`

Yes. Until we can improve VLDocking to make it possible for it to install its defaults in the LAF defaults, we will need to document that third-party LAFs must use this workaround.

amake (approved):
I am 99% satisfied with the state of this PR. Are there any other changes you'd like to make before I merge it?

The last 1% is that I would like to double-check some things with a third-party LAF. Could you make a new JAR of your plugin available?

src/org/omegat/gui/theme/DefaultFlatTheme.java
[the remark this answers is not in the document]
miurahr: > I'm not sure what we should do with this. It's not detecting Windows the OS; it's detecting a specific Windows LAF. For other LAFs I think it's OK to pass the LAF ID and it simply won't recognize it.

It is now very clear. I've satisfied the PR 100%.

miurahr:
> Could you make a new JAR of your plugin available?

Here is it.

[omegat-flatlaf-0.0.1.jar.gz](https://github.com/omegat-org/omegat/files/6644132/omegat-flatlaf-0.0.1.jar.gz)

amake:
Thanks for your patience and all your help with this. I've squashed to `master`.

PR #102, opened 2021-05-30
fix formatting of stats table header declarations

brandelune:
As per https://sourceforge.net/p/omegat/feature-requests/1571/


amake:
I've committed to `master` with some adjustments.

PR #103, opened 2021-05-30
Topics/jc/shorten with without

brandelune:
This is for readability purposes in the stats window:
without → w/o
with → +
basically that only affects "with spaces" or "without spaces" for character counts

### DIFF
--- a/src/org/omegat/Bundle.properties
+++ b/src/org/omegat/Bundle.properties
@@ -700,7 +700,7 @@ CT_STATS_Words=Words
 
 CT_STATS_Characters_NOSP=Characters (w/o spaces)
 
-CT_STATS_Characters=Characters (+ spaces)
+CT_STATS_Characters=Characters (w/ spaces)
 
 CT_STATS_Files=#Files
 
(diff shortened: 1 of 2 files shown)


src/org/omegat/Bundle.properties
```diff
@@ -698,9 +698,9 @@ CT_STATS_Segments=Segments
 
 CT_STATS_Words=Words
 
-CT_STATS_Characters_NOSP=Characters (without spaces)
+CT_STATS_Characters_NOSP=Characters (w/o spaces)
 
-CT_STATS_Characters=Characters (including spaces)
+CT_STATS_Characters=Characters (+ spaces)
```
amake, line 703: I feel like "+" is confusing in contrast to "w/o". It makes me think that the number is not simply characters *including* spaces, but that we are "adding" spaces, or something (even if that logically doesn't make sense, it creates doubt in my mind.)

I would prefer "w/" as an abbreviation for "with".

src/org/omegat/core/statistics/StatsResult.java
```diff
@@ -42,19 +42,27 @@
...
-            OStrings.getString("CT_STATS_Characters"), OStrings.getString("CT_STATS_Files") };
+            OStrings.getString("CT_STATS_Words"),
+            OStrings.getString("CT_STATS_Characters_NOSP"),
+            OStrings.getString("CT_STATS_Characters"),
+            OStrings.getString("CT_STATS_Files") };
 
     private static final String[] HT_ROWS = { OStrings.getString("CT_STATS_Total"),
-            OStrings.getString("CT_STATS_Remaining"), OStrings.getString("CT_STATS_Unique"),
+            OStrings.getString("CT_STATS_Remaining"),
+            OStrings.getString("CT_STATS_Unique"),
```
amake, line 52: These seem to be formatting-only changes. Their presence obscures the semantic changes, so in general I recommend that you not include these in the same PR (or at least put them in their own commit that is labeled appropriately).

amake (commented):
I left the meat of my comments here:
https://sourceforge.net/p/omegat/feature-requests/1571/#e479

src/org/omegat/core/statistics/StatsResult.java
brandelune: Indeed. For some reason I forgot to remove them from the commit. Sorry.

src/org/omegat/Bundle.properties
brandelune: You're right. Thank you.

src/org/omegat/core/statistics/StatsResult.java
brandelune: I've reverted the display modifications and recommitted only the formatting modifications

amake:
I've committed the column wording changes ("w/" & "w/o") to master.

I think the other changes will need more discussion, and should probably be a separate PR.

PR #105, opened 2021-06-06
XMLmind changes

brandelune:
It looks like XMLmind settings have changed and \&quot\; is now literal, along with page width modifications that impact some URLs too.
@vbdx asked me to check if the manual was still properly built, which seems to be the case (PDF and HTML5 output), and asked me to push. I'm hoping he'll check the diff before anything happens.

### DIFF
--- a/doc_src/en/App_Dictionaries.xml
+++ b/doc_src/en/App_Dictionaries.xml
@@ -14,8 +14,9 @@
     <itemizedlist>
       <listitem>
         <para>Search for the required language combination - for instance on
-        the <ulink url="https://sourceforge.net/p/omegat/wiki/Reference%20Material/">dictionary links given by the OmegaT
-        Wiki</ulink>.</para>
+        the <ulink
+        url="https://sourceforge.net/p/omegat/wiki/Reference%20Material/">dictionary
+        links given by the OmegaT Wiki</ulink>.</para>
       </listitem>
     </itemizedlist>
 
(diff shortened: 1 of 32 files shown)


amake:
Sorry, I'm confused.

Are you talking about XMLmind the editing software? Or are you talking about the [XMLmind Web Help Compiler](https://www.xmlmind.com/ditac/whc.shtml)?

In other words is the the issue simply how the documents appear *at edit time*, or is the compiled *output* the problem?

I am wary of making this change without fully understanding the problem, because if `&apos;` is not interpreted correctly then other character entities probably won't be either, and we have a bigger problem on our hands: not only must we purge any entities, but we must somehow prevent them from being added in the future.

If possible I would prefer to fix whatever software is not handling entities correctly.

brandelune:
Vincent (Linux) and I (Mac) use XMLMind, the xml editor, for editing the docbook. The intern who works on the manual has also installed the last version on Windows, and we found that the latest default settings automatically converted the \&quot\; (and only them) and line width. Since that added a lot of noise to the interns commits (made on my fork), Vincent suggested that we make a full conversion of the manual with the XMLMind current default settings. He also decided to check with the developer (who is in France) to see what was the cause of that modification.

So, right now, just consider that a pending issue while we're waiting for the developer's answer to that.

amake:
> we found that the latest default settings automatically converted the &quot; (and only them) and line width. Since that added a lot of noise to the interns commits

Ah, OK actually that's fine then. I was worried that there was some tool in the pipeline screwing something up, but if that's not the case then I'm fine with normalizing to whatever XMLmind likes if that makes people's lives easier.

amake:
Unfortunately this PR is now conflicted. Sorry for letting it sit for so long.

If the latest version of XMLmind has a new-but-equivalent preferred representation for the XML source, then I'm fine with normalizing all the source to that if it makes anyone's life easier.

But since the source has since changed, I think the normalization needs to be redone with the latest source.

brandelune:
No issue with the delay. Thank you for looking into it. I guess Vincent and I will propose a new PR when we have the new manual. Would that work better ?

amake:
Sure. In that case should we close this PR?

PR #106, opened 2021-06-08
Update HttpConnectilUtils#downloadBinaryFile

miurahr:
- Use IOUtils.copy for downloading content
- Raise FlakyDownloadError when content size is wrong
- Change MIME type to check against

### DIFF
--- a/src/org/omegat/util/HttpConnectionUtils.java
+++ b/src/org/omegat/util/HttpConnectionUtils.java
@@ -141,18 +141,12 @@ public final class HttpConnectionUtils {
         int responseCode = httpURLConnection.getResponseCode();
         if (responseCode == HttpURLConnection.HTTP_OK) {
             String contentType = httpURLConnection.getContentType();
-            int contentLength = httpURLConnection.getContentLength();
+            long contentLength = httpURLConnection.getContentLength();
             if (contentType.equals("application/octet-stream") || contentType.equals("application/jar-archive")) {
                 try (InputStream inputStream = httpURLConnection.getInputStream();
                      FileOutputStream outputStream = new FileOutputStream(saveFilePath)) {
-                    int bytesRead = -1;
-                    int readLength = 0;
-                    byte[] buffer = new byte[BUFFER_SIZE];
-                    while ((bytesRead = inputStream.read(buffer)) != -1) {
-                        readLength += bytesRead;
-                        outputStream.write(buffer, 0, bytesRead);
-                    }
-                    assert (contentLength == readLength);
+                    long transferred = IOUtils.copy(inputStream, outputStream, BUFFER_SIZE);
+                    assert(transferred == contentLength);
                     result = true;
                 } catch (Exception ex) {
                     Log.log(ex);
--- a/src/org/omegat/util/HttpConnectionUtils.java
+++ b/src/org/omegat/util/HttpConnectionUtils.java
@@ -43,6 +43,7 @@ import java.nio.charset.StandardCharsets;
 import java.util.Base64;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 
 import org.apache.commons.io.IOUtils;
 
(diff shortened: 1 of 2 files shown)


src/org/omegat/util/HttpConnectionUtils.java
```diff
@@ -142,22 +142,29 @@ public static boolean downloadBinaryFile(final URL jarFileURL, final Map<String,
...
+            if (contentType.equals("application/octet-stream") || contentType.equals("application/jar-archive")) {
                 try (InputStream inputStream = httpURLConnection.getInputStream();
                      FileOutputStream outputStream = new FileOutputStream(saveFilePath)) {
                     int bytesRead = -1;
                     int readLength = 0;
                     byte[] buffer = new byte[BUFFER_SIZE];
                     while ((bytesRead = inputStream.read(buffer)) != -1) {
+                        readLength += bytesRead;
                         outputStream.write(buffer, 0, bytesRead);
                     }
```
amake, line 154: I bet there's something that we can replace this with from Apache IOUtils.

src/org/omegat/util/HttpConnectionUtils.java
```diff
@@ -141,23 +141,24 @@ public static boolean downloadBinaryFile(final URL jarFileURL, final Map<String,
...
                 try (InputStream inputStream = httpURLConnection.getInputStream();
                      FileOutputStream outputStream = new FileOutputStream(saveFilePath)) {
-                    int bytesRead = -1;
-                    int readLength = 0;
-                    byte[] buffer = new byte[BUFFER_SIZE];
-                    while ((bytesRead = inputStream.read(buffer)) != -1) {
-                        outputStream.write(buffer, 0, bytesRead);
-                    }
+                    long transferred = IOUtils.copy(inputStream, outputStream, BUFFER_SIZE);
+                    assert(transferred == contentLength);
```
amake, line 149: What is your intent for this assert?

Asserts don't run unless you explicitly enable them, and we don't enable them, so this won't do anything in a release build.

Is that OK? Did you want to raise an exception, or log something instead?

src/org/omegat/util/HttpConnectionUtils.java
amake: Thanks!

src/org/omegat/util/HttpConnectionUtils.java
miurahr: Oh, this is python way. I'd like to raise exception

amake: Sounds good. FYI see [here](https://docs.oracle.com/javase/7/docs/technotes/guides/language/assert.html#enable-disable) for details on Java assertions.

src/org/omegat/util/HttpConnectionUtils.java
```diff
@@ -141,23 +141,26 @@ public static boolean downloadBinaryFile(final URL jarFileURL, final Map<String,
...
                 try (InputStream inputStream = httpURLConnection.getInputStream();
                      FileOutputStream outputStream = new FileOutputStream(saveFilePath)) {
-                    int bytesRead = -1;
-                    int readLength = 0;
-                    byte[] buffer = new byte[BUFFER_SIZE];
-                    while ((bytesRead = inputStream.read(buffer)) != -1) {
-                        outputStream.write(buffer, 0, bytesRead);
+                    long transferred = IOUtils.copy(inputStream, outputStream, BUFFER_SIZE);
+                    if (transferred != contentLength) {
+                        throw new FlakyDownloadException("Downloaded file length and content length of header is differ.");
```
amake, line 150:
```suggestion
                        throw new FlakyDownloadException("Downloaded file length differs from expected content length reported in header.");
```

miurahr:
I recognize that downloadBinaryFile() method say 'binary file' but 1st argument is 'jarFileUrl'.
It checks against MIME types with one when jar file is downloaded.

It is better to change argument name,  and add argument `mime` that is  `Set<String>` that is used to check against.

amake:
The code was getting complicated with many levels of nesting and error handling far removed from the checks.

I refactored into a guard/early-return style. Please let me know if this is working for you. If so, I will squash to master.

miurahr:
Thank you. It makes sense.

amake:

PR #108, opened 2021-06-12
WIP: Introduce RemotePluginInformation, container of remote plugins metadata

miurahr:
- RemotePluginInformation extends PluginInformation that can hold
  URL of plugin file to be downloaded, and its SHA256 checksum.
- Add primitive test for class.

This is spin-off of PR #69

### DIFF
--- a/src/org/omegat/core/data/RemotePluginInformation.java
+++ b/src/org/omegat/core/data/RemotePluginInformation.java
@@ -1,0 +1,59 @@
+/**************************************************************************
+ OmegaT - Computer Assisted Translation (CAT) tool
+          with fuzzy matching, translation memory, keyword search,
+          glossaries, and translation leveraging into updated projects.
+
+ Copyright (C) 2021 Hiroshi Miura
+               Home page: http://www.omegat.org/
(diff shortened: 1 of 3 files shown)


src/org/omegat/core/data/RemotePluginInformation.java
```diff
@@ -0,0 +1,59 @@
...
+    private final String remoteJarFileUrl;
+    private final String jarFilename;
+    private final String sha256Sum;
+
+    public RemotePluginInformation(String className, Manifest manifest) {
+        super(className, manifest);
+        Attributes attrs = manifest.getMainAttributes();
+        remoteJarFileUrl = attrs.getValue(PLUGIN_JAR_URL);
+        jarFilename = attrs.getValue(PLUGIN_JAR_FILENAME);
+        sha256Sum = attrs.getValue(PLUGIN_SHA256SUM);
```
miurahr, line 45: Because these values are given from external source, we need to validate and null-check.

miurahr:
I've changed PluginInformation data structure in #69 , so I'd like to abandon and close here.

PR #109, opened 2021-06-12
Feature: Provide UI to install a plugin file from local file system (RFE#1548)

miurahr:
Users can install and upgrade plugin from jar and zip file on dialog launched at plugins preference screen.

It allow users
- Add install from disk button on preference->plugins
- Allow upgrade plugin from dialog, when upgrading, remove old plugin on exit.
- Allow install plugin from zip archive that contains jar.

Related [RFE#1548](https://sourceforge.net/p/omegat/feature-requests/1548/)  that ask a plugin manager and installer

## there is several test cases to be checked

- [x] install new plugin jar file.
- [x] install new plugin from zip file.
- [x] upgrade newer version of plugin from jar file.
- [x] upgrade newer version of plugin from zip file.
- [x] downgrade plugin from jar file.
- [x] downgrade plugin from zip file.
- [x] override same version from jar file.
- [x] override same version from zip file.


src/org/omegat/gui/preferences/view/PluginsPreferencesController.java
```diff
@@ -68,6 +75,29 @@ private void initGui() {
                         OStrings.getString("ERROR_TITLE"), JOptionPane.ERROR_MESSAGE);
             }
         });
+        panel.restartOmegatButton.setText(OStrings.getString("PREFS_PLUGINS_RESTART"));
+        panel.restartOmegatButton.addActionListener(e -> {
+            String projectDir = Core.getProject().isProjectLoaded()
+                    ? Core.getProject().getProjectProperties().getProjectRoot()
+                    : null;
+            MainWindowMenuHandler.prepareForExit((MainWindow) Core.getMainWindow(), () -> {
```
miurahr, line 83: Should it be `Core.getMainWindow().getApplicationFrame()`?

miurahr:
I'd like to refactor utility methods,...

miurahr:
I'd like to refactor and move PluginInstaller into org.omegat.utils static utility.

miurahr:
Now ready to merge.

miurahr:
re-based on master

miurahr:
Any blocker for merge?

miurahr:
Re-based master and adjust to use PluginInformation.Builder.

miurahr:
OmegaT show dialog when upgrade/downgrade/override plugin with versions before and after.
It always say "upgrade?" even when downgrade. It should check version string and switch message "upgrade" "downgrade" and "override"

miurahr:
As you know OmegaT plugin should have manifest that have entries.

Here is a developer manual that OmegaT distribution provided.
https://github.com/omegat-org/omegat/blob/master/docs_devel/HowToCreateFilterPlugin.md#manifest


Attribute | Manifest entry (pick one)
-- | --
Name | Plugin-Name, Bundle-Name, Implementation-Title
Version | Plugin-Version, Bundle-Version, Implementation-Version
Author | Plugin-Author, Implementation-Vendor, Built-By
Description | Plugin-Description
Link | Plugin-Link
License | Plugin-License
Category | Plugin-Category

Stax filter has manifest as follows
```
Manifest-Version: 1.0
Ant-Version: Apache Ant 1.10.12
Created-By: 1.8.0_342-8u342-b07-0ubuntu1~22.04-b07 (Private Build)
OmegaT-Plugin: true

Name: org.omegat.filters4.xml.xliff.Xliff1Filter
OmegaT-Plugin: filter

Name: org.omegat.filters4.xml.xliff.Xliff2Filter
OmegaT-Plugin: filter

Name: org.omegat.filters4.xml.xliff.SdlXliff
OmegaT-Plugin: filter

Name: org.omegat.filters4.xml.xliff.SdlProject
OmegaT-Plugin: filter

Name: org.omegat.filters4.xml.xliff.MqXliff
OmegaT-Plugin: filter

Name: org.omegat.filters4.xml.openxml.MsOfficeFileFilter
OmegaT-Plugin: filter
```

There are no compliant entries. The plugin manager should show warning and abort installation.

miurahr:
The patch parses a plugin jar file manifest to look for "OmegaT-Plugins" to get class name,  and also parse, "name", "version" properties from manifest. It shows these values on dialog, and Stax plugin don't have any of these properties, so dialog show wrong name, and version.

miurahr:
I'd like to add following clause in developer manual.

```
OmegaT 5.8.0 can install a plugin file (jar or zip) from GUI, that has mandatory information.
You **should** provide name, and version.

OmegaT 6.1.0 is going to be able to install plugins from remote.
In order to support remote installation, You **should** provide all requested information.
The manifest should be a form of OmegaT 3.0.1 and up compliant.
```

Any comment?

miurahr:
> And the plugin does not appear in the plugin list.

Do you restart OmegaT?

brandelune:
> > And the plugin does not appear in the plugin list.
> 
> Do you restart OmegaT?

No. I did not expect it to be necessary for the plugin to appear in the list. Thank you for pointing that out.

For a better UX we thus need proper error messages for incompatible plugins *but* since install is just copying the plugin to the plugin/ folder, I don't think we should have a problem installing "old" plugins.

OmegaT should just know when a plugin is compatible and handle it regardless of the provided information fields.

Maybe that could be a requirement for remote install, for security reasons, but I think a plugin that has been downloaded by the user should be accepted as valid, regardless of the meta information.

miurahr:
I've updated to accept old plugin on the GUI installer feature.
@brandelune  could you test it again?

I've also added several tests against related methods.

miurahr:
Now it raises error
```
02404: Error: Plugin installation failed: Mandatory propety is not found in jar file. Is it really an OmegaT plugin? (PREFS_PLUGINS_INSTALLATION_FAILED)
```
uum..

brandelune:
I don't have any error message but it is not displayed in the list, even after relaunch, and it is not copied anywhere I can tell.

I tried with FolderMenu.jar, time-tracker-2.1.1.jar and I get the same result: no confirmation after selecting the downloaded file, the list of OmegaT plugins is displayed again but without any new plugin appearing in the list.

Could you give me the address of a plugin that works for your ?

miurahr:
I've fixed the error. Please pull a commit.

> Now it raises error
> 
> ```
> 02404: Error: Plugin installation failed: Mandatory propety is not found in jar file. Is it really an OmegaT plugin? (PREFS_PLUGINS_INSTALLATION_FAILED)
> ```

src/org/omegat/Bundle.properties
```diff
@@ -2876,6 +2883,13 @@ PREFERENCES_WARNING_NEEDS_RELOAD=Reload the current project to apply these chang
 PREFS_TITLE_PLUGINS=Plugins
 PREFS_PLUGINS_BROWSE_ONLINE=&Browse Online
 PREFS_PLUGINS_AVAILABLE_ONLINE=There are many plugins available for OmegaT online.
+PREFS_PLUGINS_INSTALL_FROM_DISK=&Install plugin from file
+PREFS_PLUGINS_TITLE_CONFIRM_INSTALLATION=Install plugin file
+PREFS_PLUGINS_CONFIRM_UPGRADE=Upgrade/override plugin {0}\n\
+   current: {1}\n\
+   new version: {2}
+PREFS_PLUGINS_CONFIRM_INSTALL=Install plugin {0} {1}
```
miurahr, line 2891:
```suggestion
PREFS_PLUGINS_CONFIRM_INSTALL=Install `{0}` plugin, version `{1}`
```

It is better to insert "version" because when old plugin loaded , version number become "unknown".
When old plugin loaded, it will be like  "Install `example filter` plugin, version `unkown`"

miurahr:
> I don't have any error message but it is not displayed in the list, even after relaunch, and it is not copied anywhere I can tell.

It seems better we show dialog with error message, instead of only leaving a log message when installation failed, or finished.

miurahr:
I've squashed commits with small refinement

brandelune:
Ok, it builds now. Thank you.

miurahr:
I believe now it is an enough quality to merge into 5.8.0/6.0.0.

brandelune:
Ok, I'm moving the milestone to 5.8.

PR #111, opened 2021-06-15
Fix coding style

miurahr:
- Fix indent that is tab character
- Fix trailing space, spaces both side of operators


amake:

PR #112, opened 2021-06-22
Setup encoding sig of new writable glossary

miurahr:
- Detect glossary encoding as UTF-8 when find a line `-*- coding: utf-8 -*-`.
- New project create writable glossary with encoding signature
- Team init tool also create it with encoding signature

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -104,7 +104,6 @@ dependencies {
     } else {
         implementation 'commons-io:commons-io:2.7'
         implementation 'commons-lang:commons-lang:2.6'
-        implementation 'com.google.guava:guava:16.0.1'
         api 'org.slf4j:slf4j-api:1.7.25'
         implementation 'org.slf4j:slf4j-jdk14:1.7.25'
 
--- a/build.gradle
+++ b/build.gradle
@@ -104,6 +104,7 @@ dependencies {
     } else {
         implementation 'commons-io:commons-io:2.7'
         implementation 'commons-lang:commons-lang:2.6'
+        implementation 'com.google.guava:guava:16.0.1'
         api 'org.slf4j:slf4j-api:1.7.25'
         implementation 'org.slf4j:slf4j-jdk14:1.7.25'
 
--- a/src/org/omegat/core/data/RealProject.java
+++ b/src/org/omegat/core/data/RealProject.java
@@ -35,7 +35,6 @@
 
 package org.omegat.core.data;
 
-import java.io.BufferedWriter;
 import java.io.File;
 import java.io.IOException;
 import java.io.RandomAccessFile;
@@ -61,9 +60,10 @@ import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
-import gen.core.filters.Filters;
 import org.madlonkay.supertmxmerge.StmProperties;
 import org.madlonkay.supertmxmerge.SuperTmxMerge;
+import org.xml.sax.SAXParseException;
+
 import org.omegat.CLIParameters;
 import org.omegat.core.Core;
 import org.omegat.core.CoreEvents;
@@ -85,6 +85,7 @@ import org.omegat.filters2.IAlignCallback;
 import org.omegat.filters2.IFilter;
 import org.omegat.filters2.master.FilterMaster;
 import org.omegat.gui.glossary.GlossaryEntry;
+import org.omegat.gui.glossary.GlossaryManager;
 import org.omegat.gui.glossary.GlossaryReaderTSV;
 import org.omegat.tokenizer.DefaultTokenizer;
 import org.omegat.tokenizer.ITokenizer;
@@ -104,7 +105,8 @@ import org.omegat.util.StringUtil;
 import org.omegat.util.TMXReader2;
 import org.omegat.util.TagUtil;
 import org.omegat.util.gui.UIThreadsUtil;
-import org.xml.sax.SAXParseException;
+
+import gen.core.filters.Filters;
 
 /**
  * Loaded project implementation. Only translation could be changed after project will be loaded and set by
@@ -279,14 +281,9 @@ public class RealProject implements IProject {
             createDirectory(config.getDictRoot(), OConsts.DEFAULT_DICT);
             createDirectory(config.getTargetRoot(), OConsts.DEFAULT_TARGET);
             //createDirectory(m_config.getTMOtherLangRoot(), OConsts.DEFAULT_OTHERLANG);
-            saveProjectProperties();
+            GlossaryManager.createNewWritableGlossaryFile(config.getWritableGlossaryFile());
 
-            File glossaries = config.getWritableGlossaryFile().getAsFile();
-            if (glossaries.createNewFile()) {
-                try (BufferedWriter writer = Files.newBufferedWriter(glossaries.toPath())) {
-                    writer.write("# Glossaries with tab separation -*- coding: utf-8 -*-");
-                }
-            }
+            saveProjectProperties();
 
             // Set project specific segmentation rules if they exist, or
             // defaults otherwise.
--- a/src/org/omegat/core/data/RealProject.java
+++ b/src/org/omegat/core/data/RealProject.java
@@ -60,7 +60,6 @@ import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
-import gen.core.filters.Filters;
 import org.madlonkay.supertmxmerge.StmProperties;
 import org.madlonkay.supertmxmerge.SuperTmxMerge;
 import org.omegat.CLIParameters;
@@ -104,6 +103,8 @@ import org.omegat.util.TMXReader2;
 import org.omegat.util.TagUtil;
 import org.omegat.util.gui.UIThreadsUtil;
 import org.xml.sax.SAXParseException;
+
+import gen.core.filters.Filters;
 
 /**
  * Loaded project implementation. Only translation could be changed after project will be loaded and set by
--- a/src/org/omegat/core/data/RealProject.java
+++ b/src/org/omegat/core/data/RealProject.java
@@ -35,6 +35,7 @@
 
 package org.omegat.core.data;
 
+import java.io.BufferedWriter;
 import java.io.File;
 import java.io.IOException;
 import java.io.RandomAccessFile;
@@ -60,6 +61,7 @@ import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
+import gen.core.filters.Filters;
 import org.madlonkay.supertmxmerge.StmProperties;
 import org.madlonkay.supertmxmerge.SuperTmxMerge;
 import org.omegat.CLIParameters;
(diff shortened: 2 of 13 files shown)


build.gradle
```diff
@@ -104,6 +104,7 @@ dependencies {
     } else {
         implementation 'commons-io:commons-io:2.7'
         implementation 'commons-lang:commons-lang:2.6'
+        implementation 'com.google.guava:guava:16.0.1'
```
miurahr, line 107: Guava has already been in dependencies through languagetools.

src/org/omegat/util/EncodingDetector.java
```diff
@@ -54,17 +56,32 @@ public static String detectEncoding(File inFile) throws IOException {
      * @see <a href="https://github.com/amake/juniversalchardet">Fork</a>
      */
     public static String detectEncoding(InputStream stream) throws IOException {
-        UniversalDetector detector = new UniversalDetector(null);
 
+        byte[] signature = {'-', '*', '-', ' ', 'c', 'o', 'd', 'i', 'n', 'g',
+        ':', ' ', 'u', 't', 'f', '-', '8', ' ', '-', '*', '-'};  //  -*- coding: utf-8 -*-
```
miurahr, line 61: Currently only detect utf-8 but it can be extend to detect arbitrary encoding

src/org/omegat/util/EncodingDetector.java
```diff
@@ -54,17 +56,32 @@ public static String detectEncoding(File inFile) throws IOException {
      * @see <a href="https://github.com/amake/juniversalchardet">Fork</a>
      */
     public static String detectEncoding(InputStream stream) throws IOException {
-        UniversalDetector detector = new UniversalDetector(null);
```
amake, line 57: I'm not sure I like mixing this new encoding marker scheme in with `EncodingDetector`.

The approaches are very different, and I don't think it makes sense to do them in the same place, or with every file.

Instead, I think I would suggest that for glossary only (like in `GlossaryReaderTSV`) we do:

1. Have a separate detecting method that reads a single line of the file as ASCII and looks for the magic comment
2. If that fails, detect with `EncodingDetector` as we already do

src/org/omegat/core/data/RealProject.java
```diff
@@ -279,9 +279,15 @@ public void createProject() {
...
-
             saveProjectProperties();
 
+            File glossaries = config.getWritableGlossaryFile().getAsFile();
+            if (glossaries.createNewFile()) {
+                try (BufferedWriter writer = Files.newBufferedWriter(glossaries.toPath())) {
+                    writer.write("# Glossaries with tab separation -*- coding: utf-8 -*-");
+                }
+            }
+
```
amake, line 290: This should probably be abstracted out to something like `GlossaryReaderTSV#createEmpty`.

miurahr:
Now ready for review.

src/org/omegat/core/data/RealProject.java
```diff
@@ -279,6 +281,7 @@ public void createProject() {
             createDirectory(config.getDictRoot(), OConsts.DEFAULT_DICT);
             createDirectory(config.getTargetRoot(), OConsts.DEFAULT_TARGET);
             //createDirectory(m_config.getTMOtherLangRoot(), OConsts.DEFAULT_OTHERLANG);
+            GlossaryManager.createNewWritableGlossaryFile(config.getWritableGlossaryFile());
```
amake, line 284: This is completely new, right? Before we were only creating the writable glossary file on-demand, but now we are doing it proactively on project creation.

Is there a reason to do it here instead of on-demand?

build.gradle
amake, line 107: I think you're not using this anymore. If so, it should be removed.

src/org/omegat/util/StaticUtils.java
```diff
@@ -579,4 +580,19 @@ public static boolean isProjectDir(File f) {
...
+        }
+        return null;
+    }
+
+    private static boolean isAsciiBytes(byte[] data, int size) {
+        for (int i = 0; i < size; i++) {
+            if (data[i] < 0) return false;
+        }
+        return true;
+    }
```
amake, line 597: I think it's a lot easier to use a `Reader` to read in the first line. Is there a reason you went with this implementation instead?

Also I think you're not buffering the reading, so this might be hitting the filesystem up to `maxLength` times(?).

src/org/omegat/core/data/RealProject.java
miurahr: Oh, I do it for test. I'd like to remove.

miurahr: But it may be better to create on project creation because now we put a help comment.
Reported bug may happen when user create a file,  but it is not able to detect file encoding.

src/org/omegat/util/StaticUtils.java
miurahr: When using Reader, spotBugs warns it is bad when calling Reader without Encoding.
Here is a place to detect encoding, so I read it as bytes.

src/org/omegat/gui/glossary/GlossaryReaderTSV.java
```diff
@@ -67,10 +78,29 @@ public static String getFileEncoding(final File file, String defaultEncoding) th
...
+            }
+        } catch (IOException e) {
+            // ignore
+        }
+        if (detected == null)
+            try (FileInputStream stream = new FileInputStream(inFile)) {
+                detected = EncodingDetector.detectEncoding(stream);
+            } catch (IOException ignored) {
+                // ignore exception here.
+            }
```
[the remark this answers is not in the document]
miurahr: revert to original one.

src/org/omegat/util/StringUtil.java
```diff
@@ -987,4 +993,60 @@ public static String getTailSegments(String str, int separator, int segments) {
         }
         return str.substring(start + Character.charCount(separator), str.length());
     }
+
+    private static final String magicString = "([^\\s\'\":;]+)\\s*:\\s*(\"(?:\\\\.|[^\"])*\"|[^\"\\s;]+)[\\s;]*";
+    private static final Pattern magicRegexp = Pattern.compile(magicString);
```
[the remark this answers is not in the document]
miurahr: Refactoring and add javadoc how parse an input string.

src/org/omegat/util/StaticUtils.java
amake: > Here is a place to detect encoding, so I read it as bytes.

But ultimately you didn't read it as bytes: you read it as an ASCII string. That's the same as using a `Reader` with ASCII encoding.

src/org/omegat/util/StringUtil.java
```diff
@@ -987,4 +991,35 @@ public static String getTailSegments(String str, int separator, int segments) {
...
+        HashMap<String, String> result = new HashMap<>();
+        if (str == null || str.length() < 11) {  // minimum = "-*- a:b -*-".length()
+            return result;
+        }
+        int start = str.indexOf("-*-");
+        int end;
+        if (start < 0 || (end = str.substring(start + 3).indexOf("-*-")) < 0) {
+            return result;
+        }
+        Pattern regex = Pattern.compile("([\\w-]+)\\s*:\\s*([\\w-]+)[\\s;]*");
```
amake, line 1016: This pattern is not dynamic; I would make it a `static final` constant.

src/org/omegat/util/StringUtil.java
```diff
@@ -987,4 +991,35 @@ public static String getTailSegments(String str, int separator, int segments) {
...
+     *  key and value: string w/ alphabet, number, underscore, hyphen
+     * </p>
+     * @param str input string.
+     * @return Key-Value map of String.
+     */
+    public static Map<String, String> parseMagicComment(final String str) {
+        HashMap<String, String> result = new HashMap<>();
+        if (str == null || str.length() < 11) {  // minimum = "-*- a:b -*-".length()
+            return result;
+        }
```
amake, line 1010:
```suggestion
        if (str == null || str.length() < 11) {  // minimum = "-*- a:b -*-".length()
            return Collections.emptyMap();
        }
        HashMap<String, String> result = new HashMap<>();
```

amake: Oops, actually you can do the same thing with the next early-return as well.

test/src/org/omegat/util/StringUtilTest.java
```diff
@@ -406,4 +408,17 @@ public void testCaseConversion() {
         assertEquals("Istanbul", StringUtil.replaceCase("\\uistanbul", Locale.ENGLISH)); // English version
         assertEquals("\u0130stanbul", StringUtil.replaceCase("\\uistanbul", new Locale("tr"))); // Turkish version
     }
+
+    @Test
+    public void testParseMagicComment() {
+        // TODO: junit5 parameterized test.
+        List<String> cases = Arrays.asList(
+                "# -*- coding: UTF-8 -*-",
+                "# comment -*- coding: UTF-8 -*-"
```
amake, line 417: Could we get a case where `coding: UTF-8` is just one of multiple entries? And where there is a trailing `;`?

- `-*- coding: UTF-8; foo: bar -*-`
- `-*- foo: bar; coding: UTF-8 -*-`
- `-*- coding: UTF-8; -*-`

It would also be good to have some tests for examples that *don't* parse. For instance:
- Missing the first `-*-`
- Missing the final `-*-` (?)
- ?

src/org/omegat/gui/glossary/GlossaryReaderTSV.java
```diff
@@ -67,8 +78,27 @@ public static String getFileEncoding(final File file, String defaultEncoding) th
...
+                    detected = value;
+                }
+            }
+        } catch (IOException e) {
+            // ignore
+        }
+        if (detected == null) {
+            return EncodingDetector.detectEncodingDefault(inFile, defaultEncoding);
         }
+        return detected;
```
amake, line 101:
```suggestion
        if (detected != null) {
            return detected;
        }
        return EncodingDetector.detectEncodingDefault(inFile, defaultEncoding);
```

src/org/omegat/util/StaticUtils.java
```diff
@@ -578,5 +578,4 @@ public static boolean isProjectDir(File f) {
         }
         return true;
     }
```
amake, line 580:
```suggestion
    }

```

test/src/org/omegat/util/StringUtilTest.java
amake: I added some new test cases.

amake:
I've made some additional changes. Please let me know if it looks OK, and I will squash to `master`.

src/org/omegat/util/MagicComment.java
```diff
@@ -0,0 +1,115 @@
...
+     * @return
+     * @throws IOException
+     */
+    public static Map<String, String> parse(File file) throws IOException  {
+        String line;
+        try (BufferedReader reader = new BufferedReader(new InputStreamReader(
+                new BOMInputStream(new FileInputStream(file)), StandardCharsets.US_ASCII))) {
+            line = reader.readLine();
+        }
+        if (line.startsWith("#")) {
```
miurahr, line 78: When we have parser class, it is better to be configurable for comment char.

amake: I think this is OK for now.

miurahr: It is OK for now. We can change when other module also want to add magic comment in future.

src/org/omegat/core/data/RealProject.java
amake: We create the file on-demand when appending via the Create Glossary Entry dialog. I changed it so that it creates the file with your new method, so it will have the magic comment.

amake:
Thanks for your work on this. I've squashed to `master`.

PR #113, opened 2021-06-22
Gradle: allow GnuPG for signing

miurahr:
It is convenient for developer who use GnuPG and want to publish OmegaT to localMaven to develop plugins.
It is even safer, because Gradle ask password interactively when running publish, and no need to write key password on property file.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1074,7 +1074,11 @@ publishing {
 
 signing {
     sign publishing.publications.mavenJava
+    if (!findProperty("signing.keyId") || findProperty("signing.keyId").equals("")) {
+        useGpgCmd()
+    }
 }
+
 
 remotes {
     sourceforgeWeb {
--- a/build.gradle
+++ b/build.gradle
@@ -1074,7 +1074,7 @@ publishing {
 
 signing {
     sign publishing.publications.mavenJava
-    if (!findProperty("signing.keyId") || findProperty("signing.keyId").equals("")) {
+    if (!findProperty("signing.keyId")) {
         useGpgCmd()
     }
 }
--- a/local.properties.example
+++ b/local.properties.example
@@ -41,6 +41,10 @@ signing.keyId=
 signing.password=
 # Exported as e.g. `gpg --export-secret-keys > secring.gpg`
 signing.secretKeyRingFile=
+## When you want to use GnuPG command instead of exported keyring
+## uncommented and add key ID
+#signing.gnupg.keyName=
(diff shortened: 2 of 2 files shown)


build.gradle
```diff
@@ -1074,8 +1074,12 @@ publishing {
 
 signing {
     sign publishing.publications.mavenJava
+    if (!findProperty("signing.keyId") || findProperty("signing.keyId").equals("")) {
```
amake, line 1077:
```suggestion
    if (!findProperty("signing.keyId")) {
```

In Groovy, both null and the empty string are falsy.

build.gradle
```diff
@@ -1074,8 +1074,12 @@ publishing {
...
+        useGpgCmd()
+    }
 }
 
+
```
amake, line 1082: This extra line can be reverted.

local.properties.example
```diff
@@ -41,6 +41,10 @@ signing.keyId=
 signing.password=
 # Exported as e.g. `gpg --export-secret-keys > secring.gpg`
 signing.secretKeyRingFile=
+## When you want to use GnuPG command instead of exported keyring
+## uncommented and add key ID
+#signing.gnupg.keyName=
+#signing.gnupg.executable=gpg
```
amake, line 47: I think we should improve the comments to explain better what is required and used for what.

Are `signing.{keyId,password,secretKeyRingFile}` used at all when configured with `useGpgCmd()`?

miurahr: > Are signing.{keyId,password,secretKeyRingFile} used at all when configured with useGpgCmd()?

Yes.

These properties are documented at [gradle manual](https://docs.gradle.org/current/userguide/signing_plugin.html#sec:sec:gnupg_signatory_configuration)

miurahr: I've updated comments.

amake:
As far as I can tell, it seems like `useGpgCmd()` doesn't use any of the regular `signing.{keyId,password,secretKeyRingFile}` properties; it only uses the `signing.gnupg.*` properties.

Since there are so many `signing.gnupg.*` properties and using gpg-agent seems like a bit of a power-user option, I simplified the comments a bit.

Let me know if this is OK for you.

miurahr:
> using gpg-agent seems like a bit of a power-user option

It may be a true, I don't know.Using Java signatory is very weird  for me who has experience Using PGP and GnuPG in past 30 years. :-)
Yes, I'm power-user. It is enough comment.

amake:

PR #115, opened 2021-06-23
GlossarySearcher: Fix Japanese hash collisions and wrong search

miurahr:
- English  search case
- Japanese  search case reported in [BUG#1034](https://sourceforge.net/p/omegat/bugs/1034/)
- Test Token class equality

### DIFF
--- a/src/org/omegat/gui/glossary/GlossarySearcher.java
+++ b/src/org/omegat/gui/glossary/GlossarySearcher.java
@@ -148,7 +148,7 @@ public class GlossarySearcher {
         return true;
     }
 
-    private static boolean isCjkMatch(String fullText, String term) {
+    protected static boolean isCjkMatch(String fullText, String term) {
         // This is a CJK word and our source language is not space-delimited, so include if
         // word appears anywhere in source string.
         IProject project = Core.getProject();
--- a/src/org/omegat/gui/glossary/GlossarySearcher.java
+++ b/src/org/omegat/gui/glossary/GlossarySearcher.java
@@ -131,7 +131,17 @@ public class GlossarySearcher {
                 Preferences.GLOSSARY_NOT_EXACT_MATCH_DEFAULT);
         List<Token[]> foundTokens = DefaultTokenizer.searchAll(fullTextTokens, glosTokens, notExact);
         foundTokens.removeIf(toks -> !keepMatch(toks, fullText, term));
+        foundTokens.removeIf(toks -> !rawMatch(toks, fullText, term));
         return foundTokens;
+    }
+
+    private static boolean rawMatch(Token[] tokens, String srcTxt, String term) {
+        for (Token token : tokens) {
+            if (token.getTextFromString(srcTxt).equals(term.substring(token.getLength()))) {
+                return true;
+            }
+        }
+        return false;
     }
 
     private static boolean keepMatch(Token[] tokens, String srcTxt, String locTxt) {
--- a/src/org/omegat/gui/glossary/GlossarySearcher.java
+++ b/src/org/omegat/gui/glossary/GlossarySearcher.java
@@ -131,7 +131,9 @@ public class GlossarySearcher {
                 Preferences.GLOSSARY_NOT_EXACT_MATCH_DEFAULT);
         List<Token[]> foundTokens = DefaultTokenizer.searchAll(fullTextTokens, glosTokens, notExact);
         foundTokens.removeIf(toks -> !keepMatch(toks, fullText, term));
-        foundTokens.removeIf(toks -> !rawMatch(toks, fullText, term));
+        if (StringUtil.isCJK(term)) {
+            foundTokens.removeIf(toks -> !rawMatch(toks, fullText, term));
+        }
         return foundTokens;
     }
 
--- a/test/src/org/omegat/core/data/TokenTest.java
+++ b/test/src/org/omegat/core/data/TokenTest.java
@@ -1,4 +1,3 @@
-
 /**************************************************************************
  OmegaT - Computer Assisted Translation (CAT) tool
           with fuzzy matching, translation memory, keyword search,
@@ -73,4 +72,3 @@ public class TokenTest extends TestCore {
         assertNotEquals(strTokens[0], glosTokens[0]);
     }
 }
-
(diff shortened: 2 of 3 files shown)


miurahr:
Failed `GlossarySearcher#isCjkMatch`
```
/home/vsts/work/1/s/test/src/org/omegat/gui/glossary/GlossarySearcherTest.java:63: warning: [static] static method should be qualified by type name, org.omegat.gui.glossary.GlossarySearcher, instead of by an expression
        assertTrue(searcher.isCjkMatch(sourceText, sourceText));
```

src/org/omegat/gui/glossary/GlossarySearcher.java
```diff
@@ -131,9 +131,19 @@ private boolean isTokenMatch(Token[] fullTextTokens, String fullText, String ter
                 Preferences.GLOSSARY_NOT_EXACT_MATCH_DEFAULT);
         List<Token[]> foundTokens = DefaultTokenizer.searchAll(fullTextTokens, glosTokens, notExact);
         foundTokens.removeIf(toks -> !keepMatch(toks, fullText, term));
+        foundTokens.removeIf(toks -> !rawMatch(toks, fullText, term));
         return foundTokens;
     }
 
+    private static boolean rawMatch(Token[] tokens, String srcTxt, String term) {
+        for (Token token : tokens) {
+            if (token.getTextFromString(srcTxt).equals(term.substring(token.getLength()))) {
```
amake, line 140: Do I understand this right?

- `token.getTextFromString(srcTxt)` is the text that was matched in the source text
- `term.substring(token.getLength())` is a prefix of the original term

You are rejecting any matched term where none of the tokens is a prefix of the term.

This makes some severe assumptions about how stemming/lemmatizing works. I don't think the token is guaranteed to be a prefix, or even necessarily present in the original source string.

test/src/org/omegat/core/data/TokenTest.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    /**
+     * Test case of BUG#1034
+     */
+    @Test(expected = AssertionError.class)
+    public void testGlossaryTokenEqualityJapanese() {
+        ITokenizer tok = new LuceneJapaneseTokenizer();
+        String str = "\u5834\u6240";
+        String glos = "\u5857\u5E03";
```
amake, line 68: I know we have a test that requires source files to be in ASCII.

I think that can probably be changed to require ASCII *or* UTF-8.

(That's not something you need to do in this PR)

test/src/org/omegat/gui/glossary/GlossarySearcherTest.java
```diff
@@ -0,0 +1,272 @@
...
+    @Test
+    public void testGlossarySearcherEnglish() {
+        String sourceText = "source";
+        String translationText = "translation";
+        String commentText = "comment";
+        ITokenizer tok = new LuceneEnglishTokenizer();
+        Language language = new Language("en");
+        setupProject(language);
+        List<GlossaryEntry> entries = new ArrayList<>();
+        entries.add(new GlossaryEntry(sourceText, translationText, commentText, true, "origin"));
```
amake, line 65: As long as you don't need it to be mutable, you can initialize and populate a list in one statement with `Arrays#asList`.

test/src/org/omegat/gui/glossary/GlossarySearcherTest.java
```diff
@@ -0,0 +1,272 @@
...
+            public Language getSourceLanguage() {
+                return language;
+            }
+
+            public Language getTargetLanguage() {
+                return new Language("pl");
+            }
+        };
+
+        Core.setProject(new IProject() {
```
amake, line 136: You could use an anonymous subclass of `NotLoadedProject` to avoid having to define a lot of the `IProject` methods; see e.g. `ExternalTMFactoryTest`.

test/src/org/omegat/gui/glossary/GlossarySearcherTest.java
```diff
@@ -0,0 +1,272 @@
...
+            public void compileProjectAndCommit(String sourcePattern, boolean doPostProcessing, boolean commitTargetFiles) {
+            }
+        });
+    }
+
+    private List<GlossaryEntry> glossarySearcherCommon(String sourceText, ITokenizer tok, Language language,
+                                                       List<GlossaryEntry> entries) {
+        EntryKey key = new EntryKey("file", sourceText, "id", "prev", "next", "path");
+        String[] prop = new String[]{};
+        SourceTextEntry ste = new SourceTextEntry(key, 1, prop, sourceText, new ArrayList<>());
```
amake, line 267: When you need an empty list, it's good to use `Collections#emptyList`.

Same goes for sets and maps.

src/org/omegat/gui/glossary/GlossarySearcher.java
miurahr: > This makes some severe assumptions about how stemming/lemmatizing works. I don't think the token is guaranteed to be a prefix, or even necessarily present in the original source string.

I'm also afraid this but I don't know well about stemming/lemmatizing behavior.
Actually I do only on CJK  in my first trial implementation.
IMHO, the assumption is OK on CJK languages.

amake: >  IMHO, the assumption is OK on CJK languages.

I think either we need some sort of positive confirmation that it would work with all languages, or to limit this change to only CJK.

miurahr: Because the issue, hash collision,  is reported on Japanese, it would be better to limit the change only to CJK.

test/src/org/omegat/gui/glossary/GlossarySearcherTest.java
miurahr: isCjkMatch expect `isProjectLoaded()`. Does `NotLoadedproject` work as stub? 

```
  protected static boolean isCjkMatch(String fullText, String term) {
        // This is a CJK word and our source language is not space-delimited, so include if
        // word appears anywhere in source string.
        IProject project = Core.getProject();
        return project.isProjectLoaded() && ...
```

amake:
Thank you! I've squashed to `master`.

src/org/omegat/gui/glossary/GlossarySearcher.java
amake: I only just realized that our `isCJK` method is very, very stupid. I'm now worried that this will apply to many more cases than it should have.

PR #117, opened 2021-06-24
Preference: font selection to limit one can show emoji Unicode

miurahr:
- Implement the requirement from BUG#1043
- Update preference -> font selection UI to limit Font comboBox to show only fonts that can display emoji characters

### DIFF
--- a/src/org/omegat/util/StaticUtils.java
+++ b/src/org/omegat/util/StaticUtils.java
@@ -145,10 +145,10 @@ public final class StaticUtils {
     public static String[] getUnicodeFontNames() {
         int sp;
         Font[] fonts = GraphicsEnvironment.getLocalGraphicsEnvironment().getAllFonts();
-        final int dspcp = "\uD83D\uDE03".codePointAt(0); // facemark in Emoji-1.0 specification
+        final String dspstr = "a\uD83D\uDE03"; // facemark in Emoji-1.0 specification
         Set<String> fontNames = new HashSet<>();
         for (Font font : fonts)
-            if (font.canDisplay(dspcp)) fontNames.add(font.getFamily());
+            if (font.canDisplay(dspstr.codePointAt(0)) && font.canDisplay(dspstr.codePointAt(1))) fontNames.add(font.getFamily());
         return fontNames.toArray(new String[0]);
     }
 
--- a/src/org/omegat/util/StaticUtils.java
+++ b/src/org/omegat/util/StaticUtils.java
@@ -31,6 +31,7 @@
 
 package org.omegat.util;
 
+import java.awt.Font;
 import java.awt.GraphicsEnvironment;
 import java.awt.event.KeyEvent;
 import java.io.File;
@@ -42,8 +43,10 @@ import java.net.URISyntaxException;
 import java.net.URL;
 import java.nio.file.Paths;
 import java.util.ArrayList;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Objects;
+import java.util.Set;
 import java.util.function.Predicate;
 import java.util.regex.Pattern;
 import java.util.zip.ZipEntry;
@@ -133,6 +136,20 @@ public final class StaticUtils {
         GraphicsEnvironment graphics;
         graphics = GraphicsEnvironment.getLocalGraphicsEnvironment();
         return graphics.getAvailableFontFamilyNames();
+    }
+
+    /**
+     * Returns the names of all font which can display emoji.
+     * @return Set of names.
+     */
+    public static String[] getUnicodeFontNames() {
+        int sp;
+        Font[] fonts = GraphicsEnvironment.getLocalGraphicsEnvironment().getAllFonts();
+        final int dspcp = "\uD83D\uDE03".codePointAt(0); // facemark in Emoji-1.0 specification
+        Set<String> fontNames = new HashSet<>();
+        for (Font font : fonts)
+            if (font.canDisplay(dspcp)) fontNames.add(font.getFamily());
+        return fontNames.toArray(new String[0]);
     }
 
     /** Caching install dir */
--- a/src/org/omegat/util/StaticUtils.java
+++ b/src/org/omegat/util/StaticUtils.java
@@ -43,6 +43,7 @@ import java.net.URISyntaxException;
 import java.net.URL;
 import java.nio.file.Paths;
 import java.util.ArrayList;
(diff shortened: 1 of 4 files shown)


src/org/omegat/util/StaticUtils.java
```diff
@@ -135,6 +138,20 @@ public static boolean isKey(KeyEvent e, int code, int modifiers) {
         return graphics.getAvailableFontFamilyNames();
     }
 
+    /**
+     * Returns the names of all font which can display emoji.
+     * @return Set of names.
+     */
+    public static String[] getUnicodeFontNames() {
```
amake, line 145: There are several issues:

1. This is incredibly slow. It takes 33 seconds on my machine (MacBook Pro 13" 2020, 2 GHz Quad-Core Intel Core i5)
2. This doesn't give meaningful results

On my machine there are three fonts that explicitly contain glyphs for U+1F603 😃:

- Apple Color Emoji
- BabelStone Han
- HanaMinA

The only one that *actually* supports emoji in the sense that users would expect is Apple Color Emoji. However Java Swing does not support Apple Color Emoji and probably never will.

It also doesn't make sense to choose Apple Color Emoji as your editor's font. That's just not how text rendering is intended to work in the modern era; instead the text rendering engine is supposed to appropriately fall back to appropriate fonts on a grapheme cluster level. As I understand it, it is a limitation on Java Swing that prevents this from working correctly.

The other fonts contain text (black & white outline) representations of U+1F603 😃 and, I suppose if you had very strange taste, could be appropriate for use as the editor font. However they don't and never will actually support emoji as users would expect: Java Swing is still incapable of displaying ZWJ-sequence emoji and other composed emoji like region flags.

I think the concept behind this PR (filtering the font list to enable the user to choose a font that will allow emoji support) is barking up the wrong tree. I think the ultimate goal of enabling better or easier emoji display is basically impossible while we are still using Java Swing (maybe SWT does better).

In my opinion the best we can do is the Aggressive Font Fallback feature. This basically applies your logic ("find a font that can show the character") to every character in the editor, separately. This mimics modern font fallback within the constraints of Swing's rendering capabilities.

(Note: I could be wrong about any of the above; if so I welcome corrections. Also the above might be specific to macOS; in that case it would be helpful to know where I am wrong with respect to other operating systems.)

miurahr: > The other fonts contain text (black & white outline) representations of U+1F603 smiley and, I suppose if you had very strange taste, could be appropriate for use as the editor font. However they don't and never will actually support emoji as users would expect: Java Swing is still incapable of displaying ZWJ-sequence emoji and other composed emoji like region flags.

I think Java Swing can display Unicode emoji-1.0 standard / Unicode 6 level of SMP characters, and I can display all emoji in a file https://sourceforge.net/p/omegat/bugs/_discuss/thread/e814c543af/66e8/attachment/emoicon.txt on OmegaT on my Mint Linux.

> This is incredibly slow. It takes 33 seconds on my machine (MacBook Pro 13" 2020, 2 GHz Quad-Core Intel Core i5)

Uum, core of the patch is code snippet here. It run a check loop for all installed fonts. 
https://sourceforge.net/p/omegat/bugs/1043/#9a7d/7438
Actually it is O(n) but a number of fonts may be a range of several hundreds to several thousand, and it seems no big issue. 
How do you have number of fonts? When I tested the patch on Linux, there are about 2600 of fonts, works in a second.
We can have a list of valid fonts for major platforms, macOS, Windows and Linux,  and we can check it whether installed or not. It would be not slow.

amake: > How do you have number of fonts?

The number of fonts inspected by the loop is 2,167.

I suspect that this is a difference between Linux and Mac: maybe Linux has some caching mechanism to describe which fonts contain glyphs for which codepoints, whereas on Mac each font has to actually be loaded into memory?

> We can have a list of valid fonts for major platforms

This is not a reasonable approach.

> I can display all emoji in a file https://sourceforge.net/p/omegat/bugs/_discuss/thread/e814c543af/66e8/attachment/emoicon.txt on OmegaT on my Mint Linux.

On Java 8 with Aggressive Font Fallback enabled and my editor font set to Dialog I get this:

image

Note that whether a glyph can be displayed seems to have little to do with which version of Unicode the codepoint was assigned in.

Note also that with Aggressive Font Fallback I can enjoy my preferred editor font (in this case Piazzolla) and still see emoji:

image

On Java 11 it seems that `Font#canDisplay` is broken: the Aggressive Font Fallback mechanism never kicks in, and your code reports that all of my fonts support emoji.

Here's what I get when I choose BabelStone Han as my editor font on Java 11:

image

miurahr: > However they don't and never will actually support emoji as users would expect: Java Swing is still incapable of displaying ZWJ-sequence emoji and other composed emoji like region flags.

Yes, it is right.

Java 8 support Unicode 6.2 that does not include these sequence emojis,  but include simple emojis in the standard.
https://openjdk.java.net/jeps/133

Java 11 support Unicode 10 but UTS #51, Unicode Emoji, is out of scope of the JEP
https://openjdk.java.net/jeps/327

I think OmegaT can handle these characters in UI and editor when user has a  font.

Unicode 6.1：😀😗😙😑😮😯😴😛😕😟

OmegaT w/ Java 8  may not handle properly for following emojis even when user has a font.
OmegaT w/ Java 11 may show these emojis. 

Unicode 7.0：🙂🙁🕵🗣🕴🖕🖖🖐
Unicode 8.0：🤗🤓🤔🙄🤐🙃🤑🤒🤕🤖
Unicode 9.0：🤣🤠🤡🤥🤤🤢🤧🤴🤶🤵🤷
Unicode 10.0：🤩🤨🤯🤪🤬🤮🤫🤭🧐🧒

OmegaT can process other sequence of emoji/flags  as a tag if input filter is aware of it.

miurahr: `getAllFonts()`
>Returns an array **containing a one-point size instance** of all fonts available in this GraphicsEnvironment. Typical usage would be to allow a user to select a particular font. 

This is a reason it is slow on macOS.

miurahr: FYI: https://unicode.org/Public/emoji/13.1/ is a place where all Unicode 13.1 emoji test data(text file).

miurahr:
I'd like to close here as invalid.

PR #119, opened 2021-06-27
Fix error on Java 16

miurahr:
When loading GTK LaF on Java16 / Linux by Class#forName method,
it cause error "java.desktop does not export com.sun.java.swing.plaf.gtk"  (Bug#1052) and break UI/Theme.

I know OmegaT  support Java 8 and 11, not Java 16,
but this reduces a known issue.

### DIFF
--- a/src/org/omegat/gui/theme/DelegatingLookAndFeel.java
+++ b/src/org/omegat/gui/theme/DelegatingLookAndFeel.java
@@ -62,8 +62,20 @@ public abstract class DelegatingLookAndFeel extends LookAndFeel {
     protected final LookAndFeel systemLookAndFeel;
 
     public DelegatingLookAndFeel() throws Exception {
-        systemLookAndFeel = (LookAndFeel) Class.forName(UIManager.getSystemLookAndFeelClassName())
-                .getDeclaredConstructor().newInstance();
+        // When loading GTK LaF on Java16 / Linux by Class#forName method,
+        // it cause error "java.desktop does not export com.sun.java.swing.plaf.gtk"
+        // (Bug#1052)
+        String systemLaF = UIManager.getSystemLookAndFeelClassName();
+        if (systemLaF.equals("com.sun.java.swing.plaf.gtk.GTKLookAndFeel")) {
+             try {
+                UIManager.setLookAndFeel(systemLaF);
+             } catch (Exception ignore) {
+             }
+             systemLookAndFeel = UIManager.getLookAndFeel();
+        } else {
+            systemLookAndFeel = (LookAndFeel) Class.forName(UIManager.getSystemLookAndFeelClassName())
+                    .getDeclaredConstructor().newInstance();
+        }
     }
 
     @Override
--- a/src/org/omegat/gui/theme/DelegatingLookAndFeel.java
+++ b/src/org/omegat/gui/theme/DelegatingLookAndFeel.java
@@ -62,20 +62,12 @@ public abstract class DelegatingLookAndFeel extends LookAndFeel {
     protected final LookAndFeel systemLookAndFeel;
 
     public DelegatingLookAndFeel() throws Exception {
-        // When loading GTK LaF on Java16 / Linux by Class#forName method,
-        // it cause error "java.desktop does not export com.sun.java.swing.plaf.gtk"
-        // (Bug#1052)
-        String systemLaF = UIManager.getSystemLookAndFeelClassName();
-        if (systemLaF.equals("com.sun.java.swing.plaf.gtk.GTKLookAndFeel")) {
-             try {
-                UIManager.setLookAndFeel(systemLaF);
-             } catch (Exception ignore) {
-             }
-             systemLookAndFeel = UIManager.getLookAndFeel();
-        } else {
-            systemLookAndFeel = (LookAndFeel) Class.forName(UIManager.getSystemLookAndFeelClassName())
-                    .getDeclaredConstructor().newInstance();
-        }
+        // Get System LaF instance.
+        // Calling Class.forName(UIManager.getSystemLookAndFeelClassName()).newInstance()
+        // on Java16 throw error "java.desktop does not export com.sun.java.swing.plaf.gtk"
+        // on Linux, and for aqua LaF on macOS. (Bug#1052)
+         UIManager.setLookAndFeel(UIManager.getSystemLookAndFeelClassName());
+         systemLookAndFeel = UIManager.getLookAndFeel();
     }
 
     @Override
--- a/src/org/omegat/gui/theme/DelegatingLookAndFeel.java
+++ b/src/org/omegat/gui/theme/DelegatingLookAndFeel.java
@@ -66,8 +66,8 @@ public abstract class DelegatingLookAndFeel extends LookAndFeel {
         // Calling Class.forName(UIManager.getSystemLookAndFeelClassName()).newInstance()
         // on Java16 throw error "java.desktop does not export com.sun.java.swing.plaf.gtk"
         // on Linux, and for aqua LaF on macOS. (Bug#1052)
-         UIManager.setLookAndFeel(UIManager.getSystemLookAndFeelClassName());
-         systemLookAndFeel = UIManager.getLookAndFeel();
+        UIManager.setLookAndFeel(UIManager.getSystemLookAndFeelClassName());
+        systemLookAndFeel = UIManager.getLookAndFeel();
     }
 
     @Override
--- a/src/org/omegat/gui/theme/DelegatingLookAndFeel.java
+++ b/src/org/omegat/gui/theme/DelegatingLookAndFeel.java
@@ -34,7 +34,9 @@ import javax.swing.JComponent;
 import javax.swing.LayoutStyle;
 import javax.swing.LookAndFeel;
 import javax.swing.UIManager;
+import javax.swing.UIManager.LookAndFeelInfo;
 
+import org.madlonkay.desktopsupport.DesktopSupport;
 import org.omegat.util.gui.ResourcesUtil;
 
 /**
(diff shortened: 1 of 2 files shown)


src/org/omegat/gui/theme/DelegatingLookAndFeel.java
```diff
@@ -62,8 +62,20 @@ protected static ImageIcon getIcon(String iconName) {
     protected final LookAndFeel systemLookAndFeel;
 
     public DelegatingLookAndFeel() throws Exception {
-        systemLookAndFeel = (LookAndFeel) Class.forName(UIManager.getSystemLookAndFeelClassName())
-                .getDeclaredConstructor().newInstance();
+        // When loading GTK LaF on Java16 / Linux by Class#forName method,
+        // it cause error "java.desktop does not export com.sun.java.swing.plaf.gtk"
+        // (Bug#1052)
+        String systemLaF = UIManager.getSystemLookAndFeelClassName();
+        if (systemLaF.equals("com.sun.java.swing.plaf.gtk.GTKLookAndFeel")) {
```
amake, line 69: I get the same thing on macOS for `com.apple.laf.AquaLookAndFeel`.

I don't think a workaround that hard-codes platform-specific LAF classnames makes sense.

Isn't `UIManager.setLookAndFeel(systemLaF)` always going to fail as well?

What is the default value of `UIManager.getLookAndFeel()`? Is it different from the LAF represented by `UIManager.getSystemLookAndFeelClassName()`?

miurahr: Because I found this, it check name first. It can always load LaF.

https://github.com/openjdk/jdk/blob/739769c8fc4b496f08a92225a12d07414537b6c0/src/jdk.jconsole/share/classes/sun/tools/jconsole/JConsole.java#L56-L70

miurahr: I removed a condition to do it on all LaFs.

When calling `UIManager.getSystemLookAndFeel()` without  `UIManager.setLookAndFeel(UIManager.getSystemLookAndFeelClassName());` , LaF went to fall back to Metal LaF (multi-platform LaF). So both lines are mandatory.

src/org/omegat/gui/theme/DelegatingLookAndFeel.java
```diff
@@ -62,8 +62,12 @@ protected static ImageIcon getIcon(String iconName) {
     protected final LookAndFeel systemLookAndFeel;
 
     public DelegatingLookAndFeel() throws Exception {
-        systemLookAndFeel = (LookAndFeel) Class.forName(UIManager.getSystemLookAndFeelClassName())
-                .getDeclaredConstructor().newInstance();
+        // Get System LaF instance.
+        // Calling Class.forName(UIManager.getSystemLookAndFeelClassName()).newInstance()
+        // on Java16 throw error "java.desktop does not export com.sun.java.swing.plaf.gtk"
+        // on Linux, and for aqua LaF on macOS. (Bug#1052)
+        UIManager.setLookAndFeel(UIManager.getSystemLookAndFeelClassName());
```
amake, line 69: I like that we're not hard-coding the class name, but this now has the side effect of changing the active LAF.

It looks like there is an API in Java 9+ to instantiate a LAF without installing it:
http://hg.openjdk.java.net/jdk9/jdk9/jdk/rev/174b223645b7#l1.32
https://docs.oracle.com/javase/9/docs/api/javax/swing/UIManager.html#createLookAndFeel-java.lang.String-

Is it impossible to call this API via reflection?

In [DesktopSupport](https://github.com/amake/DesktopSupport) I do some very tricky things to bridge a Java 8 and Java 9+ API; I suppose that would be possible here but very complicated to do.

amake: Actually I think this API is a good candidate to be added to DesktopSupport.

amake: I just added a wrapper for `UIManager#createLookAndFeel` in DesktopSupport, but I haven't tested it yet. If you wanted to try you could build and install it locally.

amake: Bad news: I still get the error with the wrapped `UIManager#createLookAndFeel`:

```
18206: Error: java.lang.reflect.InvocationTargetException 
18206: Error:   at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method) 
18206: Error:   at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:78) 
18206: Error:   at java.base/jdk.internal.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45) 
18206: Error:   at java.base/java.lang.reflect.Constructor.newInstanceWithCaller(Constructor.java:499) 
18206: Error:   at java.base/java.lang.reflect.Constructor.newInstance(Constructor.java:480) 
18206: Error:   at org.omegat.util.gui.UIDesignManager.setTheme(UIDesignManager.java:81) 
18206: Error:   at org.omegat.util.gui.UIDesignManager.initialize(UIDesignManager.java:100) 
18206: Error:   at org.omegat.core.Core.initializeGUI(Core.java:237) 
18206: Error:   at org.omegat.Main.runGUI(Main.java:314) 
18206: Error:   at org.omegat.Main.main(Main.java:186) 
18206: Error: Caused by: javax.swing.UnsupportedLookAndFeelException: com.apple.laf.AquaLookAndFeel 
18206: Error:   at java.desktop/javax.swing.UIManager.createLookAndFeel(UIManager.java:547) 
18206: Error:   at org.madlonkay.desktopsupport.impl.Java9DesktopSupportImpl.createLookAndFeel(Java9DesktopSupportImpl.java:300) 
18206: Error:   at org.omegat.gui.theme.DelegatingLookAndFeel.<init>(DelegatingLookAndFeel.java:67) 
18206: Error:   at org.omegat.gui.theme.DefaultFlatTheme.<init>(DefaultFlatTheme.java:62) 
18206: Error:   ... 10 more 
```

amake: Actually I got it: I misunderstood that `createLookAndFeel` takes the LAF name, not the classname.

I've published a new version of DesktopSupport and pushed some changes to use it. This works for me on Mac with Java 8 and Java 16.

I note that the Motif theme doesn't work in Java 16 with this approach, but the other built-in LAFs do (Metal, Nimbus).

miurahr:
On Mint Linux 20.01, Java 11


```
47685: 情報: OmegaT-5.6.0_0_be9ded292 (Wed Jun 30 09:10:13 JST 2021)  Locale ja_JP 
47685: 情報: Java：Ubuntu バージョン 11.0.11 が /usr/lib/jvm/java-11-openjdk-amd64 から実行されました (LOG_STARTUP_INFO)
--<snip>--
47685: エラー: java.lang.NullPointerException 
47685: エラー: 	at org.omegat.gui.theme.DelegatingLookAndFeel.isSupportedLookAndFeel(DelegatingLookAndFeel.java:88) 
47685: エラー: 	at java.desktop/javax.swing.UIManager.setLookAndFeel(UIManager.java:572) 
47685: エラー: 	at org.omegat.util.gui.UIDesignManager.setTheme(UIDesignManager.java:81) 
47685: エラー: 	at org.omegat.util.gui.UIDesignManager.initialize(UIDesignManager.java:100) 
47685: エラー: 	at org.omegat.core.Core.initializeGUI(Core.java:237) 
47685: エラー: 	at org.omegat.Main.runGUI(Main.java:314) 
47685: エラー: 	at org.omegat.Main.main(Main.java:186) 
```

miurahr:
And the same NPE on Java8

```
14547: 情報: OmegaT-5.6.0_0_be9ded292 (Wed Jun 30 09:15:05 JST 2021)  Locale ja_JP 
14547: 情報: Java：Amazon.com Inc. バージョン 1.8.0_292 が /home/miurahr/.jdks/corretto-1.8.0_292/jre から実行されました (LOG_STARTUP_INFO)
```

miurahr:
On Java 16, Mint Linux 20.1, slightly different

```
16659: エラー: java.lang.NullPointerException: Cannot invoke "javax.swing.LookAndFeel.isSupportedLookAndFeel()" because "this.systemLookAndFeel" is null 
16659: エラー: 	at org.omegat.gui.theme.DelegatingLookAndFeel.isSupportedLookAndFeel(DelegatingLookAndFeel.java:88) 
16659: エラー: 	at java.desktop/javax.swing.UIManager.setLookAndFeel(UIManager.java:577) 
16659: エラー: 	at org.omegat.util.gui.UIDesignManager.setTheme(UIDesignManager.java:81) 
16659: エラー: 	at org.omegat.util.gui.UIDesignManager.initialize(UIDesignManager.java:100) 
16659: エラー: 	at org.omegat.core.Core.initializeGUI(Core.java:237) 
16659: エラー: 	at org.omegat.Main.runGUI(Main.java:314) 
16659: エラー: 	at org.omegat.Main.main(Main.java:186) 
```

amake:
It looks like all of your NPEs are because `this.systemLookAndFeel` is null. The only way I can see that happening would be if `DesktopSupport.instance().createLookAndFeel(...)` returned null.

The only way that could happen is if the concrete `IDesktopSupport` implementation returned by `DesktopSupport.instance()` was `DummyDesktopSupportImpl`.

I realized that the way I was detecting Java 9 was not good: I tried calling `Taskbar.getTaskbar()`, which is only available in Java 9, but it can also fail when the desktop environment simply doesn't support a Taskbar. I guess Linux is one example.

amake:
Actually my mistake was even dumber: the Java 8 implementation of `createLookAndFeel` needs to be in both the Apple implementation and the dummy implementation, otherwise someone will always get null.

amake:
I've bumped DesktopSupport. Please let me know if this fixes it.

miurahr:
now working on Linux,  java 8, 11 and 16.

amake:

PR #120, opened 2021-06-28
Fix ClassCastException when run integration test on Java 11

miurahr:
Fix [BUG#1053](https://sourceforge.net/p/omegat/bugs/1053/)


amake:
Thanks for looking into this.

It still works on Java 8, right?

miurahr:
> It still works on Java 8, right?

Yes working on Java 8 at least 100 sec. Tested on corretto-1.8.0_292

amake:
Thanks, perfect. I've applied to `master` with a small additional change: the `merge` method was no longer used so I removed it.

PR #123, opened 2021-06-29
Add test cases for MagicComment class

miurahr:
- Add some corner cases
- Set many as skipped  to reuse in future
- Fix expectation typo


amake:
Thanks.

I realized most of the test methods were missing the `@Test` annotation so they weren't being run 😱 

I've squashed to `master`.

PR #129, opened 2021-07-03
Set color scheme defaults on theme class

miurahr:
Implement [RFE#1580](https://sourceforge.net/p/omegat/feature-requests/1580/)

### DIFF
--- a/src/org/omegat/gui/theme/DefaultFlatTheme.java
+++ b/src/org/omegat/gui/theme/DefaultFlatTheme.java
@@ -42,10 +42,12 @@ import javax.swing.border.CompoundBorder;
 import javax.swing.border.EmptyBorder;
 import javax.swing.border.MatteBorder;
 
+import org.omegat.util.Log;
 import org.omegat.util.OStrings;
 import org.omegat.util.Platform;
 import org.omegat.util.gui.ResourcesUtil;
 import org.omegat.util.gui.RoundedCornerBorder;
+import org.omegat.util.gui.UIDesignManager;
 
 
 public class DefaultFlatTheme extends DelegatingLookAndFeel {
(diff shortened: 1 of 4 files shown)


src/org/omegat/gui/theme/DefaultFlatTheme.java
```diff
@@ -107,6 +108,13 @@ public static UIDefaults setDefaults(UIDefaults defaults, String lafId) {
         // #EEEEEE -> #575757; Darkest border
         Color statusAreaColor = adjustRGB(standardBgColor, 0x57 - 0xEE);
 
+        // load default colors
+        try {
+            loadDefaultColors(defaults);
+        } catch (IOException e) {
+            e.printStackTrace();
```
amake, line 115: I know the codebase is not very consistent, but we should probably use `Log.log(e);`

src/org/omegat/gui/theme/DelegatingLookAndFeel.java
```diff
@@ -78,6 +82,95 @@ public DelegatingLookAndFeel() throws Exception {
         systemLookAndFeel = DesktopSupport.getSupport().createLookAndFeel(systemLafName);
     }
 
+    /**
```
amake, line 85: I think your overall idea is good:

1. We should only apply the default colors for the default themes
2. But it should be easy for third-party themes to make use of the default colors as well

But I don't think this code belongs in `DelegatingLookAndFeel`. I think it belongs in one of two places:

- An `OmegaTBaseTheme` that inherits from `DelegatingLookAndFeel`, and from which both default themes inherit from, OR
- It stays in `UIDesignManager` and is called statically from the themes

Since these methods are static, I don't think it makes sense to insert them in the class hierarchy; I feel like they make more sense in `UIDesignManager`.

miurahr: > It stays in UIDesignManager and is called statically from the themes

I like it.

amake:

PR #130, opened 2021-07-04
Dictionary: Avoid short freeze when showing large text

miurahr:
Fix [BUGS#1056](https://sourceforge.net/p/omegat/bugs/1056/).

This is dirty and quick hack but works.

### DIFF
--- a/src/org/omegat/gui/dictionaries/DictionariesTextArea.java
+++ b/src/org/omegat/gui/dictionaries/DictionariesTextArea.java
@@ -89,6 +89,7 @@ public class DictionariesTextArea extends EntryInfoThreadPane<List<DictionaryEnt
         implements IDictionaries, IPaneMenu {
 
     private static final String EXPLANATION = OStrings.getString("GUI_DICTIONARYWINDOW_explanation");
+    private static final int textBatch = 30;
 
     protected final DictionariesManager manager = new DictionariesManager(this);
 
@@ -241,9 +242,9 @@ public class DictionariesTextArea extends EntryInfoThreadPane<List<DictionaryEnt
         if (!data.isEmpty() && Preferences.isPreference(Preferences.NOTIFY_DICTIONARY_HITS)) {
             scrollPane.notify(true);
         }
-        StringBuilder txt = new StringBuilder();
         boolean wasPrev = false;
-        int textLength = 0;
+        int pos = 0;
+        StringBuilder txt = new StringBuilder();
         for (int i = 0, size = data.size(); i < size; i++) {
             DictionaryEntry de = data.get(i);
             if (wasPrev) {
@@ -258,32 +259,25 @@ public class DictionariesTextArea extends EntryInfoThreadPane<List<DictionaryEnt
 
             displayedWords.add(de.getWord().toLowerCase());
             i++;
-
-            if (i % 30 == 0) {
-                String out = txt.toString();
-                if (textLength > 0) {
-                    try {
-                        getDocument().insertString(textLength, out, SimpleAttributeSet.EMPTY);
-                    } catch (BadLocationException e) {
-                        Log.log(e);
-                    }
-                } else {
-                    setText(out);
-                }
+            if (i % textBatch == 0) {
+                pos = pos + addText(pos, txt.toString());
                 txt = new StringBuilder();
-                textLength = out.length();
             }
         }
-        String out = txt.toString();
-        if (textLength > 0) {
+        addText(pos, txt.toString());
+    }
+
+    private int addText(final int pos, final String txt) {
+        if (pos > 0) {
             try {
-                getDocument().insertString(textLength, out, SimpleAttributeSet.EMPTY);
+                getDocument().insertString(pos, txt, SimpleAttributeSet.EMPTY);
             } catch (BadLocationException e) {
                 Log.log(e);
             }
         } else {
-            setText(out);
+            setText(txt);
         }
+        return txt.length();
     }
 
     protected final transient MouseAdapter mouseCallback = new MouseAdapter() {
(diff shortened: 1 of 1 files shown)


src/org/omegat/gui/dictionaries/DictionariesTextArea.java
```diff
@@ -257,8 +258,26 @@ protected void setFoundResult(final SourceTextEntry se, final List<DictionaryEnt
...
+                }
+                txt = new StringBuilder();
+                textLength = out.length();
+            }
+        }
+        if (textLength > 0) {
+            moveCaretPosition(textLength);
+            replaceSelection(txt.toString());
+        } else {
+            setText(txt.toString());
```
amake, line 278: I would rather do something smarter, like working directly with the underlying `Document`: instead of doing a replacement you can call `Document#insertString` with the offset set to the length of the document.

How important is it to batch the appending (like you are doing here in chunks of 30)? If we can get away with not batching, I would prefer to remove it because it complicates the code.

miurahr: I'm not succeeded other than this change. 30 is a heuristics. when appending every line, things went worse.

src/org/omegat/gui/dictionaries/DictionariesTextArea.java
```diff
@@ -257,8 +259,25 @@ protected void setFoundResult(final SourceTextEntry se, final List<DictionaryEnt
...
+    }
+
+    private int addText(final int pos, final String txt) {
+        if (pos > 0) {
+            try {
+                getDocument().insertString(pos, txt, SimpleAttributeSet.EMPTY);
+            } catch (BadLocationException e) {
+                Log.log(e);
+            }
+        } else {
```
amake, line 277: Do we need this `else`? Can't the `pos == 0` case also use `insertString`?

src/org/omegat/gui/dictionaries/DictionariesTextArea.java
```diff
@@ -257,8 +259,25 @@ protected void setFoundResult(final SourceTextEntry se, final List<DictionaryEnt
...
             i++;
+            if (i % textBatch == 0) {
+                pos = pos + addText(pos, txt.toString());
+                txt = new StringBuilder();
+            }
+        }
+        addText(pos, txt.toString());
+    }
+
+    private int addText(final int pos, final String txt) {
```
amake, line 270: We are always appending to the end of the text, so I think it should be fast enough to simply call `document.getLength()` and always append at that position. Then you can remove the `pos` bookkeeping. The method could be called `appendText`.

src/org/omegat/gui/dictionaries/DictionariesTextArea.java
```diff
@@ -88,10 +89,11 @@
         implements IDictionaries, IPaneMenu {
 
     private static final String EXPLANATION = OStrings.getString("GUI_DICTIONARYWINDOW_explanation");
+    private static final int textBatch = 30;
```
amake, line 92: This should be UPPER_SNAKE_CASE.

src/org/omegat/gui/dictionaries/DictionariesTextArea.java
miurahr: > Can't the pos == 0 case also use insertString?

It breaks output.

src/org/omegat/gui/dictionaries/DictionariesTextArea.java
miurahr: When using document.getLength(), I got index error.

src/org/omegat/gui/dictionaries/DictionariesTextArea.java
miurahr: image

src/org/omegat/gui/dictionaries/DictionariesTextArea.java
miurahr: It seems working now.

src/org/omegat/gui/dictionaries/DictionariesTextArea.java
miurahr: It seems that JTextPane/Document need to call `setText(txt)` with significant size of html text once, then `insetText(txt)` works unexpectedly.

amake: Thanks. That's weird. I tried to debug why that is, but it's too much for me.

src/org/omegat/gui/dictionaries/DictionariesTextArea.java
```diff
@@ -240,11 +242,10 @@ protected void setFoundResult(final SourceTextEntry se, final List<DictionaryEnt
         if (!data.isEmpty() && Preferences.isPreference(Preferences.NOTIFY_DICTIONARY_HITS)) {
             scrollPane.notify(true);
         }
-
-        StringBuilder txt = new StringBuilder();
         boolean wasPrev = false;
-        int i = 0;
-        for (DictionaryEntry de : data) {
+        StringBuilder txt = new StringBuilder();
+        for (int i = 0, size = data.size(); i < size; i++) {
```
amake, line 247: You changed the loop to a manually counting loop, but you left the existing increment to `i` so you actually skip some dictionary entries now.

amake:

src/org/omegat/gui/dictionaries/DictionariesTextArea.java
miurahr: I have not fixed yet.

miurahr: thanks.

PR #131, opened 2021-07-06
Reload and Restart rephrasing

brandelune:
Why not use an imperative form ?


miurahr:
A reason is that user should select a `Project->Restart` menu  manually. A string is not on the button to do so.

brandelune:
But Reload is a menu item and Restart has become one, so it seems to me that using the same terms makes the string easier to understand and act on. But I'm neither a native speaker nor a UX specialist, so I won't insist.

miurahr:
> using the same terms makes the string easier to understand and act on.

I agree on it.

> But I'm neither a native speaker nor a UX specialist, so I won't insist.

Same here.

amake:
My impression was that the existing passive form was more common, but some very unscientific Googling indicates that the imperative form is also common.

Honestly I don't think it makes a significant difference, but if you like this better then let's do it.

amake:
One reason to prefer passive is to avoid using the application name in the string. Technically the application name could be customized, so we'd need to interpolate it in.

However I don't think that's a big problem.

I've pushed to `master`.

PR #132, opened 2021-07-16
TM: Prevent showing externalTM's sourceText as translation when foreign language allowed (RFE#1578)

miurahr:
Resolve [RFE#1578](https://sourceforge.net/p/omegat/feature-requests/1578/)

When `ExternalTMFactory.TMXLoader.PROP_FOREIGN_MATCH` is `true`, TM has an entry translation text is as same as source text  that source language is as same as target language.

### DIFF
--- a/src/org/omegat/core/data/ExternalTMFactory.java
+++ b/src/org/omegat/core/data/ExternalTMFactory.java
@@ -136,6 +136,11 @@ public final class ExternalTMFactory {
 
                     // Keep all the Tuvs matching at least the target language
                     for (TMXReader2.ParsedTuv tuvTarget2 : tu.tuvs) {
+                        // skip an entry that is same as tuvSource.
+                        // @see discussion on https://github.com/omegat-org/omegat/pull/132
+                        if (tuvTarget2 == tuvSource) {
+                            continue;
+                        }
                         // Skip entries from the same source language/country
                         if (sourceLang.equals(new Language(tuvTarget2.lang))) {
                             continue;
--- a/src/org/omegat/util/TMXReader2.java
+++ b/src/org/omegat/util/TMXReader2.java
@@ -604,7 +604,9 @@ public class TMXReader2 {
                 tuvLC = tuv;
             } else {
                 // other country
-                tuvLW = tuv;
+                if (tuvLW == null) {  // take first occurrence
+                    tuvLW = tuv;
+                }
             }
         }
         ParsedTuv bestTuv;
--- a/test/data/tmx/en-US_en-GB_fr_sr.tmx
+++ b/test/data/tmx/en-US_en-GB_fr_sr.tmx
@@ -1,0 +1,21 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<!DOCTYPE tmx SYSTEM "tmx14.dtd">
+<tmx version="1.4">
+  <header creationtool="Translate Toolkit" creationtoolversion="2.5.0" segtype="sentence" o-tmf="UTF-8" adminlang="en" srclang="en-US" datatype="PlainText"/>
+  <body>
+    <tu>
+      <tuv lang="en-US">
+        <seg>XXx</seg>
+      </tuv>
+      <tuv lang="en-GB">
+        <seg>XXX</seg>
+      </tuv>
+      <tuv lang="fr">
+        <seg>YYY</seg>
+      </tuv>
+      <tuv lang="sr">
+        <seg>ZZZ</seg>
+      </tuv>
+    </tu>
+  </body>
+</tmx>
--- a/test/data/tmx/en-US_sr.tmx
+++ b/test/data/tmx/en-US_sr.tmx
@@ -1,0 +1,15 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<!DOCTYPE tmx SYSTEM "tmx14.dtd">
+<tmx version="1.4">
+  <header creationtool="Translate Toolkit" creationtoolversion="2.5.0" segtype="sentence" o-tmf="UTF-8" adminlang="en" srclang="en-US" datatype="PlainText"/>
+  <body>
+    <tu>
+      <tuv xml:lang="en-US">
+        <seg>XXX</seg>
+      </tuv>
+      <tuv xml:lang="sr">
+        <seg>YYY</seg>
+      </tuv>
+    </tu>
+  </body>
+</tmx>
--- a/test/src/org/omegat/core/data/ExternalTMFactoryTest.java
+++ b/test/src/org/omegat/core/data/ExternalTMFactoryTest.java
@@ -165,25 +165,21 @@ public class ExternalTMFactoryTest extends TestCore {
         tmx = ExternalTMFactory.load(tmxFile);
 
         // All foreign translations are present
-        assertEquals(14, tmx.getEntries().size());
+        assertEquals(11, tmx.getEntries().size());
 
         matchingEntries = tmx.getEntries().stream().filter(t -> t.getSourceText().equals("Hello World!"))
                 .collect(Collectors.toList());
-        assertEquals(8, matchingEntries.size());
+        assertEquals(7, matchingEntries.size());
 
         matchingEntries = tmx.getEntries().stream().filter(t -> t.getSourceText().equals("This is an english sentence."))
                 .collect(Collectors.toList());
-        assertEquals(3, matchingEntries.size());
+        assertEquals(2, matchingEntries.size());
 
         ITMXEntry entry = matchingEntries.get(0);
-        assertEquals("EN-US", entry.getPropValue(ExternalTMFactory.TMXLoader.PROP_TARGET_LANGUAGE));
-        assertEquals("true", entry.getPropValue(ExternalTMFactory.TMXLoader.PROP_FOREIGN_MATCH));
-
-        entry = matchingEntries.get(1);
         assertEquals("DE", entry.getPropValue(ExternalTMFactory.TMXLoader.PROP_TARGET_LANGUAGE));
         assertEquals("true", entry.getPropValue(ExternalTMFactory.TMXLoader.PROP_FOREIGN_MATCH));
 
-        entry = matchingEntries.get(2);
+        entry = matchingEntries.get(1);
         assertEquals("ES", entry.getPropValue(ExternalTMFactory.TMXLoader.PROP_TARGET_LANGUAGE));
         assertEquals("true", entry.getPropValue(ExternalTMFactory.TMXLoader.PROP_FOREIGN_MATCH));
     }
--- a/test/src/org/omegat/core/statistics/FindMatchesTest.java
+++ b/test/src/org/omegat/core/statistics/FindMatchesTest.java
@@ -1,0 +1,220 @@
+/**************************************************************************
+ OmegaT - Computer Assisted Translation (CAT) tool
+          with fuzzy matching, translation memory, keyword search,
+          glossaries, and translation leveraging into updated projects.
+
+ Copyright (C) 2021 Hiroshi Miura
+               Home page: http://www.omegat.org/
+               Support center: https://omegat.org/support
+
+ This file is part of OmegaT.
+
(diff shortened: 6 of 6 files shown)


miurahr:
The problem is because `ExternalTMFactory` don't check with TMX srcLang with TUV source language when importing.
Current implementation compare Project source Language with TUV source language.
It lead a problem when project language code is different from TMX language, such as `en` against `en-US`

A patch I proposed is not direct solution but it works as workaround for that `ExternalTMFactory` should not import source text with a language which is as same as TMX header's srcLang, as the translation text.

miurahr:
I tried to fix the root cause to detect region difference in #133.

src/org/omegat/core/statistics/FindMatches.java
```diff
@@ -233,15 +233,22 @@ public void iterate(EntryKey source, TMXEntry trans) {
...
+
+                if (requiresTranslation) {
+                    if (tmen.translation == null) {
+                        continue;
+                    }
+                    String sourceLang = tmen.getPropValue(ExternalTMFactory.TMXLoader.PROP_TARGET_LANGUAGE);
+                    String targetLang = tmen.getPropValue(ExternalTMFactory.TMXLoader.PROP_SOURCE_LANGUAGE);
+                    if (sourceLang.equals(targetLang)) {
+                        continue;
+                    }
```
amake, line 245: I don't understand how this solves the issue. Aren't you comparing the source and target of the TM entry itself? Don't we need to compare to the *project's* source language?

miurahr: A discussion here is on assumption of  `ExternalTMFactory.TMXLoader.PROP_FOREIGN_MATCH` is `true`

When _project_ 's source language is as same as _TM_ 's source language(ex. both is "en-US"), this comparison always become `false`.

When _project_ 's source language is different from _TM_ 's source language, such as
  _project_'s source language does not have region code but _TM_ source language has. 
      (ex. project source is "en" , and  TM source is "en-US" and target is "es" )

 For example,
 - TM has entries from "en-US" to "es" ,  
 - project source language is "en". 
 - TM has an entry  "en-US" : "Hello World",  "es": "Hola Mundo"

 `TMXReader2` produce  two entries
- sourceLang = "en-US", targetLang= "en-US" sourceEntry="Hello World" targetEntry="Hello World"
- sourceLang= "en-US", targetLang="es" sourceEntry="Hello World" targetEntry="Hola Mundo" 

This output is tested on 
https://github.com/miurahr/omegat/blob/master/test/src/org/omegat/core/data/ExternalTMFactoryTest.java#L178-L180

So This comparison drop a case `sourceLang="en-US", targetLang="en-US" sourceEntry="Hello World", targetEntry="Hello World"` from displaying in the pain then solve the problem.

miurahr: > Don't we need to compare to the project's source language?

_Project_'s source language is compared with TM source language at 
https://github.com/omegat-org/omegat/blob/18db07c24b44158b79cb3d29995ba4caf68ded1b/src/org/omegat/core/data/ExternalTMFactory.java#L137-L149

Here is a place the issue is caused.

When project's source language is not exactly same as TM's source language, the issue is happened.
In addition to the fact,  `ExternalTMFactory` does not check TM's _header_ srcLang and `tuv`'s lang.

It is useful when translator want to use reverse TM , for example TM source is "EN" and target is "ES" but project source is "ES" and target "EN".

miurahr: further comment?

miurahr:
I'd like to add test case what this patch is solved.

miurahr:
Test failed because added case is not complete. I'd like to ask advise and suggestions how to implement the case RFE#1578 and what's wrong. @amake could you give me comments?

miurahr:
@amake you can observe a behavior and conditions of the issue from reproduce test case. 73def88 become failure as expected, and 19865f9 fixes the issue.

t-cordonnier (commented):
First comment: the option  "Include matches from other languages" is implemented in ExternalTMFactory, so during the load of the file. But you try to solve the problem in FindMatches, meaning : during display.
That makes a difference: doing it in ExternalTMFactory means that the undesired entries will simply never be in memory, while doing it in FindMatches means that they are in memory but always filtered ... except in Search class, which does not use FindMatches.
The consequence is an inconsistency: entry with other language will not be loaded in memory, so won't appear neither in matches nor in searches, while entry with source language will disappear from matches but not from searches!

t-cordonnier:
In the RFE I agree more with what you wrote here: https://sourceforge.net/p/omegat/feature-requests/1578/#9e5d

See ExternalTMFactory, line 140:
`                        if (sourceLang.equals(new Language(tuvTarget2.lang))) { `
Here if sourceLang = "en" and tuvTarget2.lang = "en-US", or the contrary, the entry is considered as foreign, while it should not, except eventually in one case: if the TMX entry contains a tuv with `lang="en"` and _another_ one with `lang="en-US"`: then maybe it makes sense to consider that the one which does not exactly match source language should be displayed. If only one of the two exists, it is necessarily identical to the source so it does not make sense to display it

But you can ask Aaron and Manuel (the requester) if they approve this

miurahr:
@msoutopico @amake 

@t-cordonnier  said
> But you can ask Aaron and Manuel (the requester) if they approve this

Your comments?

msoutopico:
> @msoutopico @amake
> 
> @t-cordonnier said
> 
> > But you can ask Aaron and Manuel (the requester) if they approve this
> 
> Your comments?

Sorry, I saw this today. Probably it's too late for any comments. I will test it and report. 
Please let me know if any specific question is pending an answer.

msoutopico:
> @msoutopico I'd like to make requirements clear. Could you read [#132 (review)](https://github.com/omegat-org/omegat/pull/132#pullrequestreview-921871511) and [#132 (comment)](https://github.com/omegat-org/omegat/pull/132#issuecomment-1079193144) opinion from @t-cordonnier and share your feedback whether it is fit for your requirements?

First of tall, thanks Hiroshi for your work and thanks Thomas for your comments. 

I agree with what Thomas says in [#132 (review)](https://github.com/omegat-org/omegat/pull/132#pullrequestreview-921871511): I would expect consistency in results displayed in the Matches pane and in the Find dialog.

I also have an opinion about [#132 (comment)](https://github.com/omegat-org/omegat/pull/132#issuecomment-1079193144) but I need to run some tests and think of different scenarios. On Monday I could also talk with some colleagues, to confirm or fine-tune my opinion. If you can wait, I'll get back to you on this on Monday.

miurahr:
@t-cordonnier how do you handle  a mandatory attribute `srclang` of `<header>` in TMX file?
I think  a source language of TMX file is defined in header. Isn't it?

miurahr:
I've revert the change and update test case.
You can try your idea on it.

miurahr:
I've revert the change and update test case.
You can try your idea on it.

t-cordonnier:
> @t-cordonnier how do you handle a mandatory attribute `srclang` of `<header>` in TMX file? I think a source language of TMX file is defined in header. Isn't it?

This is the source from the point of view of the author of the TMX. But if you use a TMX from outside your company, the author cannot know what is the source of your project.
Honnestly I never saw any CAT tool doing anything from this info. TMX is fundamentally unordered format, where each language can be considered as source or target depending on the tool using it.

t-cordonnier:
> Current logic in ExternalTMXFactory is
>` 
>                         if (sourceLang.equals(new Language(tuvTarget2.lang))) {
>                             continue;
>                         }
> `
> 
> A root cause may be here, boolean `new Language("en").equals(new Language("en-US"))` is false.

Yes, that is it. I just thought that a good solution would be to write instead
`                         if (tuvTarget2 == tuvSource) {
                             continue;
                         }` 
tuvSource is filled with the tuv having either lang.equals(sourceLang), or, if not found, the first language variant which is found in the TMX for this language. This is done in TMXReader2 before calling this method.
So, what would happen is
- if TMX contains a value for srcLang, use it as source and exclude it from the loop
- if TMX contains a value for one variant of srcLang, use it as source and exclude it from the loop
- if TMX contains a value for multiple variants of srcLang (for example srcLang = en, TMX contains en-US and en-GB but not en) then the first one will be in tuvSource, will be used as source and excluded from the source, then the others are not excluded from the loop

I did not test yet, and since I answered yesterday to this question from @msoutopico, let's wait his confirmation that this would be an acceptable behavior

miurahr:
> * if TMX contains a value for multiple variants of srcLang (for example srcLang = **en**, TMX contains **en-US** and **en-GB** but not **en**) then the first one will be in tuvSource, will be used as source and excluded from the loop, then the others are not excluded from the loop

I've added the case that import a TMX file with **en-US**, **en-GB**, **fr**, and **sr**.
This will illustrate a solution capability and feasibility.

t-cordonnier:
> ```
>                         // Skip source tuv.
>                         if (tuvSource.lang.equals(tuvTarget2.lang)) {
>                             continue;
>                         }
> ```
> 
> This is basically as same as your idea.
> 

Not exactly: there is no check that tuvTarget2 != null (there was a check for tuvSource a few lines before) so you take the risk of an exception.

miurahr:
@t-cordonnier your proposal?

miurahr:
According to @t-cordonnier suggestions in https://github.com/omegat-org/omegat/pull/132#issuecomment-1096743242 
- update a test expectation cf0eff62cae1b76f64d2be1573b20215dbcc8615  (must be failed)
- add a logic suggested deb0b0e

miurahr:
A test case `ExternalTMFactoryTest#testFuzzyMultipleTuv` failed for  deb0b0e because expectation is 14 entries includes three EN-US entry but the logic changed a behavior and that is now 11 entries.

miurahr:
I've updated an expectations.

t-cordonnier:
> A test case `ExternalTMFactoryTest#testFuzzyMultipleTuv` failed for [deb0b0e](https://github.com/omegat-org/omegat/commit/deb0b0e9dcc7ed8e17f399c5047d96cc70d46045) because expectation is 14 entries includes three EN-US entries but 11 entries.
> 
> and assertion
> 
> ```
>         ITMXEntry entry = matchingEntries.get(0);
>         assertEquals("EN-US", entry.getPropValue(ExternalTMFactory.TMXLoader.PROP_TARGET_LANGUAGE));
>         assertEquals("true", entry.getPropValue(ExternalTMFactory.TMXLoader.PROP_FOREIGN_MATCH));
> ```
> 
> also failed, because there is no "EN-US" entry.
> 
> @t-cordonnier is this as same as your expectation?

Yes, I just reviewed the test and agree that it corresponds to what I expected.

Another use case we should eventually add in this test is a new TU containing both en-US and en-GB: in such a case, the entry en-US=>en-US should disappear but not the entry en-US=>en-GB !

miurahr:
I found there is mistake in the test case tmx file. I've update the file and expectations that is as same as commit comment.

The test should be failed because OmegaT cannot handle the case properly in current code.

> Another use case we should eventually add in this test is a new TU containing both en-US and en-GB: in such a case, the entry en-US=>en-US should disappear but not the entry en-US=>en-GB !

> Well, this use case is covered by your test in FindMatchesTest
One comment however, in the code of the test I read
assertEquals("XXx", result.get(0).source); // en-US, because header's srclang=en-US
If I am not wrong, nowhere in the code there is any call to header's srclang. And as you asked me the question, my impression is that header's attributes refer to something else and we should not take them in account.
Here the source language is en-US simply because this is the first language which appears in the TU and which is a sublanguage of the project's source language. This has nothing to do with header attributes.

 I agree that "nowhere in the code there is any call to header's srclang."

> " header's attributes refer to something else and we should not take them in account."

Unfortunately or fortunately OmegaT  ALWAYS ignore header's attribute when detect source language. Is it correct behavior?

miurahr:
@t-cordonnier  Now the change pass the test expectation. It take a first occurrence of source language entry that has same language code.

miurahr:
Now adjusted all the comments. Added Tests are passed with correct expectation.
I'd like to merge it soon.

PR #133, opened 2021-07-17
External TM read from different source region

miurahr:
- (change behavior) Skip TM file that has different source language from project such as `en` and `fr`
- Read TM with different source region and same language   when Preference.EXT_TMX_KEEP_DIFFERENT_SOURCE_REGION is set
   - Default true
   - for example, `en-US` and `en-GB` or `en`
- Related with [RFE#1578](https://sourceforge.net/p/omegat/feature-requests/1578/)
- Preference GUI is not implemented because of [BUG#1057](https://sourceforge.net/p/omegat/bugs/1057/) prevent code modification.


miurahr:
The behavior change may prevent utilize reverse TM; for example the project is `en` -> `es` but external TM is `es`->`en`.

PR #134, opened 2021-07-19
Support tracking branch name for remote git repository (RFE#918)

miurahr:
This allow project manager to maintain their team repository with specific branch name other than 'main' or 'master' default branch.

User can specify branch name when downloading project from dialog.

- [RFE#918](https://sourceforge.net/p/omegat/feature-requests/918/)

TODO:

- [x] field test 
- [x] integration test

### DIFF
--- a/src/org/omegat/core/team2/impl/GITRemoteRepository2.java
+++ b/src/org/omegat/core/team2/impl/GITRemoteRepository2.java
@@ -39,7 +39,9 @@ import java.util.stream.StreamSupport;
 
 import javax.xml.namespace.QName;
 
+import org.eclipse.jgit.api.CheckoutCommand;
 import org.eclipse.jgit.api.CloneCommand;
+import org.eclipse.jgit.api.CreateBranchCommand;
 import org.eclipse.jgit.api.Git;
 import org.eclipse.jgit.api.LsRemoteCommand;
 import org.eclipse.jgit.api.ResetCommand.ResetType;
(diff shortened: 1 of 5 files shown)


src/org/omegat/core/team2/impl/GITRemoteRepository2.java
```diff
@@ -143,6 +158,12 @@ public void init(RepositoryDefinition repo, File dir, ProjectTeamSettings teamSe
             }
             repository = Git.open(localDirectory).getRepository();
             try (Git git = new Git(repository)) {
+                if (trackBranch) {
+                    git.branchCreate().setName(branch).setUpstreamMode(CreateBranchCommand.SetupUpstreamMode.TRACK).setStartPoint("origin/" + branch).call();
```
miurahr, line 162: `"origin" + branch` part should be `String.join("/", REMOTE, branch)`

amake:
> This allow project manager to maintain their team repository with specific branch name using "#" for target URL.
> such as "https://github.com/team/repo.git#translate"

This is inventing a new URL format to indicate the git URL + branch. I don't think we should do that.

If there is a standard for such a thing, we should use the standard (I don't think there is).

Otherwise we should either a) allow the user to provide it separately or, preferably, b) automatically detect it.

miurahr:
This has been inspired from [RFE#918](https://sourceforge.net/p/omegat/feature-requests/918/)  and just found in my local working branch today.

>  a) allow the user to provide it separately 

I agree with it.

miurahr:
Because #174 breaks this improvement,  I'd like to update #174 to implement RFE#918.

miurahr:
Superseded by #179

PR #137, opened 2021-07-20
Utilize Jackson library for XML parser

miurahr:
Jackson library (Apache2 license) has a good compatibility with JAXB standard.

Because we now discuss a migration to Jackson for JSON parser in RFE#1575, it is good chance to consider also to switch to Jackson for XML.

- Implement [RFE#1582](https://sourceforge.net/p/omegat/feature-requests/1582/)


amake:
I guess my question is "why?" What is the benefit of making this change?

miurahr:
> Why

When RFE#1575 is accepted, a consolidation to jackson  may help reducing distribution size (reducing bundled jar files)

amake:
OK, that could be nice. If it does produce some size savings, I'd like to know exactly how much.

Also keep in mind that we need to be able to *generate* with JAXB as well. The files in `src/gen` are supposed to be generated by JAXB from the files in `src/schemas`. This might be broken right now though.

miurahr:
> Also keep in mind that we need to be able to generate with JAXB as well. The files in src/gen are supposed to be generated by JAXB from the files in src/schemas. This might be broken right now though.

Oh, actually these are broken. I'd like to close as invalid.

miurahr:
> Also keep in mind that we need to be able to generate with JAXB as well. 

The issue for generating JAXB java is not direct problem on jackson, it is rather Java11 deprecation of xjc.
I've reopen the PR.

amake:
Yes I am working on generating with JAXB but without xjc. We should be free to use other solutions at runtime.

miurahr:
I see this plugin
https://github.com/bjornvester/xjc-gradle-plugin#generate-resources-with-different-configurations

And I found we need to use Gradle 7.0 for the solution.

amake:
I would like to minimize use of plugins because they frequently fall behind Gradle and prevent us from making other changes. For instance the JNLP plugin is currently incompatible with Gradle 7.0, so we can't upgrade yet.

miurahr:
> I'd like to know exactly how much.

jaxb-api:             128 kb
jaxb-*(w/o api): 1252 kb
Jackson:           1290 kb
Jackson(-xml):    121 kb

build.gradle
```diff
@@ -111,9 +111,13 @@ dependencies {
...
 
+        // XML/JSON parser
         api 'javax.xml.bind:jaxb-api:2.3.1'
-        implementation 'com.sun.xml.bind:jaxb-core:2.3.0.1'
-        implementation 'com.sun.xml.bind:jaxb-impl:2.3.2'
+        implementation "com.fasterxml.jackson.core:jackson-core:2.12.4"
+        implementation "com.fasterxml.jackson.core:jackson-annotations:2.12.4"
+        implementation "com.fasterxml.jackson.core:jackson-databind:2.12.4"
+        implementation "com.fasterxml.jackson.dataformat:jackson-dataformat-xml:2.12.4"
+        implementation "com.fasterxml.jackson.module:jackson-module-jaxb-annotations:2.12.4"
```
miurahr, line 120: jackson-annotations and jackson-module-jaxb-annotations may not be required when we do not use annotations

miurahr:
> Yes I am working on generating with JAXB but without xjc.

I'd like to wait its progress...

amake:
I've pushed the changes to JAXB codegen in https://github.com/omegat-org/omegat/commit/42eb82ab188f66d75a0579fde89400c18156b517

Note also that it turns out we don't need `com.sun.xml.bind:jaxb-core` at all anymore per https://github.com/omegat-org/omegat/commit/cc8b99a03c89f16e83eb3709651286701d9dc531

Do I understand that your proposal is essentially to replace the `com.sun.xml.bind:jaxb-impl` implementation of JAXB with the Jackson implementation?

miurahr:
> Do I understand that your proposal is essentially to replace the com.sun.xml.bind:jaxb-impl implementation of JAXB with the Jackson implementation?

Yes it is. So I'd like to replace JAXB bulider  with jackson.

amake:
I'm still fuzzy on the benefit of doing so. Per your previous comment, it seems like there's almost no difference in binary size?

miurahr:
> Per your previous comment, it seems like there's almost no difference in binary size?

Yes, there are similar size of jars in two implementations.

When we migrate JSON parser to Jackson,  dist increase 1.3Mb.
This commit will decrease 1.3Mb.

miurahr:
Failures of tests are caused by difference in behavior. I now think it is a bad attempt to switch JAXB implementation to Jackson.
I'd like to give it up and close here.

amake:
Should this be closed?

miurahr:
Closed.

PR #138, opened 2021-07-21
Replace runtime, compile dependency keywords for Gradle 7.0+

miurahr:
- implicit extension is deprecated: explicit definition of testIntegration configuration
- runtime deprecated: use runtimeOnly


build.gradle
```diff
@@ -90,6 +90,8 @@ configurations {
         exclude group: 'org.apache.lucene', module: 'lucene-core'
     }
     [testRuntime, testCompile]*.exclude group: 'org.languagetool', module: 'language-all'
+    testIntegration
+    testIntegraitonImplementation.extendsFrom testImplementation
```
amake, line 94: Typo: Integraiton

build.gradle
```diff
@@ -343,7 +345,7 @@ distributions {
                 exclude '**/build/**', 'doc_src/**/pdf/**', 'local.properties'
             }
             into('lib/provided') {
-                from configurations.runtime
+                from configurations.runtimeOnly
```
amake, line 348: Did you test that this actually results in all of the dependencies we need being copied? Because I think `runtimeOnly` is not equivalent to `runtime`. I think we need to include `compileOnly` as well.

amake (commented):
Thank you, it's great to prepare for Gradle 7.0.

Unfortunately we can't move to Gradle 7.0 until the JNLP plugin is updated, but that's blocked per https://github.com/tschulte/gradle-jnlp-plugin/issues/58#issuecomment-851332165

build.gradle
miurahr: runtimeOnly + compileOnly = Implementation.

I'd like to use implementation here.

miurahr: Here is a place for source distribution. I may need to test whether  source distribution  compile well.

miurahr:
This PR also fix [BUGS#1058](https://sourceforge.net/p/omegat/bugs/1058/) that source zip lacks provided libraries.

build.gradle
amake: Thanks for looking into it. Ideally we should test that the source distro compiles as part of our CI.

build.gradle
```diff
@@ -343,7 +347,8 @@ distributions {
                 exclude '**/build/**', 'doc_src/**/pdf/**', 'local.properties'
             }
             into('lib/provided') {
-                from configurations.runtime
+                from configurations.implementation, configurations.api
+
```
amake, line 351: According to [this](https://docs.gradle.org/current/userguide/java_plugin.html#tab:configurations) we can do
```suggestion
                from configurations.runtimeClasspath
```

This works in my testing.

build.gradle
```diff
@@ -995,7 +1000,7 @@ task jar2(type: Sync) {
     description = 'Create minimal executable output for debugging.'
     destinationDir = file("${buildDir}/install/jar2")
     from jar
-    from(configurations.runtime) {
+    from(configurations.runtimeOnly) {
```
amake, line 1003: I will simply remove `jar2`. It is a remnant of something we had when we were using Ant. Didier requested we have this because it's supposedly faster, but I find that `./gradlew installDist` is plenty fast.

amake:
I have applied changes in this PR, or equivalent ones, to `master`.

I still can't get Gradle to compile the source distribution's build script, however:

```
% ./gradlew

FAILURE: Build failed with an exception.

* Where:
Build file '/Users/amake/Code/omegat/build/install/OmegaT-source/build.gradle' line: 463

* What went wrong:
A problem occurred evaluating root project 'OmegaT'.
> Expected configuration ':runtimeClasspath' files to contain exactly one file, however, it contains no files.

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 1s
```

The problem is the Mac-specific codesigning mechanism needs to identify certain JARs (which contain binaries needing special handling) and those JARs can't be identified when the dependencies are included as `implementation fileTree(...)`.

I'm not sure of the best way to handle this right now.

miurahr:
> The problem is the Mac-specific codesigning mechanism needs to identify certain JARs (which contain binaries needing special handling) and those JARs can't be identified when the dependencies are included as implementation fileTree(...).
> 
> I'm not sure of the best way to handle this right now.

We can put dependencies in folder `provided/compile` and `provided/runtime` then loading the former as `implement` and the latter as `runtimeOnly`. It may help the code find a hunspell library.

Here is another idea to use folder `provided` and `provided/native`, then introduce `configurations.native` to handle it.

I prefer 'native' configuration.

amake:
When getting dependencies from `lib/provided`, the content of `configurations.runtimeClasspath.fileCollection` is a single item, an instance of [SelfResolvingDependency](https://docs.gradle.org/current/javadoc/org/gradle/api/artifacts/SelfResolvingDependency.html). I don't think any of your suggestions would change that. The only thing that could work is somehow resolving the dependency, but a) I'm not sure how, and b) that needs a different code path from the default case, so I'm thinking the entire approach is bad.

build.gradle
```diff
@@ -459,9 +469,7 @@ installMacDist {
     }
 }
 
-def hunspellJar = configurations.runtimeClasspath.fileCollection {
```
miurahr, line 462: runtimeClasspath can be dictionary [group, name, version], or set of File by fileTree()
When created y fileTree(), it does not have name.

miurahr:
We can also take a way **not** to provide dependencies libraries in `lib/provided`.

amake:
> We can also take a way **not** to provide dependencies libraries in `lib/provided`.

The reason we create `lib/provided` is to guard against dependencies disappearing from Maven Central in the future. It was a concern of some of the other developers when we switched to Gradle (before that, we were storing all dependency JARs in version control).

Someone suggested we don't even need the source distribution anymore. If we got rid of it, then all of these issues would be moot. I'm not sure what I want to do yet, though.

build.gradle
```diff
@@ -459,9 +462,7 @@ installMacDist {
...
-    it.name == 'hunspell-native-libs'
-}.singleFile
+def hunspellJar = configurations.hunspell.files.asType(List).get(0)
```
miurahr, line 465: I cannot find a better way to do this.

miurahr:
IMHO 6a70ce7 is small enough for the work around. Does it works on mac?

build.gradle
miurahr: This is the way;

```def hunspellJar = configurations.runtimeClasspath.files.find {it.path.contains("hunspell-native-libs")}```

miurahr:
c9982ac can produce a sourceJar which  run  gradle without the problem.

amake:
Thanks very much for finding this approach. I made a small tweak and pushed to `master`.

amake:
I think all of the changes from this PR have been incorporated; if there's nothing else then can we close?

miurahr:
Yes I'd like to close.

PR #139, opened 2021-07-21
Add config folder in source zip

miurahr:
Fix [BUG#1059](https://sourceforge.net/p/omegat/bugs/1059/)


amake:

PR #140, opened 2021-07-31
Feature: Add TMXprop what MT engine's translation is used (RFE#1584)

miurahr:
Automatically add property which MT engine's translation is used.
It is optional and introduce options to set a behavior in MT preference.

Implement [RFE1584](https://sourceforge.net/p/omegat/feature-requests/1584/)
Related [RFE1467](https://sourceforge.net/p/omegat/feature-requests/1467/)
Related [RFE1477](https://sourceforge.net/p/omegat/feature-requests/1477/)


miurahr:
>  The mechanism we have for storing metadata about a translation is TMXEntry. I don't think it's a good idea to add an mtEngine property to TMXEntry though. I think we should consider some sort of extensible property container like a hashmap.
> 
> We can write such properties to TMX as <prop>s. In that case we need to remember to read them as well.

We may want to add TMX prop data class and utility to add/modify it.
If there has already been the one, could you suggest it for me?

amake:
> We may want to add TMX prop data class and utility to add/modify it.
> If there has already been the one, could you suggest it for me?

There is a `TMXProp` class, but it is only involved in reading/writing actual TMX files.

The representation for our in-memory working TM is `ProjectTMX` and this contains `TMXEntry`s, but actually both of these don't really have anything to do with the TMX format, so the naming is quite bad.

So we already have most of the requirements for reading/writing actual TMX files. But we don't have a general metadata facility for the in-memory TM (`ProjectTMX`/`TMXEntry`).

miurahr:
> So we already have most of the requirements for handling arbitrary properties when reading/writing actual TMX files. But we don't have a general metadata facility for the in-memory TM (ProjectTMX/TMXEntry).

Yes, `TMXEntry` only have String `note` , not a list of properties with key and value pair.
Can we extend it with prop which have `note` key in default?

miurahr:
Now I can save MT source name into `project_save.tmx` and read it again.
When entry has `Prop.MTSOURCE` other property, property pane can show it.

But I'm wondering why OmegaT does not show the property.
I've checked `ProjectTMX#defaults` has a entry which has MT source property.
Any suggestions?

src/org/omegat/core/data/ProjectTMX.java
```diff
@@ -66,6 +66,7 @@
     protected static final String PROP_XICE = "x-ice";
     protected static final String PROP_X100PC = "x-100pc";
     protected static final String PROP_XAUTO = "x-auto";
+    protected static final String PROP_MTSOURCE = "mtsource";
```
miurahr, line 69: Is `x-mtsource` better?

src/org/omegat/util/TMXWriter2.java
```diff
@@ -221,6 +221,14 @@ public void writeEntry(String source, String translation, String note, String cr
             xml.writeCharacters(lineSeparator);
         }
 
+        if (!StringUtil.isEmpty(mtsource)) {
+            xml.writeCharacters("      ");
+            xml.writeStartElement("mtsource");
```
miurahr, line 226: It should be constant variable

PR #141, opened 2021-07-31
Introduce internal id

miurahr:
Implement [RFE#1565](https://sourceforge.net/p/omegat/feature-requests/1565/)

Introduce internal_id even when filter does not provide segment unit id.
It is generated from a hashCode of source text, creator and create date of the default entry. Alternative entry also have a same id,  that will help OmegaT find a matching segment.

RFE creator: @msoutopico

### DIFF
--- a/src/org/omegat/core/data/DataUtils.java
+++ b/src/org/omegat/core/data/DataUtils.java
@@ -25,12 +25,7 @@
 
 package org.omegat.core.data;
 
-import java.util.Set;
-import java.util.TreeSet;
-
 public final class DataUtils {
-
-    private static final Set<String> INTERNAL_IDS = new TreeSet<>();
 
     private DataUtils() {
     }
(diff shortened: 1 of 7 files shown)


src/org/omegat/util/StaticUtils.java
```diff
@@ -579,4 +579,9 @@ public static boolean isProjectDir(File f) {
         return true;
     }
 
+    public static String getInternalId(String source, String creator, long created) {
```
miurahr, line 582: The name should be `generate...` not `get...`

miurahr: and it can be in ProjectTMX class.

src/org/omegat/core/data/DataUtils.java
```diff
@@ -25,12 +25,34 @@
 
 package org.omegat.core.data;
 
+import java.util.Set;
+import java.util.TreeSet;
+
 public final class DataUtils {
 
+    private static final Set<String> INTERNAL_IDS = new TreeSet<>();
```
miurahr, line 33: Now this is alive during OmegaT application run.
This should be a project unique so it should be in ProjectTMX or RealProject class.

miurahr:
I found the approach is wrong. I'd like to close it.

PR #143, opened 2021-07-31
Add extra ignore patterns for ".git", ".hg", and ".repositories"

zephyrus00jp:
Today people use source code version control systems such as GIT, MERCURIUM and others.
So we should add DEFAULT IGNORE PATTERNS used for remote repositories other than ".cvs", etc.

This patch does exactly that.

It also adds ".repositories" because an incorrect repository specification that includes ".repositories"  as part of "source" directory used by OmegaT by mistake is so confusing due to the listing of many files underneath ".repositories".


amake:
Thanks, I've pushed to `master`.

PR #144, opened 2021-09-04
DictionaryPane performance hack

miurahr:
This improve performance BUGS#1069

- Fix Vertical scrollbar policy to always show during update document
- Create and set new Document instead of updating document

Call tree caputures, before
capture_before

after
capture_improved

Both are captured by  same seanario.

### DIFF
--- a/src/org/omegat/gui/dictionaries/DictionariesTextArea.java
+++ b/src/org/omegat/gui/dictionaries/DictionariesTextArea.java
@@ -36,6 +36,9 @@ import java.awt.event.ActionListener;
 import java.awt.event.MouseAdapter;
 import java.awt.event.MouseEvent;
 import java.io.File;
+import java.io.IOException;
+import java.io.Reader;
+import java.io.StringReader;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
@@ -49,6 +52,7 @@ import javax.swing.JScrollPane;
 import javax.swing.SwingUtilities;
 import javax.swing.text.BadLocationException;
 import javax.swing.text.Document;
+import javax.swing.text.EditorKit;
 import javax.swing.text.Element;
 import javax.swing.text.html.HTMLDocument;
 import javax.swing.text.html.StyleSheet;
@@ -89,7 +93,6 @@ public class DictionariesTextArea extends EntryInfoThreadPane<List<DictionaryEnt
         implements IDictionaries, IPaneMenu {
 
     private static final String EXPLANATION = OStrings.getString("GUI_DICTIONARYWINDOW_explanation");
-    private static final int TEXT_BATCH = 30;
 
     protected final DictionariesManager manager = new DictionariesManager(this);
 
@@ -261,10 +264,6 @@ public class DictionariesTextArea extends EntryInfoThreadPane<List<DictionaryEnt
 
             displayedWords.add(de.getWord().toLowerCase());
             i++;
-            if (i % TEXT_BATCH == 0) {
-                appendText(txt.toString());
-                txt = new StringBuilder();
-            }
         }
         appendText(txt.toString());
         scrollPane.setVerticalScrollBarPolicy(old);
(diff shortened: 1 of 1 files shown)


miurahr:
By adding `<html> .... </html>` JTextPane can handle several type of tags properly, such as  `<image src=..` tag.
It help dictionary plugins to show image content on dictionary text area. 
In my case, I'd like to put Gaiji glyph image by omegat-plugin-epwing plugin.

src/org/omegat/gui/dictionaries/DictionariesTextArea.java
```diff
@@ -242,42 +256,50 @@ protected void setFoundResult(final SourceTextEntry se, final List<DictionaryEnt
...
+        try {
+            Reader r;
+            if (doc.getLength() == 0) {
+                r = new StringReader(txt);
+            } else {
+                StringBuilder sb = new StringBuilder(doc.getText(0, doc.getLength())).append(txt);
+                r = new StringReader(sb.toString());
+            }
             // Appending to an empty document results in treating HTML tags as
             // plain text for some reason
```
amake, line 296: I think this comment is not needed anymore?

src/org/omegat/gui/dictionaries/DictionariesTextArea.java
```diff
@@ -242,42 +256,50 @@ protected void setFoundResult(final SourceTextEntry se, final List<DictionaryEnt
...
-            setText(txt);
-        } else {
-            try {
-                doc.insertString(doc.getLength(), txt, null);
-            } catch (BadLocationException e) {
-                Log.log(e);
-            }
+            doc = kit.createDefaultDocument();
+            ((HTMLDocument) doc).setPreservesUnknownTags(false);
```
amake, line 298: A comment explaining this line would also be nice.

src/org/omegat/gui/dictionaries/DictionariesTextArea.java
```diff
@@ -141,8 +150,9 @@ public void setFont(Font font) {
                 + " font-style: " + (font.getStyle() == Font.BOLD ? "bold"
                         : font.getStyle() == Font.ITALIC ? "italic" : "normal") + "; "
                 + " color: " + EditorColor.COLOR_FOREGROUND.toHex() + "; "
-                + " background: " + EditorColor.COLOR_BACKGROUND.toHex() + "; "
-                + " }");
+                + " background: " + EditorColor.COLOR_BACKGROUND.toHex() + ";} "
+                + ".word {font-size: " + (2 + font.getSize()) + "; font-style: bold; }"
```
amake, line 154: I don't see any effect from this change. In fact it looks like with this PR we are losing the effect of this stylesheet entirely. I guess it's because we are recreating the document every time. I think we need to re-set the stylesheet every time as well.

amake (commented):
Thanks, this approach looks nicer than the old batching strategy.

src/org/omegat/gui/dictionaries/DictionariesTextArea.java
```diff
@@ -187,12 +197,16 @@ protected void callDictionary(String word) {
...
+            // start position to be visible
+            Rectangle rect1 = Java8Compat.modelToView(this, el.getStartOffset());
+            // end position of article
+            Rectangle rect2 = Java8Compat.modelToView(this, el.getEndOffset());
+            // to show maximum text of selected, by moving to end of article then show start position,
+            if (rect2 != null) {
+                scrollRectToVisible(rect2);
+            }
+            if (rect1 != null) {
+                scrollRectToVisible(rect1);
```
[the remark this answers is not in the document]
miurahr: When article is larger than view window, the code works well, scroll to end of block, then move to top of block.
>Did you try scrolling to `rect1.union(rect2)`?

No, not tested.  I'm not sure  how the suggested code works.

src/org/omegat/gui/dictionaries/DictionariesTextArea.java
miurahr: Yes, we can drop it.

src/org/omegat/gui/dictionaries/DictionariesTextArea.java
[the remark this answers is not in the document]
amake: `rect1.union(rect2)` makes a single rectangle that includes both `rect1` and `rect2`, which would be the entire area covered by the element. Thus `scrollRectToVisible` would try to make the entire element visible if possible.

amake:
I think this is ready. Anything else?

miurahr:
LGTM. I also think it is ready.

PR #145, opened 2021-09-04
WIP: Dictionary: dsl: process tags properly

miurahr:

### DIFF
--- a/src/org/omegat/core/dictionaries/LingvoDSL.java
+++ b/src/org/omegat/core/dictionaries/LingvoDSL.java
@@ -34,14 +34,15 @@ import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Locale;
-import java.util.regex.Pattern;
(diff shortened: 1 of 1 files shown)


miurahr:
Now ready for review.

src/org/omegat/core/dictionaries/LingvoDSL.java
```diff
@@ -96,11 +107,19 @@ private void readDslFile(File file) throws IOException {
             }
         }
 
+        private String replaceTag(final String line) {
+            String result = line;
+            for (RE re : RE_LIST) {
+                result = result.replaceAll(re.regex, re.replacement);
```
miurahr, line 113: is it better to use `re.pattern.matcher(result).replaceAll(re.replacement)` with pre-compiled regexes?

miurahr:
Need to change test expectations.

src/org/omegat/core/dictionaries/LingvoDSL.java
miurahr: done

miurahr:
Now ready.

miurahr:
An error is not caused by the code.

amake:

PR #147, opened 2021-09-10
Support using arbitrary git branch name

brandelune:
readme update


amake:
Thanks, I've applied to `master`.

PR #150, opened 2021-09-13
Introduce URLProtocolHandler for "data:;base64,xxx"

miurahr:
This allow DictionaryPane shows images of ditionaries.

Resolve BUG#1069

omegat-chameleon

### DIFF
--- a/src/org/omegat/core/Core.java
+++ b/src/org/omegat/core/Core.java
@@ -37,6 +37,7 @@ import java.util.concurrent.locks.ReentrantLock;
 import org.omegat.core.data.EntryKey;
 import org.omegat.core.data.IProject;
 import org.omegat.core.data.NotLoadedProject;
+import org.omegat.core.protocol.URLProtocolHandler;
 import org.omegat.core.segmentation.Segmenter;
 import org.omegat.core.spellchecker.ISpellChecker;
 import org.omegat.core.spellchecker.SpellChecker;
@@ -258,6 +259,7 @@ public final class Core {
         filterMaster = new FilterMaster(Preferences.getFilters());
 
         // 4. Initialize other components. They add themselves to the main window.
+        URLProtocolHandler.install();
         editor = new EditorController(me);
         tagValidation = new TagValidationTool();
         issuesWindow = new IssuesPanelController(me);
--- a/src/org/omegat/gui/dictionaries/DictionariesTextArea.java
+++ b/src/org/omegat/gui/dictionaries/DictionariesTextArea.java
@@ -245,6 +245,7 @@ public class DictionariesTextArea extends EntryInfoThreadPane<List<DictionaryEnt
         StringBuilder txt = new StringBuilder();
         boolean wasPrev = false;
         int i = 0;
+        txt.append("<html>");
         for (DictionaryEntry de : data) {
             if (wasPrev) {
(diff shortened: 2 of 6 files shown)


miurahr:
EPWING dictionary plugin supports graphics contents of dictionary, and GAIJI image.

miurahr:
DSL dictionary also have a capability to hold image data.

miurahr:
Now protocol handler is an individual free library and set it as a dependency.

src/org/omegat/core/Core.java
```diff
@@ -258,6 +260,7 @@ public static void initializeGUI(ClassLoader classLoader, final Map<String, Stri
         filterMaster = new FilterMaster(Preferences.getFilters());
 
         // 4. Initialize other components. They add themselves to the main window.
+        URLProtocolHandler.install();
```
amake, line 263: It seems like it would be nice to initialize this package more like an OmegaT plugin: have a class that implements `loadPlugins` and `unloadPlugins`. That way the core doesn't need to know about it.

amake (commented):
Thanks for this. It looks like a nice improvement. And I like that you've generalized it as a library.

src/org/omegat/gui/dictionaries/DictionariesTextArea.java
```diff
@@ -245,6 +245,7 @@ protected void setFoundResult(final SourceTextEntry se, final List<DictionaryEnt
         StringBuilder txt = new StringBuilder();
         boolean wasPrev = false;
         int i = 0;
+        txt.append("<html>");
```
amake, line 248:
```suggestion
        StringBuilder txt = new StringBuilder("<html>");
        boolean wasPrev = false;
        int i = 0;
```

src/org/omegat/core/Core.java
miurahr: I'd like to try.

miurahr: Aah, get it. I can implement it as in-source plugin.

src/org/omegat/gui/dictionaries/DictionaryPaneExtension.java
```diff
@@ -0,0 +1,50 @@
...
+ along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ **************************************************************************/
+package org.omegat.gui.dictionaries;
+
+import tokyo.northside.protocol.URLProtocolHandler;
+
+/**
+ * @author Hiroshi Miura
+ */
+public class DictionaryPaneExtension {
```
amake, line 32: Sorry for the bikeshedding, but this naming implies that it could do all sorts of things to the Dictionary pane. I would rather have separate extensions for separate concerns.

Could we call it something more focused? Like `DictionaryMediaSupport` or something?

amake: Actually, doesn't this affect more than just the Dictionary pane? It should affect all `JTextPane`s that have HTML documents, right? So we should probably change the package as well. What about `org.omegat.util.gui.HtmlMediaSupport`?

miurahr: > org.omegat.util.gui.HtmlMediaSupport

I'd like to take it.

PR #151, opened 2021-09-15
DefaultGlossaryRenderer: set attribute to comma separator

miurahr:
resolve [RFE#1586](https://sourceforge.net/p/omegat/feature-requests/1586/)

### DIFF
--- a/src/org/omegat/gui/glossary/DefaultGlossaryRenderer.java
+++ b/src/org/omegat/gui/glossary/DefaultGlossaryRenderer.java
@@ -57,17 +57,16 @@ public class DefaultGlossaryRenderer implements IGlossaryRenderer {
             if (i > 0 && targets[i].equals(targets[i - 1])) {
                 if (!comments[i].equals("")) {
                     commentsBuf.append("\n");
-                    commentsBuf.append(String.valueOf(commentIndex));
+                    commentsBuf.append(commentIndex);
                     commentsBuf.append(". ");
                     commentsBuf.append(comments[i]);
                 }
                 continue;
             }
-            if (i > 0) {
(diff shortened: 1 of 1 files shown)


amake:

t-cordonnier:
Good that you could find the time to validate this, sorry that it was not my case.
Only one comment then. The color for comma was intentionally left to black, because in my spirit, this comma is a separator _between_ translations, and not a part of the translation itself. Let's not forget that a translation can also contain a comma, so leaving it black makes clearly distinction between comma as a separator and comma in translation. But I let you and your users decide what they prefer.

amake:
@t-cordonnier Thanks for clarifying. It looked like an oversight.

I can see your point that it would be useful to distinguish between an "in-translation comma" and the "translation-separating comma". However I think we will go with this change for the time being because:

- We had a bug report from a user complaining that his custom color was not applied to the separating comma. Clearly he expected that it *would* be applied.
- Introducing yet another customizable color to apply to just the comma seems like overkill
- In-translation commas are probably relatively rare
- It's probably still clear enough even when there *is* an in-translation comma

We can always revisit this issue if there is additional feedback.

t-cordonnier:
No problem Aaron. I will send the clarification to the user as well, and put it in the initial ticket.  
I fully agree that a dedicated color for comma (or for anything which is neither source nor translation) is an overkill. I had created the groovy-based method because I was pretty sure that there would always be such small details for which there will never be a full consensus. The advantage of using Groovy would be that in such a case, I simply send the "correction" to the user and every individual user can decide which one of the groovy scripts he prefers. This is more difficult with hardcoded style. Now wait and see...

PR #152, opened 2021-09-27
remove/adapt references to yahoo.com

brandelune:


release/changes.txt
```diff
@@ -8660,6 +8660,7 @@ Bug fix release:
 
 - Fixed Bad Segmentation of some files,
   as documented in: http://groups.yahoo.com/group/OmegaT/message/672
+  (https://sourceforge.net/p/omegat/mailman/message/36811754/)
```
miurahr, line 8663: IMHO, yahoo URL can be removed because old URL had already been unreachable.

brandelune: I preferred to leave it for historical purpose, but you're right.

test/src/org/omegat/tokenizer/TokenizerTest.java
```diff
@@ -188,6 +188,7 @@ public void testChinese() {
      * @see <a href=
      *      "https://groups.yahoo.com/neo/groups/OmegaT/conversations/messages/28395">
      *      User group discussion</a>
+     * @see <a href="https://sourceforge.net/p/omegat/mailman/message/36839317/">Sourceforge archive</a>
```
miurahr, line 191: We can remove yahoo URL here, too.

release/changes.txt
amake: Presumably the original link was to a threaded view. The equivalent view now would be https://sourceforge.net/p/omegat/mailman/omegat-users/thread/cflh3i%2Bkbe1%40.../#msg36811754

test/src/org/omegat/tokenizer/TokenizerTest.java
brandelune: Sure, but this one I was not 100% sure that it was pointing at the right mail.

release/changes.txt
miurahr: >  for historical purpose,

IMHO, historical URL can be seen in git history when necessary.

test/src/org/omegat/tokenizer/TokenizerTest.java
brandelune: You're right. I just did not feel comfortable removing information. If you think it is not relevant, I am totally fine with your decision.

amake:

PR #153, opened 2021-09-29
LingvoDSL: Support image/video/sound media

miurahr:
Add a feature discussed in BUGS#1065

- Depends: #144 #145 
- Resolve [BUGS#1065](https://sourceforge.net/p/omegat/bugs/1065/)

### DIFF
--- a/src/org/omegat/core/dictionaries/LingvoDSL.java
+++ b/src/org/omegat/core/dictionaries/LingvoDSL.java
@@ -189,9 +189,9 @@ public class LingvoDSL implements IDictionaryFactory {
         TAG_REPLACEMENTS.put(Pattern.compile(
                 "\\[s](?<media>.+?\\.wav)\\[/s]"), "<a href=\"file://" + DICT_DIR_TAG + "/${media}\">${media}</a>");
         TAG_REPLACEMENTS.put(Pattern.compile(
-                "\\[s](?<media>.+?\\.jpg)\\[/s]"), "<img src=\"file://"+ DICT_DIR_TAG + "/${media}\"/>");
+                "\\[s](?<media>.+?\\.jpg)\\[/s]"), "<img src=\"file://" + DICT_DIR_TAG + "/${media}\"/>");
         TAG_REPLACEMENTS.put(Pattern.compile(
-                "\\[s](?<media>.+?\\.png)\\[/s]"), "<img src=\"file://"+ DICT_DIR_TAG + "/${media}\"/>");
+                "\\[s](?<media>.+?\\.png)\\[/s]"), "<img src=\"file://" + DICT_DIR_TAG + "/${media}\"/>");
         TAG_REPLACEMENTS.put(Pattern.compile(
                 "\\[video](?<media>.+?)\\[/video]"), "<a href=\"file://" + DICT_DIR_TAG + "/${media}\">${media}</a>");
         TAG_REPLACEMENTS.put(Pattern.compile(
--- a/src/org/omegat/core/dictionaries/LingvoDSL.java
+++ b/src/org/omegat/core/dictionaries/LingvoDSL.java
@@ -68,7 +68,7 @@ public class LingvoDSL implements IDictionaryFactory {
     // An ordered list of Pair of Regex pattern and replacement string
     private static final TreeMap<Pattern, String> TAG_REPLACEMENTS = new TreeMap<>(
             Comparator.comparing(Pattern::pattern));
-
+    private static final String DICT_DIR_TAG = "@dir@";
     private static final int BLOCKSIZE = 8192;
 
     @Override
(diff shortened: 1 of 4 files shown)


src/org/omegat/core/dictionaries/LingvoDSL.java
```diff
@@ -130,4 +134,91 @@ private void loadData(Stream<String> stream) {
                     .collect(Collectors.toList());
         }
     }
+
+    @SuppressWarnings("visibilitymodifier")
+    static class RE {
```
amake, line 139: I'm not sure why we need this class. Can't we just use a `HashMap<Pattern, String>` to store the patterns as keys and replacements as values?

If order is important it can be a `TreeMap`.

src/org/omegat/core/dictionaries/LingvoDSL.java
```diff
@@ -130,4 +134,91 @@ private void loadData(Stream<String> stream) {
...
+            // The following line tries to replace [c value]text[/c] with text colored as per the value.
+            // Since the color names are plain words like 'red', or 'blue', or 'steelgray' etc.,
+            // FIXME: I use the ([a-z]+?) regular expression, but am not sure if it is correct.
+            reList.add(new RE("\\[c\\s([a-z]+?)\\](.+?)\\[/c\\]", "<span style='color:$1'>$2</span>"));
+            reList.add(new RE("\\[com\\]", ""));
+            reList.add(new RE("\\[/com\\]", ""));
+            reList.add(new RE("\\[ex\\]", ""));
+            reList.add(new RE("\\[/ex\\]", ""));
+            reList.add(new RE("\\[lang\\]", ""));
+            reList.add(new RE("\\[/lang\\]", ""));
```
amake, line 181: You can combine these into a single regex.

src/org/omegat/core/dictionaries/LingvoDSL.java
```diff
@@ -99,19 +102,20 @@ private void readDslFile(File file) throws IOException {
...
-                } else {
-                    if (word.length() > 0) {
-                        data.add(word.toString(), trans.toString());
-                        word.setLength(0);
-                        trans.setLength(0);
-                    }
-                    word.append(line);
-                }
-            });
+            stream.filter(line -> !line.isEmpty()).filter(line -> !line.startsWith("#"))
```
amake, line 105:
```suggestion
            stream.filter(line -> !line.isEmpty())
                .filter(line -> !line.startsWith("#"))
```

src/org/omegat/core/dictionaries/LingvoDSL.java
```diff
@@ -130,4 +134,91 @@ private void loadData(Stream<String> stream) {
...
+            reList.add(new RE("\\[m7\\](.+?)\\[/m\\]", "<div style=\"text-indent: 90px\">$1</div>"));
+            reList.add(new RE("\\[m8\\](.+?)\\[/m\\]", "<div style=\"text-indent: 90px\">$1</div>"));
+            reList.add(new RE("\\[m9\\](.+?)\\[/m\\]", "<div style=\"text-indent: 90px\">$1</div>"));
+            reList.add(new RE("\\[p\\]", ""));
+            reList.add(new RE("\\[/p\\]", ""));
+            reList.add(new RE("\\[preview\\]", ""));
+            reList.add(new RE("\\[/preview\\]", ""));
+            reList.add(new RE("\\[ref\\]", ""));
+            reList.add(new RE("\\[/ref\\]", ""));
+            reList.add(new RE("\\[s\\](.+?\\.wav)\\[/s\\]", "<a href=\"file://@dir@/$1\">SOUND: $1</a>"));
```
amake, line 198: It would be nice to have a comment explaining how `@dir@` is resolved.

src/org/omegat/core/dictionaries/LingvoDSL.java
```diff
@@ -130,4 +135,91 @@ private void loadData(Stream<String> stream) {
...
+        public Pattern pattern;
+        public String replacement;
+
+        public RE(final String regex, final String replacement) {
+            pattern = Pattern.compile(regex);
+            this.replacement = replacement;
+        }
+    }
+
+    static class LingvoDSLTag {
```
amake, line 150: I don't think we really need this class either.

src/org/omegat/core/dictionaries/LingvoDSL.java
amake: Actually for consistency wouldn't it make sense to do these like the other ones?

```java
reList.add(new RE("\\[lang\\](.+?)\\[/lang\\]", "$1"));
```

amake: - In fact you could combine all of the "simple replacement" ones like below
- I highly recommend using [named capturing groups](https://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html#groupname)

```java
reList.add(new RE("\\[(?<tag>com|ex|lang|...)\\](?<content>.+?)\\[/\ktag\\]", "${content}"));
```

miurahr:
> I made some small tweaks. Please let me know if it still works for you.

Thank you, it works for me.

amake:

PR #155, opened 2021-10-10
LingvoDSL: update copyright holder name in source header

miurahr:
The author contributed a lot for researching DSL tags specification,
made a proposal of all the regex replacements, and tested with
actual dictionaries. The improvements are not able to be realized
without the efforts.


amake:

PR #158, opened 2021-10-12
DictionaryPane: open hyper link in external browser

miurahr:


amake:

PR #159, opened 2021-10-17
WIP: Team: Git: Support id_ed25519 and id_ecdsa key algorithms

miurahr:
- Support eclipse curve ciphers, ed25519 and ecdsa
- Migrate to apache MINA sshd to support it instead of dead/unmaintained jsch
- 

### DIFF
--- a/src/org/omegat/core/team2/impl/GITCredentialsProvider.java
+++ b/src/org/omegat/core/team2/impl/GITCredentialsProvider.java
@@ -27,6 +27,7 @@
 
 package org.omegat.core.team2.impl;
 
+import java.io.File;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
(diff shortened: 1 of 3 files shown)


miurahr:
> Why was it working without this before? What changed?

I just tried with OmegaT 5.5.0 and I observed the same error on Windows. -- it is also not working before.
The improvement here may be to fix an another issue than BUG#1075 reported.

The condition and observation when I try to reproduce the issue:

- I use id_ecdsa key for github
- without `.ssh/config`, "connection error" is produced.
- A `.ssh/config`  entry resolve the issue, the file is like 
```
Host github.com
  IdentityFile ~/.ssh/id_github
```


The PR here add a support for key algorithms  "id_ecdsa" and "id_ed25519", and explicitly add "id_rsa" and "id_dsa" default names, when ssh agent is not found.

ak-mdufour:
> Why was it working without this before? What changed?

The recent OpenSSH 8.8 release has disabled support for RSA keys by default. See: https://www.openssh.com/txt/release-8.8

The latest git-for-windows includes this version, and this affects more than just OmegaT!

amake:
Thanks, but I don't think OmegaT itself uses OpenSSH in any capacity.

miurahr:
I'd like to close here and  ask as another PR for review.

miurahr:
I opened #161

amake:
Thanks, I think the approach in #161 is much better.

miurahr:
I opened #162 for the next step and working solution.

It seems to be necessary to watch a progress for ssh agent support on apache mina to complete  #161

PR #162, opened 2021-11-04
Team: ssh+git: Support elliptic curve cryptography  ed25519 and ecdsa keys (RFE#1598)

miurahr:
- Add id_ecdsa and id25519 as default user key
- Drop weak id_dsa from default key, that still usable when user set it in .ssh/config
- Default keys are listed as a constant of OConsts
- Ssh library JSch has been unmaintained and dead, so replace with [forked version of Jsch](https://github.com/mwiede/jsch)
- [mwiede/jsch](https://github.com/mwiede/jsch) supports ssh-agent natively so drop ssh-agent plugins
- Work around for [BUGS#1075](https://sourceforge.net/p/omegat/bugs/1075/): when ssh server reject environment varialbe GIT_PROTOCOL=version=2,
  it throw exception but it should continues to process with protocol version = 0.
  This behavior is a bug of JSch library, and avoid it without setting timeout.
- Implement [RFE#1598](https://sourceforge.net/p/omegat/feature-requests/1598/) 
    - For future action: migrate to apache mina sshd
        - JGit will drop support for JSch in JGit 6.0 and recommend Apache MINA SSHD.

### DIFF
--- a/src/org/omegat/core/team2/impl/GITCredentialsProvider.java
+++ b/src/org/omegat/core/team2/impl/GITCredentialsProvider.java
@@ -28,8 +28,6 @@
 package org.omegat.core.team2.impl;
 
 import java.io.File;
-import java.io.FileInputStream;
-import java.io.IOException;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
@@ -79,7 +77,7 @@ import org.omegat.util.OStrings;
 public class GITCredentialsProvider extends CredentialsProvider {
 
     static {
-        JschConfigSessionFactory sessionFactory = new JschConfigSessionFactory() {
+        JschConfigSessionFactory sessionFactory = new OJschConfigSessionFactory() {
 
             @Override
             protected void configure(OpenSshConfig.Host host, Session session) {
@@ -106,49 +104,27 @@ public class GITCredentialsProvider extends CredentialsProvider {
             /**
              * Default JSch instance creator.
              *
-             * default key names; identity, id_rsa, id_ecdsa, and id_ed25519.
-             * It solves BUG#1075, and realizes RFE#1598
+             * Add id_ecdsa and id_ed25519 as default key.
+             * Resolve RFE#1598
              */
             @Override
-            protected JSch createDefaultJSch(FS fs) throws JSchException {
-                // Note: it should not call super.createDefaultJSch().
-                // we create instance from scratch to avoid
-                // the forked JSch and JGit combination issue that
-                // lead NPE.
-                JSch jsch = new JSch();
-                configureJSch(jsch);
-                knownHosts(jsch, fs);
-                identities(jsch, fs);
+            protected JSch createDefaultJSch(final FS fs) throws JSchException {
+                JSch jsch = super.createDefaultJSch(fs);
+                overrideIdentities(jsch, fs);
                 return jsch;
             }
 
-            private void knownHosts(JSch jsch, FS fs) throws JSchException {
-                final File home = fs.userHome();
-                if (home == null)
-                    return;
-                final File known_hosts = new File(new File(home, ".ssh"), "known_hosts");
-                try (FileInputStream in = new FileInputStream(known_hosts)) {
-                    jsch.setKnownHosts(in);
-                } catch (IOException ignored) {
-                }
-            }
-
-            private void identities(final JSch jsch, final FS fs) {
+            private void overrideIdentities(final JSch jsch, final FS fs) {
                 final File home = fs.userHome();
                 if (home == null) {
                     return;
                 }
                 final File sshdir = new File(home, ".ssh");
-                if (sshdir.isDirectory()) {
-                    //  Add more default key names; id_ecdsa and id_ed25519
-                    loadIdentity(jsch, new File(sshdir, "identity"));
-                    loadIdentity(jsch, new File(sshdir, "id_rsa"));
-                    loadIdentity(jsch, new File(sshdir, "id_ecdsa"));
-                    loadIdentity(jsch, new File(sshdir, "id_ed25519"));
-                }
+                loadIdentity(jsch, new File(sshdir, "id_ecdsa"));
+                loadIdentity(jsch, new File(sshdir, "id_ed25519"));
             }
 
-            private void loadIdentity(JSch sch, File priv) {
+            private void loadIdentity(final JSch sch, final File priv) {
                 if (priv.isFile()) {
                     try {
                         sch.addIdentity(priv.getAbsolutePath());
--- a/src/org/omegat/core/team2/impl/GITCredentialsProvider.java
+++ b/src/org/omegat/core/team2/impl/GITCredentialsProvider.java
@@ -27,6 +27,9 @@
 
 package org.omegat.core.team2.impl;
 
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.IOException;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
@@ -35,6 +38,13 @@ import java.util.regex.Pattern;
 
 import javax.swing.JOptionPane;
 
+import com.jcraft.jsch.AgentIdentityRepository;
+import com.jcraft.jsch.AgentProxyException;
+import com.jcraft.jsch.IdentityRepository;
+import com.jcraft.jsch.JSch;
+import com.jcraft.jsch.JSchException;
+import com.jcraft.jsch.SSHAgentConnector;
+import com.jcraft.jsch.Session;
 import org.eclipse.jgit.errors.UnsupportedCredentialItem;
 import org.eclipse.jgit.transport.CredentialItem;
 import org.eclipse.jgit.transport.CredentialsProvider;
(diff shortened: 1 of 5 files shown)


src/org/omegat/core/team2/impl/GITRemoteRepository2.java
```diff
@@ -126,7 +126,10 @@ public void init(RepositoryDefinition repo, File dir, ProjectTeamSettings teamSe
             CloneCommand c = Git.cloneRepository();
             c.setURI(repositoryURL);
             c.setDirectory(localDirectory);
-            c.setTimeout(TIMEOUT);
+            // BUGS#1075 - When set timeout, JSch expect to check result of setEnv('GIT_PROTOCOL=version=2')
+            // and throw exception when failed, but it should ignored.
+            // It is JSch bug but we avoid it by not setting timeout here.
+            // c.setTimeout(TIMEOUT);
```
miurahr, line 132: A fix is requested on mwiede/jsch project.
https://github.com/mwiede/jsch/pull/92

src/org/omegat/core/team2/impl/GITCredentialsProvider.java
```diff
@@ -90,23 +90,32 @@ protected void configure(OpenSshConfig.Host host, Session session) {
...
-                try {
-                    if (SSHAgentConnector.isConnectorAvailable()) {
-                        USocketFactory usf = new JNAUSocketFactory();
-                        con = new SSHAgentConnector(usf);
-                    } else {
-                        ConnectorFactory cf = ConnectorFactory.getDefault();
-                        con = cf.createConnector();
+           		final JSch jsch = new JSch();
+                final File home = fs.userHome();
+                if (home != null) {
```
amake, line 95: Thanks for looking into this.

I understand that this is some setup boilerplate that is probably mostly equivalent to what we were doing before, but

- Is this documented anywhere?
- Do we really have to do all this? The library won't look for known hosts and supported private keys on its own?

src/org/omegat/core/team2/impl/GITRemoteRepository2.java
amake: Thanks for taking this upstream. I see they've merged the fix.

src/org/omegat/core/team2/impl/GITCredentialsProvider.java
amake: Sorry I should also say that this level of manual configuration still seems wrong to me.

I have my keys named after the service I created them for, like `~/.ssh/gitlab{,.pub}`. Enumerating supported key names just doesn't make sense to me. Shouldn't the library be transparently consulting `~/.ssh/config`?

src/org/omegat/core/team2/impl/GITRemoteRepository2.java
miurahr: The fix is merged.
https://github.com/mwiede/jsch/pull/90

src/org/omegat/core/team2/impl/GITCredentialsProvider.java
miurahr: JGit/JSch looks `~/.ssh/config` and when configuration is there, it respect it.
I also have a habit to generate keys with name and configure `~/.ssh/config` with it.
The code here has no effect when user put key configuration.
It is just for default.

> Is this documented anywhere?

No. I understand things from JGit source code.

> Do we really have to do all this?

I want to override the behavior doing less than original, so I need to do it.
JSch don't have `removeIdentity` method, so I don't want to call super as previously did. 
And utility functions, such as knownHosts` are private so I need to write it.

The chunk of code doing similar things as [JGit method with same name](https://git.eclipse.org/r/plugins/gitiles/jgit/jgit/+/refs/tags/v5.11.1.202105131744-r/org.eclipse.jgit.ssh.jsch/src/org/eclipse/jgit/transport/JschConfigSessionFactory.java#399) because it override it.

```
	protected JSch createDefaultJSch(FS fs) throws JSchException {
		final JSch jsch = new JSch();
		JSch.setConfig("ssh-rsa", JSch.getConfig("signature.rsa")); //$NON-NLS-1$ //$NON-NLS-2$
		JSch.setConfig("ssh-dss", JSch.getConfig("signature.dss")); //$NON-NLS-1$ //$NON-NLS-2$
		configureJSch(jsch);
		knownHosts(jsch, fs);
		identities(jsch, fs);
		return jsch;
	}
	private static void knownHosts(JSch sch, FS fs) throws JSchException {
		final File home = fs.userHome();
		if (home == null)
			return;
		final File known_hosts = new File(new File(home, ".ssh"), "known_hosts"); //$NON-NLS-1$ //$NON-NLS-2$
		try (FileInputStream in = new FileInputStream(known_hosts)) {
			sch.setKnownHosts(in);
		} catch (FileNotFoundException none) {
			// Oh well. They don't have a known hosts in home.
		} catch (IOException err) {
			// Oh well. They don't have a known hosts in home.
		}
	}
	private static void identities(JSch sch, FS fs) {
		final File home = fs.userHome();
		if (home == null)
			return;
		final File sshdir = new File(home, ".ssh"); //$NON-NLS-1$
		if (sshdir.isDirectory()) {
			loadIdentity(sch, new File(sshdir, "identity")); //$NON-NLS-1$
			loadIdentity(sch, new File(sshdir, "id_rsa")); //$NON-NLS-1$
			loadIdentity(sch, new File(sshdir, "id_dsa")); //$NON-NLS-1$
		}
	}
	private static void loadIdentity(JSch sch, File priv) {
		if (priv.isFile()) {
			try {
				sch.addIdentity(priv.getAbsolutePath());
			} catch (JSchException e) {
				// Instead, pretend the key doesn't exist.
			}
		}
	}
```

miurahr: Sorry JSch have `remoteIdentity` method. I'd like to rewrite here.

miurahr: After rewriting `createDefaultJSch` method, OmegaT failed to clone. Now investigating...

src/org/omegat/core/team2/impl/GITRemoteRepository2.java
miurahr: The fix is released and worked.

src/org/omegat/core/team2/impl/GITCredentialsProvider.java
miurahr: [JGit original `createDefaultJSch` method](https://git.eclipse.org/r/plugins/gitiles/jgit/jgit/+/refs/tags/v5.11.1.202105131744-r/org.eclipse.jgit.ssh.jsch/src/org/eclipse/jgit/transport/JschConfigSessionFactory.java#399) is like

```java
 	protected JSch createDefaultJSch(FS fs) throws JSchException {
	       final JSch jsch = new JSch();
               JSch.setConfig("ssh-rsa", JSch.getConfig("signature.rsa")); //$NON-NLS-1$ //$NON-NLS-2$
               JSch.setConfig("ssh-dss", JSch.getConfig("signature.dss")); //$NON-NLS-1$ //$NON-NLS-2$
```

Forked version of JSch already have "ssh-rsa" config, but it does not have "signature.rsa", "signature.dsa" so JSch.getConfig("signature.rsa") become null .
It override "ssh-rsa" configuration by null, then JSch class throws NPE at finally and failed to clone repository.

These two lines are merged at ["Ensure that JSch knows HostKeyAlgorithms ssh-rsa and ssh-dss"](https://git.eclipse.org/r/plugins/gitiles/jgit/jgit/+/09753591fae875cee71eb5674a0103844351e508)
which seems intend to work around original JSch problem.

The easiest way to avoid it is to implement whole functions by ourselves, which I did before.

miurahr: The issue has already reported to forked jsch project.

- https://github.com/mwiede/jsch/issues/85

It is also filed on JGit project.

- [Bug 576604 - JGit should not overwrite JSch properties, if they are already set.](https://bugs.eclipse.org/bugs/show_bug.cgi?id=576604)
- [Bug 576605 - Replace original JSch with an updated Fork](https://bugs.eclipse.org/bugs/show_bug.cgi?id=576605)

src/org/omegat/core/team2/impl/GITCredentialsProvider.java
```diff
@@ -80,36 +79,83 @@
...
+            @Override
+            protected JSch createDefaultJSch(FS fs) throws JSchException {
+                // Note: it should not call super.createDefaultJSch().
+                // we create instance from scratch to avoid
+                // the forked JSch and JGit combination issue that
+                // lead NPE.
+                JSch jsch = new JSch();
+                configureJSch(jsch);
+                knownHosts(jsch, fs);
+                identities(jsch, fs);
```
miurahr, line 121: There is a copy from parent class to avoid the JGit bug.
It may be better to have an intermediate class to proxy `JschConfigSessionFactory` and change behavior by back porting of the fix. 
I'd like to leave only a required code when next JGit version, 5.13.x,  is released with the fix.

When the fix released, we can remove intermediate patching class.

miurahr: done

miurahr:
@amake do you want to wait a backport of fix in upstream jgit project?

PR #163, opened 2021-11-04
Team: git: improve commit signing (BUGS#1077)

miurahr:
## Fix throwing exception

JGit 5.3 and later support gpg signing but OmegaT 5.6 does not bundle singing service library.
This add a dependency for bouncycastle gpg signing to support commit signing.

## how to configure team project

Users can set global git configs to sign their commits.

git config --global user.signingkey  0123456789ABCDEF
git config --global user.program gpg
git config --global gpg.program /usr/bin/gpg

and then user can config omegat translation team repository

git config commit.gpgsign true

With this, when omegat save translations, omegat will ask gpg passphrase.

## Allow external gpg command

It provides a preference checkbox to allow user to ignore above configuration.
It also provides a preference checkbox to allow user to select external gpg command.

## Reference

- Resolve [BUGS#1077](https://sourceforge.net/p/omegat/bugs/1077/)
- Use code from EGit's external gpg command support (Apache-2)

### DIFF
--- a/src/org/omegat/Bundle.properties
+++ b/src/org/omegat/Bundle.properties
@@ -3012,8 +3012,6 @@ THEME_OMEGAT_DEFAULT_DESCRIPTION=OmegaT default theme
 THEME_OMEGAT_CLASSIC_NAME=OmegaT Classic
 THEME_OMEGAT_CLASSIC_DESCRIPTION=OmegaT classic theme
 
-PREFS_TEAM_IGNORE_GIT_SIGN=Ignore signing for git commit even when commit.gpgsign=true
-PREFS_TEAM_USE_EXTERNAL_GPG=Use external gpg/gpg.exe command for signing
 GPG_EXTERNAL_SIGNER_BUFFER_ERROR=Error while getting output of external program
 GPG_EXTERNAL_SIGNER_CANNOT_SEARCH=Cannot find GPG executable on the bash $PATH
 GPG_EXTERNAL_SIGNER_ENVIRONMENT_ERROR=Cannot prepare GPG environment properly
--- a/src/org/omegat/Bundle.properties
+++ b/src/org/omegat/Bundle.properties
@@ -3012,6 +3012,8 @@ THEME_OMEGAT_DEFAULT_DESCRIPTION=OmegaT default theme
 THEME_OMEGAT_CLASSIC_NAME=OmegaT Classic
 THEME_OMEGAT_CLASSIC_DESCRIPTION=OmegaT classic theme
 
+PREFS_TEAM_IGNORE_GIT_SIGN=Ignore signing for git commit even when commit.gpgsign=true
+PREFS_TEAM_USE_EXTERNAL_GPG=Use external gpg/gpg.exe command for signing
 GPG_EXTERNAL_SIGNER_BUFFER_ERROR=Error while getting output of external program
 GPG_EXTERNAL_SIGNER_CANNOT_SEARCH=Cannot find GPG executable on the bash $PATH
 GPG_EXTERNAL_SIGNER_ENVIRONMENT_ERROR=Cannot prepare GPG environment properly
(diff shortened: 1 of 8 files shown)


MarioLiebisch:
Might be worth mentioning/testing: How does this handle a key with a passphrase with auto-saving enabled? This might get tedious, **if** you get a prompt every 2-3 minutes.

miurahr:
Passphrase caching has been discussed but has not resolved yet
https://bugs.eclipse.org/bugs/show_bug.cgi?id=543437

miurahr:
For reference;

EGit supports external GPG command instead of BouncyCastle library by configuration options.
This gives users to improve security and usage of "gpg-agent" for free.
ref: https://git.eclipse.org/c/egit/egit.git/commit/?h=stable-5.11&id=2b66e9b8baa2104c320c301daea9147227ef39fe

miurahr:
> Might be worth mentioning/testing: How does this handle a key with a passphrase with auto-saving enabled? This might get tedious, if you get a prompt every 2-3 minutes.

It does not seem ask passphrase recurring. Do you experienced it?

miurahr:
Now adding support for external gpg command. You can use external gpg.exe when it can be found in PATH.
@MarioLiebisch could you try?

MarioLiebisch:
> It does not seem ask passphrase recurring. Do you experienced it?
If it's using native gpg, it will have its own cache. Depending on how long between commits, there shouldn't be an extra prompt. Haven't had a chance to try this yet.

> @MarioLiebisch could you try?
Will have to see if/when I can find the time to set up a dev environment.

amake:
To be honest this is an extreme power-user thing. I don't really want to add any config UI to OmegaT if it's configurable purely in git.

miurahr:
>To be honest this is an extreme power-user thing. I don't really want to add any config UI to OmegaT if it's configurable purely in git.

Thank you for feedback. I'd like to revert  d1423886bd91d3a2fe8a102e21f46aa625b6fa9d and now OmegaT respect 
git config

MarioLiebisch:
Just wanted to mention that I agree, this isn't really necessary, since it's all already handled in Git.

miurahr:
@MarioLiebisch could you tell me how to detect user to configure git to use external gpg command?
Current PR always set external gpg as default signer.

1. check `program.gpg` is not empty
2. check `user.program` is gpg or not

which one?

miurahr:
take
> 2. check user.program is gpg or not

miurahr:
Now ready for review and merge. @amake

src/org/omegat/Bundle.properties
```diff
@@ -3011,3 +3011,14 @@ THEME_OMEGAT_DEFAULT_NAME=OmegaT Default
 THEME_OMEGAT_DEFAULT_DESCRIPTION=OmegaT default theme
 THEME_OMEGAT_CLASSIC_NAME=OmegaT Classic
 THEME_OMEGAT_CLASSIC_DESCRIPTION=OmegaT classic theme
+
+GPG_EXTERNAL_SIGNER_BUFFER_ERROR=Error while getting output of external program
+GPG_EXTERNAL_SIGNER_CANNOT_SEARCH=Cannot find GPG executable on the bash $PATH
```
amake, line 3016: PATH is not a bash-specific concept, so this message seems odd. Did it also come from EGit?

```suggestion
GPG_EXTERNAL_SIGNER_CANNOT_SEARCH=Cannot find GPG executable
```

miurahr: I cannot remember but I might pick it from EGit.

src/org/omegat/core/team2/impl/GITExternalGpgSigner.java
```diff
@@ -0,0 +1,471 @@
...
+            code = result.getRc();
+            if (code != 0) {
+                if (stderr != null) {
+                    stderr.accept(result.getStderr());
+                }
+                throw new IOException(
+                        MessageFormat.format(
+                                OStrings.getString(ExternalGpgSigner_processFailed),
+                                command, code + ": "
+                                        + toString(result.getStderr())));
```
amake, line 104: There is an override of `OStrings.getString` to which you can pass additional arguments that will be put into the string via `MessageFormat.format`.

If you prefer to keep this style (using `MessageFormat.format` on the "outside") because it's closer to the original EGit code, then that's OK.

amake:
Actually, forget my comment. Also I understand now why the message mentioned the "bash PATH".

I've squashed to `master`. Thanks as always for your hard work on these issues.

PR #164, opened 2021-11-07
Work around for menu focus issue (BUGS#1073)

miurahr:
This intend to fix an issue that the focus is sticky on edit pane when calling menu by Alt/Key combination, and
cannot select menu item with keyboard shortcut or cursor keys.
It is caused when default LaFs are selected.

The bug is exist on default LaFs are inherited from LookAndFeel abstract class, not from BasicLookAndFeel abstract
class. All known system LaFs, such as gtk, metal, nimbus, and aqua, are descendants of BasicLookAndFeel
that is direct known subclass of LookAndFeel class.

These classes implement serializable and incompatible with known MultiLookAndFeel subclass of LookAndFeel.
We brings limitation for users to allow system LaFs only a descendant of BasicLookAndFeel.

Now users have a way to use arbitrary theme/LaF as plugin. So it is not a problem.

- resolve [BUGS#1073](https://sourceforge.net/p/omegat/bugs/1073/)

### DIFF
--- a/src/org/omegat/gui/theme/DefaultClassicTheme.java
+++ b/src/org/omegat/gui/theme/DefaultClassicTheme.java
@@ -44,6 +44,7 @@ import org.omegat.util.OStrings;
 import org.omegat.util.gui.UIDesignManager;
 
 
+@SuppressWarnings("serial")
 public class DefaultClassicTheme extends DelegatingLookAndFeel {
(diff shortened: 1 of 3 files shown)


amake:
Thank you for figuring this out. I've applied to `master`.

PR #168, opened 2021-11-24
DSL: remove lang and orphan trn tag

miurahr:
resolve a bug reported in omegat-users ML
https://sourceforge.net/p/omegat/mailman/message/37388173/


src/org/omegat/core/dictionaries/LingvoDSL.java
```diff
@@ -170,6 +170,8 @@ private static String replaceTag(final String line) {
         TAG_REPLACEMENTS.put(Pattern.compile("\\[\\[(?<content>.+?)]]"), "&#91;${content}&#93;");
         TAG_REPLACEMENTS.put(Pattern.compile(Pattern.quote("\\[")), "&#91;");
         TAG_REPLACEMENTS.put(Pattern.compile(Pattern.quote("\\]")), "&#93;");
+        // lang tag
+        TAG_REPLACEMENTS.put(Pattern.compile("\\[lang id=(.+?)]"), "");
```
amake, line 174: Wouldn't there be a closing `[/lang]` tag as well?

Also [these docs](http://lingvo.helpmax.net/en/troubleshooting/dsl-compiler/dsl-tags/) say that the attribute is `name` not `id`. We can just catch anything by doing `\\[lang[^]]*]`.

amake: Also we handle `lang` in `ignoreTags` below. If we're going to handle it here instead, we should remove it from `ignoreTags` for clarity.

src/org/omegat/core/dictionaries/LingvoDSL.java
```diff
@@ -220,5 +222,7 @@ private static String replaceTag(final String line) {
         for (String tag : ignoreTags) {
             TAG_REPLACEMENTS.put(Pattern.compile("\\[(?<tag>" + tag + ")](?<content>.+?)\\[/\\k<tag>]"), "${content}");
         }
+        // remove orphan trn tag
+        TAG_REPLACEMENTS.put(Pattern.compile("\\[trn]"), "");
```
amake, line 226: Have we seen orphan `trn` in the wild? Do we know why only `trn` needs this treatment?

miurahr: Susan report it in omegat-users ML. I see it in capture.

src/org/omegat/core/dictionaries/LingvoDSL.java
miurahr: Does this handle closing tag as well?
`TAG_REPLACEMENTS.put(Pattern.compile("\\[/?lang[^]]*]"), "");`

miurahr: Also adding test and passed it.

src/org/omegat/core/dictionaries/LingvoDSL.java
miurahr: Any tag of `ignoreTags` may be able to be orphan.

miurahr:
#169 will supersede this.

PR #169, opened 2021-11-30
LingvoDSL: Utilize DSL4j dictionary loader/parser

miurahr:
DSL4j is real parser for LingvoDSL

Todos:

- [x] Handles `[lang id=1]` case

### DIFF
--- a/src/org/omegat/core/dictionaries/LingvoDSL.java
+++ b/src/org/omegat/core/dictionaries/LingvoDSL.java
@@ -35,6 +35,7 @@ import java.util.stream.Collectors;
 import io.github.eb4j.dsl.DslArticle;
 import io.github.eb4j.dsl.DslDictionary;
 import io.github.eb4j.dsl.visitor.DslVisitor;
+import org.apache.commons.io.FilenameUtils;
 
 /**
  * Dictionary implementation for Lingvo DSL format.
@@ -90,7 +91,7 @@ public class LingvoDSL implements IDictionaryFactory {
      */
     public static class HtmlVisitor extends DslVisitor<String> {
 
-        private static final String[] IMAGE_EXTS = new String[] {".png", ".jpg", ".PNG", ".JPG"};
+        private static final String[] IMAGE_EXTS = new String[] {"png", "jpg", "PNG", "JPG"};
 
         private StringBuilder sb;
         private boolean delayText;
@@ -178,8 +179,9 @@ public class LingvoDSL implements IDictionaryFactory {
         }
 
         private boolean isMediaImage() {
-            for (String ext: IMAGE_EXTS) {
-                if (previousText.endsWith(ext)) {
+            String ext = FilenameUtils.getExtension(previousText);
+            for (String e : IMAGE_EXTS) {
+                if (e.equals(ext)) {
                     return true;
                 }
             }
--- a/src/org/omegat/core/dictionaries/LingvoDSL.java
+++ b/src/org/omegat/core/dictionaries/LingvoDSL.java
@@ -29,11 +29,13 @@ package org.omegat.core.dictionaries;
 
 import java.io.File;
 import java.io.IOException;
+import java.util.ArrayList;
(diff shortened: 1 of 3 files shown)


test/src/org/omegat/core/dictionaries/LingvoDSLTest.java
```diff
@@ -86,28 +80,29 @@ public void testReadArticleRussian() throws Exception {
...
     @Test
     public void testReadArticleFontStyles() throws Exception {
-        LingvoDSLDict dict = (LingvoDSLDict) new LingvoDSL().loadDict(TEST_DICT, ENGLISH);
+        LingvoDSLDict dict = (LingvoDSLDict) new LingvoDSL().loadDict(TEST_DICT);
         String word = "italic";
         List<DictionaryEntry> result = dict.readArticles(word);
         assertFalse(result.isEmpty());
         assertEquals(word, result.get(0).getWord());
-        assertEquals("Here is an <span style='font-style: italic'>italic</span> <strong>word</strong>.\n",
+        assertEquals("<p>Here is an <span style='font-style: italic'>italic</span> <strong>word</strong>.</p>\n",
```
miurahr, line 99: Here is difference. When parser visit `[/m]`, we put `</p>` for `[m1]` - `[m9]` so we put `<p>` also for `[m]`

src/org/omegat/core/dictionaries/LingvoDSL.java
```diff
@@ -64,161 +52,226 @@
...
+
+        LingvoDSLDict(File file) throws Exception {
+            data = DslDictionary.loadDictionary(file);
+            htmlVisitor = new HtmlVisitor(file.getParent());
+        }
+
+        @Override
+        public List<DictionaryEntry> readArticles(String word) {
+            return data.lookup(word).getEntries(htmlVisitor).stream()
+                    .map(e -> new DictionaryEntry(e.getKey(), e.getValue()))
```
miurahr, line 77: Here is a place we can see how to use the parser.

- `DslDictionary#loadDictionary` is a builder method to load `dsl` file.
- `DslDictionary#lookup` returns `DslResult` object that hold result articles' AST.
- `DslResult#getEntries` returns `List<Entry<String, T>>` when pass `visitor<T>` object.

miurahr:
Now ready for review.

src/org/omegat/core/dictionaries/LingvoDSL.java
amake: Thanks.

Two minor thoughts:

1. As I understand it, Streams are significantly less efficient (make much more garbage) than simple `for` loops for this kind of trivial transformation. You might consider switching to a `for` loop if you're at all worried about that.
2. The only difference between `readArticles` and `readArticlesPredictive` is the method called on `this.data`. You could consolidate the transformation to `List<DictionaryEntry>`.

src/org/omegat/core/dictionaries/LingvoDSL.java
```diff
@@ -64,161 +52,226 @@
...
+         *
+         * @return result.
+         */
         @Override
-        public List<DictionaryEntry> readArticlesPredictive(String word) {
-            return data.lookUpPredictive(word).stream().map(e -> new DictionaryEntry(e.getKey(), e.getValue()))
-                    .collect(Collectors.toList());
+        public String getObject() {
+            if (sb == null) {
+                return "";
```
amake, line 240: Just an idea: You can make calling `getObject` before `start` an error and throw an exception in that case.

amake (approved):
Thanks, this looks quite nice. I'm glad we now have a rigorous approach.

Just let me know what you think about my minor comments.

src/org/omegat/core/dictionaries/LingvoDSL.java
miurahr: Thank you for the feedback. I'd like to try both ideas.

amake:

PR #170, opened 2021-12-09
Dictionary: fix to point correct word(BUGS#1074)

miurahr:
- Correct scroll position
   - when using prefix search
   - when title word is different with index word
- resolve [BUGS#1074](https://sourceforge.net/p/omegat/bugs/1074/)
- Fix bug raising and ignoring an exception when move to last entry

### DIFF
--- a/src/org/omegat/core/dictionaries/DictionaryEntry.java
+++ b/src/org/omegat/core/dictionaries/DictionaryEntry.java
@@ -37,9 +37,7 @@ public class DictionaryEntry implements Comparable<DictionaryEntry> {
     private final String article;
 
     public DictionaryEntry(final String word, final String article) {
-        this.query = word.toLowerCase();
-        this.word = word;
-        this.article = article;
+        this(word.toLowerCase(), word, article);
     }
 
     public DictionaryEntry(final String query, final String word, final String article) {
--- a/src/org/omegat/core/dictionaries/DictionaryEntry.java
+++ b/src/org/omegat/core/dictionaries/DictionaryEntry.java
@@ -4,6 +4,7 @@
           glossaries, and translation leveraging into updated projects.
 
  Copyright (C) 2009 Alex Buloichik
+               2021 Hiroshi Miura
                Home page: http://www.omegat.org/
                Support center: https://omegat.org/support
 
@@ -31,12 +32,24 @@ package org.omegat.core.dictionaries;
  * @author Alex Buloichik (alex73mail@gmail.com)
  */
 public class DictionaryEntry implements Comparable<DictionaryEntry> {
+    private final String query;
     private final String word;
     private final String article;
 
     public DictionaryEntry(final String word, final String article) {
+        this.query = word.toLowerCase();
         this.word = word;
         this.article = article;
+    }
+
+    public DictionaryEntry(final String query, final String word, final String article) {
+        this.query = query;
+        this.word = word;
+        this.article = article;
+    }
+
+    public String getQuery() {
+        return query;
     }
 
     public String getWord() {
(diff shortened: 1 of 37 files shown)


src/org/omegat/core/dictionaries/DictionaryEntry.java
```diff
@@ -31,14 +31,26 @@
...
     private final String word;
     private final String article;
 
     public DictionaryEntry(final String word, final String article) {
+        this.query = word.toLowerCase();
         this.word = word;
         this.article = article;
     }
 
+    public DictionaryEntry(final String query, final String word, final String article) {
```
miurahr, line 44: question: how to allow plugin developer to use this safely with compatible older omegat.

miurahr: ref https://wiki.eclipse.org/Evolving_Java-based_APIs

miurahr: It works file for plugins built with OmegaT 5.7. 
When dictionary plugin adopt new interface, the plugin will only work with newer OmegaT.

src/org/omegat/core/dictionaries/DictionaryEntry.java
```diff
@@ -31,14 +32,26 @@
  * @author Alex Buloichik (alex73mail@gmail.com)
  */
 public class DictionaryEntry implements Comparable<DictionaryEntry> {
+    private final String query;
     private final String word;
     private final String article;
 
     public DictionaryEntry(final String word, final String article) {
+        this.query = word.toLowerCase();
```
miurahr, line 40: This is for backward compatibility

src/org/omegat/gui/dictionaries/DictionariesTextArea.java
```diff
@@ -197,16 +198,26 @@ protected void callDictionary(String word) {
...
+                return;
+            }
         }
         Element el = doc.getElement(Integer.toString(index));
         if (el == null) {
             return;
         }
         int start = el.getStartOffset();
-        int end = el.getEndOffset();
+        int end = el.getEndOffset() - 1;
```
miurahr, line 220: This is one of places which lead exception. When try to select last word, Swing cannot make rectangle if end pointing really last character. this  change help endRect will be a point the last character when item is last.

miurahr:
@amake could you leave a comment?
I've suggested to extend DictionaryEntry class, a public API.

miurahr:
Resolved the merge conflict.

src/org/omegat/core/dictionaries/DictionaryEntry.java
amake, line 40: I would have this constructor call the other constructor rather than (mostly) duplicate its body.

src/org/omegat/core/dictionaries/DictionaryEntry.java
```diff
@@ -49,6 +62,10 @@ public String getArticle() {
 
     @Override
     public int compareTo(DictionaryEntry o) {
-        return word.compareTo(o.word);
+        if (query.equals(o.query)) {
+            return word.compareTo(o.word);
+        } else {
+            return query.compareTo(o.query);
+        }
```
amake, line 69: You could slightly optimize this like

```suggestion
        int queryComparison = query.compareTo(o.query);
        if (queryComparison != 0) {
            return queryComparison;
        } else {
            return word.compareTo(o.word);
        }
```

amake (commented):
Mostly LGTM, just some small comments

src/org/omegat/core/dictionaries/DictionaryEntry.java
miurahr: resolved.

src/org/omegat/core/dictionaries/DictionaryEntry.java
miurahr: accepted the suggestion.

miurahr:
Now ready to merge.

PR #172, opened 2021-12-11
POFilter: fix empty entry hack (BUGS#1082)

miurahr:
POFilter option `disallowEditingBlackSegment` does not work properly.
This patch change option to be positive `allowEditingBlackSegment` and handle header properly.

Resolve [BUGS#1082](https://sourceforge.net/p/omegat/bugs/1082/).


src/org/omegat/filters2/po/PoFilter.java
```diff
@@ -325,7 +326,7 @@ public void processFile(File inFile, File outFile, FilterContext fc) throws IOEx
 
         String disallowEditingBlankSegmentStr = processOptions.get(OPTION_ALLOW_EDITING_BLANK_SEGMENT);
         allowEditingBlankSegment = disallowEditingBlankSegmentStr == null
-                || disallowEditingBlankSegmentStr.equalsIgnoreCase("true");
+                || disallowEditingBlankSegmentStr.equalsIgnoreCase("false");
```
amake, line 329: Instead of inverting this, we should invert `disallowEditingBlankSegmentStr` (change it to `allowEditingBlankSegmentStr` and adjust all logic to match) because everywhere else in the program the setting is phrased as a positive ("allow"), not a negative ("disallow").

miurahr:
Now I introduce PoOptions class to handle options, and add test case.
PoOptions class handle options properly and also see `disallow*` options as backward compatibility.

miurahr:
@amake I've adjusted for review comment. Do you have further ideas?

miurahr:
I believe it is worth merging now.

src/org/omegat/filters2/po/PoFilter.java
miurahr: I've changed it to ``allowEditingBlackSegmentStr`` and keep it to compare with "true" as same as the suggestion.

miurahr:
I've added a test case that not only reproduce a bug also check a ``allowEditBlankSegment`` option behaviors.
I've changed variable names to be positive notation.

I'd like to squash and mege it into ``master``.

PR #173, opened 2021-12-12
HTML2FilterTest: fix failure when locale is not EN

miurahr:
Because comments can be localized, an expectation is also localized.
Fix [BUGS#1066](https://sourceforge.net/p/omegat/bugs/1066/)


amake:
Thanks, I've merged this to `master`.

PR #176, opened 2021-12-13
update macos build instructions (and other little things)

brandelune:
The readme needed some updates. I've only changed the macos part and a few generic things.

### DIFF
--- a/doc_src/Readme.md
+++ b/doc_src/Readme.md
@@ -108,8 +108,8 @@ The location of dbk must also be set in `docbook-utf8.xsl`:
 
     <xsl:import href="file:///c:\dbk\html\docbook.xsl"/>
 
-Both `doc_src_paths.xml` and `docbook-utf8.xsl` must be copied at the same level
-as your OmegaT development folder.
+Both `doc_src_paths.xml` and `docbook-utf8.xsl` must be copied at the
+same level as your OmegaT development folder.
 
     C:\dev\doc_src_paths.xml
     C:\dev\docbook-utf8.xsl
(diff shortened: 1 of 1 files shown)


miurahr:
LGTM.

doc_src/Readme.md
```diff
@@ -106,44 +118,48 @@ The ones available in `doc_src` are only there for reference.
 
 ### macOS
 
-All the dependencies can be installed through MacPorts except for Saxon 6-5-5:
+`ant` can be installed with the Xcode tools:
+    `$ xcode-select --install`
+	
+All the other dependencies can be installed through [MacPorts](https://www.macports.org/) except for saxon and whc.
+    `$ sudo port install docbook-xsl-nons, docbook-xml-4.5 @5.0_1, fop @1.1_1, libxml2 @2.9.12_1`
```
amake, line 125: 1. You don't need backticks around indented code blocks
2. None of the other CLI commands in this document are prefixed with the shell prompt; we should at 
3. The syntax of the `port` command is wrong:
  - You can't suffix the port names with commas
  - Specifying the version is meaningless at best because MacPorts only lets you install the current port version
4. Does `ant` still come with Xcode? Where is it on your system? I don't have it except for through MacPorts.

    I suspect it hasn't been bundled with Xcode for a long time. We should just instruct people to get it from MacPorts.

doc_src/Readme.md
```diff
@@ -106,44 +118,48 @@ The ones available in `doc_src` are only there for reference.
...
+- unzip the `docbook-xml-4.5-5.0_1.darwin_20.noarch.tbz2` package found
+inside.
+- rename the `4.5` folder found in `opt/local/share/xml/docbook/4.5/`
+  to `/docbook-xml-4.5`.
+- move that folder to the OmegaT source tree, to the folder "above" `doc_src`:
+    - /path/to/omegat/doc_src/en/
+    - /path/to/docbook-xml-4.5/
+- `doc_src_paths.xml` and `docbook-utf8.xsl` must be put at the same
+level as `docbook-xml-4.5`. The ones available in `doc_src` are only
+there for reference.
```
amake, line 162: 1. Wouldn't it be easier to tell people to symlink to the installed files at `/opt/local/share/xml/docbook/4.5/`?
2. Sequential steps would probably be better as an ordered list an an unordered list

brandelune:
The port install syntax was supposed to be corrected before pushing. Sorry. I'll rewrite the thing. Thank you for checking.

brandelune:
I've just pushed a correction.

brandelune:
Thank you the modifications. I was not looking at the right location for the DTD... It looks good to me.

amake:

PR #177, opened 2021-12-14
Glossary: shorter indent for dictionary rendering

miurahr:
- resolve RFE#1588 and RFE#1587

### DIFF
--- a/src/org/omegat/gui/glossary/DictionaryGlossaryRenderer.java
+++ b/src/org/omegat/gui/glossary/DictionaryGlossaryRenderer.java
@@ -57,11 +57,8 @@ public class DictionaryGlossaryRenderer implements IGlossaryRenderer {
         for (int i = 0; i < targets.length; i++) {
             if (i == 0 || (!targets[i].equals(targets[i - 1]))) {
                 if (hasComments) {
-                    if (trg instanceof HtmlTarget) {
-                        trg.append("\n<span style=\"text-indent: 10px\">");
-                    } else {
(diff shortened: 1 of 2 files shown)


src/org/omegat/gui/glossary/DictionaryGlossaryRenderer.java
```diff
@@ -57,8 +57,11 @@ public void render(GlossaryEntry entry, IRenderTarget<?> trg) {
         for (int i = 0; i < targets.length; i++) {
             if (i == 0 || (!targets[i].equals(targets[i - 1]))) {
                 if (hasComments) {
-                    trg.append("\n\t");
-                    hasComments = false;
+                    if (trg instanceof HtmlTarget) {
```
amake, line 60: This smells like a problem. We shouldn't have to know about the type of `trg`. The render target itself should know what to do here. Like maybe `IRenderTarget` should have methods like `startIndent` and `endIndent` or something?

t-cordonnier: Yes I agree with that. We cannot be sure that there will not be another renderer in the future.
Two possibilities: either methods return a string and we write trg.append(trg.startIndent()) or we have a method trg.appendStartIndent(), as you prefer

miurahr: Add `trg.appendStartIndent()`

amake:
Thanks. I've squashed this to `master`.

PR #174, opened 2021-12-12
Team: Git: migrate to Apache MINA sshd (RFE#1598)

miurahr:
JSch has been unmaintained and JGit goes to Apache MINA sshd for its target.
This PR migrate ssh transport to use Apache MINA sshd.

**THIS IS A CHANGE for OmegaT@6.1.0**

- Related PR #240 for OmegaT@5.8.0

- Bump JGit@6.1.0
- Fix [BUGS#1075](https://sourceforge.net/p/omegat/bugs/1075/)
- Support elliptic curve cryptography ed25519 and ecdsa keys [RFE#1598](https://sourceforge.net/p/omegat/feature-requests/1598/)
- Remove dependency for JSch and JSch agentproxy utilities.
- Use Apache MINA sshd for ssh connection and agent proxy.

Future tasks
- [x] Support ssh agent

ssh-agent support by JGit through apache MINA sshd is merged into JGit 6.0 (minimum Java 11)
https://git.eclipse.org/c/jgit/jgit.git/commit/?id=634302d2da74226cff9f78e121ad5b8216c476e6

The PR is successor of  PR #161 and #162.

PR #179, opened 2021-12-16
Team: Git: Support tracking non-default branch name for remote git repository (RFE#918) on JGit/MINA-SSH base

miurahr:
This allow project manager to maintain their team repository with specific branch name other than 'main' or 'master' default branch. User can specify branch name when downloading team project.

- Resolve RFE#918
- Successor of PR #134 
- Depends on PR #174

Modifications over #174:
- Add field to enter branch name in new team dialog
- Add feature to track non-default branch name in GITRemoteRepository2 class

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -165,12 +165,13 @@ dependencies {
         implementation "org.apache.lucene:lucene-analyzers-stempel:${luceneVersion}"
 
         // Team project server support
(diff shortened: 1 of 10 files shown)


miurahr:
- Re-based on updated #174 
- run genJAXB

miurahr:
I'd like to consolidate it into #174

PR #180, opened 2021-12-19
Migrate MT to use Jackson/Json parser (RFE#1575) 

miurahr:
- Support Java 11
  - Drop dependency for deprecated javascript engine in Java 11 
  - [JEP 335: Deprecate the Nashorn JavaScript Engine](https://openjdk.java.net/jeps/335)
- All MTs migrate to Jackson
- This is successor of #136
- Implement [RFE#1575](https://sourceforge.net/p/omegat/feature-requests/1575/)


src/org/omegat/core/machinetranslators/ApertiumTranslate.java
```diff
@@ -135,31 +137,23 @@ protected String translate(Language sLang, Language tLang, String text) throws E
...
-        } else {
-            return OStrings.getString("APERTIUM_CUSTOM_SERVER_INVALID");
-        }
-
-        if (rootNode.containsKey("responseData")) {
-            Map<String, Object> data = (Map<String, Object>) rootNode.get("responseData");
-            tr = (String) data.get("translatedText");
-        }
+        Integer code = rootNode.get("responseStatus").asInt();
+        // return OStrings.getString("APERTIUM_CUSTOM_SERVER_INVALID");
```
amake, line 150: The old code checked that the `responseStatus` key was present. The new code assumes it is.

I don't know enough about Apertium to know if this is a safe assumption to make.

If it's OK to assume this, then we should remove the code. If not, then we should check for the key.

src/org/omegat/core/machinetranslators/DeepLTranslate.java
```diff
@@ -147,22 +148,17 @@ private String unescapeHTML(String text) {
...
 
         // { "translations": [ { "detected_source_language": "DE", "text": "Hello World!" } ] }
-        try {
-            List<Object> translationsList = (List<Object>) rootNode.get("translations");
-            Map<String, String> translationNode = (Map<String, String>) translationsList.get(0);
-            return translationNode.get("text");
-        } catch (NullPointerException e) {
-            return null;
-        }
+        return rootNode.get("translations").get(0).get("text").asText();
```
amake, line 161: The old code allowed (in a dumb way) for something to be null. How does the new code behave if there is a null?

amake: (I don't know anything about Jackson but I assume this blows up with an NPE)

src/org/omegat/core/machinetranslators/Google2Translate.java
```diff
@@ -155,22 +156,15 @@ private String unescapeHTML(String text) {
...
-
-        try {
-            Map<String, Object> dataNode = (Map<String, Object>) rootNode.get("data");
-            List<Object> translationsList = (List<Object>) dataNode.get("translations");
-            Map<String, String> translationNode = (Map<String, String>) translationsList.get(0);
-            return translationNode.get("translatedText");
-        } catch (NullPointerException e) {
-            return null;
-        }
+        return rootNode.get("data").get("translations").get(0).get("translatedText").asText();
```
amake, line 167: This also has the same change in null handling as above.

amake (commented):
Thanks for doing this work.

I'm a little concerned about the change in null handling, as I commented a couple times.

Did you test all of these connectors? I know that would be difficult, so I'm OK even if you didn't test them all. We can call for users to test the nightly builds after we land this PR.

src/org/omegat/core/machinetranslators/Google2Translate.java
miurahr: Because I found API document, I'd like to write response model and parse it explicitly.

miurahr:
> Thanks for doing this work.
> 
> I'm a little concerned about the change in null handling, as I commented a couple times.

> Did you test all of these connectors?

I don't test it because I don't know these, except for MyMemory. I've tested with MyMemory free account.

src/org/omegat/core/machinetranslators/ApertiumTranslate.java
miurahr: I've added test code and improve the check.

src/org/omegat/core/machinetranslators/Google2Translate.java
miurahr: Done.

src/org/omegat/core/machinetranslators/DeepLTranslate.java
miurahr: Updated.

miurahr:
> Did you test all of these connectors?

I've added tests for Google, Apertium and DeepL parser to check it works.

amake:
Thank you for this. I've squashed to `master`.

Since you weren't able to test with the actual services (which is very difficult, I know), we should probably put a call out for users to test with the nightly builds.

PR #135, opened 2021-07-20
LanguageToolNetworkBridge migrate to Jackson json parser (RFE#1575)

miurahr:
* Migration to Jackson for JSON parser
* [RFE#1575](https://sourceforge.net/p/omegat/feature-requests/1575/)

Signed-off-by: Hiroshi Miura <miurahr@linux.com>

PR #181, opened 2021-12-20
Migrate languagetool to use Jackson/Json parser(RFE#1575)

miurahr:
- Support Java 11
  - Drop dependency for deprecated javascript engine in Java 11 
  - [JEP 335: Deprecate the Nashorn JavaScript Engine](https://openjdk.java.net/jeps/335)
- LanguageToolNetworkBridge to use Jackson
- Implement [RFE#1575](https://sourceforge.net/p/omegat/feature-requests/1575/)
- related PR #180


miurahr:
Duplicated with #135

PR #182, opened 2021-12-24
Refactoring PrepareTMXEntry and TMXEntry classes

miurahr:
Refactoring TMXEntry and PrepareTMXEntry classes based on [omegat-development ML discussion](https://sourceforge.net/p/omegat/mailman/omegat-development/thread/ab740468-a692-1cc6-dc9f-5540928da99f%40northside.tokyo/#msg37405978)

### Changes

- I don't change existent PrepareTMXEntry/TMXEntry public API
- add @Deprecate IProject#setTranslation with PrepareTMXEntry as argument
- `TMXEntry` and `PrepareTMXEntry`  now has common interface `ITranslationEntry`
- Extend `TMXEntry` class to hold `List<TMXProp> otherproperties` 
-  `ExternalTMX` hold immutable `TMXEntry` instead of mutable `PrepareTMXEntry`
    -  The change is motivated by a comment of @t-cordonnier that external TMXs are basically immutable.
    - `ExternalTMXFactory` uses TMXEntry::Builder
- Searcher:
```
private void searchEntries(Collection < PrepareTMXEntry > tmEn, final String tmxID) throws SearchLimitReachedException
private void searchEntriesAlternative(Collection < TMXEntry > tmEn, final String tmxID) throws SearchLimitReachedException
```
consolidated into
```
private void searchEntries(Iterable < ITMXEntry > tmEn, final String tmxID) throws SearchLimitReachedException
```

### A present for holiday season! 🎁

### DIFF
--- a/src/org/omegat/core/data/ExternalTMFactory.java
+++ b/src/org/omegat/core/data/ExternalTMFactory.java
@@ -4,6 +4,7 @@
           glossaries, and translation leveraging into updated projects.
 
  Copyright (C) 2017 Aaron Madlon-Kay
+               2021 Hiroshi Miura
                Home page: http://www.omegat.org/
                Support center: https://omegat.org/support
 
@@ -27,7 +28,6 @@ package org.omegat.core.data;
 
 import java.io.File;
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.List;
 import java.util.Locale;
 
@@ -123,8 +123,8 @@ public final class ExternalTMFactory {
             return new ExternalTMX(file.getName(), loadImpl(sourceLang, targetLang));
         }
 
-        private List<PrepareTMXEntry> loadImpl(Language sourceLang, Language targetLang) throws Exception {
-            List<PrepareTMXEntry> entries = new ArrayList<>();
+        private List<TMXEntry> loadImpl(Language sourceLang, Language targetLang) throws Exception {
+            List<TMXEntry> entries = new ArrayList<>();
 
             TMXReader2.LoadCallback loader = new TMXReader2.LoadCallback() {
                 public boolean onEntry(TMXReader2.ParsedTu tu, TMXReader2.ParsedTuv tuvSource,
@@ -166,22 +166,21 @@ public final class ExternalTMFactory {
                             tuvSource.text, targetLang, tuvTarget.text, sources, targets);
 
                     for (int i = 0; i < sources.size(); i++) {
-                        PrepareTMXEntry te = new PrepareTMXEntry();
-                        te.source = sources.get(i);
-                        te.translation = targets.get(i);
-                        te.changer = changer;
-                        te.changeDate = changed;
-                        te.creator = creator;
-                        te.creationDate = created;
-                        te.note = tu.note;
-                        te.otherProperties = new ArrayList<TMXProp>(tu.props);
-                        te.otherProperties.add(new TMXProp(PROP_SOURCE_LANGUAGE, tuvSource.lang));
-                        te.otherProperties.add(new TMXProp(PROP_TARGET_LANGUAGE, tuvTarget.lang));
+                        TMXEntry.Builder builder = new TMXEntry.Builder()
+                                .setSource(sources.get(i))
+                                .setCreator(creator)
+                                .setCreationDate(created)
+                                .setTranslation(targets.get(i))
+                                .setChanger(changer)
+                                .setChangeDate(changed)
+                                .setNote(tu.note)
+                                .setProperties(tu.props)
+                                .setProperty(PROP_SOURCE_LANGUAGE, tuvSource.lang)
+                                .setProperty(PROP_TARGET_LANGUAGE, tuvTarget.lang);
                         if (nonTarget) {
-                            te.otherProperties.add(new TMXProp(PROP_FOREIGN_MATCH, "true"));
+                            builder.setProperty(PROP_FOREIGN_MATCH, "true");
                         }
-
-                        entries.add(te);
+                        entries.add(builder.build());
                     }
                 }
             };
@@ -231,8 +230,8 @@ public final class ExternalTMFactory {
             return new ExternalTMX(file.getName(), loadImpl(sourceLang, targetLang));
         }
 
-        private List<PrepareTMXEntry> loadImpl(Language sourceLang, Language targetLang) throws Exception {
-            List<PrepareTMXEntry> entries = new ArrayList<>();
+        private List<TMXEntry> loadImpl(Language sourceLang, Language targetLang) throws Exception {
+            List<TMXEntry> entries = new ArrayList<>();
             ParseEntryResult throwaway = new ParseEntryResult();
             Core.getFilterMaster().loadFile(file.getPath(),
                     new FilterContext(sourceLang, targetLang, true).setRemoveAllTags(removeTags),
@@ -299,7 +298,7 @@ public final class ExternalTMFactory {
 
     public static final class Builder {
         private final String name;
-        private final List<PrepareTMXEntry> entries = new ArrayList<>();
+        private final List<TMXEntry> entries = new ArrayList<>();
 
         public Builder(String name) {
             this.name = name;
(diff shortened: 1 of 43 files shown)


src/org/omegat/core/data/ExternalTMFactory.java
```diff
@@ -168,17 +166,17 @@ private void addTuv(TMXReader2.ParsedTu tu, TMXReader2.ParsedTuv tuvSource,
...
                         te.creationDate = created;
-                        te.note = tu.note;
-                        te.otherProperties = new ArrayList<TMXProp>(tu.props);
-                        te.otherProperties.add(new TMXProp(PROP_SOURCE_LANGUAGE, tuvSource.lang));
-                        te.otherProperties.add(new TMXProp(PROP_TARGET_LANGUAGE, tuvTarget.lang));
+                        te.setTranslation(targets.get(i));
+                        te.setChanger(changer);
+                        te.setChangeDate(changed);
+                        te.setNote(tu.note);
+                        te.otherProperties = new ArrayList<>(tu.props);
```
miurahr, line 175: Is it necessary to add `setProperties(props)` method?

src/org/omegat/core/data/ExternalTMFactory.java
```diff
@@ -168,17 +166,17 @@ private void addTuv(TMXReader2.ParsedTu tu, TMXReader2.ParsedTuv tuvSource,
                     for (int i = 0; i < sources.size(); i++) {
                         PrepareTMXEntry te = new PrepareTMXEntry();
                         te.source = sources.get(i);
```
miurahr, line 168: Is it necessary to add `PrepareTMXEntry#setSource`, `setCreator`, and `SetCreationDate` method?

src/org/omegat/core/data/IProject.java
```diff
@@ -139,6 +139,14 @@ void compileProjectAndCommit(String sourcePattern, boolean doPostProcessing, boo
      */
     List<SourceTextEntry> getAllEntries();
 
+    /**
+     * Set translation for entry.
+     *
+     * @param ste source text entry.
+     * @param tmxEntry tmx entry.
+     */
+    void setTranslation(SourceTextEntry ste, TMXEntry tmxEntry);
```
miurahr, line 148: Should `setTranslation(SourceTextEntry ste, TMXEntry tmxEntry)` be `setTranslation(SourceTextEntry ste, ITMXEntry entry)`?

src/org/omegat/core/data/ITMXEntry.java
```diff
@@ -0,0 +1,59 @@
...
+
+    boolean hasNote();
+
+    boolean hasProperties();
+
+    String getPropValue(String propType);
+
+    boolean hasPropValue(String propType, String propValue);
+
+    Iterable<TMXProp> getProperties();
```
miurahr, line 58: Should expose `List<TMXProp>` instead of `Iterable<>` ? or provide `getRawProperties` method?

src/org/omegat/core/data/ImportFromAutoTMX.java
```diff
@@ -169,20 +168,13 @@ private boolean altTranslationMatches(PrepareTMXEntry entry, EntryKey key) {
                 path = p.getValue();
             }
         }
-        return key.equals(new EntryKey(file, entry.source, id, prev, next, path));
+        return key.equals(new EntryKey(file, entry.getSource(), id, prev, next, path));
     }
 
-    private void setTranslation(SourceTextEntry entry, PrepareTMXEntry trans, boolean defaultTranslation,
+    private void setTranslation(SourceTextEntry entry, MutableTMXEntry trans, boolean defaultTranslation,
```
miurahr, line 174: TODO: looks mistake.

src/org/omegat/core/data/PrepareTMXEntry.java
```diff
@@ -28,23 +29,19 @@
...
 import org.omegat.util.TMXProp;
 
 /**
- * Class for prepare TMXEntry content before save unchangeable copy in the ProjectTMX. We can't use just
- * parameters in the setTranslation() method since count of parameters is too much. Structure of this class is
- * almost the save like TMXEntry.
- *
- * Instead, we will set all parameters into this class, then ProjectTMX will convert in into TMXEntry than
- * save internally.
+ * Class for *Mutable* version of TMXEntry.
```
miurahr, line 38: Try update javadoc. better wording?

src/org/omegat/core/data/PrepareTMXEntry.java
```diff
@@ -28,23 +29,19 @@
...
  *
  * @author Alex Buloichik (alex73mail@gmail.com)
  * @author Guido Leenders
  * @author Aaron Madlon-Kay
  */
-public class PrepareTMXEntry {
+public class PrepareTMXEntry extends MutableTMXEntry implements ITMXEntry {
```
miurahr, line 44: Explicitly define `PrepareTMXEntry` is mutable here.

src/org/omegat/core/data/PrepareTMXEntry.java
```diff
@@ -57,14 +54,18 @@
...
 
-    public PrepareTMXEntry(TMXEntry e) {
-        source = e.source;
-        translation = e.translation;
-        changer = e.changer;
-        changeDate = e.changeDate;
-        creator = e.creator;
-        creationDate = e.creationDate;
-        note = e.note;
+    public PrepareTMXEntry(SourceTextEntry e) {
```
miurahr, line 57: This can be a constructor which take `source`, `creator` and `createDate`.

src/org/omegat/util/TMXWriter.java
```diff
@@ -67,7 +68,7 @@ private TMXWriter() {
      * @throws IOException
      */
     public static void buildTMXFile(final String filename, final boolean forceValidTMX,
-            final boolean levelTwo, final ProjectProperties config, final Map<String, PrepareTMXEntry> data)
+            final boolean levelTwo, final ProjectProperties config, final Map<String, TMXEntry> data)
```
miurahr, line 71: Should it be `final Map<String, ITMXEntry> data` ?

test/src/org/omegat/languagetools/FalseFriendsTest.java
```diff
@@ -147,6 +147,9 @@ public ProjectProperties getProjectProperties() {
                 return null;
             }
 
+            public void setTranslation(SourceTextEntry ste, TMXEntry tmxEntry) {
```
miurahr, line 150: Should it be `setTranslation(SourceTextEntry ste, ITMXEntry entry)`?

miurahr (commented):
My questions and findings.

src/org/omegat/core/data/ExternalTMX.java
```diff
@@ -43,9 +43,9 @@
 
     private final String name;
 
-    private final List<PrepareTMXEntry> entries;
+    private final List<ITranslationEntry> entries;
 
-    ExternalTMX(String name, List<PrepareTMXEntry> entries) {
+    ExternalTMX(String name, List<ITranslationEntry> entries) {
```
miurahr, line 48: This can be break plugins when plugin use `ExternalTMX` constructor. Any plugin use it?

miurahr (commented):
Find a possibility for breakage of public API.

src/org/omegat/core/data/ExternalTMX.java
```diff
@@ -43,9 +44,9 @@
 
     private final String name;
 
-    private final List<PrepareTMXEntry> entries;
+    private final List<TMXEntry> entries;
 
-    ExternalTMX(String name, List<PrepareTMXEntry> entries) {
+    ExternalTMX(String name, List<TMXEntry> entries) {
```
miurahr, line 49: Here is a place where public API is changed.

miurahr:
I'm wondering whether is ITranslationEntry interface really necessary or not.

src/org/omegat/core/data/ITMXEntry.java
t-cordonnier: Personally in interfaces I prefer to use Iterable instead of List. I have users who do not hesitate to use gigabytes of translation memories in a project, for that reason I have some implementations of my plugins which do not really load everything in memory, and for that reason sometimes the implementation of such a method is not necessarily a collection.

t-cordonnier:
> I'm wondering whether is ITranslationEntry interface really necessary or not.

ITranslationEntry (or ITMXEntry in your code) is useful to have common callers, remember my sample for Searcher:
private void searchEntries(Collection<PrepareTMXEntry> tmEn, final String tmxID) throws SearchLimitReachedException
private void searchEntriesAlternative(Collection<TMXEntry> tmEn, final String tmxID) throws SearchLimitReachedException 
which I would simplify to
private void searchEntries(Iterable<ITMXEntry> tmEn, final String tmxID) throws SearchLimitReachedException

In my side I did a small review of the code, I saw that 90% of the calls of these two classes are read-only, so good candidates for replacement with the interface having only getters.

On the other hand I do not see any good reason why to have the abstract class for mutable entry: not only because this abstract class will probably have only one implementation, but because I do not see in the code any algorithm which can modify both internal and external entries a similar way.

src/org/omegat/core/data/ITMXEntry.java
t-cordonnier: That said, TMXEntry does not store properties and there is a reason: if you look in project_save.tmx you will see always the same properties which are stored in the entry key, not in the entry itself. So, no reason to have these methods in the interface (unless you see in the code agorithms which are using the properties and could be used for both chasses: in such a case, implementation in InternalTMXEntry should always be null)

t-cordonnier:
I started my own implementation as well, but not yet published, not enough mature. Let's wait for next week and we can compare after.

src/org/omegat/core/data/ExternalTMX.java
t-cordonnier: Actually I don't think that OmegaT plugins are using these class. For DGT it is different because I also have translation memory plugins.
That said, to be more generic you can write
ExternalTMX(String name, Iterable<? extends ITranslationEntry> entries) {
then, callers based on List<PrepareTMXEntry> should continue to work.

src/org/omegat/core/data/IProject.java
t-cordonnier: If it only uses the read methods of tmxEntry I would  say yes. Instead, not sure

miurahr:
@t-cordonnier thanks for opinions. I'd like to rewrite a change history in order to remove try-and-error commits.

t-cordonnier:
Don't go too fast. As I said previously I started to do something similar and I would like that we work together, so that I can reflect similar changes later in DGT and keep compatibility between the tools. But I am not working today, sorry.

miurahr:
I'd like to close here. You can compare  #184 and #185 how we can improve TMXEntry/PrepareTMXEntry

PR #184, opened 2021-12-28
Introduce TMXEntry builder in minimalist approach

miurahr:
This is minimalist approach of #182
Here does same as #182,  other than introduction of ITranslationEntry,  its getters and replacement for getters.

## Proposed changes

This includes following commits that every changes work well.

### Part 0. Preparation

1. Tweaks AutoTmxTest and PoFilterTtest
    Hide explicit variable of PrepareTMXEntry, for future replacement to TMXEntry
2. TMXEntry hold otherProperties
    TMXEntry also hold otherProperties which PrepareTMXEntry has. This make TMXEntry  immutable version of PrepareTMXEntry, and allow using it for replacement of PrepareTMXEntry in some place
 
### Part 1 . Introduce builder class

3. Introduce TMXEntry::Builder
4. MergeTest to test with TMXEntry::Builder
    MergeTest tests equivalence of objects. For future replacement from PrepareTMXEntry to TMXEntry, this checks 
### part 2. add Public API with TMXEntry and use it

5. Add `IProject#setTranslation` methods with TMXEntry object
    this prepare future replacement of the method equivalence.

### part 3. use Builder in classes

6. ProjectTMX: use TMXEntry::Builder
    This looks many changes, but it carefully remove PreapreTMXEntry object used as an intermediate variable.
7. ReplaceFilter: use TMXEntry::Builder

### ~~part 4. Improve TMXWriter2~~

~~TMXWriter is for PrepareTMXEntry which is used in command line mode.~~
~~We can migrate writing project file with TMXWriter2 in CLI mode.~~

~~8. Introduce TMXWrite2#buildTMXFile static method~~
~~9. Main: use TMXWriter2 for CLI TMX generations~~

### part 5. use Builder in tests

10. TMXComplianceTest: use TMXEntry::Builder
11. integrationTest: use TMXEntry::Builder
     the test code depends IProject#setTranslation

### part 6. ExternalTM should have immutable entry

12. ExternalTMFactory: return immutable TMXEntry
     this changes public API of external TMX file handlings.  

### changed
This series of patches make public API modification at part1. 6

* `ExternalTMX` has a  `List<TMXEntry>` field instead of `List<PrepareTMXEntry>`, and changes constructor and return type of `getEntries`.

### added

There are several public API additions in part 2

* `IProject#setTranslation` which accept TMXEntry

### DIFF
--- a/src/org/omegat/core/data/ExternalTMFactory.java
+++ b/src/org/omegat/core/data/ExternalTMFactory.java
@@ -4,6 +4,7 @@
           glossaries, and translation leveraging into updated projects.
 
  Copyright (C) 2017 Aaron Madlon-Kay
+               2021 Hiroshi Miura
                Home page: http://www.omegat.org/
                Support center: https://omegat.org/support
 
@@ -27,7 +28,6 @@ package org.omegat.core.data;
 
 import java.io.File;
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.List;
 import java.util.Locale;
 
@@ -123,8 +123,8 @@ public final class ExternalTMFactory {
             return new ExternalTMX(file.getName(), loadImpl(sourceLang, targetLang));
         }
 
-        private List<PrepareTMXEntry> loadImpl(Language sourceLang, Language targetLang) throws Exception {
-            List<PrepareTMXEntry> entries = new ArrayList<>();
+        private List<TMXEntry> loadImpl(Language sourceLang, Language targetLang) throws Exception {
+            List<TMXEntry> entries = new ArrayList<>();
 
             TMXReader2.LoadCallback loader = new TMXReader2.LoadCallback() {
                 public boolean onEntry(TMXReader2.ParsedTu tu, TMXReader2.ParsedTuv tuvSource,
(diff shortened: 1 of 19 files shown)


t-cordonnier:
Generally, what you propose is an alternative way to ensure that TMXEntry remains immutable.
As I said in my own pull request, before searching a way to make TMXEntry immutable we should ask ourselves _why_ it is immutable: if the goal is to ensure changes from everywhere (such as plugins), the immutable views which I did in my own pull request may be a better alternative.
For me, your TMX builder shares an inconvenience with PrepareTMXEntry: each time you need an entry you must create two objects (the builder and finally the entry) and loose one of them. Costs time and memory. Having mutable objects hidden through immutable interfaces does not cause this problem.

miurahr:
Because #187 is proposed, I'd like to drop related change from here and do force-push.

miurahr:
#187 suggests me that `PrepareTMXEntry#getTMXEntry` small builder utility method help changes keeping small,  and emphasize `PrepareTMXEntry` is a *Preparation* class for build `TMXEntry` object.
It may appeal to some developers.

miurahr:
I'd like to put this away because I've proposed the successor PR #191 that intend to keep binary compatibility for plugins.

PR #185, opened 2021-12-29
Refactor tmx entry

t-cordonnier:
Introduce interfaces to TMXEntry

These interfaces have multiple goals.

In short term, they enable to merge functions actually duplicated, as you can see in my second patch (in Searcher class)

In middle term (I did only one example in patch 3) they enable to create an immutable view to (maybe) mutable entries: now users of ExternalTMX cannot modify its contents. We could do the same for ProjectTMX: store mutable entries in the maps, but return ITMXEntry in getDefaultTranslation/getMultipleTranslation
As @amake mentionned in the discussion list, PrepareTMXEntry actually looks like mutable variant of TMXEntry. That was initial meaning, but if you look where they are used, you will see that since ExternalTMX continued to use PrepareTMXEntry, it has been used as something similar to external TMX entry.
One other difference between these two classes is that PrepareTMXEntry can contain any kind of properties (as it can be read from any kind of TMX) while TMXEntry stores only properties a project_save.tmx should contain (actually no property but already stores externalLinked): this enforces the idea that TMXEntry means, in reality, internal TMX entry.
Then we can ask ourselves: why does ExternalTMX store mutable entries while internal (project) TMX contains immutable entries? Shouldn't the contrary be more logical?

In long term, these two interfaces could ensure stability of the code even if you change the definition of TMXEntry/PrepareTMXEntry (as @miurahr wanted to do), even if you finally remove one of them or rewrite it completely: all code using the interfaces would not need to be changed. 
But it requires that we inventory where the classes are used, and for each use case, whenever it requires writing (else it should use interface instead).

### DIFF
--- a/src/org/omegat/core/data/ITMXEntry.java
+++ b/src/org/omegat/core/data/ITMXEntry.java
@@ -25,6 +25,9 @@
 
 package org.omegat.core.data;
 
+import java.util.List;
+
+import org.omegat.util.TMXProp;
 
 /**
  * Common interface for any object storing a pair source / translation text
@@ -62,4 +65,14 @@ public interface ITMXEntry extends ITranslationEntry {
     default boolean hasNote() {
         return getNote() != null;
     }
+
+    /* --------------------- Properties ------------ */
+
+    boolean hasProperties();
+
+    String getPropValue(String propType);
+
+    boolean hasPropValue(String propType, String propValue);
+
+    List<TMXProp> getProperties();
 }
--- a/src/org/omegat/core/data/ITMXEntry.java
+++ b/src/org/omegat/core/data/ITMXEntry.java
@@ -1,0 +1,65 @@
+/**************************************************************************
+ OmegaT - Computer Assisted Translation (CAT) tool
+          with fuzzy matching, translation memory, keyword search,
+          glossaries, and translation leveraging into updated projects.
+
+ Copyright (C) 2021 Thomas Cordonnier
+               Home page: http://www.omegat.org/
+               Support center: https://omegat.org/support
+
+ This file is part of OmegaT.
+
+ OmegaT is free software: you can redistribute it and/or modify
+ it under the terms of the GNU General Public License as published by
+ the Free Software Foundation, either version 3 of the License, or
+ (at your option) any later version.
+
+ OmegaT is distributed in the hope that it will be useful,
+ but WITHOUT ANY WARRANTY; without even the implied warranty of
+ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ GNU General Public License for more details.
+
+ You should have received a copy of the GNU General Public License
+ along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ **************************************************************************/
+
+package org.omegat.core.data;
+
+
+/**
+ * Common interface for any object storing a pair source / translation text
+ * with date and author
+ *
+ * @author Thomas Cordonnier
+ */
+public interface ITMXEntry extends ITranslationEntry {
+
+    /**
+     * Gets the initial creator of the entry
+     */
+    String getCreator();
+
+    /**
+     * Gets the initial creation date as an EPOCH timestamp
+     */
+    long getCreationDate();
+
+    /**
+     * Gets the author of last change in the entry
+     */
+    String getChanger();
+
+    /**
+     * Gets the EPOCH timestamp for last change in this entry
+     */
+    long getChangeDate();
+
+    /**
+     * Gets text note (markup &lt;note&gt; in TMX format)
+     */
+    String getNote();
+
+    default boolean hasNote() {
+        return getNote() != null;
+    }
+}
--- a/src/org/omegat/core/data/ImportFromAutoTMX.java
+++ b/src/org/omegat/core/data/ImportFromAutoTMX.java
@@ -33,7 +33,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 
-import org.omegat.util.StringUtil;
 import org.omegat.util.TMXProp;
 
 /**
@@ -66,8 +65,8 @@ public class ImportFromAutoTMX {
      */
     void process(ExternalTMX tmx, boolean isEnforcedTMX) {
 
-        for (PrepareTMXEntry e : tmx.getEntries()) { // iterate by all entries in TMX
-            List<SourceTextEntry> list = existEntries.get(e.source);
+        for (ITMXEntry e : tmx.getEntries()) { // iterate by all entries in TMX
+            List<SourceTextEntry> list = existEntries.get(e.getSourceText());
             if (list == null) {
                 continue; // there is no entries for this source
             }
@@ -113,13 +112,13 @@ public class ImportFromAutoTMX {
                             || existTranslation.linked == TMXEntry.ExternalLinked.x100PC) {
                         // already contains x-ice
                         if (hasICE
-                                && !Objects.equals(existTranslation.translation, e.translation)) {
+                                && !Objects.equals(existTranslation.getTranslationText(), e.getTranslationText())) {
                             setTranslation(ste, e, false, TMXEntry.ExternalLinked.xICE);
                         }
                     } else if (existTranslation.linked == TMXEntry.ExternalLinked.x100PC) {
                         // already contains x-100pc
                         if (has100PC
-                                && !Objects.equals(existTranslation.translation, e.translation)) {
+                                && !Objects.equals(existTranslation.getTranslationText(), e.getTranslationText())) {
                             setTranslation(ste, e, false, TMXEntry.ExternalLinked.x100PC);
                         }
                     }
@@ -128,13 +127,13 @@ public class ImportFromAutoTMX {
         }
     }
 
-    private boolean isAltTranslation(PrepareTMXEntry entry) {
-        if (entry.otherProperties == null) {
+    private boolean isAltTranslation(ITMXEntry entry) {
+        if (!entry.hasProperties()) {
             return false;
         }
         boolean hasFileProp = false;
         boolean hasOtherProp = false;
-        for (TMXProp p : entry.otherProperties) {
+        for (TMXProp p : entry.getProperties()) {
             if (p.getType().equals(ProjectTMX.PROP_FILE)) {
                 hasFileProp = true;
             } else if (p.getType().equals(ProjectTMX.PROP_ID)
@@ -148,12 +147,12 @@ public class ImportFromAutoTMX {
         return EntryKey.isIgnoreFileContext() ? hasOtherProp : hasFileProp;
     }
 
-    private boolean altTranslationMatches(PrepareTMXEntry entry, EntryKey key) {
-        if (entry.otherProperties == null) {
+    private boolean altTranslationMatches(ITMXEntry entry, EntryKey key) {
+        if (!entry.hasProperties()) {
             return false;
         }
         String file = null, id = null, next = null, prev = null, path = null;
-        for (TMXProp p : entry.otherProperties) {
+        for (TMXProp p : entry.getProperties()) {
             if (ProjectTMX.PROP_FILE.equals(p.getType())) {
                 file = p.getValue();
             } else if (ProjectTMX.PROP_ID.equals(p.getType())) {
@@ -169,20 +168,14 @@ public class ImportFromAutoTMX {
                 path = p.getValue();
             }
         }
-        return key.equals(new EntryKey(file, entry.source, id, prev, next, path));
+        return key.equals(new EntryKey(file, entry.getSourceText(), id, prev, next, path));
     }
 
-    private void setTranslation(SourceTextEntry entry, PrepareTMXEntry trans, boolean defaultTranslation,
+    private void setTranslation(SourceTextEntry entry, ITMXEntry trans, boolean defaultTranslation,
             TMXEntry.ExternalLinked externalLinked) {
-        if (StringUtil.isEmpty(trans.note)) {
-            trans.note = null;
-        }
-
-        trans.source = entry.getSrcText();
-
         TMXEntry newTrEntry;
 
-        if (trans.translation == null && trans.note == null) {
+        if ((!trans.isTranslated()) && (!trans.hasNote())) {
             // no translation, no note
             newTrEntry = null;
         } else {
--- a/src/org/omegat/core/data/PrepareTMXEntry.java
+++ b/src/org/omegat/core/data/PrepareTMXEntry.java
@@ -95,6 +95,14 @@ public class PrepareTMXEntry implements ITMXEntry {
         return note;
     }
 
+    public boolean hasProperties() {
+        return (otherProperties != null) && (otherProperties.size() > 0);
+    }
+
+    public List<TMXProp> getProperties() {
+        return otherProperties;
+    }
+
     public String getPropValue(String propType) {
         if (otherProperties == null) {
             return null;
--- a/src/org/omegat/core/data/PrepareTMXEntry.java
+++ b/src/org/omegat/core/data/PrepareTMXEntry.java
@@ -44,7 +44,7 @@ import org.omegat.util.TMXProp;
  * @author Guido Leenders
  * @author Aaron Madlon-Kay
  */
-public class PrepareTMXEntry {
+public class PrepareTMXEntry implements ITMXEntry {
     public String source;
     public String translation;
     public String changer;
@@ -65,6 +65,34 @@ public class PrepareTMXEntry {
         creator = e.creator;
         creationDate = e.creationDate;
         note = e.note;
+    }
+
+    public String getSourceText() {
+        return source;
+    }
+
+    public String getTranslationText() {
+        return translation;
+    }
+
+    public String getCreator() {
+        return creator;
+    }
+
+    public long getCreationDate() {
+        return creationDate;
+    }
+
+    public String getChanger() {
+        return changer;
+    }
+
+    public long getChangeDate() {
+        return changeDate;
+    }
+
+    public String getNote() {
+        return note;
     }
 
     public String getPropValue(String propType) {
--- a/src/org/omegat/core/data/TMXEntry.java
+++ b/src/org/omegat/core/data/TMXEntry.java
@@ -28,7 +28,10 @@
 
 package org.omegat.core.data;
 
+import java.util.List;
 import java.util.Objects;
+
+import org.omegat.util.TMXProp;
 
 /**
  * Storage for TMX entry.
@@ -58,14 +61,14 @@ public class TMXEntry implements ITMXEntry {
     public final boolean defaultTranslation;
     public final ExternalLinked linked;
 
-    TMXEntry(PrepareTMXEntry from, boolean defaultTranslation, ExternalLinked linked) {
-        this.source = from.source;
-        this.translation = from.translation;
-        this.changer = from.changer;
-        this.changeDate = from.changeDate;
-        this.creator = from.creator;
-        this.creationDate = from.creationDate;
-        this.note = from.note;
+    TMXEntry(ITMXEntry from, boolean defaultTranslation, ExternalLinked linked) {
+        this.source = from.getSourceText();
+        this.translation = from.getTranslationText();
+        this.changer = from.getChanger();
+        this.changeDate = from.getChangeDate();
+        this.creator = from.getCreator();
+        this.creationDate = from.getCreationDate();
+        this.note = from.getNote();
 
         this.defaultTranslation = defaultTranslation;
         this.linked = linked;
@@ -163,5 +166,21 @@ public class TMXEntry implements ITMXEntry {
             return false;
         }
         return true;
+    }
+
+    public boolean hasProperties() {
+        return false;
+    }
+
+    public String getPropValue(String propType) {
+        return null; // for the moment internal entries do not store properties
+    }
+
+    public boolean hasPropValue(String propType, String propValue) {
+        return false; // for the moment internal entries do not store properties
+    }
+
+    public List<TMXProp> getProperties() {
+        return null; // for the moment internal entries do not store properties
     }
 }
--- a/src/org/omegat/core/data/TMXEntry.java
+++ b/src/org/omegat/core/data/TMXEntry.java
@@ -42,7 +42,7 @@ import java.util.Objects;
  * @author Guido Leenders
  * @author Aaron Madlon-Kay
  */
-public class TMXEntry {
+public class TMXEntry implements ITMXEntry {
     public enum ExternalLinked {
         // declares how this entry linked to external TMX in the tm/auto/
         xICE, x100PC, xAUTO, xENFORCED
@@ -71,12 +71,32 @@ public class TMXEntry {
         this.linked = linked;
     }
 
-    public boolean isTranslated() {
-        return translation != null;
+    public String getSourceText() {
+        return source;
     }
 
-    public boolean hasNote() {
-        return note != null;
+    public String getTranslationText() {
+        return translation;
+    }
+    
+    public String getCreator() {
+        return creator;
+    }
+
+    public long getCreationDate() {
+        return creationDate;
+    }
+
+    public String getChanger() {
+        return changer;
+    }
+
+    public long getChangeDate() {
+        return changeDate;
+    }
+
+    public String getNote() {
+        return note;
     }
 
     @Override
--- a/src/org/omegat/gui/glossary/GlossaryEntry.java
+++ b/src/org/omegat/gui/glossary/GlossaryEntry.java
@@ -31,6 +31,7 @@ import java.util.Arrays;
 import java.util.Objects;
 import java.util.stream.Stream;
 
+import org.omegat.core.data.ITranslationEntry;
 import org.omegat.util.StringUtil;
 
 /**
@@ -40,7 +41,7 @@ import org.omegat.util.StringUtil;
  * @author Aaron Madlon-Kay
  * @author Alex Buloichik
  */
-public class GlossaryEntry {
+public class GlossaryEntry implements ITranslationEntry {
     public GlossaryEntry(String src, String[] loc, String[] com, boolean[] fromPriorityGlossary, String[] origins) {
         mSource = StringUtil.normalizeUnicode(src);
         mTargets = loc;
@@ -60,6 +61,16 @@ public class GlossaryEntry {
         return mSource;
     }
 
+    /* Method for ITranslationEntry */
+    public String getSourceText() {
+        return mSource;
+    }
+    
+    /* Method for ITranslationEntry */
+    public String getTranslationText() {
+        return mTargets.length > 0 ? mTargets[0] : "";
+    }
+    
     /**
      * Return the first target-language term string.
      *
--- a/src/org/omegat/core/data/ExternalTMX.java
+++ b/src/org/omegat/core/data/ExternalTMX.java
@@ -43,7 +43,7 @@ public class ExternalTMX {
 
     private final String name;
 
-    private final List<PrepareTMXEntry> entries;
+    private final List<? extends ITMXEntry> entries;
 
     ExternalTMX(String name, List<PrepareTMXEntry> entries) {
         this.name = name;
@@ -54,7 +54,7 @@ public class ExternalTMX {
         return name;
     }
 
-    public List<PrepareTMXEntry> getEntries() {
+    public List<ITMXEntry> getEntries() {
         return Collections.unmodifiableList(entries);
     }
 }
--- a/src/org/omegat/core/data/ITranslationEntry.java
+++ b/src/org/omegat/core/data/ITranslationEntry.java
@@ -1,0 +1,53 @@
+/**************************************************************************
+ OmegaT - Computer Assisted Translation (CAT) tool
+          with fuzzy matching, translation memory, keyword search,
+          glossaries, and translation leveraging into updated projects.
+
+ Copyright (C) 2021 Thomas Cordonnier
+               Home page: http://www.omegat.org/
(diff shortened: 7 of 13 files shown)


miurahr:
I agree that there are several code duplication in order to handle both TMXEntry and PrepareTMXEntry for same purpose, such as `searchEntreis` method. The PR here resolve it by common interface contract among two classes.

I found there is similar problem in two classes; TMXWrite class and TMXWrite2 class. 
Does the proposal also resolve the duplication on two classes?  TMXWrite handles PrepareTMXEntry and called from CLI methods. TMXWrite2 handles TMXEntry and called from GUI parts. Both class has same function to write project TMX file.

miurahr:
You can take [this change for test cases from my PR](https://github.com/omegat-org/omegat/pull/184/commits/750b6a0594c0966e27a8a43b55d2fc4e2f211159) to avoid making TMXEntry/PrepareTMXEntry object variable, to resolve check error without any side effect.

t-cordonnier:
> I found there is similar problem in two classes; TMXWrite class and TMXWrite2 class. 
> Does the proposal also resolve the duplication on two classes? 

Yes, removing all duplications is one of the goals of this work. But I maybe forgot to say that I did not consider it as finished, there is much more work to be done but I wanted to expose it to explain the principle
So yes, modifying the TMXWriter pair is in the next steps, I did not have the time to do it, I wanted to concentrate on other aspects before.

Today I published new changes. Now TMXEntry is not completely immutable: when a user wants to change the note of current translation, we modify it in TMXEntry rather than creating a new object each time. Second patch shows that I would like to do the same for setTranslation, but this one is only a prototype because I did not have the time to modify every method which calls it.
Once again, if there was a real reason why TMXEntry was immutable, I ask @amake  and @briacp  to explain why. Until that, I consider that modifying the object takes less time and memory than creating a new one on each modification.

t-cordonnier:
> You can take [this change for test cases from my PR](https://github.com/omegat-org/omegat/pull/184/commits/750b6a0594c0966e27a8a43b55d2fc4e2f211159) to avoid making TMXEntry/PrepareTMXEntry object variable, to resolve check error without any side effect.

You are true I had completely forgotten the tests. Now I pushed a correction with --force.

miurahr:
> > I found there is similar problem in two classes; TMXWrite class and TMXWrite2 class.
> > Does the proposal also resolve the duplication on two classes?
> 
> Yes, removing all duplications is one of the goals of this work. But I maybe forgot to say that I did not consider it as finished, there is much more work to be done but I wanted to expose it to explain the principle So yes, modifying the TMXWriter pair is in the next steps, I did not have the time to do it, I wanted to concentrate on other aspects before.

OK, I'd like to work for it in another PR without dependency for #184 and #185. 
When The PR #185 completed and provide backward compatibility,  my proposal can be still working.

miurahr:
> > > I found there is similar problem in two classes; TMXWrite class and TMXWrite2 class.
> > > Does the proposal also resolve the duplication on two classes?
> > 
> > 
> > Yes, removing all duplications is one of the goals of this work. But I maybe forgot to say that I did not consider it as finished, there is much more work to be done but I wanted to expose it to explain the principle So yes, modifying the TMXWriter pair is in the next steps, I did not have the time to do it, I wanted to concentrate on other aspects before.
> 
> OK, I'd like to work for it in another PR without dependency for #184 and #185. When The PR #185 completed and provide backward compatibility, my proposal can be still working.

Done in #187

miurahr:
> As @amake mentionned in the discussion list, PrepareTMXEntry actually looks like mutable variant of TMXEntry. That was initial meaning, but if you look where they are used, you will see that since ExternalTMX continued to use PrepareTMXEntry, it has been used as something similar to external TMX entry.
One other difference between these two classes is that PrepareTMXEntry can contain any kind of properties (as it can be read from any kind of TMX) while TMXEntry stores only properties a project_save.tmx should contain (actually no property but already stores externalLinked): this enforces the idea that TMXEntry means, in reality, internal TMX entry.
Then we can ask ourselves: why does ExternalTMX store mutable entries while internal (project) TMX contains immutable entries? Shouldn't the contrary be more logical?

I don't think it is good idea to keep using `PrepareTMXEntry` other than *Preparetion* class of `TMXEntry`. This is *the*  reason of my frustration. 

 A design which allow `PrepareTMX `as yet another container and allow getting values  from it bring unnecessary complexity in the project.
I'd like to propose an extension of TMXEntry which *can* have properties. Only an `ExternalTMFactory` want to bring it into `TMXEntry` object. Other parts can be as-is. This is what #188 doing. 

If we want to have  the class which is `TMXEntry` variant *with properties*, we can introduce `ExternalTMXEntry` for the purpose, that seems have a same ancestor with `TMXEntry`,  and make PrepareTMXEntry as for *preparation*.
There can be `ExternalTMXEntry extends TMXEntry` with properties, and `ExternalTMFactor` use it.

t-cordonnier:
@miurahr Yes, the final idea (remember that I said my work is not finished) is something like ExternalTMXEntry with capacity to store variable list of properties while InternalTMXEntry would store only what a project_save.tmx is susceptible to contain. This is better than storing always maps because you don't need to store the same key thousands of times, and data can have a type (for example ExternalLinked)
In my logic PrepareTMXEntry will also disappear but I need some time to produce a patch, even as a prototype like the last one, to show you what I want to do as an alternative.
In any case the most important for me is the pair of interfaces: even if later the core team totally rejects the rest of the code, using these interfaces will ensure compatibility with later implementations.

t-cordonnier:
As promised, I removed all work about (im)mutability but also introduced ExternalTMXEntry
Now both internal (project) and external TMX entries are immutable 
PrepareTMXEntry still exists, usage as ITMXEntry is permitted for compatibility with plugins only: in the code itself, it is used as a builder for Project/External TMX entries

miurahr:
Looks good but CI test failed.

It looks good direction. I think there is no big problem for change core data in OmegaT OSS ecosystem, because there is no obvious effect/objection by plugin maintainers, so source compatibility may be enough.

miurahr:
@amake could you rerun CI test? It seems  an issue in Azure-Pipeline service, not on us.

src/org/omegat/core/data/ExternalTMXEntry.java
```diff
@@ -0,0 +1,125 @@
...
+ * unchangeable.
+ *
+ * Only RealProject can create and change TMXEntry objects.
+ *
+ * @author Alex Buloichik (alex73mail@gmail.com)
+ * @author Guido Leenders
+ * @author Aaron Madlon-Kay
+ */
+public class ExternalTMXEntry extends TMXEntry {
+    public List<TMXProp> otherProperties;
```
miurahr, line 49: I think this can be and should be private.

src/org/omegat/core/data/ExternalTMXEntry.java
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(changeDate / 1000, creationDate / 1000, translation, note, changer, creator,
+                source);
+    }
+    
+    public List<TMXProp> getProperties() {
+        return otherProperties;
```
miurahr, line 92: This should return a copy of properties. For example, 
```java
return Collections.unmodifiableList(otherProperties)
```

src/org/omegat/core/data/ImportFromAutoTMX.java
```diff
@@ -93,48 +93,48 @@ void process(ExternalTMX tmx, boolean isEnforcedTMX) {
...
                     }
                 }
             }
         }
     }
 
-    private boolean isAltTranslation(PrepareTMXEntry entry) {
-        if (entry.otherProperties == null) {
+    private boolean isAltTranslation(ITMXEntry entry) {
+        if (entry.getProperties() == null) {
```
miurahr, line 132: we should introduce `boolean hasProperties()` method then we can encapsulate properties field in class.

src/org/omegat/core/data/ProjectTMXEntry.java
```diff
@@ -0,0 +1,116 @@
+/**************************************************************************
+ OmegaT - Computer Assisted Translation (CAT) tool
+          with fuzzy matching, translation memory, keyword search,
+          glossaries, and translation leveraging into updated projects.
+
+ Copyright (C) 2010 Alex Buloichik
+               2012 Guido Leenders, Thomas Cordonnier
+               2013 Aaron Madlon-Kay
+               2014 Alex Buloichik, Aaron Madlon-Kay
```
miurahr, line 9: Your copyright here?

miurahr (changes requested):
A total direction looks good but there are several point where you can update, and some suggestions.

src/org/omegat/core/data/ExternalTMXEntry.java
t-cordonnier: I did it like that only because all other fields remained public in the other classes. And setting other fields as private requires lot of changes in the code. But I agree that it can be discutable. Tell me what you think

src/org/omegat/core/data/ExternalTMXEntry.java
t-cordonnier: Agree, change pushed today

src/org/omegat/core/data/ImportFromAutoTMX.java
t-cordonnier: Agree, change pushed today

src/org/omegat/core/data/ProjectTMXEntry.java
t-cordonnier: Agree, change pushed today

src/org/omegat/core/data/ExternalTMXEntry.java
miurahr: My idea is...
 As we discussed in developer mailing list, List<> is flexible but not optimal for memory consumption.
This is only chance to change access modifier for `otherProperties` because `ExternalTMXEntry` is new class. Other fields should public for backward compatibility.
 After we introduce `private otherProperties` and access methods, we can change it into other representation.
If we put it as `public`, we lost a chance to change it in future.

miurahr:
There are over handful checkstyle warnings, actually 15, for new class `ExternalTMXEntry`.
You can take care for warnings when you introduced new code. I think we should care it at least when touched.
Same for ITMXEntry, `PrepareTMXEntry` getters, `ProjectTMXEntry` getters and `TMXEntry` getters.

OmegaT dev community dislike a style only patch, we should be careful and check checkstyle warnings when adding classes and methods.

src/org/omegat/core/data/TMXEntry.java
```diff
@@ -28,51 +28,66 @@
 
 package org.omegat.core.data;
 
+import java.util.List;
```
miurahr, line 31: This is unused import.

src/org/omegat/core/data/TMXEntry.java
```diff
@@ -28,51 +28,66 @@
...
 import java.util.Objects;
 
+import org.omegat.util.TMXProp;
```
miurahr, line 34: Unused import.

src/org/omegat/core/data/TMXEntry.java
```diff
@@ -110,21 +125,12 @@ public boolean equals(Object obj) {
...
-            return false;
-        }
         if (!Objects.equals(source, other.source)) {
             return false;
         }
         return true;
     }
 
-    @Override
-    public int hashCode() {
```
miurahr, line 123: When you define `equals` you should also define `hashCode`, this is one of checkstyle warning.

src/org/omegat/core/data/ITMXEntry.java
```diff
@@ -0,0 +1,74 @@
...
+     * Gets text note (markup &lt;note&gt; in TMX format)
+     */
+    String getNote();
+    
+    /* --------------------- Properties ------------ */
+    
+    boolean hasProperties();
+    
+    String getPropValue(String propType);
+
```
miurahr, line 70: white space in empty line.

miurahr (changes requested):
There is several obvious coding style issues.

miurahr:
I've tried to merge https://github.com/omegat-org/omegat/pull/187  , I found it conflict with introduction of `ProjectTMXEntry`.

I need to add method `buildTMXFile` that take `Map<String, TMXEntry>` as source data, that need to change `Map<String, ProjectTMXEntry>` but package-private `PrepareTMXEntry#toProjectTMXEntry` can not be looked from Main class.

```java
    public static void buildTMXFile(final String filename, final boolean forceValidTMX,
                                    final boolean levelTwo, final ProjectProperties config,
                                    final Map<String, ProjectTMXEntry> data) throws Exception {
        TMXWriter2 wr = new TMXWriter2(new File(filename), config.getSourceLanguage(), config.getTargetLanguage(),
                config.isSentenceSegmentingEnabled(), levelTwo, forceValidTMX);
        List<String> p = new ArrayList<>();
        try {
            for (Map.Entry<String, ProjectTMXEntry> en : data.entrySet()) {
                TMXEntry transEntry = en.getValue();
                wr.writeEntry(en.getKey(), transEntry.getTranslationText(), transEntry.getNote(),
                        transEntry.getCreator(), transEntry.getCreationDate(), transEntry.getChanger(),
                        transEntry.getChangeDate(), p);
            }
        } finally {
            wr.close();
        }
    }
```

src/org/omegat/core/data/PrepareTMXEntry.java
```diff
@@ -107,5 +147,12 @@ public String toString() {
                 .append(", otherProperties=").append(otherProperties).append("]");
         return builder.toString();
     }
+    
+    ExternalTMXEntry toExternalTMXEntry() {
+        return new ExternalTMXEntry(this);
+    }
 
+    ProjectTMXEntry toProjectTMXEntry(boolean defaultTranslation, ProjectTMXEntry.ExternalLinked linked) {
```
miurahr, line 155: Could this change to `public` because  [TMXWriter2 improvement conflicts with this](https://github.com/omegat-org/omegat/pull/185#issuecomment-1017136973)

miurahr: There is another idea to introduce `static ProjectTMX#createPseudoTranslateTMX` in `org.omegat.core.data` package instead of change to `public`.

t-cordonnier:
> public static void buildTMXFile(final String filename, final boolean forceValidTMX,
>                                    final boolean levelTwo, final ProjectProperties config,
>                                    final Map<String, ProjectTMXEntry> data) 
For me this method should take Map<String, ? extends ITMXEntry> data instead: now we should try to have methods as generic as possible
> List<String> p = new ArrayList<>();
> [...]
> wr.writeEntry(en.getKey(), transEntry.getTranslationText(), transEntry.getNote(),
> [...]
 >                                               transEntry.getChangeDate(), p);
Strange implementation here. For genericity, this method should take the properties from the entry itself (now ITMXEntry has properties method, which TMXEntry did not)

You may say that there are situations where we want to write other properties than the ones of the entry itself: for example project_save.tmx contains key properties which are not registered in ProjectTMXEntry. I will study this case but I suppose that we can use the already existing method in this case:
    public void writeEntry(String source, String translation, String note, String creator, long creationDate,
            String changer, long changeDate, List<String> propValues)

src/org/omegat/core/data/PrepareTMXEntry.java
t-cordonnier: This method is package-protected because previous constructor TMXEntry(PrepareTMXEntry) was like that and the comments in the top of this class said that it was intentional.
Now look to my answer to your previous comment and tell me if it is still required

miurahr:
Thank you for the suggestion. I'd like to move the TMX writing logic in Main class,  and remove dependency to  `ProjectTMXEntry` class.

t-cordonnier:
Yes I am working on it, I will add a new patch when I have time

miurahr:
Similar issue on `runConsoleAlign`:
It can be 

```        TMXWriter2 wr = new TMXWriter2(new File(tmxFile), config.getSourceLanguage(), config.getTargetLanguage(),
                config.isSentenceSegmentingEnabled(), false, false);
        Map<String, ProjectTMXEntry> data = p.align(p.getProjectProperties(), new File(dir));
        try {
            for (Map.Entry<String, ProjectTMXEntry> en : data.entrySet()) {
                wr.writeEntry(en.getKey(), en.getValue().getTranslationText(), en.getValue(), null);
            }
        } finally {
            wr.close();
        }
```

It looks clean now.

miurahr:
I've updated #187 .
Here is an updated merge https://github.com/miurahr/omegat/blob/topic/t-cordonnier/refactor-tmxEntry/src/org/omegat/Main.java#L447-L508
please check and compare with yours.

src/org/omegat/core/data/RealProject.java
```diff
@@ -1814,7 +1814,7 @@ protected String getSegmentTranslation(String id, int segmentIndex, String segme
             this.config = props;
         }
 
-        Map<String, TMXEntry> data = new HashMap<>();
+        Map<String, ProjectTMXEntry> data = new HashMap<>();
```
miurahr, line 1817: Let's change this to TreeMap, based on discussion https://github.com/omegat-org/omegat/commit/e740a9753201228a08048a3296377583ac2312bd

miurahr:
Test failed because there is a missing update for integration test code.
```
test-integration/src/org/omegat/core/data/TestTeamIntegrationChild.java:668: error: incompatible types: TMXEntry cannot be converted to ProjectTMXEntry
            mergedTMX.setTranslation(ste, en, true);
```

t-cordonnier:
> Test failed because there is a missing update for integration test code.

Correct, should be ok now

src/org/omegat/core/data/TMXEntry.java
t-cordonnier: This warning is a little bit stupid: the method has been moved to concrete classes (Project,External)TMXEntry because the implementation differs, due to different internal structure.
Should we define an abstract method only to make the style checker happy?

miurahr: You can add `@SuppressWarnings` annotation here

src/org/omegat/gui/glossary/GlossaryEntry.java
```diff
@@ -58,8 +59,18 @@ public GlossaryEntry(String src, String loc, String com, boolean fromPriorityGlo
 
     public String getSrcText() {
         return mSource;
-    }
+    }
```
miurahr, line 62: Codestyle: spaces on end of line

src/org/omegat/core/data/PrepareTMXEntry.java
```diff
@@ -33,18 +33,22 @@
...
 /**
- * Class for prepare TMXEntry content before save unchangeable copy in the ProjectTMX. We can't use just
+ * Class for prepare TMXEntry content before save unchangeable copy anywhere. We can't use just
  * parameters in the setTranslation() method since count of parameters is too much. Structure of this class is
  * almost the save like TMXEntry.
  *
  * Instead, we will set all parameters into this class, then ProjectTMX will convert in into TMXEntry than
  * save internally.
  *
+ * In the past, entries read from tm/ folder were kept as instances of PrepareTMXEntry.
```
miurahr, line 43: codingstyle:  Line has trailing spaces.

src/org/omegat/core/data/ProjectTMXEntry.java
```diff
@@ -0,0 +1,120 @@
...
+    ProjectTMXEntry(ITMXEntry from, boolean defaultTranslation, ExternalLinked linked) {
+        super(from);
+
+        this.defaultTranslation = defaultTranslation;
+        this.linked = linked;
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (! super.equals(obj)) {
```
miurahr, line 66: Many checkstyle warnings to be fixed to fix based on coding guide on OmegaT.
```
[ant:checkstyle] [WARN] /home/vsts/work/1/s/src/org/omegat/core/data/ProjectTMXEntry.java:64:5: Class 'ProjectTMXEntry' looks like designed for extension (can be subclassed), but the method 'equals' does not have javadoc that explains how to do that safely. If class is not designed for extension consider making the class 'ProjectTMXEntry' final or making the method 'equals' static/final/abstract/empty, or adding allowed annotation for the method. [DesignForExtension]
[ant:checkstyle] [WARN] /home/vsts/work/1/s/src/org/omegat/core/data/ProjectTMXEntry.java:66:14: '!' is followed by whitespace. [NoWhitespaceAfter]
[ant:checkstyle] [WARN] /home/vsts/work/1/s/src/org/omegat/core/data/ProjectTMXEntry.java:80:5: Class 'ProjectTMXEntry' looks like designed for extension (can be subclassed), but the method 'hashCode' does not have javadoc that explains how to do that safely. If class is not designed for extension consider making the class 'ProjectTMXEntry' final or making the method 'hashCode' static/final/abstract/empty, or adding allowed annotation for the method. [DesignForExtension]
[ant:checkstyle] [WARN] /home/vsts/work/1/s/src/org/omegat/core/data/ProjectTMXEntry.java:82:42: '1000' is a magic number. [MagicNumber]
[ant:checkstyle] [WARN] /home/vsts/work/1/s/src/org/omegat/core/data/ProjectTMXEntry.java:82:63: '1000' is a magic number. [MagicNumber]
[ant:checkstyle] [WARN] /home/vsts/work/1/s/src/org/omegat/core/data/ProjectTMXEntry.java:91:14: '!' is followed by whitespace. [NoWhitespaceAfter]
[ant:checkstyle] [WARN] /home/vsts/work/1/s/src/org/omegat/core/data/ProjectTMXEntry.java:104: Line has trailing spaces. [RegexpSingleline]
[ant:checkstyle] [WARN] /home/vsts/work/1/s/src/org/omegat/core/data/ProjectTMXEntry.java:105:5: Class 'ProjectTMXEntry' looks like designed for extension (can be subclassed), but the method 'hasProperties' does not have javadoc that explains how to do that safely. If class is not designed for extension consider making the class 'ProjectTMXEntry' final or making the method 'hasProperties' static/final/abstract/empty, or adding allowed annotation for the method. [DesignForExtension]
[ant:checkstyle] [WARN] /home/vsts/work/1/s/src/org/omegat/core/data/ProjectTMXEntry.java:108: Line has trailing spaces. [RegexpSingleline]
[ant:checkstyle] [WARN] /home/vsts/work/1/s/src/org/omegat/core/data/ProjectTMXEntry.java:109:5: Class 'ProjectTMXEntry' looks like designed for extension (can be subclassed), but the method 'getPropValue' does not have javadoc that explains how to do that safely. If class is not designed for extension consider making the class 'ProjectTMXEntry' final or making the method 'getPropValue' static/final/abstract/empty, or adding allowed annotation for the method. [DesignForExtension]
[ant:checkstyle] [WARN] /home/vsts/work/1/s/src/org/omegat/core/data/ProjectTMXEntry.java:112: Line has trailing spaces. [RegexpSingleline]
[ant:checkstyle] [WARN] /home/vsts/work/1/s/src/org/omegat/core/data/ProjectTMXEntry.java:113:5: Class 'ProjectTMXEntry' looks like designed for extension (can be subclassed), but the method 'hasPropValue' does not have javadoc that explains how to do that safely. If class is not designed for extension consider making the class 'ProjectTMXEntry' final or making the method 'hasPropValue' static/final/abstract/empty, or adding allowed annotation for the method. [DesignForExtension]
[ant:checkstyle] [WARN] /home/vsts/work/1/s/src/org/omegat/core/data/ProjectTMXEntry.java:116: Line has trailing spaces. [RegexpSingleline]
[ant:checkstyle] [WARN] /home/vsts/work/1/s/src/org/omegat/core/data/ProjectTMXEntry.java:117:5: Class 'ProjectTMXEntry' looks like designed for extension (can be subclassed), but the method 'getProperties' does not have javadoc that explains how to do that safely. If class is not designed for extension consider making the class 'ProjectTMXEntry' final or making the method 'getProperties' static/final/abstract/empty, or adding allowed annotation for the method. [DesignForExtension]
```

miurahr (changes requested):
There are new raising checkstyle warnings.
Please take care  especially for added new classes.

src/org/omegat/core/data/ProjectTMXEntry.java
t-cordonnier: I converted the PR to draft so that we stop concentrate on checkstyle; I have still some points to test before considering everything is correct, and come back to the style before moving again to a real pull request

src/org/omegat/core/data/TMXEntry.java
t-cordonnier: Delayed, was true in a patch which will be moved to another pull request

src/org/omegat/core/data/TMXEntry.java
t-cordonnier: Patched moved to other PR, to be considered later

src/org/omegat/gui/glossary/GlossaryEntry.java
t-cordonnier: Solved

src/org/omegat/core/data/PrepareTMXEntry.java
t-cordonnier: Patch moved to other PR

src/org/omegat/core/data/ProjectTMXEntry.java
t-cordonnier: Definition of ProjectTMXEntry moved to another PR

t-cordonnier:
Regarding the problem of backward or upward compatibility with plugins, I had sent the results of my tests here:
https://sourceforge.net/p/omegat/mailman/message/37593315/

I did also test against my translation memories API and come to exactly same conclusion: the 3 first patches have zero impact on plugins, 3 last patches can have an impact.

For all that reasons, I remove the draft status for the 3 first patches, and will open a new pull request with other patches, which can be treated later. Only problem is that GitHub provides no way to make clear that the second PR depends on this one.

src/org/omegat/core/data/ImportFromAutoTMX.java
```diff
@@ -169,20 +169,14 @@ private boolean altTranslationMatches(PrepareTMXEntry entry, EntryKey key) {
...
-        if (StringUtil.isEmpty(trans.note)) {
-            trans.note = null;
-        }
-
-        trans.source = entry.getSrcText();
-
         TMXEntry newTrEntry;
 
-        if (trans.translation == null && trans.note == null) {
+        if (trans.getTranslationText() == null && (trans.getNote() == null || trans.getNote().length() == 0)) {
```
amake, line 179: I would be OK with a `hasNote()` method that computes `note != null && !note.isEmpty()`

amake (approved):
LGTM! I have given @t-cordonnier and @miurahr write access to the SourceForge git repo, so please push when you are ready.

src/org/omegat/core/data/ImportFromAutoTMX.java
t-cordonnier: OK, indeed the method was in TMXEntry, so I have moved it to the interface so that we are sure to have a consistent behaviour everywhere we would need it.
(in particular this line was not correct since it seems that OmegaT considers that empty note is different from no note at all, so now it is corrected)

miurahr (approved):

miurahr:
LGTM. I think it is ready to merge.

t-cordonnier:
@amake, @miruahr
Thanks for your review and for writing permissions
One last point before validating this PR

As you mentioned previously the checkstyle results, I see that the new methods often add the following warning:
"Class 'PrepareTMXEntry' looks like designed for extension (can be subclassed), but the method 'getPropValue' does not have javadoc that explains how to do that safely. If class is not designed for extension consider making the class 'PrepareTMXEntry' final or making the method 'getPropValue' static/final/abstract/empty, or adding allowed annotation for the method."
Setting the methods as final makes this warning disappear.
However, while it sounds clear for me that method like getCreator() or getNote() can be final, some other methods like getPropValue(String)  may be overridden, in particular in case of TMXEntry (look in PR #202, even if I know that this is a draft with errors, for example)
What makes also the warning to disappear is to add any kind of javadoc, such as /** do what you want **/

I also notice that this warning is present in lot of classes, and if we take care of it, it will be the first time.

So, the question is, should I
- ignore and validate as is
- add final attribute where I think it is correct (in accessors, but not in methods which may be modified)
- add a javadoc in methods which cannot be final, only to make the checkstyle happy, even if I don't know what to say...

Tell me what you think and we can validate the PR then.

amake:
My stance on CheckStyle is that formatting-level things like spaces and line length should be fixed unless there is a good reason not to.

On the other hand, things that affect semantics can generally be ignored. I have considered disabling the "designed for extension" rule entirely, but I left it on because it's sometimes good food for thought.

In this case I think

> - ignore and validate as is

is fine.

t-cordonnier:
Since both Aaron and Hiroshi confirmed, I did the merge.
Thanks to Aaron who gave me rights to do so

miurahr:
@t-cordonnier you merge here to github's mirror,  not source-forge's our origin source of truth.

@amake can I push mirror onto origin w/o rebase ?

miurahr:
I've just pushed PR #208 to express merge strategy and procedure in docs_devel.

t-cordonnier:
> @t-cordonnier you merge here to github's mirror, not source-forge's our origin source of truth.

Thanks for the info. Indeed I only clicked on the button in GitHub's web interface, was there another procedure?

amake:
> Thanks for the info. Indeed I only clicked on the button in GitHub's web interface, was there another procedure?

It must not have been clear, but I thought I had mentioned that the source of truth is the SourceForge git repo. The GitHub mirror is synced on-demand, so any changes made only here will be overwritten.

I've squashed your changes to SourceForge, and they have been synced here now: https://github.com/omegat-org/omegat/commit/569ebf2f8cc15f864354afdc456fadf24555097c

PR #186, opened 2021-12-30
Update RealProject tests

miurahr:
Test more methods on RealProject
   - setTranslation
   - saveProjectProperties
   - getTranslationInfo
   - setNote

The PR intend to help improvement effort of core data structure, such as #185  #188 #191


amake:
Thanks very much! We definitely need more tests, especially for important core classes like `RealProject`.

I have merged this to `master`.

PR #187, opened 2021-12-31
CLI: use TMXWriter2 for producing pseudo project TMX and  aligner

miurahr:
This is a proposal to migrate TMXWriter2 to produce pseudo/empty project TMX in Main CLI function.

### Problem 

GUI mode use TMXWriter2 to write project TMX file, but CLI use old TMXWriter. This code duplication and separation can be a nest of future bugs.

### Background

~TMXEntry is package-private and all fields are public. Main cannot call TMXEntry constructor, so it use its writer.~

Calling `TMXWriter2#writeEntry` from Main class solves the issue. 

### Approach

~1. Introduce `TMXWriter2#buildTMXFile` that has same functionality with `TMXWriter#buildTMXFile`~
~2. Introduce `PrepareTMXEntry#getTMXEntry` small builder utility method that work as a proxy for TMXEntry constructor.~
3. Update `Main#runCreatePseudoTranslateTMX` and `Main#runConsoleAlign`
4. Add `@Deprecated` annotation to `TMXWriter` class

### DIFF
--- a/src/org/omegat/Main.java
+++ b/src/org/omegat/Main.java
@@ -59,11 +59,9 @@ import org.omegat.convert.ConvertConfigs;
 import org.omegat.core.Core;
 import org.omegat.core.CoreEvents;
 import org.omegat.core.data.NotLoadedProject;
-import org.omegat.core.data.PrepareTMXEntry;
 import org.omegat.core.data.ProjectProperties;
 import org.omegat.core.data.RealProject;
 import org.omegat.core.data.SourceTextEntry;
-import org.omegat.core.data.TMXEntry;
 import org.omegat.core.events.IProjectEventListener;
 import org.omegat.core.tagvalidation.ErrorReport;
 import org.omegat.core.team2.TeamTool;
@@ -81,7 +79,7 @@ import org.omegat.util.Preferences;
 import org.omegat.util.ProjectFileStorage;
 import org.omegat.util.RuntimePreferences;
 import org.omegat.util.StringUtil;
-import org.omegat.util.TMXWriter;
+import org.omegat.util.TMXWriter2;
 import org.omegat.util.gui.OSXIntegration;
 
 import com.vlsolutions.swing.docking.DockingDesktop;
@@ -444,25 +442,19 @@ public final class Main {
             fname = "";
         }
 
-        // prepare tmx
-        Map<String, PrepareTMXEntry> data = new HashMap<>();
-        for (SourceTextEntry ste : entries) {
-            PrepareTMXEntry entry = new PrepareTMXEntry();
-            entry.source = ste.getSrcText();
-            switch (pseudoTranslateType) {
-            case EQUAL:
-                entry.translation = ste.getSrcText();
-                break;
-            case EMPTY:
-                entry.translation = "";
-                break;
+        // Write OmegaT-project-compatible TMX:
+        try (TMXWriter2 wr = new TMXWriter2(new File(fname), config.getSourceLanguage(), config.getTargetLanguage(),
+                config.isSentenceSegmentingEnabled(), false, false)) {
+            for (SourceTextEntry ste : entries) {
+                switch (pseudoTranslateType) {
+                    case EQUAL:
+                        wr.writeEntry(ste.getSrcText(), ste.getSrcText(), null, null, 0, null, 0, null);
+                        break;
+                    case EMPTY:
+                        wr.writeEntry(ste.getSrcText(), "", null, null, 0, null, 0, null);
+                        break;
+                }
             }
-            data.put(ste.getSrcText(), entry);
-        }
-
-        try {
-            // Write OmegaT-project-compatible TMX:
-            TMXWriter.buildTMXFile(fname, false, false, config, data);
         } catch (IOException e) {
             Log.logErrorRB("CT_ERROR_CREATING_TMX");
             Log.log(e);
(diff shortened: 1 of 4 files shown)


t-cordonnier:
About the general idea: yes, in the past TMXWriter was much more used than now, you found the last location where it was used, so if we can use only one writer, that is potentially a good idea.

I was not sure it would be possible to do it independently from change of TMXEntry definition, but since it uses lot of parameters instead of one, yes, it should work.

Note that the method is not strictly identical to what TMXWriter does: initial method wrote strings, this one is using StaX. Before validating, you should try to create some files with both methods and **check if there is no significant difference** susceptible to imply different behavior when the produced files are used in OmegaT or another tool. Code is important, but before all, what is important is its result...

miurahr:
I've added test case to check difference.

t-cordonnier:
> I've added test case to check difference.

Your test is based on Main class, so indirectly from TMXWriter2. This only shows whenever the change works.
Now to show the difference you should also implement a call to TMXWriter, and check whenever the two return the same contents in the file.

miurahr:
Only a difference between TMXWriter and TMXWriter2 is an attribute `creationtoolversion` string that is OmegaT version string. #190 change TMXWriter to produce same value as TMXWriter2.

miurahr:
I've updated it to reflect discussion on #185

miurahr:
> So, considering the fact that this class is probably the only location where there is any call to align method.

Yes, It is. Even though it is only place to caller, I think there is another class we should know to improve `align`.

There is a class shared between `RealProject#align` and `TmxComplianceBase#align`, `RealProject$AlignFilesCallback`,
that calls `IFilter#alignFile` plugin API, and constructs  tmx map, `Map<String, TMXEntry>`, for caller.

A method is `AlignFilesCallback#addTranslation`, it is in RealProject#L1821

t-cordonnier:
> There is a class shared between RealProject#align and TmxComplianceBase#align, RealProject$AlignFilesCallback,
that calls IFilter#alignFile plugin API, and constructs tmx map, Map<String, TMXEntry>, for caller.

This class is a test, it is not used by OmegaT itself. Main class remains the only location where OmegaT calls aligner. But we may eventually have a look to scripts and plugins.
More precisely TmxComplianceTests seems to be _the_ test set for aligner, so it is logical that it is using it!
So, first thing to do is probably to run this test, to check that your change does not cause a regression.

miurahr:
> So, first thing to do is probably to run this test, to check that your change does not cause a regression.

CI/CD does not report any regression.

t-cordonnier:
Then the last test to do is to try to do an alignment with old OmegaT and with the one including your patch. If the result is similar we can consider that this pull request is now mature and can be applied (actually, as it does not use TMXEntry or its variants, it can be considered as totally independent from the other pull requests)

miurahr:
> we can consider that this pull request is now mature and can be applied (actually, as it does not use TMXEntry or its variants, it can be considered as totally independent from the other pull requests)

Now remove dependency for TMXEntry or its variants from Main.

t-cordonnier (commented):
Instead of `void writeEntries(Map<String, TMXEntry> entries)` I would prefer `void writeEntries(Map<String, ? extends ITMXEntry> entries)` so that the method can be re-used elsewhere in the code, in location where other classes are used.

t-cordonnier:
Other possibility, even more generic: `void writeEntries(Iterable<? extends ITMXEntry> entries)`  and call with map.getValues(). But before doing this I will do tests to check whenever TreeMap.getValues() returns them in correct order or not.

Of course you will say that you used the key string as source but the real question is: in which cases would the key differ from entry.getSourceText()? Honnestly, as well as you had doubts about the use of PrepareTMXEntry, I also saw lot of pieces of code with Map<String,TMXEntry> where we use the key as source and everything else from the entry, while normally the source should be identical. This is one of the points where I said that I still hesitate about what to do. That is one of the reasons why I said that my pull request is not finished.

miurahr:
> This is one of the points where I said that I still hesitate about what to do. That is one of the reasons why I said that my pull request is not finished.

You can do it as another activity than introduction of ProjectTMXEntry.
Please don't make things too big. You can walk step by step, and  take one by one, solve this and there.

t-cordonnier:
> Please don't make things too big. You can walk step by step, and take one by one, solve this and there.

I agree in principle but the problem is the dependency between pull requests, while GitHub provides no way to say that one pull request depends on another (or if it does, I don't know)

What I could do is for example reduce my actual pull request to the two first patches, for which I think I did everything possible, and put the next ones in a distinct pull request kept as a draft. But how should I make clear that second PR depends on first one?

amake:
> But how should I make clear that second PR depends on first one?

If we had a "sane" workflow, you would do that by branching off of the first PR's branch, and targeting that same PR's branch with your PR (if that makes sense). GitHub is quite smart and will e.g. retarget your PR to `master` when the first PR is merged.

Unfortunately since our source of truth is SourceForge and we don't use GitHub's merging mechanism, I think it wouldn't work correctly for us.

amake:
Is this blocked on anything?

miurahr:
> Is this blocked on anything?

I don't have any work items.

t-cordonnier:
> Is this blocked on anything?

I would not say there is anything blocking. But since the pull request was made by Hiroshi, I thought it was up to him to validate it finally. Or maybe you prefer that we validate each other?

That said, let's come back to the remarks I made on Feb 5th:
https://github.com/omegat-org/omegat/pull/187#pullrequestreview-873912481
When I made them, I later realized that the interface ITMXEntry was not yet validated, making this remark irrelevant. But now it is integrated, so the question can come back

So, @amake  and @miurahr have a look to my two comments
https://github.com/omegat-org/omegat/pull/187#pullrequestreview-873912481
https://github.com/omegat-org/omegat/pull/187#issuecomment-1030573500

And let's decide together what we can do

miurahr:
I've done a rebase  and change writeEntries to accept Map<String, ? extends ITMXEntry> entries.

miurahr:
I've consolidated fixes into three individual commits. I think it is now ready.

miurahr:
I've merged into [master](8bc1ab58d3c82dde59371edb91356ec68939a6f6)

PR #189, opened 2021-12-31
MainTest: add console-align and pseudoTranslatetmx tests

miurahr:


miurahr:
I don't know why results of test cases have empy tmx body in CI environment.
It works on Java 11 on Linux, any difference?

miurahr:
This is very hard to test because configuration path is class variable so it can be changed during running test

PR #190, opened 2022-01-02
TMXWriter to produce omegat version string as same as TMXWriter2

miurahr:
Related #187


amake:
If we are deprecating and removing TMXWriter then do we need to make this change? It seems like if we do #187 then we don't need this one.

PR #191, opened 2022-01-13
Introduce ExternalTMXEntry

miurahr:
- PrepareTMXEntry: update as TMXEntry and ExternalTMXEntry factory method
- ITMXEntry common interface of TMXEntry/ExternalTMXEntry
- ExternalTMXEntry inherit PrepareTMXEntry for binary compatibility
    - IProject#getTransMemories() returns Map<String, ExternalTMX> and
      ExternalTMX has entries that is List<ExternalTMXEntry>
    - Intended to be compatible with previous versions


miurahr:
> Unfortunately what you do will not ensure compatibility with plugins because they will continue to create "new PrepareTMXEntry": interfaces are good to factorize methods but not constructors (new).

>  but not constructors (new).

What do you mean? Plugins can create PrepareTMXEntry object with constructor and access fields through public fields interface in Phase1. We should recommend plugins authors (incl. me :) ) and script authors like kos to rewrite their products until Phase 3 reached. 

> Another point is that your ExternalTMXEntry is mutable, while we had discussed that mutability is more pertinent for internal (project tmx) entries.

ExternalTMXEntry is mutable for smooth migration and keep compatibility in Phase 1. It finally become immutable in Phase 3.
I design ExternalTMXEntry as immutable, but I do not do it for the migration strategy.

miurahr:
> I just updated my own pull request (#185). In my release, ExternalTMXEntry is immutable as well as internal one, and both are built upon PrepareTMXEntry. They differ by definition of properties.
PrepareTMXEntry remains here as a builder and is acceptable as ITMXEntry only for compatibility with plugins: look for the comment in the class itself, I hope it is clear.

Unfortunately #185 still break binary compatibility for plugins.
It is because plugins build against OmegaT 4.3 or OmegaT 5.7 does not support  ITMXEntry.

miurahr:
I've updated description of PR to put my post to dev ML, and express this PR is for phase 1 of proposed migration plan.

miurahr:
>  and is acceptable as ITMXEntry only for compatibility with plugins: look for the comment in the class itself, I hope it is clear.

This breaks backward binary compatibility for plugins. Change to accept ITMXEntry DO break API binary compatibility. We should keep accepting PrepareTMXEntry for plugins. You can AMMEND method to accept ITMXEntry without breaking API  binary compatibility.

see. [Achieving API Binary Compatibility](https://wiki.eclipse.org/Evolving_Java-based_APIs_2)

t-cordonnier:
> Yes, I don't ENSURE for forward compatibility, but I keep BACKWARD compatibility of plugins released against OmegaT 4.3.x and OmegaT 5.x.

I hope to ensure both. Plugins using PrepareTMXEntry will remain compatible since I did not REMOVE anything in this class, I only added implementation of ITMXEntry

Let's say that in OmegaT 4 we have an API containing a method like this:
public abstract List<PrepareTMXEntry> getEntries(...)
which  I rewrote like this
public abstract List<? extends ITMXEntry> getEntries(...)
Now, in a plugin the implementation is
public List<PrepareTMXEntry> getEntries(...) { ... }
As you can see, the implementation is valid according both to old and new contract.

I already encountered situations like that in DGT-OmegaT. In such a case, we need to rebuild the plugin because else it will detect that PrepareTMXEntry has changed. But most important: we don't need to change the code. Strictly speaking I was wrong when I spoke about _binary_ compatibility but it is still _source_ compatible.

> It is because plugins build against OmegaT 4.3 or OmegaT 5.7 does not support ITMXEntry.

Looking at the code in previous paragraph, you can see that a plugin which returns List<PrepareTMXEntry>
- will respect the old contract, so will be compliant with OmegaT 4.3 and 5.7
- wil respect the new contract, so will be compliant with OmegaT 5.8

Know what? I will try next week to be more concrete and try with real plugins.
I had asked in one message whenever Aaron was able to tell us about an existing plugin, but I saw no answer.
Actually I don't think standard OmegaT has such plugins. But DGT does.
I have an API which enables to create translation memory plugins. There is a port to OmegaT 5 (I wanted to propose it to Aaron for integration but as you know he never found the time to consider it seriously):
http://www.silvestris-lab.org/node/43
So, I will try to rebuild my API against both my and your approach. And then we will see together which one has less impact on plugins.
Is it ok like that?

amake:
>  I had asked in one message whenever Aaron was able to tell us about an existing plugin, but I saw no answer.

Sorry! I am not aware of any plugins that use these APIs, but I have not searched very hard.

t-cordonnier:
> You breaks backward binary compatibility for plugins. Change to accept ITMXEntry DO break API binary compatibility. [...]
> see. Achieving API Binary Compatibility

I had a look in your link and I saw this(section "Evolving API Interfaces"):
Expand superinterface set (direct or inherited) | - | Binary compatible
In PrepareTMXEntry I added "implements ITMXEntry", so if I correctly understand, what I did is "expand subinterface set". It should be binary compatible. But as said imy previous message, I will check it with concrete plugins next week.

t-cordonnier:
Thanks for the search, Aaron. I am pretty sure that there are no plugins in actual OmegaT which make use of TMX entries. But in my own API (see the link) I have such plugins. So I will do the test with my plugins and then we will see whenever @miurahr  or me was right.

PR #193, opened 2022-01-24
Record and display what MT engine's translation is used (RFE#1584)

miurahr:
Re-implement RFE#1584 on top of PR#185


t-cordonnier:
I don't understand why you want _all_ TMX classes to have this property. ExternalTMX is here to store data as they are in the TMX, if MT engine is present it is one property between others.

ITMXEntry should remain unchanged: this is the common part between all implementations so should contain only the most generic part
In ExternalTMXEntry, no reason why to have a dedicated property: some TMX will have it, some not
In ProjectTMXEntry the property is stored as a string, OK for a method getMTSource() but methods returning properties should return a collection where this value is one of the results (actually this is a singleton list, but it may change if we add other properties later): then TMXWriter2 will write the correct property

t-cordonnier:
Oh, and reading second patch, did I correctly understand that you want to add a _markup_ (start and end element) with name "mtsource"? If you do so, the file is not anymore a valid TMX and you loose compatibility with other tools!
The property must be stored as standard markup, as it is already done for other properties. Only its "type" indicates that the property is the MT engine for OmegaT

miurahr:
RFE#1467, RFE#1477 and RFE#1584 raised to request getting MT source name and  handle it in OmegaT screen.
It can be stored in project TMX for further review process.

@amake suggest utilize of TMXProp for this in discussion on #140 and try it. And @t-cordonnier suggest an inefficiency of properties when extend TMXEntry to hold it as a `ArrayList`.

TMXProp is stored in ExternalTMXEntry in #185 but feature-request ask to store it in somewhere in ProjectTMX.

I've re-implemented #140 as
1. extend ProjectTMXEntry hold mtsource field as similar as note.
2. ProjectTMXEntry may have mtsource string when necessary
3. Write/read mtsource string value as a TMX `<prop type="x-mtsource"> value </prop>`in porject_save.tmx
4. OmegaT does not export it in level_? TMX files.

> The property must be stored as standard markup, as it is already done for other properties. Only its "type" indicates that the property is the MT engine for OmegaT

This PR do this.

> ITMXEntry should remain unchanged: this is the common part between all implementations so should contain only the most generic part

OK, I'd like to adjust it.

miurahr:
@t-cordonnier 
> In ProjectTMXEntry the property is stored as a string, OK for a method getMTSource() but methods returning properties should return a collection where this value is one of the results

Do you suggest that MT source string value stores as TMXProp, type of mtsource, in ProjectTMXEntry.properties?

t-cordonnier:
> @t-cordonnier
> Do you suggest that MT source string value stores as TMXProp, type of mtsource, in ProjectTMXEntry.properties?

What I suggest is that ProjectTMX 
- contains a _field_ mtEngine, typed String (or maybe IMachineTranslation)
- contains a method getMTSource
- the method getProperties() includes a TMXProp, type mtsource (but this is created only when you call the method)
So, instead of returning null as actually, it should actually return a singleton (unless the field is null). But other properties may be added later.

t-cordonnier:
Any reason why PROP_MTSOURCE is in TMXWriter2, which is not using it? My impression is that correct location is ProjectTMX, as for other OmegaT-specific properties. TMXWriter2 is a generic class for any ITMXEntry, it should not contain OmegaT-specific properties.

t-cordonnier:
More general suggestion, for the full pull request: instead of a property for MT source, I would create a property "origin": if in the future somebody wants to store the fact that a segment comes from translation memory, then one common property "origin" is enough since a segment cannot come at same time from TM and MT.

miurahr:
> That also means that you have two ways to get the MT engine (either via getMTSource or via getProperties(mt_source) : the first method is preferred when you are sure that you are using a ProjectTMXEntry, the second one is ideal for generic calls (when you only know that the object is an ITMXEntry, without being sure it is a ProjectTMXEntry or not)
> 
> PS. I had not seen that you corrected this in last patch. So, now it seems that we have the same idea.

I've take generic way to get it via `getPropValue("origin")`

miurahr:
I've adjusted all the comment from @t-cordonnier 
I think this is worth merging.

If there is no more comments, I'd like to rebase to create clean history on master:HEAD

miurahr:
re-based onto updated #185

miurahr:
When translator use MT result by selecting menu item `OverwriteMachineTranslationMenuItem` or `Ctrl-M`, then add TMXEntry with MT name and replace edit with MT result.
When editor post-edit translation, recorded origin will be erased and recognize translation is made by manual or unknown.

src/org/omegat/core/data/TMXEntry.java
```diff
@@ -171,20 +175,48 @@ public boolean equalsTranslation(TMXEntry other) {
         }
         return true;
     }
-    
+
+    /**
+     * We only hold origin property in ProjectTMXEntry.
```
t-cordonnier, line 180: Only one small detail: this comment still refers to ProjectTMXEntry which is not anymore in the pull request

t-cordonnier (approved):
About the code, except the small comment just added, I did all my comments long time before and you corrected everything.

I did a small test of the feature and it seems to work correctly. I checked in particular that the origin correctly disappears if you change anything in a segment.
Maybe I would change the added value to "MT:[engine name]" instead of only the engine name: remember that I suggested change to "origin" to enable, in the future, to do the same with TM or any other potential origin.

Except of that, all seems correct to me now.

t-cordonnier:
I will also consider the opportunity to apply your changes also in DGT-OmegaT.

In DGT-OmegaT origin is currently displayed with a color (http://185.13.37.79/?q=/node/4) but it is actually only valid between the time you click on CTRL+M and the validation, precisely because until now I did not save the value. This was not a limitation, the client simply did not request for it ... but it has been a subject of discussion in the OmegaT mailing list, years ago (https://sourceforge.net/p/omegat/mailman/message/36854293/). If I keep the origin, then it would be logical that in DGT-OmegaT the color is also maintained.

t-cordonnier:
Oh, sorry one last point I forgot
In Options => Editor, OmegaT has an option "Save auto-populated status"

Shouldn't we also consider the opportunity either to add an option to save origin or not?

miurahr:
>  Maybe I would change the added value to "MT:[engine name]" instead of only the engine name: remember that I suggested change to "origin" to enable, in the future, to do the same with TM or any other potential origin.

OK, I'd like to change it to "MT:[engine name]"


>but it has been a subject of discussion in the OmegaT mailing list, years ago (https://sourceforge.net/p/omegat/mailman/message/36854293/).

Some of issues are requested in RFE#1467 and RFE#1477 then change is proposed in https://github.com/omegat-org/omegat/pull/205

From your mail message
> However, this solution implies a question: how to identify which engine is currently the one which has just been inserted? And indeed, as soon as the patch was sent, the question was asked. Now it is still discussed and I follow Didier's suggestion to move the discussion here, to get other user's opinion.

#205 show which MT is selected and consistent result order by sort cycling order by name.

miurahr:
> Oh, sorry one last point I forgot In Options => Editor, OmegaT has an option "Save auto-populated status"

It is `AutoBehaviorPanel#cbSaveAutoStatus` and `Preferences.SAVE_AUTO_STATUS`
 
> Shouldn't we also consider the opportunity either to add an option to save origin or not?

Giving opportunity to  configuration is always good idea but I think we can give another options. There are  three possible behaviors.

1. Record MT origin and keep it after post-edit.
2. Record MT origin but remove/reset it when post-edit.
3. Show MT origin and keep it during session,  but does not save for next session.

t-cordonnier:
>     It is AutoBehaviorPanel#cbSaveAutoStatus and Preferences.SAVE_AUTO_STATUS

Yes, I mentioned it to emphase the fact that the idea was to be consistent with other options (if we can decide whenever we want to save auto status, that should be the same for origin)

 > Record MT origin but remove/reset it when post-edit.

In my mind when you post-edit, your segment does not come from MT any more. But I admit that others may think differently

t-cordonnier:
> #205 show which MT is selected and consistent result order by sort cycling order by name.
Yes, that is true. In the discussion I am referring to, people wanted to see it also in the MT pane, by setting it as bold. Personally I was not really in favour of this solution because bold has another meaning in Matches pane, so it could be source of confusion. But I remain open to other's opinion

t-cordonnier:
"This branch has conflicts that must be resolved"

Apparently you did not merge against the very last release of #185; in the meantime I added a few methods requested by Aaron.

miurahr:
Rebased on `master`

miurahr:
A commit link is gone to 404. What happened?

ref: https://github.com/omegat-org/omegat/pull/213#issuecomment-1102758758

miurahr:
seems 0961df01f2f97d9387b5e563bde

PR #194, opened 2022-01-30
Bump DSL4j@0.4.3 and DictZip@0.11.1

miurahr:
- LingvoDSL support
  previous support read all contents into memory when loading.
  Now dsl4j make article index at first time and save index into file.
  A performance after second time is very improved because just loading index.
  formatting in HTML is done when query word.
- DictZip
  Support mark/reset methods for InputStream interface.


amake:
Thanks. I've pushed to `master`.

PR #196, opened 2022-02-04
Bump dsl4j@0.4.5 and dictzip@0.11.2

miurahr:
- Fix broken article when using open-dsl-dict data.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -132,9 +132,9 @@ dependencies {
         implementation 'net.loomchild:maligna:3.0.1'
 
         // Dictionary
-        implementation 'io.github.dictzip:dictzip:0.11.1'
+        implementation 'io.github.dictzip:dictzip:0.11.2'
         implementation 'com.github.takawitter:trie4j:0.9.8'
-        implementation 'io.github.eb4j:dsl4j:0.4.4'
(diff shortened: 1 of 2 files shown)


amake:
I notice that running the unit tests results in two temporary files that are not cleaned up:

- test/data/dicts-lingvo-dz/test.dsl.dz.idx
- test/data/dicts-lingvo/test.dsl.idx

Are these created by the dsl4j library? If so, could you please make sure to clean them up?

miurahr:
> I notice that running the unit tests results in two temporary files that are not cleaned up:
> 
> * test/data/dicts-lingvo-dz/test.dsl.dz.idx
> * test/data/dicts-lingvo/test.dsl.idx
> 
> Are these created by the dsl4j library? If so, could you please make sure to clean them up?

These are generated index file that will load second time and later to speed up loading dictionary.
A file name is specified by OmegaT side.
When OmegaT pass null for loader, no index file is not created.

The files are generated by library using Protocol Buffer library, and check integrity, dictionary file name, size, mtime,  when loading by the library.

OmegaT can specify index file path other than the same directory of dictionary, for example, $HOME/.cache/omegat/bra/bra

amake:
OK. Users may find that useful, but they need to be cleaned up after unit tests.

PR #198, opened 2022-02-06
dictionary: javadoc: Update stardict format spec URL

miurahr:
- Use github repository url.
- google code was gone.
- Stardict.org domain was gone.


amake:

PR #199, opened 2022-02-08
dictionary:   Improve StarDict access with stardict4j

miurahr:
- Use stardict4j access library
  - Stardict4j is created from OmegaT stardict code. 
  - Stardict4j supports .dict and dict.dz, .idx and idx.gz, as same as current OmegaT does
  - It can handle .syn synonym file.
  - It can handle 64bit index file as same as 32bit index which OmegaT can handle.
  - It can handle entry type such as 'm', 'x' and 'h'. 
  - It has true LRU cache for articles expiry by number of items and  time (depends Caffeine)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -135,6 +135,7 @@ dependencies {
         implementation 'io.github.dictzip:dictzip:0.11.2'
         implementation 'com.github.takawitter:trie4j:0.9.8'
         implementation 'io.github.eb4j:dsl4j:0.4.5'
+        implementation 'io.github.eb4j:stardict4j:0.3.0'
 
         // Encoding dectection
         implementation 'com.github.albfernandez:juniversalchardet:2.4.0'
--- a/src/org/omegat/core/dictionaries/StarDict.java
+++ b/src/org/omegat/core/dictionaries/StarDict.java
@@ -6,6 +6,7 @@
  Copyright (C) 2009 Alex Buloichik
                2015-2016 Hiroshi Miura, Aaron Madlon-Kay
                2020 Suguru Oho, Aaron Madlon-Kay
+               2022 Hiroshi Miura
                Home page: http://www.omegat.org/
                Support center: https://omegat.org/support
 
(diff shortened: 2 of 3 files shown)


src/org/omegat/gui/preferences/view/DictionaryPreferencesPanel.form
```diff
@@ -47,5 +47,16 @@
         <AuxValue name="JavaCodeGenerator_VariableModifier" type="java.lang.Integer" value="0"/>
       </AuxValues>
     </Component>
+    <Component class="javax.swing.JCheckBox" name="condensedViewCheckBox">
+      <Properties>
+        <Property name="text" type="java.lang.String" value="Use &amp;Condensed view"/>
+      </Properties>
```
miurahr, line 53: The property should be localized such as;

```xml
        <Property name="text" type="java.lang.String" editor="org.netbeans.modules.i18n.form.FormI18nStringEditor">
          <ResourceString bundle="org/omegat/Bundle.properties" key="PREFS_DICTIONARY_CONDENSED" replaceFormat="OStrings.getString(&quot;{key}&quot;)"/>
        </Property>
```

src/org/omegat/gui/preferences/view/DictionaryPreferencesPanel.java
```diff
@@ -60,10 +61,23 @@ private void initComponents() {
...
+                condensedViewCheckBoxActionPerformed(evt);
+            }
+        });
+        add(condensedViewCheckBox);
     }// </editor-fold>//GEN-END:initComponents
 
+    private void condensedViewCheckBoxActionPerformed(java.awt.event.ActionEvent evt) {//GEN-FIRST:event_condensedViewCheckBoxActionPerformed
+        // TODO add your handling code here:
+    }//GEN-LAST:event_condensedViewCheckBoxActionPerformed
+
```
miurahr, line 77: should not have action listener.

miurahr (commented):
Need more change

amake:
I am willing to take the stardict4j part as-is, but I would prefer the other parts to be a separate PR.

miurahr:
> I am willing to take the stardict4j part as-is, but I would prefer the other parts to be a separate PR.

Done.

amake:
Thanks. I've pushed to `master`.

PR #201, opened 2022-02-14
Improve robustness for loading plugins

miurahr:
Catch exceptions precisely and protect another plugins.
Don't skip registering base plugins when 3rd party raises ClassNotFound error.

Related [RFE#1607 Keep OmegaT default colors when using glossary layout plugins](https://sourceforge.net/p/omegat/feature-requests/1607/)
Related https://github.com/omegat-org/glossary-renderer-plugin-skeleton/issues/1

### DIFF
--- a/src/org/omegat/filters2/master/PluginUtils.java
+++ b/src/org/omegat/filters2/master/PluginUtils.java
@@ -121,6 +121,7 @@ public final class PluginUtils {
                         // found main manifest - not in development mode
                         foundMain = true;
                     }
+                    loadFromManifest(m, pluginsClassLoader);
                     if ("theme".equals(m.getMainAttributes().getValue("Plugin-Category"))) {
                         String target = mu.toString();
                         for (URL url : urlList) {
@@ -129,7 +130,6 @@ public final class PluginUtils {
                             }
                         }
                     }
-                    loadFromManifest(m, pluginsClassLoader);
                 } catch (ClassNotFoundException e) {
                     Log.log(e);
                 }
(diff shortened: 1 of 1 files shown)


t-cordonnier (approved):
I did a small test with your change.
Remember my MT plugin? To do the test I created 3 versions of this plugin which return the same with a prefix (1, 2, 3). Then I modified the manifest of item 2 so that the class name is wrong.
With OmegaT 5.7 : none of the plugins is loaded
With your change : release 1 and 3 are correctly loaded, release 2 not. That was the expected behavior.
So, I confirm that your change has an added value here.

src/org/omegat/filters2/master/PluginUtils.java
```diff
@@ -116,15 +123,21 @@ public static void loadPlugins(final Map<String, String> params) {
...
                         String target = mu.toString();
-                        for (URL url : urls) {
+                        for (URL url : urlList) {
                             if (target.contains(url.toString())) {
                                 THEME_PLUGIN_JARS.add(url);
                             }
                         }
                     }
                     loadFromManifest(m, pluginsClassLoader);
+                } catch (ClassNotFoundException e) {
```
miurahr, line 133: A protection here is not perfect.
When target plugin is for theme and `loadFromManifest` raises exception,  `THEME_PLUGIN_JARS` contains the broken plugin. `if ("theme"....` block should be after `loadFromManifest()` to avoid the case.

t-cordonnier (commented):
looks good to me

amake:

PR #202, opened 2022-02-14
Refactor tmx entry part2

t-cordonnier:
This Pull Request is continuation of #185 

The role of these patches is to solve concerns mentioned by @miurahr about the role of PrepareTMXEntry class
This is of course one possible solution, not exclusive.

While patches from #185 were here for ensure plugins compatibility even in case of later changes inside OmegaT, the last patches here are known to break backward compatibility (see https://sourceforge.net/p/omegat/mailman/message/37593315/)
For that reason this pull request remains as a draft.

### DIFF
--- a/src/org/omegat/Main.java
+++ b/src/org/omegat/Main.java
@@ -61,6 +61,7 @@ import org.omegat.core.CoreEvents;
 import org.omegat.core.data.NotLoadedProject;
 import org.omegat.core.data.PrepareTMXEntry;
 import org.omegat.core.data.ProjectProperties;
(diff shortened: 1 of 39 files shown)


miurahr:
Progress?

t-cordonnier:
No, I do not work on this right now. I had divided my work in two PR, last time, precisely to have the possibility to postpone this second part.

brandelune:
@t-cordonnier Was the first part that you mention merged?

brandelune:
Do we set a milestone here ?

miurahr:
No need.

miurahr:
stalled.

PR #203, opened 2022-02-14
Deprecate JsonParser

miurahr:


amake:
Thanks. I've merged to `master`.

Note that you don't need to remove the unit test. You simply need to deprecate any methods that call deprecated methods.

PR #205, opened 2022-02-16
MT pane: sorted by translator name and highlight selected translation(RFE#1477 RFE#1467)

miurahr:
Implement [RFE#1477  Notify which Machine Translation was picked](https://sourceforge.net/p/omegat/feature-requests/1477/) and [RFE#1467 : always sort translations order by MT engine name](https://sourceforge.net/p/omegat/feature-requests/1467/)

- Change MachineTranslationPane to show contents with HTML editor kit.
- Add hightlight for selected translation text.
- Cyclic get translations that order is based on translator name.
  for example, when use IBMWatson and DeepL for MT, first Ctrl-M take a result of DeepL then IBMWatson.


src/org/omegat/gui/exttrans/MachineTranslateTextArea.java
```diff
@@ -105,17 +115,54 @@ public MachineTranslateTextArea(IMainWindow mw) {
...
+        if (displayed.isEmpty()) {
             return null;
         }
-        return cycle.next().result;
+        if ((cycle > displayed.size() - 1)) {
+            cycle = 0;
+        }
+        MachineTranslationInfo info = displayed.get(cycle);
+        flashSelected();
+        cycle++;
```
amake, line 147: I think "cycle" is a bit cryptic. Would "selectedIndex" or something be better?

Also a less verbose, common idiom for cycling an index would be

```
selectedIndex = (selectedIndex + 1) % displayed.size();
```

src/org/omegat/gui/exttrans/MachineTranslateTextArea.java
```diff
@@ -149,16 +196,25 @@ protected void setFoundResult(final SourceTextEntry se, final MachineTranslation
         UIThreadsUtil.mustBeSwingThread();
 
         if (data != null && data.result != null) {
-            displayed.add (data);
-            setText(getText() + data.result + "\n<" + data.translatorName + ">\n\n");
+            displayed.add(data);
+            StringBuilder sb = new StringBuilder();
+            for (int i = 0; i < displayed.size(); i++) {
```
amake, line 201: If this method can only be called sequentially on the main thread, then I guess it's OK to mutate and iterate `displayed` like this, but it makes me nervous.

Then if using `displayed` in this way is OK, it seems like it should be OK to append each incoming result to the pane instead of rebuilding the text each time.

miurahr: I think `displayed` can be sorted by `translatorName` and always display translations in same order in future.
[RFE#1467 Machine translation results sortin ](https://sourceforge.net/p/omegat/feature-requests/1467/) request this.
I'd like to try it now.

miurahr: When we move to Java 11, we can use SortedList here.

miurahr: > Then if using displayed in this way is OK, it seems like it should be OK to append each incoming result to the pane instead of rebuilding the text each time.

I'd like to try the idea adding data without rebuilding the text each time.

miurahr: It is not a really optimal and safe to inserting content by ordering. It can be like as follows changing `displayd = new TreeSet`
```java
            displayed.add(data);
            // build content
            StringBuilder sb = new StringBuilder();
            sb.append("<div id=\"").append(data.translatorName).append("\">");
            sb.append(data.result);
            sb.append("<div class=\"engine\">&lt;");
            sb.append(data.translatorName);
            sb.append("&gt;</div></div>");
            //
            HTMLDocument doc = (HTMLDocument) getDocument();
            HTMLEditorKit kit = (HTMLEditorKit) getEditorKit();
            MachineTranslationInfo previous = displayed.lower(data);
            try {
                if (previous == null) {
                    kit.insertHTML(doc, 0,  sb.toString(), 0, 0, HTML.Tag.DIV);
                } else {
                    Element el = doc.getElement(previous.translatorName);
                    kit.insertHTML(doc, el.getEndOffset(), sb.toString(), 0, 0, HTML.Tag.DIV);
                }
            } catch (BadLocationException | IOException e) {
                Log.log(e);
            }
```
But `translatorName` is not unique, because multiple Ctrl-M produce multiple results from same translator, logic is complex and difficult to handle exceptions. @amake  any suggestions?

miurahr: > If this method can only be called sequentially on the main thread, then I guess it's OK to mutate and iterate `displayed` like this, but it makes me nervous.

This method is guarded by `UIThreadsUtil.mustBeSwingThread();` Is it not enough?

src/org/omegat/gui/exttrans/MachineTranslateTextArea.java
```diff
@@ -68,20 +78,31 @@
...
-    protected Set<MachineTranslationInfo> displayed = new HashSet<>();
+    /**
+     *  displayed hold entries.
+     *  <p>
+     *  It keeps the order by translatorName, MachineTranslationInfo is defined as Comparable&le;E&gt; that
+     *  prioritize translatorName. {@see MachineTranslationInfo}
+     *  A value should be unique and used for identify HTML element by id attribute.
+     *  We use an index counter for identifier of elements.
+     */
+    protected Map<MachineTranslationInfo, String> displayed = new TreeMap<>();
```
amake, line 97: I missed this before, but if you use `MachineTranslationInfo` as a key in a Set or Map then it needs to implement `equals` and `hashCode`.

src/org/omegat/gui/exttrans/MachineTranslationInfo.java
```diff
@@ -32,12 +32,26 @@
...
+     * Returns a negative integer, zero, or a positive integer
+     * as this object is less than, equal to, or greater than the specified object.
+     */
+    @Override
+    public int compareTo(final MachineTranslationInfo o) {
+        int compare = this.translatorName.compareTo(o.translatorName);
+        if (compare == 0) {
+            return this.result.compareTo(o.result);
+        }
+        return compare;
```
amake, line 55: I feel like usually this is done with the comparison flipped:

```suggestion
        int compare = translatorName.compareTo(o.translatorName);
        if (compare != 0) {
            return compare;
        }
        return result.compareTo(o.result);
```

(Also you don't need `this` for the current object's members since there is no shadowing.)

src/org/omegat/gui/exttrans/MachineTranslateTextArea.java
```diff
@@ -147,18 +205,34 @@ private void startSearchThread(SourceTextEntry newEntry, boolean force) {
     @Override
     protected void setFoundResult(final SourceTextEntry se, final MachineTranslationInfo data) {
         UIThreadsUtil.mustBeSwingThread();
-
         if (data != null && data.result != null) {
-            displayed.add (data);
-            setText(getText() + data.result + "\n<" + data.translatorName + ">\n\n");
+            String indexString = String.format("%d", index);
```
amake, line 209: This is an extremely expensive way of stringifying an int (`String#valueOf` would be better), and it's not even necessary: you can simply do `StringBuilder#append` with the int and the result will be the same.

src/org/omegat/gui/exttrans/MachineTranslateTextArea.java
```diff
@@ -147,18 +205,34 @@ private void startSearchThread(SourceTextEntry newEntry, boolean force) {
...
+            sb.append(data.result);
+            sb.append("<div class=\"engine\">&lt;");
+            sb.append(data.translatorName);
+            sb.append("&gt;</div></div>");
+            try {
+                HTMLDocument doc = (HTMLDocument) getDocument();
+                HTMLEditorKit kit = (HTMLEditorKit) getEditorKit();
+                kit.insertHTML(doc, doc.getLength(), sb.toString(), 0, 0, null);
+                displayed.put(data, indexString);
+                index++;
```
amake, line 222: I don't like having `index` as separate state in the instance. If I understand correctly, `index` will always be the same as `displayed.size()`.

In fact you don't need `displayed` to be a Map at all if the value is just the stringified index. You could just use a List and in `flashSelected` use `List#indexOf`. The list will almost always be so small that a linear search is plenty fast enough, so there is no need for a Set or Map.

src/org/omegat/gui/exttrans/MachineTranslateTextArea.java
```diff
@@ -105,17 +126,54 @@ public MachineTranslateTextArea(IMainWindow mw) {
...
         if (!cycle.hasNext()) { // only possible if displayed.isEmpty()
             return null;
         }
-        return cycle.next().result;
+        MachineTranslationInfo info = cycle.next();
+        flashSelected(info);
+        return info.result;
+    }
+
+    private void flashSelected(final MachineTranslationInfo info) {
```
amake, line 161: Does this actually *flash* (点滅)? Or does it merely *highlight*?

src/org/omegat/gui/exttrans/MachineTranslateTextArea.java
miurahr: Adjusted.

src/org/omegat/gui/exttrans/MachineTranslateTextArea.java
miurahr: It does highlight. I'd like to change a method name.

src/org/omegat/gui/exttrans/MachineTranslateTextArea.java
miurahr: Now use `insertHTML` instead of rebuilding the text each time.
``` 
kit.insertHTML(doc, doc.getLength(), sb.toString(), 0, 0, null);
```

src/org/omegat/gui/exttrans/MachineTranslateTextArea.java
```diff
@@ -94,7 +94,7 @@
      *  A value should be unique and used for identify HTML element by id attribute.
      *  We use an index counter for identifier of elements.
      */
-    protected Map<MachineTranslationInfo, String> displayed = new TreeMap<>();
+    protected List<MachineTranslationInfo> displayed = new ArrayList<>();
     private int index;
```
amake, line 98: I don't think you need `index` at all. `index` is always equal to `displayed.length()`.

src/org/omegat/Bundle.properties
```diff
@@ -2722,9 +2723,9 @@ ALIGNER_PANEL_LABEL_COUNTER=Counter:
 ALIGNER_PANEL_LABEL_COMPARISON=Comparison Mode:
 ALIGNER_PANEL_LABEL_AVGSCORE=Average Score: {0}
 # BLACK UP-POINTING TRIANGLE
-ALIGNER_PANEL_BUTTON_MOVEUP=\u25b2
+ALIGNER_PANEL_BUTTON_MOVEUP=\u25B2
```
amake, line 2726: What are these capitalization changes in Unicode escapes? Did a tool do this automatically?

miurahr: A tool automatically do that.
I've pushed again without modifications.

src/org/omegat/gui/exttrans/MachineTranslateTextArea.java
```diff
@@ -68,20 +78,30 @@
...
     private static final String EXPLANATION = OStrings.getString("GUI_MACHINETRANSLATESWINDOW_explanation");
 
-    protected Set<MachineTranslationInfo> displayed = new HashSet<>();
+    /**
+     *  displayed hold entries.
+     *  <p>
+     *  It keeps the order by translatorName, MachineTranslationInfo is defined as Comparable&le;E&gt; that
+     *  prioritize translatorName. {@see MachineTranslationInfo}
+     *  A value should be unique and used for identify HTML element by id attribute.
+     *  We use an index counter for identifier of elements.
```
miurahr, line 95: It should be updated.
The PR does not sort an order of entries.

src/org/omegat/gui/exttrans/MachineTranslateTextArea.java
```diff
@@ -147,16 +205,58 @@ private void startSearchThread(SourceTextEntry newEntry, boolean force) {
...
+            StringBuilder sb = new StringBuilder();
+            sb.append("<div id=\"").append(index).append("\">");
+            sb.append(data.result);
+            sb.append("<div class=\"engine\">&lt;");
+            sb.append(data.translatorName);
+            sb.append("&gt;</div></div>");
+            try {
+                insertElement(data, sb.toString());
+            } catch (BadLocationException | IOException e) {
+                Log.log(e);
```
miurahr, line 219: When exception happened, field `displayed` become inconsistent with actual html. Should update `displayed` when all are well.

miurahr:
now ready to review again.

src/org/omegat/gui/exttrans/MachineTranslateTextArea.java
miurahr: I've removed a field `index`.  A method ` setFoundResult` get local variable `int index = displayed.size()`

src/org/omegat/gui/exttrans/MachineTranslationInfo.java
miurahr: I've accepted the proposal.
These `hashCode` and `compareTo` were generated by IDE, intelliJ.

src/org/omegat/gui/exttrans/MachineTranslateTextArea.java
miurahr: I change here to use `StringBuilder#append`.

miurahr:
I changed a line from `Element element = doc.getElement(String.valueOf(index));` to
`Element element = doc.getElement(rootElement, HTML.Attribute.ID,  String.valueOf(index));`

 [javadoc of `HtmlDocument#getElement` ](https://docs.oracle.com/javase/8/docs/api/javax/swing/text/html/HTMLDocument.html#getElement-java.lang.String-)  explains

> Returns the element that has the given id Attribute. If the element can't be found, null is returned. This is a convenience method for getElement(RootElement, HTML.Attribute.id, id). This is not thread-safe.

I use the latter syntax to express explicitly use `id` attribute to get element, which attribute `id` set in 

```
`protected void setFoundResult(final SourceTextEntry se, final MachineTranslationInfo data) {
        UIThreadsUtil.mustBeSwingThread();
        if (data != null && data.result != null) {
            int index = displayed.size();
            StringBuilder sb = new StringBuilder();
            sb.append("<div id=\"").append(index).append("\">");
```

Because `getElement` is not thread-safe, a logic is guarded with `UIThreadsUtil.mustBeSwingThread();`

src/org/omegat/gui/exttrans/MachineTranslateTextArea.java
miurahr: https://github.com/omegat-org/omegat/pull/205/commits/8ba52eda43a2310d13bd29440214c52c6b8dfe92  fix it.

src/org/omegat/gui/exttrans/MachineTranslateTextArea.java
miurahr: Sort feature is implemented again at https://github.com/omegat-org/omegat/pull/205/commits/f606147e812e3a89e57ea174415a11f65d9f8398

miurahr:
Note: this is key point to cycle selection order in sorted  L#148
```            cycle = displayed.stream().sorted().iterator();```
An `id` and order in `displayed` is in order of arrival, but selection cycle is in sorted order.

File: src/org/omegat/gui/exttrans/MachineTranslateTextArea.java [hunk 12]
```diff
@@ -105,17 +122,58 @@ public MachineTranslateTextArea(IMainWindow mw) {
...
 
+    private void initDocument() {
+        StyleSheet baseStyleSheet = new StyleSheet();
+        HTMLEditorKit htmlEditorKit = new HTMLEditorKit();
+        baseStyleSheet.addStyleSheet(htmlEditorKit.getStyleSheet()); // Add default styles
+        Font font = getFont();
+        baseStyleSheet.addRule("body { font-family: " + font.getName() + "; "
+                + " font-size: " + font.getSize() + "; "
+                + " font-style: " + (font.getStyle() == Font.BOLD ? "bold"
+                        : font.getStyle() == Font.ITALIC ? "italic" : "normal") + "; "
```
amake, line 134: `font-style: bold` is not a thing; it's `font-weight: bold`.

amake: I see DictionariesTextArea currently uses `font-style: bold`. Maybe that's a special feature of Swing HTML?

amake:
This is actually quite complex now and I think there are some big issues.

1. Retaining an `Iterator` to indicate the selected item is an unusual approach. It is equivalent to, but much more heavyweight than, simply keeping an int index that cycles through the available list items. I would prefer a simple index.
2. The update mechanism in `insertElement` is complex and error-prone. Instead I would much prefer a solution where the single source of truth is the `displayed` list (if it's to be sorted, then it should always be sorted), and the pane content is completely rebuilt every time it changes.

    Only if that approach has some major downside (flashing on update?) should we try to incrementally update the content.

src/org/omegat/gui/exttrans/MachineTranslateTextArea.java
amake: I double-checked with the following Groovy program and I was right: `font-style: bold` has no effect.

```groovy
import javax.swing.text.html.HTMLEditorKit;
import javax.swing.text.html.StyleSheet;
import javax.swing.JTextPane;
import javax.swing.JFrame;

def pane = new JTextPane();

def stylesheet = new StyleSheet();
def editorKit = new HTMLEditorKit();
stylesheet.addRule('body { font-style: bold }')
editorKit.setStyleSheet(stylesheet);
pane.setEditorKit(editorKit);
pane.setText('foo');

def frame = new JFrame();
frame.add(pane);
frame.pack();
frame.show();
```

miurahr:
> This is actually quite complex now and I think there are some big issues.
> 
> 1. Retaining an `Iterator` to indicate the selected item is an unusual approach. It is equivalent to, but much more heavyweight than, simply keeping an int index that cycles through the available list items. I would prefer a simple index.
> 2. The update mechanism in `insertElement` is complex and error-prone. Instead I would much prefer a solution where the single source of truth is the `displayed` list (if it's to be sorted, then it should always be sorted), and the pane content is completely rebuilt every time it changes.
>    Only if that approach has some major downside (flashing on update?) should we try to incrementally update the content.

These are basically equivalent with my previous approach. 
I'd like to revert changes and update some.

miurahr:
Now updated and I believe it is ready to merge.

src/org/omegat/gui/exttrans/MachineTranslateTextArea.java
miurahr: I've updated a stylesheet

miurahr:
I think now all comments are adjusted.

miurahr:
Squashed changes to prepare for merge.

miurahr:
Squashed changes to prepare for merge.

src/org/omegat/gui/exttrans/MachineTranslateTextArea.java
```diff
@@ -68,20 +77,26 @@
...
 @SuppressWarnings("serial")
 public class MachineTranslateTextArea extends EntryInfoThreadPane<MachineTranslationInfo>
         implements IPaneMenu {
 
     private static final String EXPLANATION = OStrings.getString("GUI_MACHINETRANSLATESWINDOW_explanation");
 
-    protected Set<MachineTranslationInfo> displayed = new HashSet<>();
+    /**
+     *  List displayed hold entries. An index shall be as same as ID attribute value of HTML.
+     *  Actual displayed entries are sorted, and the order is different from the List.
```
amake, line 90: I think the second line of this comment is no longer true.

src/org/omegat/gui/exttrans/MachineTranslateTextArea.java
```diff
@@ -147,18 +201,29 @@ private void startSearchThread(SourceTextEntry newEntry, boolean force) {
     @Override
     protected void setFoundResult(final SourceTextEntry se, final MachineTranslationInfo data) {
         UIThreadsUtil.mustBeSwingThread();
-
         if (data != null && data.result != null) {
-            displayed.add (data);
-            setText(getText() + data.result + "\n<" + data.translatorName + ">\n\n");
+            displayed.add(data);
+            displayed.sort(Comparator.comparing(info -> info.translatorName));
```
amake, line 206: Implementing Comparable on MachineTranslationInfo suggests you have to define comparability over all members, but that doesn't make a lot of sense (why should the translation itself be sorted lexicographically?).

Since we only care about sorting by translator name, I used this formulation and removed Comparable.

amake:
I pushed a couple changes. Please make sure they work for you (sorry I can't test them very well; I don't have MT connectors set up right now).

If it works for you, feel free to squash to `master`.

miurahr:
These works fine except for a selection order when  engines are equals or more than 2.
I've pushed the change.

miurahr:
I've squashed and merged into [master](https://github.com/omegat-org/omegat/commit/1aad877e8b1875d10fa9499c72b3438581145637)

kosivantsov:
With this change you often get the Conflict Resolution dialog after you use one of the MT results, edit it and switch to the next segment or save. It slows down the translation pace, and for longer sentences it even creates additional problems as that dialog doesn't show you the whole segment.

PR #207, opened 2022-02-17
DictionaryPane: view slot for result display

miurahr:
Try to implement an idea posted to dev-ML

https://sourceforge.net/p/omegat/mailman/omegat-development/thread/e1e9995a-f6a4-690d-c2a8-c50df0edd3ca%40northside.tokyo/#msg37611260

There is no scroll event handler. Only when clicking source text word, redraw content. 

OmegaT_dictionary_pane_performance_improvement


miurahr:
I'd like to reopen when I find a way  working.

PR #208, opened 2022-02-20
Docs_devel: explain git merge strategy

miurahr:
- Explicitly express our preference: squash-merge or rebase-merge
- Add note that github record no-merge-close but it is not a problem


amake:
Thanks. I think it's good to document this, but I think it's very easy to overlook, and it's quite specific to the current workflow.

I think if we put it in a [PR template](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/creating-a-pull-request-template-for-your-repository) then the people who most need this information will be more likely to see it.

PR #209, opened 2022-02-26
Dictionary: condensed view

miurahr:
* Format articles based on content type
  - Support condensed view mode. Adding preference to set condensed view.
  - Allow to select custom font size for dictionary pane
* StarDict: format article in HTML
  - when text (mean and phonetic tag) article to format html tag
  - produce condensed and standard view
  - when html article to pass as is
* LingvoDSL
  - Produce condensed and standard view
  - Better handling of "*" tag
  - Produce "span" for "lang" tag
* Preference
  - Add condensed view option
  - Allow custom font size for dictionary pane

image\
image
condensed view / standard view


src/org/omegat/core/dictionaries/LingvoDSL.java
```diff
@@ -172,6 +180,9 @@ public void finish() {
          */
         @Override
         public void visit(final DslArticle.Tag tag) {
+            if (inDetails && Preferences.isPreferenceDefault(Preferences.DICTIONARY_CONDENSED_VIEW, false)) {
```
miurahr, line 183: It seems better to be get preference in constructor for consistency among processing.

miurahr:
I uses ▷ to indicate a start of paragraph,   but it would be better to use a pilcrow sign ¶ .

PR #211, opened 2022-02-26
Fully automate integration test as CI/CD friendly

miurahr:
This recipes make an integration test as fully automated and unattended, just one line command.

There are three docker configuration Dockerfile and one compose recipe.
compose create three containers.

1. keygen: generate ssh key pair to communicate test client and git server
2. server: git+ssh  and smart git https server for an identical repository,  ordinal http service for mapping. 
3. client: run integration test

There is also a developer memo how to run integration test .


miurahr:
Squashed and merged  into master.

PR #214, opened 2022-03-03
Github: Add pull request template

miurahr:
Add explanation for contributors to improve review process.

Adjust for #208 comment.


amake:
Thanks, I think this is quite useful. I've made some changes and applied to `master`.

PR #215, opened 2022-03-15
fix broken spellcheck dictionary download URL

brandelune:
I'm not sure how to create a PR that also works for "stable", which is really the place where we want to have this.


brandelune:
Closing here anf PRing from 4.3

PR #216, opened 2022-03-15
Fix #1086 bug

brandelune:


amake:
Thanks. I've applied this to `standard` with some changes.

PR #222, opened 2022-04-04
MT: fix conflict dialog when insert MT result

miurahr:
This fixes a degrade of 0961df01f2f97d9387b5e563bde84eda121fa61e that cause translation conflict dialog.

A design to hold origin is changed. Now EditorControll class keeps origin data of current edit entry, and write to TMX when commit.

## Pull request type

- [x] Bug fix
- [x] Feature
- [ ] Documentation
- [ ] Other (describe below)

## Which ticket is resolved?

- Link: https://github.com/omegat-org/omegat/pull/205#issuecomment-1087391240
- Title: 

## What does this PR change?

- A commit   0961df01f2f97d9387b5e563bde84eda121fa61e from  Pull-Request #193 cause translation conflict dialog when insert MT result.
- This fix the issue.

### DIFF
--- a/src/org/omegat/gui/editor/EditorController.java
+++ b/src/org/omegat/gui/editor/EditorController.java
@@ -117,6 +117,7 @@ import org.omegat.util.OStrings;
 import org.omegat.util.Preferences;
 import org.omegat.util.StaticUtils;
 import org.omegat.util.StringUtil;
+import org.omegat.util.TMXProp;
 import org.omegat.util.gui.DragTargetOverlay;
 import org.omegat.util.gui.DragTargetOverlay.IDropInfo;
 import org.omegat.util.gui.StaticUIUtils;
@@ -209,6 +210,9 @@ public class EditorController implements IEditor {
     private Component entriesFilterControlComponent;
 
     private SegmentExportImport segmentExportImport;
+
+    protected String currentEntryOrigin;
+    protected String translationFromOrigin;
 
     /**
      * Previous translations. Used for optimistic locking.
@@ -1183,6 +1187,12 @@ public class EditorController implements IEditor {
             } else {
                 // new translation is not empty and not equals to source - just change
                 newen.translation = newTrans;
+                if (currentEntryOrigin != null && newTrans.equals(translationFromOrigin)) {
+                    if (newen.otherProperties == null) {
+                        newen.otherProperties = new ArrayList<>();
+                    }
+                    newen.otherProperties.add(new TMXProp(ProjectTMX.PROP_ORIGIN, currentEntryOrigin));
+                }
             }
         }
 
@@ -1191,6 +1201,7 @@ public class EditorController implements IEditor {
         boolean isNewAltTrans = !defaultTranslation && oldTE.defaultTranslation;
         boolean translationChanged = !Objects.equals(oldTE.translation, newen.translation);
         boolean noteChanged = !StringUtil.nvl(oldTE.note, "").equals(StringUtil.nvl(newen.note, ""));
+        resetOrigin();
 
         if (!isNewAltTrans && !translationChanged && noteChanged) {
             // Only note was changed, and we are not making a new alt translation.
@@ -1295,6 +1306,7 @@ public class EditorController implements IEditor {
         }
 
         doc.stopEditMode();
+        resetOrigin();
     }
 
     /**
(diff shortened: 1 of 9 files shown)


kosivantsov:
Still getting the dialog. Here's a part of the log:
```31073: Info: Event: new entry activated (LOG_INFO_EVENT_ENTRY_ACTIVATED)
31073: FINER: Time for find matches: 3 
31073: Info: Click on 'editOverwriteMachineTranslationMenuItem' menu item (LOG_MENU_CLICK)
31073: Info: Event: project change - "MODIFIED" (LOG_INFO_EVENT_PROJECT_CHANGE)
31073: Info: Event: project change - "MODIFIED" (LOG_INFO_EVENT_PROJECT_CHANGE)
31073: Info: Event: project change - "MODIFIED" (LOG_INFO_EVENT_PROJECT_CHANGE)```
I don't know if it relevant, but should switching to a new segment trigger 3 project change events?

miurahr:
Thanks for feedback. It is caused by calling IProject.setTranlate() that call the trigger. I've reworked and fixed it.

t-cordonnier (commented):
Problem with a public method like setOrigin(entry, prevEntry, origin) is that it may later be called by somebody else, in another context, without any guarantee that prevEntry really contains the previous entry.

I propose as an alternative to rollback to original way (using eplaceEditText instead of setTranslation) and adding origin there, meaning:
- create method replaceEditText(text,origin)
- calling it in editOverwriteMachineTranslationMenuItemActionPerformed (if I look to the code before apply of PR #205, there was no call to setTranslation but to replaceEditText only)

Well, one difficulty will be that replaceEditText is in IEditor, not in IProject... so, you will probably need to keep origin in the editor and use it when editor finally launches setTranslation. To be tested.

miurahr:
@t-cordonnier  you may want to see #193,  not #205.

miurahr:
> Problem with a public method like setOrigin(entry, prevEntry, origin) is that it may later be called by somebody else, in another context, without any guarantee that prevEntry really contains the previous entry.

It's right.

 
> * create method replaceEditText(text,origin)
> * calling it in editOverwriteMachineTranslationMenuItemActionPerformed

Ok, I'd like to move this way.

kosivantsov:
I don't get the conflict resolution dialog anymore! From the user's perspective it's fixed now. Thank you

t-cordonnier:
> I don't get the conflict resolution dialog anymore! From the user's perspective it's fixed now. Thank you

See my remark before, the behavior of menu "Overwrite with machine translation" changed: in the past, this method modified current segment _in the editor_, without validating it. New method changes in _project memory_ and _validates_ it. Not sure it is the expected result.

kosivantsov:
>  in the past, this method modified current segment in the editor, without validating it. New method changes in project memory and validates it. Not sure it is the expected result.

From the user's perspective there's very little difference. When you exit the segment, it's validated anyway, even if you don't save. It may not be written to the `project_save.tmx` immediately, but there's no way to have such segment unsaved other than killing the app or switching back to it and removing the translation.

miurahr:
> > * create method replaceEditText(text,origin)
> > * calling it in editOverwriteMachineTranslationMenuItemActionPerformed
> 
> Ok, I'd like to move this way.

I cannot find an instant solution to implement it. I'd like to look into a design of IEditor and related classes in next few weeks/months.

> > Problem with a public method like setOrigin(entry, prevEntry, origin) is that it may later be called by somebody else, in another context, without any guarantee that prevEntry really contains the previous entry.
> 
> It's right.

I'd like to work on it for the short term solution.

t-cordonnier:
> When you exit the segment, it's validated anyway, even if you don't save. 

Yes but the user can change the segment beween the moment he types CTRL+M and the moment when he leaves the segment.
And we have this problem of conflict which I interpret as the fact that OmegaT tries to validate the segment twice - one because of the new behavior of CTRL+M and one because OmegaT always validates the segment before changing it...

> I'd like to work on it for the short term solution.

In short terms why not, but don't forget that this is not only a question of design, IEditor.replaceText is definitively not equivalent to Project.saveTranslation

miurahr:
> > > * create method replaceEditText(text,origin)
> > > * calling it in editOverwriteMachineTranslationMenuItemActionPerformed
> > 
> > 
> > Ok, I'd like to move this way.
> 
> I cannot find an instant solution to implement it. I'd like to look into a design of IEditor and related classes in next few weeks/months.

I don't understand well but it may work now. Could you review and test it?

t-cordonnier (approved):
Seems ok for me now. I did a small test, CTRL+M inserts the origin correctly, then if I modify the text between CTRL+M and validation the origin disappears, which is what I expected.
LGTM

miurahr:
@kosivantsov @t-cordonnier  I think it now work as expected, Thanks for review.
I'll squash and clean-up code soon.

miurahr:
I've merged it to 71e9bf26

PR #226, opened 2022-04-10
[#1548] Install plugins from remote repository

miurahr:

## Pull request type

- [ ] Bug fix
- [ ] Feature
- [ ] Documentation
- [ ] Other (describe below)

## Which ticket is resolved?

- Link: 
- Title: 

## What does this PR change?

-
-
-


amake:
Please don't merge this until I approve. This gets us into an entirely new area for the project (hosting a directory of plugins) so I want to take a thorough look before we commit to it.

brandelune:
Is there a reason why we don't have a milestone anymore?

miurahr:
We need to have a discussion how to manage 3rd party plugin registry in core team or community,  how to be a database format of 3rd party plugins, how to accept an addition/modification/removal request,  how to decide accept or decline, how to remove an item from the list, etc....

brandelune:
Sure, but we can have that feature included in the code, and then start the discussion based on the feature requirements, don't you think so?

Is that what is keeping you from "targeting" 6.1?

miurahr:
> Sure, but we can have that feature included in the code, and then start the discussion based on the feature requirements, don't you think so?

There is no consensus with you and core team to have an obligation of maintains certified plugin database on the web.  
There is also no discussion how we should have a database data format and its work flows.

Once the code included, above is forced to you with the assumptions of the code.

OK, PLEASE maintain the database by @brandelune !? Is it what you want?  


@amake said
> This gets us into an entirely new area for the project (hosting a directory of plugins) so I want to take a thorough look before we commit to it.

miurahr:
This is a sample how we can manage plugin database

1. Pull-Request from 3rd party plugin author with Manifest data. 
https://github.com/omegat-org/omegat-plugins/pull/24/files

2. When approved the request, automatically compile database with github actions
https://github.com/omegat-org/omegat-plugins/actions/runs/2180165236/jobs/3199284035

3. Up-to-date database is on github releases
https://github.com/omegat-org/omegat-plugins/releases/tag/continuous-release

4. a database is like this
https://github.com/omegat-org/omegat-plugins/blob/continuous-release/plugins.MF.sample

miurahr:
> 4\. a database is like this
>     https://github.com/omegat-org/omegat-plugins/blob/continuous-release/plugins.MF.sample

This is too primitive and there is no sign by omegat core team.
It is better to be JSON or XML with signature.

miurahr:
Because we publish Current OmegaT versions on OmegaT web,
https://omegat.sourceforege.io/Version-latest.properties, it may be possible to publish plugin list like https://omegat.sourceforge.io/plugins-latest.xml

brandelune:
Ok, let's have this discussion on the list.

https://sourceforge.net/p/omegat/mailman/message/37746064/

miurahr:
OkAPI filters plugin for OmegaT 1.13-1.45.0 is released with support for new Plugin configuration UI.
image

It looks better for management.

miurahr:
Are there any good way to check downloaded jar  file is not malformed using signature file jar.asc?
In current proposed procedure ask plugin developer to register their GnuPG/OpenPGP public key in GitHub account. 
Is it possible for OmegaT to detect signer account and download a public key?

miurahr:
Now "description" field is not localized.  There is no specification how to define localized description in MANIFEST.
Does anyone have idea how to realize it?

src/org/omegat/Bundle.properties
```diff
@@ -3020,14 +3020,34 @@ PREFS_SECURE_STORAGE_RESET_MASTER_PASSWORD_MESSAGE=Are you sure you want to rese
 # PREFERENCES - PLUGINS
 PREFS_TITLE_PLUGINS=Plugins
 
+# org.omegat.gui.dialogs.PluginInstallerDialog
+PREFS_PLUGINS_DETAILS_explanation=Show plugin details
+PREFS_PLUGINS_COL_STAT=Status
+PREFS_PLUGINS_COL_CATEGORY=Category
+PREFS_PLUGINS_COL_NAME=Name
+PREFS_PLUGINS_COL_VERSION=Version
+PREFS_PLUGINS_UPGRADE=Upgrade
```
Kazephil, line 3029: I think update is better than upgrade:
- PREFS_PLUGIN_UPDATE=Update

src/org/omegat/Bundle.properties
```diff
@@ -3020,14 +3020,34 @@ PREFS_SECURE_STORAGE_RESET_MASTER_PASSWORD_MESSAGE=Are you sure you want to rese
...
+PREFS_PLUGINS_INSTALL=Install
+
+PLUGIN_STATUS_BUNDLED=Bundled
+PLUGIN_STATUS_UPGRADABLE=Upgradable
```
Kazephil, line 3033: Following from the above:
- PLUGIN_STATUS_UPDATABLE=Update available

Updatable is better for the key name, but "update available" is more natural for users.

Kazephil:
More general comment: this looks much better than the current plugin UI.

One question, though. How is the initial order of the plugins shown determined? It seems completely random, but it would probably be more user friendly to sort them by either category or name at first, until the user changes the sort order.

miurahr:
> A more general comment: this looks much better than the current plugin UI.
> 
> One question, though. How is the initial order of the plugins shown determined? It seems completely random, but it would probably be more user friendly to sort them by either category or name at first, until the user changes the sort order.

The sort order is as same as current OmegaT. A order of class path.

miurahr:
> > Now "description" field is not localized. There is no specification how to define localized description in MANIFEST. Does anyone have idea how to realize it?
> 
> That might be a good question for the development list. I unfortunately don't know much about MANIFEST files, so I can't make any relevant suggestions.

In Eclipse, plugin MANIFEST is defined in XML form as a file, `plugin.xml`
https://wiki.eclipse.org/FAQ_What_is_the_plug-in_manifest_file_(plugin.xml)%3F

It also has an "externalize" string for localize such as "name=%plugin.name" and there is "fr.properties" which defines like "%plugin.name=Bonjour plugin" .

We can take a similar way, change MANIFEST.MF like "Plugin-Name: %name" and "Plugin-Description: %description", then externalize these strings in properties file such as "META-INF/fr.properties"

brandelune:
Documentation ticket:
https://sourceforge.net/p/omegat/documentation/439/

PR #229, opened 2022-06-08
EditorController: fix paragraph marks not showing when loading partial pages

Nutzer:
This fixes a bug where paragraph start markers will not show when OmegaT partially loads large documents (when scrolling up or down to load new segments), [Here](https://github.com/omegat-org/omegat/files/8863884/test.txt) is a test file.

The problem exists because `insertStartParagraphMark` is not beeing called in `loadUp()`,
It is also not setting `trustedChangesInProgress` before inserting the paragraph mark when called from `loadDown()`.

## Pull request type

- [x] Bug fix
- [ ] Feature
- [ ] Documentation
- [ ] Other (describe below)

## Which ticket is resolved?

N/A

## What does this PR change?

- EditorController:
  - Add document offset to insertStartParagraphMark to allow inserting
    at any place in the document
  - Set trustedChangesInProgress and unset caret update when inserting
    a paragraph mark so that the document listenes for changes when
    loading a partial page
  - Insert pargagraphs marks in loadUp

### DIFF
--- a/src/org/omegat/gui/editor/EditorController.java
+++ b/src/org/omegat/gui/editor/EditorController.java
@@ -404,9 +404,10 @@ public class EditorController implements IEditor {
         }
         int loadFrom = lastLoaded + 1;
         int loadTo = Math.min(m_docSegList.length - 1, loadFrom + count - 1);
+        Document3 doc = editor.getOmDocument();
         for (int i = loadFrom; i <= loadTo; i++) {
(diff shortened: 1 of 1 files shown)


brandelune:
It would be nice if you could create a ticket in SourceForge for reference here because that ticket will be referenced in the changes.txt file in the end.

Nutzer:
That's fine. The ticket is bug #1102

PR #32, opened 2019-01-27
[RFE#1408] mode console align multiples translations

keko:
https://sourceforge.net/p/omegat/feature-requests/1408/

PR #230, opened 2022-07-05
[RFE#1408] CLI: new option in mode=console-align

miurahr:
## Pull request type

- [ ] Bug fix
- [x] Feature
- [ ] Documentation
- [x] Other (describe below)

Change ``AbstractFilter#alignFile`` signature.

## Which ticket is resolved?

- Link: https://sourceforge.net/p/omegat/feature-requests/1408/
- Title: In mode=console-align, generate tmx with mutiples translations

## What does this PR change?

- Produce TMX that have same source string with multiples translations
- Change TMXWriter2#writeEntries to accept Map<EntryKey, TMXEntry> data.
- Update RealProject#align to return Map<EntryKey, TMXEntry>
- Update align to change behavior according to project configuration.

Merged the proposal #32 onto current master and adjust to current TMXEntry definition.

### DIFF
--- a/src/org/omegat/core/data/RealProject.java
+++ b/src/org/omegat/core/data/RealProject.java
@@ -12,6 +12,7 @@
                2014 Aaron Madlon-Kay, Didier Briel
                2015 Aaron Madlon-Kay
                2017-2018 Didier Briel
+               2018 Enrique Estevez Fernandez
                2019 Thomas Cordonnier
                2020 Briac Pilpre
                Home page: http://www.omegat.org/
@@ -62,6 +63,8 @@ import java.util.stream.Collectors;
 
 import org.madlonkay.supertmxmerge.StmProperties;
 import org.madlonkay.supertmxmerge.SuperTmxMerge;
+import org.xml.sax.SAXParseException;
+
 import org.omegat.CLIParameters;
 import org.omegat.core.Core;
 import org.omegat.core.CoreEvents;
@@ -102,7 +105,6 @@ import org.omegat.util.StringUtil;
 import org.omegat.util.TMXReader2;
 import org.omegat.util.TagUtil;
 import org.omegat.util.gui.UIThreadsUtil;
-import org.xml.sax.SAXParseException;
 
 import gen.core.filters.Filters;
 
@@ -443,7 +445,7 @@ public class RealProject implements IProject {
     /**
      * Align project.
      */
-    public Map<String, TMXEntry> align(final ProjectProperties props, final File translatedDir)
+    public Map<EntryKey, ITMXEntry> align(final ProjectProperties props, final File translatedDir)
             throws Exception {
         FilterMaster fm = Core.getFilterMaster();
 
@@ -456,8 +458,7 @@ public class RealProject implements IProject {
         for (File file : srcFileList) {
             // shorten filename to that which is relative to src root
             String midName = file.getPath().substring(srcRoot.length());
-
-            fm.alignFile(srcRoot, midName, translatedDir.getPath(), new FilterContext(props),
+            fm.alignFile(root.toString(), midName, translatedDir.getPath(), new FilterContext(props),
                     alignFilesCallback);
         }
         return alignFilesCallback.data;
@@ -1820,17 +1821,24 @@ public class RealProject implements IProject {
             this.config = props;
         }
 
-        Map<String, TMXEntry> data = new TreeMap<>();
-        private ProjectProperties config;
+        Map<EntryKey, ITMXEntry> data = new TreeMap<>();
+        private final ProjectProperties config;
+        List<String> sources = new ArrayList<>();
 
         @Override
-        public void addTranslation(String id, String source, String translation, boolean isFuzzy, String path,
+        public void addTranslation(String id, String source, String translation, boolean isFuzzy, String sourcePath,
                 IFilter filter) {
             if (source != null && translation != null) {
                 ParseEntry.ParseEntryResult spr = new ParseEntry.ParseEntryResult();
                 boolean removeSpaces = Core.getFilterMaster().getConfig().isRemoveSpacesNonseg();
                 String sourceS = ParseEntry.stripSomeChars(source, spr, config.isRemoveTags(), removeSpaces);
                 String transS = ParseEntry.stripSomeChars(translation, spr, config.isRemoveTags(), removeSpaces);
+                if (config.isSupportDefaultTranslations()) {
+                    if (sources.contains(sourceS)) {
+                        return;
+                    }
+                }
+                sources.add(sourceS);
 
                 PrepareTMXEntry tr = new PrepareTMXEntry();
                 if (config.isSentenceSegmentingEnabled()) {
@@ -1844,7 +1852,7 @@ public class RealProject implements IProject {
                         }
                         tr.source = sourceS;
                         tr.translation = transS;
-                        data.put(sourceS, new TMXEntry(tr, true, null));
+                        data.put(new EntryKey(sourcePath, sourceS, id, "", "", ""), tr);
                     } else {
                         for (short i = 0; i < segmentsSource.size(); i++) {
                             String oneSrc = segmentsSource.get(i);
@@ -1854,7 +1862,7 @@ public class RealProject implements IProject {
                             }
                             tr.source = oneSrc;
                             tr.translation = oneTrans;
-                            data.put(sourceS, new TMXEntry(tr, true, null));
+                            data.put(new EntryKey(sourcePath, oneSrc, id, "", "", ""), tr);
                         }
                     }
                 } else {
@@ -1863,7 +1871,7 @@ public class RealProject implements IProject {
                     }
                     tr.source = sourceS;
                     tr.translation = transS;
-                    data.put(sourceS, new TMXEntry(tr, true, null));
+                    data.put(new EntryKey(sourcePath, sourceS, id, "", "", ""), tr);
                 }
             }
         }
--- a/src/org/omegat/filters2/moodlephp/MoodlePHPFilter.java
+++ b/src/org/omegat/filters2/moodlephp/MoodlePHPFilter.java
@@ -167,10 +167,10 @@ public class MoodlePHPFilter extends AbstractFilter {
     }
 
     @Override
-    protected void alignFile(BufferedReader sourceFile, BufferedReader translatedFile, FilterContext fc)
-            throws Exception {
-        Map<String, String> source = new HashMap<String, String>();
-        Map<String, String> translated = new HashMap<String, String>();
+    protected void alignFile(BufferedReader sourceFile, BufferedReader translatedFile, FilterContext fc,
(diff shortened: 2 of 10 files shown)


src/org/omegat/core/data/RealProject.java
```diff
@@ -463,6 +465,39 @@ public Map<String, TMXEntry> align(final ProjectProperties props, final File tra
...
+        Map<EntryKey, TMXEntry> auxiliar = new HashMap<EntryKey, TMXEntry>();
+        String srcRoot = config.getSourceRoot();
+        for (File file : srcFileList) {
+            // shorten filename to that which is relative to src root
+            String midName = file.getPath().substring(srcRoot.length());
+            fm.alignFile(srcRoot, midName, translatedDir.getPath(), new FilterContext(props),
+                    alignFilesCallback);
+            // I did not know how modify the entries into the hash (missing filename)
+            // I generate the new the entries, remove the old and add the new
+            for (Map.Entry<EntryKey, ? extends TMXEntry> entry : alignFilesCallback.multiples.entrySet()) {
```
miurahr, line 488: should be  ``Map.Entry<EntryKey, ? extends ITMXEntry> entry``

miurahr:
@keko could you test your case?

src/org/omegat/core/data/RealProject.java
```diff
@@ -463,6 +465,39 @@ public Map<String, TMXEntry> align(final ProjectProperties props, final File tra
         return alignFilesCallback.data;
     }
 
+    /**
+     * Alternative Align project.
+     */
+    public Map<EntryKey, TMXEntry> alignAlt(final ProjectProperties props, final File translatedDir)
```
miurahr, line 471: It is better to consolidate function to single one with flag for behavior, because ``alignAlt`` and ``align`` share similar function.

src/org/omegat/util/TMXWriter2.java
```diff
@@ -163,6 +165,24 @@ public void writeEntries(final Map<String, ? extends ITMXEntry> entries) throws
         }
     }
 
+    /**
+     * Write entries with alternative translations.
+     * @param entries Map of EntryKey and TMXEntry to output.
+     */
+    public void writeEntriesAlt(final Map<EntryKey, ? extends ITMXEntry> entries) throws Exception {
```
miurahr, line 172: ``wirteEntries`` and ``writeEntriesAlt`` can be consolidated. We should refactoring an interface.

miurahr (commented):
These methods share same code and has similar functions. We should consolidated it.

miurahr:
This make ``align.tmx`` like such as
```
    <tu tuid="accesshelp">
      <prop type="file">admin.php</prop>
      <prop type="id">accesshelp</prop>
      <tuv xml:lang="en-US">
        <seg>Accessibility help:</seg>
      </tuv>
      <tuv xml:lang="ja">
        <seg>Axuda sobre accesibilidade:</seg>
      </tuv>
    </tu>
    <tu tuid="access">
      <prop type="file">files.php</prop>
      <prop type="id">access</prop>
      <tuv xml:lang="en-US">
        <seg>Accessibility help</seg>
      </tuv>
      <tuv xml:lang="ja">
        <seg>Axuda sobre accesibilidade</seg>
      </tuv>
    </tu>
    <tu tuid="accesshelp">
      <prop type="file">files.php</prop>
      <prop type="id">accesshelp</prop>
      <tuv xml:lang="en-US">
        <seg>Accessibility help</seg>
      </tuv>
      <tuv xml:lang="ja">
        <seg>Axuda de accesibilidade</seg>
      </tuv>
    </tu>
```

src/org/omegat/core/data/RealProject.java
```diff
@@ -1848,7 +1856,7 @@ public void addTranslation(String id, String source, String translation, boolean
                             }
                             tr.source = oneSrc;
                             tr.translation = oneTrans;
-                            data.put(sourceS, new TMXEntry(tr, true, null));
+                            data.put(new EntryKey(sourcePath, sourceS, id, "", "", oneSrc), new TMXEntry(tr, true, null));
```
t-cordonnier, line 1859: I discovered a small bug here when I was reviewing this: https://sourceforge.net/p/omegat/bugs/1104/ 
I think we should decide what to do in this case before considering this patch.

In any case, sourceS is not the correct source (and as a consequence, not the correct parameter in EntryKey) and oneSrc is not the correct last parameter ("path") in EntryKey's constructor. Here we have no "path".

What we can do:
- if we want only one entry per paragraph, then we must concatenate all sources in tr.source and all translations in tr.translation, then the call to data.put is correct
- if we want one entry per segment, then call to data.put should be part of the loop, and each time, EntryKey uses oneSrc as a source and nothing as "path"

src/org/omegat/filters2/moodlephp/MoodlePHPFilter.java
```diff
@@ -167,10 +167,10 @@ protected void processBlock(String block, BufferedWriter out) throws IOException
     }
 
     @Override
-    protected void alignFile(BufferedReader sourceFile, BufferedReader translatedFile, FilterContext fc)
-            throws Exception {
-        Map<String, String> source = new HashMap<String, String>();
-        Map<String, String> translated = new HashMap<String, String>();
+    protected void alignFile(BufferedReader sourceFile, BufferedReader translatedFile, FilterContext fc,
+                             String sourcePath) throws Exception {
```
t-cordonnier, line 171: Maybe sourceFile or sourceFilePath instead of sourcePath, to avoid confusion with EntryKey's "path" parameter (whose meaning is not clear, I admit)

src/org/omegat/core/data/RealProject.java
```diff
@@ -1857,7 +1865,7 @@ public void addTranslation(String id, String source, String translation, boolean
                     }
                     tr.source = sourceS;
                     tr.translation = transS;
-                    data.put(sourceS, new TMXEntry(tr, true, null));
+                    data.put(new EntryKey(sourcePath, sourceS, id, "", "", ""), new TMXEntry(tr, true, null));
```
t-cordonnier, line 1868: If at line 1817/1818 you change to ITMXEntry, no need to call a constructor:
data.put(new EntryKey(sourcePath, sourceS, id, "", "", ""), tr);

src/org/omegat/core/data/RealProject.java
```diff
@@ -1814,17 +1815,24 @@ static class AlignFilesCallback implements IAlignCallback {
             this.config = props;
         }
 
-        Map<String, TMXEntry> data = new TreeMap<>();
-        private ProjectProperties config;
+        Map<EntryKey, TMXEntry> data = new TreeMap<>();
```
t-cordonnier, line 1818: I think we could use ITMXEntry here, instead of TMXEntry. Sure this interface did not exist when previous version was written. But this has some advantages, see my comment on line 1860/1868

t-cordonnier (commented):
One bug in the previous version requires your attention before applying this patch:
https://sourceforge.net/p/omegat/bugs/1104/

For the rest, only some minor suggestions.

miurahr:
@t-cordonnier Adjust for your comments.

@keko seems to want to this.
>if we want one entry per segment, then call to data.put should be part of the loop, and each time, EntryKey uses oneSrc as a source and nothing as "path"

miurahr:
rebase to master

miurahr:
@t-cordonnier when using ITMXEntry test failed with inconsistency of object types.
[02ab2a1](https://github.com/omegat-org/omegat/pull/230/commits/02ab2a1c25f19beca8975e73d3f1f7211c02939d) try to fix it.

miurahr:
rebase to master again.

miurahr:
Now I think it is ready for merge.

t-cordonnier:
> @t-cordonnier when using ITMXEntry test failed with inconsistency of object types. [02ab2a1](https://github.com/omegat-org/omegat/pull/230/commits/02ab2a1c25f19beca8975e73d3f1f7211c02939d) try to fix it.

Yes, ProjectTMX requires really to have TMXEntry because it makes distinction between default and non-default entry
Now if I look to
` tmx.defaults.put(en.getKey().sourceText, new TMXEntry(en.getValue(), false, null));`
1. boolean "default" should be true if it is a default translation
2. Are you sure you still want to use default translation here, as the goal of the PR is to accept non-default alignments? Look exaclty about what the test does, maybe you should look to the contents of the entry (for example is id = "" or not) to see whenever the entry is default or not?

miurahr:
> > @t-cordonnier when using ITMXEntry test failed with inconsistency of object types. [02ab2a1](https://github.com/omegat-org/omegat/pull/230/commits/02ab2a1c25f19beca8975e73d3f1f7211c02939d) try to fix it.
> 
> Yes, ProjectTMX requires really to have TMXEntry because contrarily to TMXWriter2, it makes distinction between default and non-default entry Now if I look to ` tmx.defaults.put(en.getKey().sourceText, new TMXEntry(en.getValue(), false, null));`
> 
> 1. boolean "default" should be true if it is a default translation
> 2. Are you sure you still want to use default translation here, as the goal of the PR is to accept non-default alignments? Look exaclty about what the test does, maybe you should look to the contents of the entry (for example is id = "" or not) to see whenever the entry is default or not?

Thank you. I changed it  `true`

miurahr:
Now breaks the feature.

```
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE tmx SYSTEM "tmx11.dtd">
<tmx version="1.1">
  <header creationtool="OmegaT" o-tmf="OmegaT TMX" adminlang="EN-US" datatype="plaintext" creationtoolversion="5.8.0_0_4f423fa6e" segtype="sentence" src
lang="en-US"/>
  <body>
    <tu>
      <tuv lang="en-US">
        <seg>Accessibility:</seg>
      </tuv>
      <tuv lang="ja">
        <seg>Accessibility3</seg>
      </tuv>
    </tu>
    <tu>
      <tuv lang="en-US">
        <seg>Accessibility help:</seg>
      </tuv>
      <tuv lang="ja">
        <seg>Accessibility help</seg>
      </tuv>
    </tu>
    <tu>
      <tuv lang="en-US">
        <seg>Accessibility</seg>
      </tuv>
      <tuv lang="ja">
        <seg>Accessibility2</seg>
      </tuv>
    </tu>
    <tu>
      <tuv lang="en-US">
        <seg>Accessibility help</seg>
      </tuv>
      <tuv lang="ja">
        <seg>Accessibility help</seg>
      </tuv>
    </tu>
  </body>
</tmx>

```

Now produces without property

t-cordonnier:
> Now breaks the feature.
> 
> 
> Now produces without property

Yes that is the reason of my point 2: the test works with default translations, if you want properties you must set alternative translations in tmx.alternatives instead of defaults in tmx.defaults

miurahr:
I set up test project with default value: `        <support_default_translations>true</support_default_translations>`
It should be `        <support_default_translations>false</support_default_translations>` when we want to get multiple translations.
Now solved.

PR #232, opened 2022-07-07
[Feature] Team: Git: Allow a branch name other than master

miurahr:
- UI: Add branch field for repo configuration
- UI: Add branch field in new team project form
- Team: Git: Tracking branch ban be other than master

## Pull request type

- [ ] Bug fix
- [x] Feature
- [ ] Documentation
- [ ] Other (describe below)

## Which issue to be solved?

Link: https://sourceforge.net/p/omegat/feature-requests/1531/
Title: Support using arbitrary git branch name

Link: https://sourceforge.net/p/omegat/feature-requests/918/
Title: Support Git branch in Team Project

## Other information

Back port the feature from the PR #174 onto Java 8.
Because it will be far from now that the migration to Java 11 become, so I'd like to merge the feature in Java 8 age.

## Checks

- [x] Integration test with main git branch
- [x] Integration test with master git branch
- [ ] Integration test with other than default branch name(*)

(*) integration test does not support other than default branch


src/org/omegat/gui/dialogs/NewTeamProject.java
```diff
@@ -304,6 +315,12 @@ public void actionPerformed(java.awt.event.ActionEvent evt) {
         gridBagConstraints.insets = new java.awt.Insets(5, 5, 5, 5);
         getContentPane().add(localFolderLabel, gridBagConstraints);
 
+        org.openide.awt.Mnemonics.setLocalizedText(jLabel1, "Branch Name");
```
miurahr, line 318: Here is incomplete localization.

miurahr:
All the possible tests are passed.

miurahr:
Current GUI is like this, and it can be improved using RadioButton.
image

to be

image

Which is better?

miurahr:
The latter should be controlled text box enable/disable according to radio box.
OK button should be enabled when custom branch name entered.

There are many controller logic methods in the form class, so we should separate into a controller class.

miurahr:
Edited commit history and rebased on current master.

PR #242, opened 2022-08-21
linux installation files by Marc Prior

brandelune:
- [ ] Bug fix
- [X] Feature
- [ ] Documentation
- [ ] Other (describe below)

## Which ticket is resolved?

https://sourceforge.net/p/omegat/feature-requests/1626/

## What does this PR change?

Marc updated his install script based on recent discussions.


brandelune:
The new linux-install.readme file should be included in the distribution but I've not yet modified the build script.

release/linux-specific/linux-install.sh
```diff
@@ -109,31 +132,17 @@ else
...
-icon_sizes=( 32 128 256 512 )
-for size in "${icon_sizes[@]}"
-do
-	sudo xdg-icon-resource install --novendor --noupdate --mode system --size $size /opt/omegat/OmegaT-default/images/OmegaT.iconset/icon_$size\x$size.png omegat
-done
-sudo xdg-icon-resource forceupdate --mode system
-
-# install desktop file
-sudo xdg-desktop-menu install --novendor --mode system /opt/omegat/OmegaT-default/omegat.desktop
+sudo cp omegat.desktop /usr/share/applications/
```
miurahr, line 146: It should use xdg for standard compliance. 
This removes standard compliant process. I'm against it.

release/linux-specific/omegat.desktop
```diff
@@ -1,24 +1,13 @@
...
-Comment[it]=lo strumento gratuito per memorie di traduzione
-Comment[es]=La herramienta TAO gratuita
-Comment[zh_CN]=免费的翻译记忆工具
-Comment[ja]=自由に使える翻訳メモリツール
-Comment[uk]=безкоштовна програма для перекладу
-Comment[fr]=l'outil libre de TAO
-Comment[ia]=Le application gratuite pro memorias de traduction
-Comment[de]=Das kostenlose Translation-Memory-Tool
-Comment[fi]=vapaa käännöstyökalu
-Comment[ca]=l'eina lliure de memòria de traducció
```
miurahr, line 19: Should not remove localization.

release/linux-specific/omegat.desktop
```diff
@@ -1,24 +1,13 @@
...
+StartupNotify=false
+Terminal=0
+TerminalOptions=
 Type=Application
-Categories=Office;
-Icon=omegat
-Exec=omegat %f
-Terminal=false
+X-KDE-SubstituteUID=false
+X-KDE-Username=
```
miurahr, line 13: This remove standard properties and add KDE specific option. It is bad.

release/linux-specific/linux-install.sh
```diff
@@ -1,6 +1,12 @@
 #!/usr/bin/env bash
 
-OMTVERSION="OmegaT_@VERSION_NUMBER_SUBST@"
+
```
miurahr, line 3: I cannot understand why it removes parameter substitution for producing script.

release/linux-specific/linux-install.readme
```diff
@@ -0,0 +1,84 @@
...
+
+omegat.kaptn should be dropped and the tcl/tk launch script (omegat.tcl) for OmegaT shipped in its place. This script is much more up to date with respect to changes in OmegaT, only requires tcl/tk to run (readily available and easily installed on Linux) and is much more intuitive for users to modify.
+
+
+Desktop integration:
+--------------------
+
+The Linux installer does not provide any desktop integration, other than instructions in the documentation.
+
+SUGGESTED CHANGE:
```
miurahr, line 79: It should not be in readme document.

miurahr (changes requested):
I don't agree with the proposed changes. It breaks many things.
I think it makes things worse.

brandelune:
Thank you. I understand. I'll see with Marc.

brandelune:
I think Marc is fine with you making sure that the script does not break backward compatibility. So if you feel like making the necessary modifications, go ahead so that we can include it in 5.8.

brandelune:
Let's move this to 6.1. I'll see with Marc that we fix it so that it's compatible with what you suggest. Thank you.

miurahr:
stalled.

PR #250, opened 2022-09-02
[OmegaT 6.0]Copyright Header: Change home page and FSF url to https

miurahr:
## Pull request type

- Bug fix

## Which ticket is resolved?

- Copyright Header GNU project URL should be https
    - https://sourceforge.net/p/omegat/bugs/1109/
 

## What does this PR change?

Update copyright header.

Before
```
 homepage: http://www.omegat.org/

 You should have received a copy of the GNU General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
```

After
```
 homepage: https://www.omegat.org/

 You should have received a copy of the GNU General Public License
 along with this program.  If not, see <https://www.gnu.org/licenses/>.
```

see https://www.gnu.org/licenses/gpl-3.0.txt


miurahr:
There is another `http` URL in header for omegat home page. It is good chance to update it.

miurahr:
I'd like to propose merge it just before 6.0.0 release. @brandelune

brandelune:
Definitely.

brandelune:
What do we do with the conflicts here ?

miurahr:
We can create the patch with simple sed command. We don't need a hand resolve.

brandelune:
> We can create the patch with simple sed command. We don't need a hand resolve.

As you think is best.

brandelune:
Is there a reason why the milestone is not 5.8?

brandelune:
Ok. I had forgotten about it.

miurahr:
remake and merged into master.

PR #273, opened 2022-09-24
Filter: ResourceBundleFilter: Add option to write in UTF-8

miurahr:
## Pull request type

- [x] Bug fix
- [ ] Feature
- [ ] Documentation
- [ ] Other (describe below)

## Which ticket is resolved?

- Link:  https://sourceforge.net/p/omegat/bugs/1116/
- Title:  BundleProperties fIlter should accept UTF-8 properties file

- LInk: https://sourceforge.net/p/omegat/feature-requests/1634/
- Title: BundleProperties Filter: write properties file in UTF-8 w/o escape in default

## What does this PR change?

- Fix BUGS#1116 https://sourceforge.net/p/omegat/bugs/1116/
- Add preference option force to write Java 8 compatible escaped properties and set default to true, escaping.
- Accept UTF-8 encoded properties file in default.
- Add Regression test for ed5e7da83b5c Reproduce reported issue
	> 46573: Error: java.io.IOException: /Users/suzume/Documents/Repositories/OmegaT plugin FR/source/omt-package/omt-package.properties
	> 46573: Error: java.nio.charset.MalformedInputException: Input length = 1

## Other information

This PR add  option for resource bundle filter to force escape output file as compatible as Java 8.
In default, it set `true` and output will be escaped as same as previous OmegaT versions.
When set it false, OmegaT can output UTF-8 encoded properties file that is compatible with Java 9 and later.

### DIFF
--- a/src/org/omegat/filters2/text/bundles/ResourceBundleFilter.java
+++ b/src/org/omegat/filters2/text/bundles/ResourceBundleFilter.java
@@ -34,13 +34,13 @@ import java.awt.Window;
 import java.io.BufferedReader;
 import java.io.BufferedWriter;
 import java.io.File;
-import java.io.FileOutputStream;
 import java.io.IOException;
+import java.io.InputStreamReader;
 import java.io.OutputStreamWriter;
-import java.io.UnsupportedEncodingException;
 import java.nio.charset.Charset;
 import java.nio.charset.CharsetEncoder;
 import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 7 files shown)


src/org/omegat/filters2/text/bundles/ResourceBundleFilter.java
```diff
@@ -134,15 +137,45 @@ public boolean isTargetEncodingVariable() {
...
+    public BufferedReader createReader(File inFile, String inEncoding) throws IOException {
+        Charset charset;
+        if (inEncoding != null) {
+            charset = Charset.forName(inEncoding);
+        } else {
+            charset = Charset.defaultCharset();
+        }
+        return new BufferedReader(new InputStreamReader(Files.newInputStream(inFile.toPath()), charset));
+    }
+
```
miurahr, line 178: ### Workaround 1.

Override createReader() with ancient IO subsystem. to accept UTF-8 characters even encoding is specified as US-ASCII that is default on OmegaT 5.7.1 or before.

src/org/omegat/filters2/text/bundles/ResourceBundleFilter.java
```diff
@@ -263,8 +296,10 @@ private String toAscii(String text, EscapeMode mode) {
                 result.append("\\=");
             } else if (mode == EscapeMode.KEY && cp == ':') {
                 result.append("\\:");
-            } else if ((cp >= 32 && cp < 127)
-                    || charsetEncoder.canEncode(text.substring(i, i + Character.charCount(cp)))) {
+            } else if (cp >= 32 && cp < 127) {
+                result.appendCodePoint(cp);
+            } else if (!forceTargetEscape
+                    && charsetEncoder.canEncode(text.substring(i, i + Character.charCount(cp)))) {
                 result.appendCodePoint(cp);
```
miurahr, line 303: ### Change 2.

We set default to be UTF-8 so `canEncode` might return TRUE but we introduce `forceTargetEscape` flag that is true in default force the behavior  as same as  OmegaT 5.7.1.

miurahr (commented):
Here is a point you should review.

PR #278, opened 2022-10-01
drop dependency from core to filter classes

miurahr:
Some core class calls or depends class or methods on filter drivers.
Because filters are build as plugins, but its (reverse) dependency breaks independence of core classes.

This is a patch to remove the dependency and improve independence of the core classes.

## Pull request type

- [ ] Bug fix
- [ ] Feature
- [ ] Documentation
- [x] Other (describe below)

Improve code structure. 

## Which ticket is resolved?

N.A.

## What does this PR change?

- Remove imports from TagValidationTool  for PoFIlter.
- Introduce org.omegat.util.HtmlUtils class as common utility from org.omegat.filters2.html2 local class
- Deprecate org.omegat.filters2.html2.HTMLUtils class


miurahr:
I'd like to move it to be merged into 5.8/6.0

PR #283, opened 2022-10-08
Update PR template not to use checklist

miurahr:
Update template not to use checklist but use Github labels

## Pull request type

- Other:  github template

## What does this PR change?

- Update template not to use checklist but use label
- Change order of ticket tiltle/url as same as change log format


miurahr:
Merged into master.

PR #284, opened 2022-10-08
Dictionary: notify when search done but no data

miurahr:
## Pull request type

- Bug fix -> [bug]
- Other (describe below)

## Which ticket is resolved?

- Title ex. fix...
  * URL ex. https://...
 
- Visual confirmation that dictionary was searched
  * https://sourceforge.net/p/omegat/feature-requests/1633/

## What does this PR change?

- Introduce preference NOTIFY_DICTIONARY_NOHIT
- When set notify even when no search result, it notify with color

### DIFF
--- a/src/org/omegat/Bundle.properties
+++ b/src/org/omegat/Bundle.properties
@@ -1849,6 +1849,7 @@ GUI_MATCHWINDOW_SUBWINDOWTITLE_Dictionary=Dictionary
 GUI_DICTIONARYWINDOW_explanation=Displays the dictionary entries found in the dictionary files located in the /dictionary/ folder if they are present in the source segment.
 
 GUI_DICTIONARYWINDOW_SETTINGS_NOTIFICATIONS=Notify Me When a Segment Has Dictionary Entries
+GUI_DICTIONARYWINDOW_SETTINGS_NOTIFICATION_NOHIT=Notify even When No Result found on Dictionary Search
 
 GUI_MATCHWINDOW_SUBWINDOWTITLE_MachineTranslate=Machine Translation
 
--- a/src/org/omegat/gui/dictionaries/DictionariesTextArea.java
+++ b/src/org/omegat/gui/dictionaries/DictionariesTextArea.java
@@ -283,12 +283,17 @@ public class DictionariesTextArea extends EntryInfoThreadPane<List<DictionaryEnt
         clear();
 
         if (data == null) {
+            if (Preferences.isPreference(Preferences.NOTIFY_DICTIONARY_NOHIT)) {
+                scrollPane.notify(false);
+            }
             return;
-        }
-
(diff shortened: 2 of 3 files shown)


brandelune:
I think this one is the direct answer to the original RFE.

If we implement this, the other one seems less useful and more like a distraction. What do you think ?

miurahr:
This is bad UI.  There is no integrity of UX rules.
The other is coherent implementation. Request is just his request. We should understand the root of the reason.

brandelune:
Thank you for your feedback.

> This is bad UI. There is no integrity of UX rules.

I think we have an issue of redundant and irrelevant notifications, see below.

> The other is coherent implementation. Request is just his request. We should understand the root of the reason.

The root of the reason is that there is no indication that a _manually_ triggered function is functioning.
That is confusing.

# This implementation
Here, we have an indication that the manually triggered function is functioning when the result is 0 data.
**→ the user knows that OmegaT did follow “orders” but did not find anything.**

That is good and relevant information provided by the system.

# The other implementation
1. In the other implementation, we have indication that an _automatically_ triggered function is functioning, but we already know that since a result is produced.
**→ the visual notification is redundant**

2. The visual notification is applied to automatically triggered functions (each time a segment is entered) even though the user has reasonable expectations that OmegaT functions.
The problem is that for automatically triggered functions like TMX/glossary fetching, the user will have a lot of visual distractions, which _**does**_ impair concentration and focus on a translation.
**→ the visual notification is a distraction that impairs concentration**

# Summary

My proposal is

**Default**:
- **When an action has provided a result**, the result is enough as a visual notification and **no extra visual notification is provided**, regardless of whether the action was automatically or manually triggered.
- When a manually triggered actions has not provided a result, a visual notification is provided.

**Option**:
- When an automatically triggered action has not produced a result, a visual notification is produced.

What do you think about that?

miurahr:
A change look like this. Is it really better than #287 ?

https://user-images.githubusercontent.com/123720/203684602-7f87ebef-c769-4182-a15c-75bdc6b5da29.mp4

brandelune:
> A change look like this. Is it really better than #287 ?

I don't have an opinion on the notification style itself.
Here, you use the highlight system, in #287 you use the gear icon.

I think this implementation is better, not in style, but in UX:

- a manual operation that has no result shows that it has no result
- an automatic operation that has results does not send redundant information
- an automatic action that has no result is expected to work and should not require a visual indication that it has worked

Considering the style here, I think it is consistent with the notification system when the pane is closed and OmegaT calls attention to the translators "you don't see the result, but there is one". Here it says something similar : "there is no result but I went to fetch one as you told me to."

As I wrote, the less redundant distractions with have the better it is to focus on the work.

I would like your opinion on the distraction aspect of the UX.

miurahr:
Thank you for UX consideration.
I'd like to merge #284 for 5.8/6.0.

brandelune:
> Thank you for UX consideration.
> I'd like to merge #284 for 5.8/6.0.

Please do so. Thank you for your work.

PR #285, opened 2022-10-08
GlossaryPane: configurable for sort order of entries 

miurahr:
## Pull request type

- Bug fix -> [bug]
- Other (describe below)

## Which ticket is resolved?

-  Glossary pane sorting
  * https://sourceforge.net/p/omegat/feature-requests/1606/

## What does this PR change?

- sort entries when result is set
-
-

### DIFF
--- a/src/org/omegat/Bundle.properties
+++ b/src/org/omegat/Bundle.properties
@@ -1869,6 +1869,7 @@ GUI_GLOSSARYWINDOW_insertselection=Insert Selection
 GUI_GLOSSARYWINDOW_SETTINGS_NOTIFICATIONS=Notify Me When a Segment Has Glossary Hits
 
 GUI_GLOSSARYWINDOW_SETTINGS_OPEN_FILE=Open Writable Glossary File
+GUI_GLOSSARYWINDOW_SETTINGS_SORT_BY_LENGTH=Sort by target term length
 
 GUI_MATCHWINDOW_SUBWINDOWTITLE_Dictionary=Dictionary
 
--- a/src/org/omegat/gui/glossary/FindGlossaryThread.java
+++ b/src/org/omegat/gui/glossary/FindGlossaryThread.java
@@ -45,7 +45,7 @@ import org.omegat.util.Preferences;
  * This process looks up the source string entries, and find matched glossary
  * entries.
  * <p>
- * Test cases wheter a glossary entry matches a string entry text:
+ * Test cases whether a glossary entry matches a string entry text:
  * <ul>
  * <li>"Edit" vs "Editing" - doesn't match
  * <li>"Old Line" vs "Hold Line" - doesn't match
--- a/src/org/omegat/gui/glossary/GlossaryAutoCompleterView.java
+++ b/src/org/omegat/gui/glossary/GlossaryAutoCompleterView.java
@@ -153,9 +153,9 @@ public class GlossaryAutoCompleterView extends AutoCompleterListView {
 
     static class GlossaryComparator implements Comparator<AutoCompleterItem> {
 
-        private boolean bySource = Preferences.isPreference(Preferences.AC_GLOSSARY_SORT_BY_SOURCE);
-        private boolean byLength = Preferences.isPreference(Preferences.AC_GLOSSARY_SORT_BY_LENGTH);
-        private boolean alphabetically = Preferences.isPreference(Preferences.AC_GLOSSARY_SORT_ALPHABETICALLY);
+        private final boolean bySource = Preferences.isPreference(Preferences.AC_GLOSSARY_SORT_BY_SOURCE);
+        private final boolean byLength = Preferences.isPreference(Preferences.AC_GLOSSARY_SORT_BY_LENGTH);
+        private final boolean alphabetically = Preferences.isPreference(Preferences.AC_GLOSSARY_SORT_ALPHABETICALLY);
 
         private final List<GlossaryEntry> entries;
         private final String matchTo;
--- a/src/org/omegat/gui/glossary/GlossaryManager.java
+++ b/src/org/omegat/gui/glossary/GlossaryManager.java
@@ -179,7 +179,7 @@ public class GlossaryManager implements DirectoryMonitor.Callback {
     }
 
     /**
-     * Loads one glossary file. It choose and calls required required reader.
+     * Loads one glossary file. It chooses and calls required reader.
      */
     private List<GlossaryEntry> loadGlossaryFile(final File file) throws Exception {
         boolean isPriority = priorityGlossary.equals(file);
@@ -210,7 +210,7 @@ public class GlossaryManager implements DirectoryMonitor.Callback {
      * @param src
      */
     public List<GlossaryEntry> getGlossaryEntries(String src) {
-        List<GlossaryEntry> result = new ArrayList<GlossaryEntry>();
+        List<GlossaryEntry> result = new ArrayList<>();
         synchronized (this) {
             for (List<GlossaryEntry> en : glossaries.values()) {
                 result.addAll(en);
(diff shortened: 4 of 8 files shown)


kosivantsov:
I can't see what needs to be enabled or configured to get the alphabetical sorting
aaeb78ce8

miurahr:
@kosivantsov  Thank you for feedback. I've changed a place where insert sort logic.
Does this solve your case?

miurahr:
On GlossarySearcher.class Line 79, I found 
```
        // After the matched entries have been tokenized and listed.
        // We reorder entries: 1) by priority, 2) by length, 3) by alphabet
        // Then remove the duplicates and combine the synonyms.
        sortGlossaryEntries(result);
```

A sort order seems to be a  designated behavior,  committed by @amake in 2017 when version 4.3.0.
@kosivantsov do you remember what omegat community discussed and decided?

miurahr:
Auto-completion has a preference configuration like as
image

Should glossary also has options for sort order?

miurahr:
Add settings for glossary pane,  1) sort by term length 2) sort by source length
@kosivantsov  Does this works for you?
image

miurahr:
Because This  may change behavior from previous versions, I'd like to keep a sort order in default.
I want to ask @brandelune  preferable whether sort order (alphabetically first)  or compatibility (length first) ?

kosivantsov:
> On GlossarySearcher.class Line 79, I found
> 
> ```
>         // After the matched entries have been tokenized and listed.
>         // We reorder entries: 1) by priority, 2) by length, 3) by alphabet
>         // Then remove the duplicates and combine the synonyms.
>         sortGlossaryEntries(result);
> ```
> 
> A sort order seems to be a designated behavior, committed by @amake in 2017 when version 4.3.0. @kosivantsov do you remember what omegat community discussed and decided?

It was a long time ago, but in any case, if it is possible to sort glossary entries alphabetically and make it one of the sorting options, it would be a beneficial addition.

kosivantsov:
> 



> Add settings for glossary pane, 1) sort by term length 2) sort by source length
> @kosivantsov Does this works for you?

Yes, it works.
`Sort by source term length` and `Sort by target term length` might be a bit more consistent with what we already have.
Still, `Sort by source term alphabetically` would be a welcome option for the glossary pane.

miurahr:
> Yes, it works. `Sort by source term length` and `Sort by target term length` might be a bit more consistent with what we already have. Still, `Sort by source term alphabetically` would be a welcome option for the glossary pane.

thanks for feedback, I'll update term.

In my sense,  `Sort by source term length`  is not very useful, and I cannot imagine a person who need it.

I've put a comment;
```
            // sort entries:
            // 1) by priority
            // 2) by length of source term (in default)
            // 3) by alphabet of source term
            // 4) by length of localized term (optional)
            // 5) by alphabet of localized term
```

I think we can remove No.2 and its related option, and No.4 is useful option as same as auto-completion's one.

brandelune:
I don't think we need to stick to an annoying previous behavior.

I'd suggest we go with alphabetical by default and accept "by order of apparition in source" as an option.

I can't see a use case where "length of term" order has any kind of usefulness, but if needed should be ok as a third option.

By the way, the options should be available from the Glossary pane tool icon, we should not need to go to the Preferences to change something like that.

kosivantsov:
Now it seems to work as expected. From my humble point of view, it can be merged with the master. Thank you so much for implementing it!

brandelune:
I get a "sort by target term length" in the glossary pane, but disabling it does not change the display. I must be missing something. Can you clarify how the setting is supposed to work?

kosivantsov:
> I get a "sort by target term length" in the glossary pane, but disabling it does not change the display. I must be missing something. Can you clarify how the setting is supposed to work?

You should have a source term that has more than one target equivalent, preferably of a different length. Once the option is enabled, you should refresh the pane by either reloading the project or switching to another segment and back.

brandelune:
> On Oct 10, 2022, at 23:09, Kos Ivantsov ***@***.***> wrote:
> 
> 
> I get a "sort by target term length" in the glossary pane, but disabling it does not change the display. I must be missing something. Can you clarify how the setting is supposed to work?
> 
> You should have a source term that has more than one target equivalent, preferably of a different length. Once the option is enabled, you should refresh the pane by either reloading the project or switching to another segment and back.

I see. I thought it was about sorting all the terms in the glossary pane.

JC

kosivantsov:
Should it be merged with the master?

brandelune:
> Should it be merged with the master?

I thought the scope of the feature was to cover all the glossary items, not only those that had multiple matches.

I think it would be best to cover all the glossary matches, for ex. right now it does not seem to be exactly alphabetical order, and it is very confusing.

Segment:
`Visiting, creating, and saving files.`

Glossary matches:
```
creating = création
1. Sun

Visiting = Consulter un fichier

saving = enregistrement
1. Sun

files = fichiers
1. @(px)ref
```

If there is no way to adapt that code to have full scope on glossary matches, I guess it is still good, though. What do you think?

miurahr:
> I thought the scope of the feature was to cover all the glossary items, not only those that had multiple matches.
> 
> I think it would be best to cover all the glossary matches, for ex. right now it does not seem to be exactly alphabetical order, and it is very confusing.
>
> If there is no way to adapt that code to have full scope on glossary matches, I guess it is still good, though. What do you think?

Which is a glossary file format to reproduce?

brandelune:
The glossary I use on that project is TXT in UTF-8.

It is the Emacs French localization team project that is hosted here:

https://forge.chapril.org/brandelune/documentation_emacs/

miurahr:
> The glossary I use on that project is TXT in UTF-8.
> It is the Emacs French localization team project that is hosted here:

Thank you for information. In my environment it looks ok.

Segment:
`Visiting, creating, and saving files.`

Glossary matches:

> 
> save = **d'enregistrer**, enregistrement, enregistrer
> 1. omegat
> 2. Sun
> 3. Sun
> 
> visit = **consulter**
> 
> create = créer
> 1. Dans certains cas, emploi du subst. "création", Sun
> 1. Sun
> 
> created = créé
> 1. Dans certains cas : "crée le" suivi de la date – infos document, Sun
> 
> creating = création
> 1. Sun
> 
> file = classer, fichier
> 1. Sun
> 2. KDE
> 2. Sun
> 
> files = fichiers
> 1. @(px)ref
> 
> Save = Enregistrer
> 1. Pas "sauver"
> 
> save file = enregistrer le fichier
> 1. Sun
> 
> saving = enregistrement
> 1. Sun
> 
> saving file = enregistrement du fichier
> 1. Sun
> 
> Visiting = Consulter un fichier
> 
> 
>

miurahr:
These are shown in bold
 
> save = **d'enregistrer**,
> visit = consulter

which are  come from writable glossary.txt that is `priority = true`.
Other localized candidates are come from read only one.
That is why, these two are shown on top of pane.

All others has same `priority = false` so these are sorted in alphabetically.

miurahr:
@brandelune 

Your result looks  as same as current Nightly does.
 
> creating => source length 8, smaller compared to "Visiting'
> Visiting => source length 8
> saving => souce length 7
> files => source length 5

brandelune:
Thank you for checking.

I'll need to sort out the info and make sure the manual reflects this. Or maybe you did that already?

JC

> On Oct 26, 2022, at 19:33, Hiroshi Miura ***@***.***> wrote:
> 
> 
> @brandelune
> 
> Your result looks as same as current Nightly does.
> 
> creating => source length 8, smaller compared to "Visiting'
> Visiting => source length 8
> saving => souce length 7
> files => source length 5
> 
> —
> Reply to this email directly, view it on GitHub, or unsubscribe.
> You are receiving this because you were mentioned.
> 

-- 
Jean-Christophe Helary @brandelune
https://traductaire-libre.org
https://mac4translators.blogspot.com
https://sr.ht/~brandelune/omegat-as-a-book/

miurahr:
@kosivantsov  has already confirms the patch works, so I'd like to merge.

I'm also going to work on update manual clause to reflect the new behavior. pls. wait a bit..

miurahr:
Squash and rebased.

PR #287, opened 2022-10-11
Show Progress Icon  on panes: fuzzy matches, machine translations, glossary and multitrans pane.

miurahr:
Show Progress Icon  on several panes

- Machine Translation pane
- Dictionary pane
- Multiple translations pane
- Fuzzy matches pane
- Glossary pane

## Pull request type

- Bug fix -> [bug]
- Other (describe below)

## Which ticket is resolved?

- Visual confirmation that dictionary was searched
    -  https://sourceforge.net/p/omegat/feature-requests/1633/
 

## What does this PR change?

-  Add capability put spin progress icon 
loading
on Panes
- Put the icon on Machine translation and dictionary pane which need some time to search in thread

## Other information

https://user-images.githubusercontent.com/123720/203661835-af08b309-4bb1-4107-bd50-4c941c6f1c3a.mp4


miurahr:
TODO:
This sometimes not working... 
- [x] When cache hit and show cached content ... no problem
- [ ] ??

miurahr:
Now ready for review.

miurahr:
here is new capture

https://user-images.githubusercontent.com/123720/198159556-4b2a02e3-58d5-4336-a509-50ffeae6e7d4.mp4

brandelune:
The problem is that match computation and display is so quick that the actual effect is that I see "Fuzzy Matches" jump on the side and come back, I see something happening, but it looks more like as a graphical display artifact than like something that was intended to carry some information.

I think the wheel position should rather be on the right hand of the string it is associated too:

```
Fuzzy Matches
⚙️ Fuzzy Matches
Fuzzy Matches
```

```
Fuzzy Matches
Fuzzy Matches ⚙️
Fuzzy Matches
```

miurahr:
A icon position  is designed in default theme, not as a feature.

brandelune:
The videos you have, show the icon moving for about 1 second. On my machine, I don't have the time to notice it. The only thing I notice is the label that suddenly moving to the right and the comes back in position.

If you can't move the icon, maybe moving the label to the right is the only solution left.

Otherwise, it only looks like a graphic bug.

vescovaccio:
Personally, I'd rather have a tick (check mark) symbol appear, in order to confirm the success of the search and retrieval, rather than a spinning progress symbol. Spinning progress symbols in UIs drive me nuts, from the days of using Windows and waiting interminably from a process to finish. I'd rather a tick to show the search has successfully completed, than a rotating thingamajig. That would also solve the problem of the label jumping to the side and back again, sometimes extremely quickly, because the label would only move once to the right. Just a suggestion!

brandelune:
I think we're losing track of the original RFE.

It was only asked to have visual confirmation that a `Edit > Search Dictionaries` action had actually been performed, when it was set to _not_ be automatic.

What we have now is visual indications that OmegaT is doing what it is supposed to do: automatically search for stuff (TM/gloss/etc.)

So the proposal:

1. is very much beyond the scope of the RFE (confirm that a _manual_ action was performed when no result appear)
2. does not provide any extra information (we know that OmegaT automatically searches for info in various places already)
3. is implemented in a way that looks like a bug (the label jumps back and forth)

My idea would be to limit the visual cues to a simple message in the dictionary pane like "OmegaT did not find any match" (we could eventually have a similar message in the glossary and TM panes, to just reassure the user that OmegaT is actually doing its part), but let's not overengineer something that could be kept super simple.

miurahr:
> My idea would be to limit the visual cues to a simple message in the dictionary pane like "OmegaT did not find any match" (we could eventually have a similar message in the glossary and TM panes, to just reassure the user that OmegaT is actually doing its part), but let's not overengineer something that could be kept super simple.

There is no way to show retrial is nothing because of multi-threading, asynchronous query.

We had been engineering to remove messages from panes, especially in MT pane. 
That is why I am against your opinion.

Current change is basically very simple in view of code. You may feel many changes in PR, but these is because of Java language,  giving common character on all search panes (there is no reason only dictionary pane changes), and handling multi-threading implementation of search.

What I add for put icon is

```
    public void startProgressIcon() {
        if (dockKey.getLocation() == Location.HIDDEN) {
            return;
        }
        Icon spinIcon = new ImageIcon("src/org/omegat/gui/resources/loading.png");
        dockKey.setIcon(spinIcon);
    }

    public void stopProgressIcon() {
        dockKey.setIcon(null);
    }
```

As you can see, there is nothing to give no place control, and no duration control.
If you feel progress icon is too high speed and shorter duration, your PC is super high performance.

There is alternative idea to notify search action completion, that is a use of notifier as same as comments and notes pane.
It is too much eyes catch to tell just a search progress.

miurahr:
> I think the wheel position should rather be on the right hand of the string it is associated too:

Does anyone know a way to change a position on Java Swing icon on title bar, please teach  me.

miurahr:
> > I think the wheel position should rather be on the right hand of the string it is associated too:
> 
> Does anyone know a way to change a position on Java Swing icon on title bar, please teach me.

@brandelune  could you ask @amake to implement an icon position control on dependency `vldocking` project or tell us a way to realize the feature?

https://github.com/omegat-org/vldocking/blob/master/src/main/java/com/vlsolutions/swing/docking/DockViewTitleBar.java#L166-L171
```
                 if(pName.equals(DockKey.PROPERTY_ICON)) {
				titleLabel.setIcon((Icon) e.getNewValue());
```

Current vldocking just set icon on the title label(JLabel).

brandelune:
> As you can see, there is nothing to give no place control, and no duration control.

Too bad.

> If you feel progress icon is too high speed and shorter duration, your PC is super high performance.

It's a 7 years old laptop that I use professionally.

Professional translators need relatively good spec machines because CAT tools are resource intensive and typical use is often to have many applications working at the same time.

It's not abnormal that OmegaT feels fast on such machines. In fact I've had a lot of co-workers who felt that OmegaT was faster than their proprietary CAT tool.

So I'm totally fine with OmegaT running fast.

Also, considering that machine specs are generally improving, it is likely that a lot of people who will run OmegaT will feel the same as I do.

So, please, let's reconsider the scope of the RFE and other ways to implement that.

> There is alternative idea to notify search action completion, that is a use of notifier as same as comments and notes pane. It is too much eyes catch to tell just a search progress.

I am not sure that it is too catchy. Because it is a human action that triggers the notification. It does not happen all the time or automatically:

The notification happens only when the user searches for matches in a dictionary **and** there are no matches.

miurahr:
> > There is alternative idea to notify search action completion, that is a use of notifier as same as comments and notes pane. It is too much eyes catch to tell just a search progress.
> 
> I am not sure that it is too catchy. Because it is a human action that triggers the notification. It does not happen all the time or automatically:
> 
> The notification happens only when the user searches for matches in a dictionary **and** there are no matches.

Unfortunately OmegaT dictionary pane (and glossary, MT etc) don't have a mechanism to know the trigger is manual or not.
So we can just switch  on or off for notifier, or progress icon.

miurahr:
> or, leave some space for the wheel on the left:
> 
> ```
>    Fuzzy Matches
> ⚙️ Fuzzy Matches
>    Fuzzy Matches
> ```


This is easy because we can just put empty(white?) icon for each panes.
Current implementation does put nothing for it.

miurahr:
> > As you can see, there is nothing to give no place control, and no duration control.
> 
> Too bad.

We can put timer to show progress icon in minimum duration like 100 msec.

miurahr:
Is it what you want to see?

image

miurahr:
Empty and spin icon size is 20x20 that is bigger than font size...

src/org/omegat/util/gui/SpinIcon.java
```diff
@@ -0,0 +1,88 @@
...
+
+public class SpinIcon implements Icon {
+    private final Icon spinIcon;
+    private final int width;
+    private final int height;
+    private double angle = 0;
+    private final Timer motionTimer;
+
+    public SpinIcon(final JComponent component) {
+        spinIcon = new ImageIcon("src/org/omegat/gui/resources/loading.png");
```
miurahr, line 47: We should get icon image from `UIManager` for theme plugins can customize it.

src/org/omegat/gui/theme/DefaultFlatTheme.java
```diff
@@ -195,7 +195,16 @@ public static UIDefaults setDefaults(UIDefaults defaults, String lafId) {
...
-        // Icons
+        // Icons for panes
+        defaults.put("OmegaTMachineTranslatesPane.icon", getIcon("appbar.empty.icon.png"));
+        defaults.put("OmegaTFuzzyMatchPane.icon", getIcon("appbar.empty.icon.png"));
+        defaults.put("OmegaTMultiTranslatesPane.icon", getIcon("appbar.empty.icon.png"));
+        defaults.put("OmegaTDictionaryPane.icon", getIcon("appbar.empty.icon.png"));
+        defaults.put("OmegaTGlossaryPane.icon", getIcon("appbar.empty.icon.png"));
+        defaults.put("OmegaTCommentsPane.icon", null);
+        defaults.put("OmegaTNotesPane.icon", null);
+        defaults.put("OmegaTSegmentPropertiesPane.icon", null);
```
miurahr, line 206:
```suggestion
        defaults.put("OmegaTSegmentPropertiesPane.icon", null);
        defaults.put("OmegaTProgressSpinIcon", getIcon("loading.png")); 
```

miurahr:
I think I need more investigation, design and engineering for the enhancements. So changing target milestone to be 6.1.

brandelune:
The result looks fine. But indeed, we should not rush development. Thank you for postponing.

miurahr:
> Fuzzy Matches
> Fuzzy Matches ⚙️
> Fuzzy Matches

Now implement this style. The issue is clearly solved.

miurahr:
If we prefer only notify no hit on dictionary pane, it is #284

miurahr:
It is a GUI on current implementation 

https://user-images.githubusercontent.com/123720/203661835-af08b309-4bb1-4107-bd50-4c941c6f1c3a.mp4

miurahr:
When search dictionary from Edit->search dictionary.

https://user-images.githubusercontent.com/123720/203662298-155aaf6e-41df-4a17-83d3-455466e80960.mp4

brandelune:
Please, let me refer you to my comment on #284 :

https://github.com/omegat-org/omegat/pull/284#issuecomment-1319351025

I still think that the visual notifications for panes that work automatically and that provide a result is redundant and a distraction.

miurahr:
> Please, let me refer you to my comment on #284 :
> 
> [#284 (comment)](https://github.com/omegat-org/omegat/pull/284#issuecomment-1319351025)
> 
> I still think that the visual notifications for panes that work automatically and that provide a result is redundant and a distraction. Also, in the second case, the Dictionary pane provides a result to the visual notification is redundant.

I've put screen capture of #284 

[#284(comment)](https://github.com/omegat-org/omegat/pull/284#issuecomment-1325886888)

Feature #284 can be configurable but here always show.
It may be better to show progress icon when the process is longer.

brandelune:
> Feature #284 can be configurable but here always show. It may be better to show progress icon when the process is longer.

But the process is never _that_ long that the user has to be aware that something is happening under the hood :-)

src/org/omegat/gui/common/EntryInfoSearchThread.java
```diff
@@ -110,10 +113,11 @@ public void run() {
         SwingUtilities.invokeLater(new Runnable() {
             public void run() {
                 if (isEntryChanged()) {
+                    pane.abortSearch(OStrings.getString("ENTRYINFOSEARCH_ENTRY_CHANGED"));
                     return;
                 }
                 if (ferror != null) {
-                    pane.setError(ferror);
+                    pane.abortSearch(ferror.getLocalizedMessage());
```
miurahr, line 120:
```suggestion
                    pane.setError(ferror);
```

change is redundant. It should be reverted.

miurahr:
> Personally, I'd rather have a tick (check mark) symbol appear, in order to confirm the success of the search and retrieval, rather than a spinning progress symbol. Spinning progress symbols in UIs drive me nuts, from the days of using Windows and waiting interminably for a process to finish. I'd rather a tick to show the search has successfully completed, than a rotating thingamajig. That would also solve the problem of the label jumping to the side and back again, sometimes extremely quickly, because the label would only move once to the right. Just a suggestion!

Good idea. I'd like to change this
-  remove gray tick icon in process, when shown
-  show gray small tick after the process finished

PR #294, opened 2022-10-19
Select source

kosivantsov:
## Which ticket is resolved?

- #1639 Add a shortcut to select all source text of the current segment
  * https://sourceforge.net/p/omegat/feature-requests/1639/
  

## What does this PR change?

- Adds an Editor shortcut to select all source text in the current segment (even with the locked cursor)


miurahr:
Is there a corresponding menu item? It would be better to have menu item  from view of DE&I.

brandelune:
Select all source is not an obvious action, so putting it in the menu is a good thing. I guess you want to use that to do searches ? Do you have other uses ?

kosivantsov:
> I guess you want to use that to do searches ? Do you have other uses ?

Asking the client or other team members about the segment, performing OmegaT searches, or running External Search, among other possible uses...

brandelune:
Does the function "only" select or does it copy as well ?

kosivantsov:
> 



> Does the function "only" select or does it copy as well ?
Only selects, because for searches copying is not required, and it might not be desired to override the clipboard. If the user wants to copy, there's a simple shortcut for that.

kosivantsov:
Sorry for the separate commits for each individual file, had to do it through the Web interface

kosivantsov:
I'm not sure what the problem with the check is, it compiles locally without any errors.

brandelune:
Excellent. It works very nicely and exactly as expected. It's a very nice workflow addition.

brandelune:
Here again, it is a `gradlew check` that fails.

damien-rembert:
The check is likely failing as the IEditor interface is modified and some of its methods are not implemented in places where they should be. The change to IEditor is not necessary (the change to EditorController is enough) and should be removed.
This PR seems to modify files that are outside of its scope, maybe it would be best to only modify EditorController here and have another PR for the other changes?

brandelune:
> The check is likely failing as the IEditor interface is modified and some of its methods are not implemented in places where they should be. [Edit] Oh I see you used the interface to bring selectSourceText to the menu. Then you need to implement selectSourceText in:
> 
> > -test-integration/src/org/omegat/core/data/TestTeamIntegrationChild.java
> > -test/src/org/omegat/core/TestCore.java
> 
> This PR seems to modify files that are outside of its scope, maybe it would be best to only modify files that need to have selectSourceText here and have another PR for the other changes?

@damien-rembert, thank you for checking. Would you mind looking into that?

miurahr:
Please rebased on current master head.

miurahr:
These all 33 commits looks not related to select source feature.

damien-rembert:
@miurahr was faster! Nice work!

miurahr:
Fixed and merged into master.

PR #295, opened 2022-10-20
Make dictionary drivers as true plugin

miurahr:
- Register dictionary drivers, LingvoDSL and StarDict through core API.
- Add class path to Plugins.properties.
- It allow clarify a core API, these drivers is not core.

## Pull request type

- Bug fix -> [bug]
- Other (describe below)

 Refactoring for future source tree management and clarify core API.

## What does this PR change?

-
-
-


miurahr:
Can we move `StarDict` and `LingvoDSL` driver class  to other than an `org.omegat.core` package?

PR #297, opened 2022-10-22
Show glossary comment on context menu candidate

miurahr:
when context menu on glossary match on source text show replace candidate, and when mouse is hover on the candidate text, OmegaT show glossary comment as tooltip when it exist.

## Pull request type

## Which ticket is resolved?

- Improving the right-click menu on underlined words
  * https://sourceforge.net/p/omegat/feature-requests/1515/

## What does this PR change?

- 

## Other information
image

### DIFF
--- a/src/org/omegat/gui/glossary/TransTipsPopup.java
+++ b/src/org/omegat/gui/glossary/TransTipsPopup.java
@@ -27,6 +27,8 @@
 
 package org.omegat.gui.glossary;
 
+import java.awt.event.MouseAdapter;
+import java.awt.event.MouseEvent;
 import java.util.HashSet;
 import java.util.Set;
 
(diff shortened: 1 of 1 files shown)


miurahr:
Now waiting feedback from the issue reporter.

brandelune:
The tooltip on the underlined term is very quick to display but the tooltip in the context menu is too slow to display. Why not use the same speed ?

miurahr:
> The tooltip on the underlined term is very quick to display but the tooltip in the context menu is too slow to display. Why not use the same speed ?

I don't know but it is depends on Java or your machine.

miurahr:
There is no feedback from issuer.

brandelune:
Do you need any more comments to decide whether the feature is to merge or not?

miurahr:
Ok, I'd like to merge it.

PR #304, opened 2022-10-29
Project>Reload  load remote project configuration when team project

miurahr:
## Pull request type

- New feature [enhancement]

## Which ticket is resolved?

- Local project does not update after remote masking some files
   https://sourceforge.net/p/omegat/bugs/1124/

## What does this PR change?

- Load remote omegat.project when project is on team
- Refactoring `ProjectUICommands` class
-

### DIFF
--- a/src/org/omegat/core/data/ProjectProperties.java
+++ b/src/org/omegat/core/data/ProjectProperties.java
@@ -52,6 +52,7 @@ import org.omegat.util.StringUtil;
 
 import gen.core.filters.Filters;
 import gen.core.project.RepositoryDefinition;
+import gen.core.project.RepositoryMapping;
 
 /**
  * Storage for project properties. May read and write project from/to disk.
@@ -433,6 +434,17 @@ public class ProjectProperties {
 
     public List<RepositoryDefinition> getRepositories() {
         return repositories;
+    }
+
+    public boolean isTeamProject() {
+        for (RepositoryDefinition repositoryDefinition: repositories) {
+            for (RepositoryMapping repositoryMapping: repositoryDefinition.getMapping()) {
+                if ("".equals(repositoryMapping.getLocal()) || "/".equals(repositoryMapping.getLocal())) {
+                    return true;
+                }
+            }
+        }
+        return false;
     }
 
     public void setRepositories(List<RepositoryDefinition> repositories) {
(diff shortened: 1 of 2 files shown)


brandelune:
Do we need a "Reload _team_ project" menu item ? Can't it be the same "Reload project" item that calls 2 different functions depending on the status of the project ?

miurahr:
see. https://sourceforge.net/p/omegat/bugs/1124/

brandelune:
Your code logic detects that a project is a team project and acts on that.

Are you suggesting that we give 2 possibilities to team members: reload the _local_ info only and also reload the _remote_ info?

miurahr:
> Your code logic detects that a project is a team project and acts on that.
> 
> Are you suggesting that we give 2 possibilities to team members: reload the _local_ info only and also reload the _remote_ info?

So we need to two menu item  reload local and reload remote

brandelune:
What about

`Reload Project`

and

`Update Remote Project Information`

?

msoutopico:
Thanks for your work, Hiroshi.

I haven't understood why the `Reload project` option cannot be used to update remote files, as JC suggests. Is it because it may be very slow? 

In fact, I wasn't aware that one could reload the project and update local files but not remote files... I had always assumed that `Reload project` (F5) (or when I use the `LOAD eventType` in a groovy script) is equivalent to closing and re-opening the project, but it seems that I got that wrong. 

Would `Reload project` be enough to fetch new or updated files from mapped repositories or would users need to run the new `Reload team project` for that?

miurahr:
> I haven't understood why the `Reload project` option cannot be used to update remote files, as JC suggests. Is it because it may be very slow?

Yes super slow. Is is really acceptable? Really no need local reload even on team project?

miurahr:
@brandelune now  Project > reload  is going to  download remote configuration.

miurahr:
merged into `master`

brandelune:
Can you check that a normal reload on a non-team project works on your side ? It looks locked on my side, but works for team-projects...

miurahr:
Thanks for the check.
`GitRemoteRepository2IT#isTeamProject` raises a null pointer exception.

miurahr:
A fix 0c8e173e is pushed to `master` on the sourceforge git repository.

PR #310, opened 2022-11-03
Change some l10n keys

miurahr:
pick form @brandelune manuel5.n branch and fix key not found build error


miurahr:
@brandelune  do you really want to change this?

PR #314, opened 2022-11-03
Update user guide HTML files (English)

miurahr:


brandelune:
It looks fin here.

brandelune:
But we need to remove the DRAFT flag from the docbook.

miurahr:
> But we need to remove the DRAFT flag from the docbook.

Could you please propose to fix it?

brandelune:
It's ok. But there is a problem that comes from the HTML. Some IDs are not defined and create artificial diffs.

So, let's fix the XML first and rebuild the files. Is that OK with you? If yes, I'll handle the XML.

miurahr:
> It's ok. But there is a problem that comes from the HTML. Some IDs are not defined and create artificial diffs.
> 
> So, let's fix the XML first and rebuild the files. Is that OK with you? If yes, I'll handle the XML.

Please.

brandelune:
done in a separate PR.

PR #316, opened 2022-11-03
Jc/topic/add i dsto examples

brandelune:
Some examples did not have proper IDs and generated random ID names at build, which created false diffs.


miurahr:
LGTM.

PR #320, opened 2022-11-06
Fix the XML to produce valid FO and output PDF

brandelune:
The DocBook file set did not output a proper FO file and did not allow the creation of a PDF file.

This fix ensures that the PDF is properly created but **does not** address PDF specific issues like fonts or layout issues.

### DIFF
--- a/doc_src/en/App_Glossaries.xml
+++ b/doc_src/en/App_Glossaries.xml
@@ -125,79 +125,80 @@
     their file extension:</para>
 
     <table id="glossary.extensions">
-      <caption>Format, extensions and expected encoding</caption>
-
+      <title>Format, extensions and expected encoding</title>
+	  <tgroup cols="3">
       <thead>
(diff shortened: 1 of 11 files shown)


brandelune:
@amake Is there a reason why you closed this without merging ? Is there a problem with this PR ?

amake:
"I" didn't delete it; it got clobbered by the git sync from SourceForge to GitHub. 

Any branch present only here and not there will be deleted by the sync. The sync happens to use a credential tied to my account, so it looks like I did it. 

Some solutions:

1. Push your branches to SF instead
2. Make your branches on your personal fork on GitHub
3. Change the behavior of the sync task. The code for this lives in Azure, but I don't know offhand how it should be changed.

brandelune:
Ok. Thank you for the information. I'll check with @miurahr how we can do 3. and I'll try 1. and 2. for now.

PR #331, opened 2022-11-12
Make sure the manual, Bundle.properties, and code all match

Kazephil:
[documentation]

## What does this PR change?

- Fix typos and in Bundle.properties and reword some phrases.
- Update manual to match the changes in Bundle.properties.
- Update identifiers in code files to match the change in both of the above.


Kazephil (commented):
Updated to fix the problem with the capitalization of "Global Post-processing Commands" pointed out by Kos.

At the same time, I reworded some of the descriptions in the "View" and "Saving and Output" sections of the Preferences chapter of the manual.

brandelune:
Sorry to come back late to this one.
Is there a necessity to modify .xml files in other languages than en ?

Kazephil:
> Is there a necessity to modify .xml files in other languages than en ?

I don't think so.

Only IDs and any phrases kept in English even in the translation would be affected, but I imagine that as long as the translators for other languages are working from the most recent batch of files committed to the repository, the .xml files they generate will contain all the updated content.

However f, for whatever reason, some people are working directly with the .xml files in other languages, changes to the IDs would have to be applied there.

brandelune:
> > Is there a necessity to modify .xml files in other languages than en ?
> 
> I don't think so.
> 
> Only IDs and any phrases kept in English even in the translation would be affected, but I imagine that as long as the translators for other languages are working from the most recent batch of files committed to the repository, the .xml files they generate will contain all the updated content.

That is my understanding also.
 
> However f, for whatever reason, some people are working directly with the .xml files in other languages, changes to the IDs would have to be applied there.

I don't think that's a pattern we have to worry about. The canonical version is the English version and the EN XML will be translated to other languages so that other languages would have a base corresponding to the corrected EN XML.

So I guess the commit contains files that we modified by mistake.

I'm not a pro about Github so my only suggestion is that you:
- update your master
- create a new branch from master
- cherry-pick only the relevant changes from this branch to that branch
- commit/PR that branch and forget this one

Kazephil:
> So I guess the commit contains files that we modified by mistake.

I think the non-English file in the commit are there because Kos started working on the Ukrainian version. One had a phrase that was still in English, the others had modifications to the IDs.

I don't think keeping the changes is a problem, since they'll be overwritten later when the final translation is generated, but if you (or Kos) thinks it might be a problem, I can try your cherry-pick suggestion.

brandelune:
> I don't think keeping the changes is a problem, since they'll be overwritten later when the final translation is generated, but if you (or Kos) thinks it might be a problem, I can try your cherry-pick suggestion.

I think it would be cleaner, just in case we need to revert something.

Kazephil:
Cleaned up and updated as PR #332.
Closing this PR as no longer necessary.

PR #336, opened 2022-11-18
fix XMLStreamReader.java error message strings

brandelune:
The XMLStreamReader had 1 badly formatted error message and 1 redundant error message.
Both are fixed here.


miurahr:
merged into master.

PR #338, opened 2022-11-18
Remove irrelevant Bidi "justification" function and related information

brandelune:
The current BiDi "Justification" function is not relevant for BiDi related tasks, information in the related chapter is obsolete and the associated shortcut confusing because not well enough documented.

I'm temporarily removing all references to that feature, it's shortcut and the related information.


src/org/omegat/gui/editor/EditorTextArea3.java
```diff
@@ -383,16 +383,6 @@ protected void processKeyEvent(KeyEvent e) {
             setSelectionStart(doc.getTranslationStart());
             setSelectionEnd(doc.getTranslationEnd());
             processed = true;
-        } else if (s.equals(KEYSTROKE_SWITCH_ORIENTATION)) {
```
miurahr, line 386: Please also remove `KEYSTROKE_SWITCH_ORIENTATION` constant static variable

miurahr (changes requested):

src/org/omegat/gui/editor/EditorTextArea3.java
brandelune: Done !

miurahr:
Thanks.
Squashed and merged into `master`

PR #300, opened 2022-10-24
Goto x

kosivantsov:
## Which ticket is resolved?

- #1641 Navigate between segments inserted from /tm/auto and /tm/enforce
  * https://sourceforge.net/p/omegat/feature-requests/1641/
 

## What does this PR change?

- It introduses 4 new Goto menu items (grouped in a submenu): goto next/prev xauto/xenforced.

PR #343, opened 2022-11-21
Recreated PR for goto x

miurahr:
Successor of #300

## Which ticket is resolved?

- #1641 Navigate between segments inserted from /tm/auto and /tm/enforce
  * https://sourceforge.net/p/omegat/feature-requests/1641/
 

## What does this PR change?

- It introduses 4 new Goto menu items (grouped in a submenu): goto next/prev xauto/xenforced.


miurahr:
Waiting @kosivantsov to reply https://github.com/omegat-org/omegat/pull/300#issuecomment-1321972111

PR #344, opened 2022-11-21
Recreated PR "Added a call to select the current source text"

miurahr:
Re-create commits and PR of #294

### DIFF
--- a/src/org/omegat/gui/editor/EditorController.java
+++ b/src/org/omegat/gui/editor/EditorController.java
@@ -1863,6 +1863,10 @@ public class EditorController implements IEditor {
         markAsComesFromMT(text);
     }
 
+    /**
+     * {@inheritDoc}
+     */
+    @Override
     public void insertTag(final String tag) {
         UIThreadsUtil.mustBeSwingThread();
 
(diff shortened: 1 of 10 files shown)


damien-rembert:
Very nice, tested on MacOS, works as expected.

src/org/omegat/gui/editor/EditorController.java
```diff
@@ -1880,6 +1880,17 @@ public void insertTag(final String tag) {
         }
     }
 
+    public void selectSourceText() {
```
miurahr, line 1883: Because this is public method, so please add javadoc comment.

You can check CI test warning
```
[ant:checkstyle] [WARN] /home/vsts/work/1/s/src/org/omegat/gui/editor/EditorController.java:1883:5: Class 'EditorController' looks like designed for extension (can be subclassed), but the method 'selectSourceText' does not have javadoc that explains how to do that safely. If class is not designed for extension consider making the class 'EditorController' final or making the method 'selectSourceText' static/final/abstract/empty, or adding allowed annotation for the method. [DesignForExtension]

```

src/org/omegat/gui/editor/IEditor.java
```diff
@@ -378,6 +378,11 @@ public static CaretPosition startOfEntry() {
      */
     String getSelectedText();
 
+    /**
+     * Select the source text of the current segment
```
miurahr, line 382:
```suggestion
     * Select the source text of the current segment.
```

miurahr (commented):
Could you improve javadoc?

PR #348, opened 2022-11-22
feature: shortcuts to navigate between Notes and Editor panel

briacp:
Unfortunately, this does not make the panel appear if they're not already visible -- it only switch focus between panels. As JC suggested, we can make the panel tab blinks and/or add menubar information.

## Which ticket is resolved?

- Shortcut to switch between Notes and Editor panels
  * https://sourceforge.net/p/omegat/feature-requests/1630

PR #349, opened 2022-11-23
Introduce requestVisible method to unhide/show hidden panes

miurahr:
Panes need to handle its visible state when user ask actions.

related #348

## Pull request type

## What does this PR change?

- Introduce `MainWIndow#getDesktop` that returns `DockingDesktop`  object.
- Introduce `StaticUIUtils#requestVisible` method

### DIFF
--- a/src/org/omegat/gui/exttrans/MachineTranslateTextArea.java
+++ b/src/org/omegat/gui/exttrans/MachineTranslateTextArea.java
@@ -91,6 +91,8 @@ public class MachineTranslateTextArea extends EntryInfoThreadPane<MachineTransla
      */
     protected List<MachineTranslationInfo> displayed = new CopyOnWriteArrayList<>();
 
+    protected final DockableScrollPane scrollPane;
+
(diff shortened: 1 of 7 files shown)


briacp (approved):
Great stuff Hiroshi, I didn't manage to make it work like you did, so good job!

I was just wondering if it was maybe possible to refactor all the `requestVisible` methods as they are mostly the same for all the components?

miurahr:
> Great stuff Hiroshi, I didn't manage to make it work like you did, so good job!

Best solution is  to show "AutoHideExpandPanel" but I didn't find an interface. So I made it to show "DockablePanel".

AutoHideExpandPanel is shown when you hover mouse on hidden pane such as Note, which will automatically hide with Timer.

It is best because when you operate note and go back to editor, it automatically hide to initial position.

> I was just wondering if it was maybe possible to refactor all the `requestVisible` methods as they are mostly the same for all the components?

Thank you for the suggestion. I finally decided to add the function in StaticUIUtils class.

PR #352, opened 2022-11-26
Fix "flagged text" string inconsistency.

Kazephil:
Fix for the discrepancy between "Flagged text" and "Remove text target" pointed out on the list by Manuel.


kosivantsov:
I've committed the proposed change to the SF repo

Kazephil:
Thank you!

PR #353, opened 2022-11-26
Remove doc expression about moses translator

miurahr:
## Pull request type

Documentation

## Which ticket is resolved?

N.A.

## What does this PR change?

- Remove table row of moses connector


miurahr:
merged into `master`

PR #356, opened 2022-11-28
Add Stax filters

t-cordonnier:
Jean-Christophe Helary asked me whenever it would be possible to include my StaX filters in the core of OmegaT. Here is the pull request where we try to do it.

## Pull request type

 [enhancement]

## What does this PR change?

Add new filters for the following formats:
- XLIFF 1.x
- XLIFF 2.x
- SDLXLIFF
- OpenXML

It also adds a global architecture to write filters for XML formats (distinct from filters3) and ZIP+XML (like OpenXML which is a zip containing XML files): with this architecture, we do direct StaX calls (https://en.wikipedia.org/wiki/StAX),
- inconvenience: the code is harder to read than the one in "filters3"
- advantage: during read of the source file, you can keep some extra data which are not part of the translatable segments (example: in OpenXML, the language which is declared in a distinct XML file)
- advantage: during write of the target file, you can do any change in the flow (example: adding a status attribute in target)

These filters are already known as StaX filters (http://185.13.37.79/?q=node/183) : the goal of the PR is to integrate them in the core of OmegaT.

What I include here is the release 2.0.6 of StaX filters: release 2.1 contains some added features which have less been tested and which are potentially controversial; the idea is to discuss them later.


miurahr:
@t-cordonnier Please fix SpotBugs violations.

miurahr (changes requested):
Need to fix the SpotBugs violations

t-cordonnier:
"Need to fix the SpotBugs violations"

Correction done, thanks for the alert I was concentrated on checkstyle and had not seen these ones.

miurahr:
Are there any test cases?

For XLIFF1 there is a test data.
It is important to test XLIFF2 filter with sample XLIFF2 data.

t-cordonnier:
I did not prepare unit tests. If needed that is an extra work. Maybe we could see with Manuel, he always has good test cases and said that he is implied in specification of XLIFF 2.

miurahr:
> For XLIFF1 there is a test data. 

I think it should pass all the current XLIFF1 test cases with Stax XLIFF1 filter.

I wandering when stax filter become genuine filter,  why not remove a current XLIFF1 mono language filter?
It is better that OmegaT community provide a current filter as a plugin for users who need an exact compatibility.

t-cordonnier:
> I think it should pass all the current XLIFF1 test cases with Stax XLIFF1 filter.

Logically this filter should cover almost all cases valid from the old filter, and additional ones. I will check when I have time.

> I wandering when stax filter become genuine filter, why not remove a current XLIFF1 mono language filter?

Decision would not be mine but I think this has already been discussed in the list: since compatibility is never ensured between two filters for the same format (always some differences almost with tags) the community will want to keep the old filter for people who continue to work with old documents.

Personally I would be in favour of moving all "filters3" old architecture to a plugin which only people working with old documents would really need (not yet, but almost after we can rewrite everything in Stax architecture, more modern). But again decision is not mine.

t-cordonnier:
> I think it should pass all the current XLIFF1 test cases with Stax XLIFF1 filter.
> 

I just did the test, in fact if I replace the filter in current test I have the following exception:
java.lang.NullPointerException
	at org.omegat.filters4.xml.xliff.Xliff1Filter.processStartElement(Xliff1Filter.java:105)

This error is **normal**: in the sample files present in the repository (/test/data/filters/xliff) all files contain almost one <trans-unit> without an id. But according to the specification this field is **REQUIRED**:
http://docs.oasis-open.org/xliff/v1.2/os/xliff-core.html#trans-unit
For that reason, I consider that this is not a bug in StaX filters but simply demonstration that old filter is more permissive against incorrect XLIFF files.

I will check if I can find examples conform to the specification and small enough to be used for tests. I remember there is such a set of files delivered with XLIFF 2 specification, or maybe I will use samples provided by Manuel

miurahr:
> This error is **normal**: in the sample files present in the repository (/test/data/filters/xliff) all files contain almost one without an id. But according to the specification this field is **REQUIRED**

Good catch. 
Could you update test expectations and cases that raise proper error or warning on wrong data, not NPE, and pass a test with correct data?

miurahr:
@t-cordonnier 
Current nightly has an improvement on plugin information handlings.
Could you rebase here on a  head of master?  And check  visibility on `Options>Prefereces..>plugins` dialog, when launching from installation? 

 The contents of plugin lists when launching from `./gradlew run`  are different from the case when installed launcher, because the former read `Plugins.properties` and the latter read jar `META-INF/MANIFEST.MF`.

You can check `MANIFEST.MF` contents that is generated into `build/tmp/jar/MANIFEST.MF`.

brandelune:
The current OmegaT XLIFF filter has been renamed "Monolingual XLIFF". It will stay in OmegaT until we take enough time to deprecate it, or we can decide to keep it for documented use cases.

Here is the thread:
https://sourceforge.net/p/omegat/mailman/omegat-development/thread/CABm46bbpSUHKaRs39yzyxz4rMMvPpz6PSf3Ob_ty48wH1evtVw%40mail.gmail.com/#msg37720772

t-cordonnier:
> Could you rebase here on a head of master? And check visibility on Options>Prefereces..>plugins dialog, when launching from installation?

I just added the info in Plugins.properties but I don't see where the change is supposed to appear in the UI, I let you check

t-cordonnier:
> > This error is **normal**: in the sample files present in the repository (/test/data/filters/xliff) all files contain almost one without an id. But according to the specification this field is **REQUIRED**
> 
> Good catch. Could you update test expectations and cases that raise proper error or warning on wrong data, **not NPE**, and pass a test with correct data?

I did not commit my unit tests yet since they are not finished. I continue to wok on it.

t-cordonnier:
Just added some unit tests, but not finished
Now after the test with XLIFF 2 I realize that I may have to review the behavior of the filter related to <sc> / <ec> specific markups. 
And I still have to add test showing translation working

miurahr:
I just find an example of XLIFF2 files at https://xmlpress.net/publications/xliff/examples/ which has permissive license.
Can we use it for test data?

t-cordonnier:
Not required, I finally use those from the specification (https://github.com/oasis-tcs/xliff-xliff-22/tree/master/xliff-21/test-suite)
I pushed yesterday some tests but it is not finished

miurahr:
> I finally use those from the specification (https://github.com/oasis-tcs/xliff-xliff-22/tree/master/xliff-21/test-suite) 

What a license of test data? I don't think it is compatible with GPL.

t-cordonnier:
Indeed the license written at https://github.com/oasis-tcs/xliff-xliff-22/blob/master/LICENSE.md does not seem very clear about what we are allowed to do with it: they seem more interested about saying how we are not allowed to contribute than about using the files.
In the README it is written "Content in this repository is intended to be part of the XLIFF TC's permanent record of activity, visible and freely available for all to use, " that remains vague.
However I could say the same about the link you sent to me, where the guy allows anybody to use it but "The only exception would be if you wish to sell or distribute the examples. " : if I use them in a Github, can't we consider that I am "distributing" the samples?

miurahr:
> However I could say the same about the link you sent to me, where the guy allows anybody to use it but "The only exception would be if you wish to sell or distribute the examples. " : if I use them in a GitHub, can't we consider that I am "distributing" the samples?

Only we can do is "a remote data" testing because authors does not allow a redistribution.

docs/index.html
```diff
@@ -35,7 +35,7 @@ <h1>OmegaT User Manual</h1>
 
     <tr>
         <td><a href="en/index.html">English</a></td>
-        <td>(<span class="out-of-date">4.3.1</span>)</td>
+        <td>(<span class="up-to-date">5.8.0</span>)</td>
```
miurahr, line 38: Please don't touch a document.

src/org/omegat/filters4/xml/openxml/MsOfficeFileFilter.java
```diff
@@ -115,6 +115,11 @@ private void defineDOCUMENTSOptions(Map<String, String> config) {
 
         TRANSLATABLE = Pattern.compile(DOCUMENTS);
     }
+    
+    public MsOfficeFileFilter() {
+        super();
+        defineDOCUMENTSOptions(new java.util.HashMap<String,String>()); // Define the documents to read
```
miurahr, line 121: Why don't import HashMap?

test/src/org/omegat/filters4/Xliff1FilterTest.java
```diff
@@ -0,0 +1,122 @@
...
+        });
+        System.err.println("Parsing file where trans-unit/@id is missing returns: "
+            + exception.getClass() + ": " + exception.getMessage());
+        assertTrue(exception.getMessage().contains("Missing attribute"));
+    }
+
+    @Test
+    public void testBilingual() throws Exception {
+        Map<String,String> result = new HashMap<String,String>();
+        Map<String,String> legacy = new HashMap<String,String>();
```
miurahr, line 70:
```suggestion
        Map<String,String> result = new HashMap<>();
        Map<String,String> legacy = new HashMap<>();
```

test/src/org/omegat/filters4/Xliff1FilterTest.java
```diff
@@ -0,0 +1,122 @@
...
+            java.util.Collections.emptyMap());
+        ParsedEntry firstEntry = entries.get(0);
+        assertEquals(firstEntry.id, "example_01");
+        assertEquals(firstEntry.path, "//interface.po");  // path is built with file name
+        ParsedEntry secondEntry = entries.get(1);
+        assertEquals("foo", secondEntry.source);
+        assertEquals("bar", secondEntry.translation);
+        ParsedEntry lastEntry = entries.get(entries.size() - 1);
+        assertEquals(lastEntry.id, "Added_1");  // next entry Added_2 is ignored as translate=no
+        assertEquals(lastEntry.path, "//interface.po/Group 3");  // path is built with file name and groups
```
miurahr, line 89:
```suggestion
        assertEquals("Added_1", lastEntry.id);  // next entry Added_2 is ignored as translate=no
        assertEquals("//interface.po/Group 3", lastEntry.path);  // path is built with file name and groups              
```

test/src/org/omegat/filters4/Xliff1FilterTest.java
```diff
@@ -0,0 +1,122 @@
...
+
+    @Test
+    public void testTranslation() throws Exception {
+        Xliff1Filter filter = new Xliff1Filter();
+        filter.translateFile(new File("test/data/filters/xliff/filters4-xliff1/en-xx.xlf"), 
+            outFile, java.util.Collections.emptyMap(), context,
+                new ITranslateCallback() {
+                    public String getTranslation(String id, String source, String path) {
+                        if ("Should translate in result.".equals(source)) {
+                            return "Devrait traduire dans le résultat.";
```
miurahr, line 100: Please escape non-ascii character here.

miurahr (changes requested):
There are a few place we can improve.

docs/index.html
t-cordonnier: OK but this is not really made by me: the problem is that build script always updates this file.
In DGT I had totally removed this file from the Git for this reason: maybe you have a better solution?

src/org/omegat/filters4/xml/openxml/MsOfficeFileFilter.java
t-cordonnier: Usually I work with a basic editor (Kate or Kwrite) so it does not automatically go to top in order to add the imports, as does Eclipse or Netbeans.
The checkstyle says nothing about this, so I did not consider it to be changed.
But I can do, if I find the time.

test/src/org/omegat/filters4/Xliff1FilterTest.java
t-cordonnier: OK, I must review  a little bit my tests (especially not finished for XLIFF 2), I will do the change.

t-cordonnier:
Just pushed corrected version, now all tests work

miurahr (approved):

brandelune:
I'll be working on the documentation. @t-cordonnier if there are documents that you think I could use, please send me pointers!

t-cordonnier:
I did not write a lot of documentation but you can look in DGT website
For OpenXML filter: http://185.13.37.79/?q=node/133
For XLIFF 1 and 2, I must explain that for my clients, the target was SDLXLIFF, while XLIFF 1 and 2 were only written as subsets. That is the reason why I documented only SDLXLIFF:
http://185.13.37.79/?q=/node/88

If you want more documentation, please point me out to the documentation of another filter, so that I can have an idea about your expectations.

brandelune:
What we have for the current "monolingual" XLIFF is the following:
https://doublet.jp/_omegat/en/chapter.appendices.html#file.filters.xliff

As for the MSO format we have this kind of format:
https://doublet.jp/_omegat/en/chapter.appendices.html#file.filter.microsoft

I checked what you have on the site and that does not give me a clear idea of how we could summarize that into the current "format".

If you could send me a draft of what you think is important, I could write the DocBook and integrate it.

brandelune:
Would it be possible to merge the code and then PR the doc?

t-cordonnier:
I have read your reference documents, they speak about the options available for each filter. My documents speak about other aspects - features which are not in other filters - that's the reason why they would not be in same location.

My filters for XLIFF and its variants currently do not have options at all. Not that I don't want it - it remains open for the future - but the actual behavior corresponds to what my clients wanted. Options will come when other users will ask for them.

My filter for Microsoft OpenXML currently has exactly the same options as the one which was already in OmegaT (in filters3). Of course there are differences between the filters but they are not yet about the options.

t-cordonnier:
> Would it be possible to merge the code and then PR the doc?

Not sure I understand what you mean. Now that all the tests are correct this PR seems ok to be merged, unless somebody says the contrary. And unless we find a good location where to put it (see previous message) I have nothing to add to the doc which could be in a separate PR.

brandelune:
> Not sure I understand what you mean. Now that all the tests are correct this PR seems ok to be merged, unless somebody says the contrary. And unless we find a good location where to put it (see previous message) I have nothing to add to the doc which could be in a separate PR.

I mean that _I_ prefer to write the doc as a separate action, so please go ahead with the merge.

miurahr:
I found missing regression test  for XLIFF1 and proposed in #416.
Please wait to merge #416 and my proposal of adding test case here.

t-cordonnier:
Let's note that #416 concerns only the monolingual filter - you should use this name instead of XLIFF1 to avoid ambiguity.
StaX filters implements status change but not yet a similar transition workflow, so I will not implement the same test.
These two PR could be independent, the only change I made related to monolingual filter is the fact that I changed the location of the sample files, because I partially re-use them in the tests for StaX filters.

I will wait, but not too long - Jean-Christophe now begins to become impatient...

brandelune:
> I will wait, but not too long - Jean-Christophe now begins to become impatient...

Not at all, really :-)

I just think we need to be able to have a "whole" 6.0 available as nightly for a while before the release (if only to leave the l10n team a bit of time to update the UI the the latest changes). And we're getting really close to that point.

miurahr:
I just merge master into here and add a regression test for XLIFF1Test.

miurahr:
@t-cordonnier could you add a test case in `Xliff1FilterTest.java` ?

```
    @Test
    public void testTranslationRFE1506() throws Exception {
        XLIFFFilterTest.checkXLiffTranslationRFE1506(new Xliff1Filter(), context, outFile, false);
    }
```

This will be passed.

t-cordonnier:
I added the test, and confirms that it works. I also left as comment the fact that NeedsTranslate is not implemented.

@miurahr @brandelune Is it ok to merge, now?

t-cordonnier:
ok, merge done in sourceforge, can close now. In case of new changes we open a new PR

miurahr:
@brandelune  Now we have merged staxfilter, with proper support for XLIFF 1.2 and 2.0, could you mind to call for test to users the Nightly with their live XLIFF files?
Also I'd like to ask you and @Kazephil to check documentation around XLIFF.

miurahr:
@t-cordonnier also could you check a behavior when user has already installed your stax filter as 3rd-party plugin?

brandelune:
https://github.com/omegat-org/omegat/issues/421

brandelune:
The plugin is properly registered in OmegaT (I can see it in the plugin list) but it does not appear in the file filters list *unless* I reinitialize the defaults, which I don't want since I have set parameters to some of my filters.

miurahr:
I'd like to be disabled monolingual xliff filter in default for 5.8/6.0. I'll be pushed another PR.
Also I'd like to i18n/l10n for filter names.

brandelune:
There are things to do that I would appreciate that you consider in your other PR, within what you think is possible:
- order the filters by alphabetical order by default
- order the filters by activated status
- disable all filters by default
- enable automatically the filters that are necessary in the project (in order of priority in the list)

I'll be working on the doc later this week. Thank you both for your work.

miurahr:
> * disable all filters by default
> * enable automatically the filters that are necessary in the project (in order of priority in the list)

Disable not all filters but duplicated one.

brandelune:
My idea was rather "only enable the filters that OmegaT needs when parsing the contents of /source". I think it is better in terms of UX than the current situation (and it removes issues with multiple filters that handle the same format).

Kazephil:
I don't know how easy it is to do, but I agree with Jean-Christophe.

From the user's point of view, it would be ideal if`:
- Only the filters required for the files in the current project are enabled.
- Those filters were sorted at the top of the list, in alphabetical order of filters used.
- The remaining list of disabled filters (including non-prioritized duplicate filters for the same file type) was in alphabetical order below that.

If that's difficult to implement, it could be something to add to the 6.1 or later wish list, though.

t-cordonnier:
Difficult or not, that is out of the scope of this PR, as it potentially concerns all filters not only StaX
Let's also not forget that the contents of /source can change in the lifecycle of a project: my clients liked the fact that they can delete/add files here while continuing to work on the same project. And the list of needed filters may change in this case.
There can also be multiple filters for a single format: which one is "needed" in this case? we must establish priorities

t-cordonnier:
> The plugin is properly registered in OmegaT (I can see it in the plugin list) but it does not appear in the file filters list _unless_ I reinitialize the defaults, which I don't want since I have set parameters to some of my filters.

Sounds like a problem in OmegaT itself, not in the StaX filters: should occur each time you introduce new filters after you made your configuration
You should probably open an RFE or a bug for this

miurahr:
> @miurahr I don't know how easy it is to do, but I agree with Jean-Christophe.
> 
> From the user's point of view, it would be ideal if`:
> 
> * Only the filters required for the files in the current project are enabled. 

It may be realized in 6.1.x.

> * Those filters were sorted at the top of the list, in alphabetical order of filters used.
> * The remaining list of disabled filters (including non-prioritized duplicate filters for the same file type) was in alphabetical order below that.
> 
> If that's difficult to implement, it could be something to add to the 6.1 or later wish list, though.

I've pushed a change that implement the proposal partially in #423

PR #357, opened 2022-11-29
new plugin api in MANIFEST.MF

miurahr:
Use new plugin descriptor in MANIFEST.MF

## Pull request type

## Which ticket is resolved?
 
- Title
  * URL

## What does this PR change?

- Use 'OmegaT-Plugins' entry for bundled plugins
    - New interface can provide human readable name, author, version etc.
-
-


miurahr:
ref #358

PR #355, opened 2022-11-27
Cmd+Shift+Q conflicts with

brandelune:
https://sourceforge.net/p/omegat/bugs/1127/

PR #363, opened 2022-12-07
Use similar shortcut on all platforms for "go to next unique"

brandelune:
Cf.
https://github.com/omegat-org/omegat/pull/355

A similar shortcut for Windows/Linux is set here.

Ctrl + Shift + K


miurahr:
Good catch. Our manual is neutral for platforms, so we should use same key for shortcut.

miurahr:
merged into `master`

PR #366, opened 2022-12-09
add panes navigation mac shortcut

brandelune:
https://github.com/omegat-org/omegat/pull/348 did not have mac specific shortcuts assigned.


miurahr:
merged into master

PR #369, opened 2022-12-10
Introduce common cache utility

miurahr:
Add common caching utility that can be used in many places in OmegaT code.

## Pull request type

Refactoring/enhancement

## Which ticket is resolved?

[Discussion on Dev ML](https://sourceforge.net/p/omegat/mailman/omegat-development/thread/b4b6s9rkcsbGbHpQJdC3Yugjsm1a3ZTRhgqnwinV2faVk6a5h_u_M6iSbxAD554ummjMZkjH-IH6mi55yQ6htZqqmPkoNhSIBfaNACvTjPs%3D%40northside.tokyo/#msg37746515)

## What does this PR change?

- Introduce LRUCache interface and LRUCacheFactory 
- Introduce CaffeineLRUCache implementation of LRUCache API
- Add SimleLRUCache that use LinkedHashMap.
- Use it on LingvoDSL dictionary driver.

## Other information

@t-cordonnier suggests DGT-OmegaT use WeakHashMap based cache implementation.
If it is also available, it will be better. Code suggestion is welcome.


src/org/omegat/util/cache/impl/CaffeineLRUCache.java
```diff
@@ -0,0 +1,171 @@
...
+
+    @Override
+    public void cleanUp() {
+
+    }
+
+    @Override
+    public void getMaxCacheSize() {
+
+    }
```
miurahr, line 97: Should be implemented.

miurahr:
It will be better to provide a preference UI how much omegat cache for dictionary, each MT

Is it really requires Caffeine cache library? When @t-cordonnier suggests weak cache, we can drop it.

miurahr:
> I also recopy here what I sent directly to @miurahr
> 
> > I'd like to propose  Caffeine  for cache library.
> > I also wanna propose Abstraction layer for caching, that hide an implementation. This helps us when we find better caching library,  we can switch it anytime.
> 
> I don't understand why you want to introduce your own abstraction layer. There is already one which exists and you quoted it yourself : [JSR-107 JCache](https://github.com/ben-manes/caffeine/wiki/JCache) And if I correctly read its documentation, Caffeine does also implement JCache, even if you are not forced to use it via JCache (in your code you called it directly)
> 
> You should study how to implement MT cache using JCache as an abstraction layer and Caffeine as default implementation. Then later we can switch from Caffeine to another implementation without rewriting code, except the code which says what is the desired implementation

JCache is Java EE specification and it is not convenient for desktop application like OmegaT.
That is why I propose a thin layer based on Abstract Factory pattern.

Do you like a complex interface for omegat cache layer?

t-cordonnier:
>      JCache [...]  is not convenient for desktop application like OmegaT.

Really? Do you have any concrete reference which says why?
Note: this is a real question, I would really be surprised that somebody thought about a generic specification for caching in web applications but nobody thought about the same for desktop ones?

> Do you like a complex interface for omegat cache layer?

If it has an added value, why not. For the moment compared to weak hash maps the only added value I see is the possibility to choose the size of the cache - but I am not sure basic users would be capable to set the best parameters for this, so not sure it is really pertinent.

In any case, if you want to have an abstraction layer, it should
- be outside the code of OmegaT itself, else you won't be able to use it in libraries which are not plugins (I think it is the case of stardict4j, which can be used in other programs, or did I misunderstand?)
- provide almost 2 implementations from the beginning, this is the only way to be sure that you don't accidentally included in the abstract layer an access to some features which exist only in Caffeine

miurahr:
I've simplified the code.  I removed `LRUCache` interface class. Now there is no abstraction layer and the factory returns standard `Map<K, V>` .

 There are two cache factory, 
1) Simple LRU cache that use java standard `LinkedHashMap` 
2) Soft referenced cache that use Caffeine library.

brandelune:
> Maybe something to be discussed with Jean-Christophe: is the notion of cache something clear for translators which do not have IT knowledge? I am not sure a simple user of the application can anticipate how may entries he can have in the cache. With WeakReference he does not have to care: the application knows when there is not enough memory anymore, and Java is known as one of the rare languages where even the developer rarely needs to take care about question of RAM quantity. Such an option should be modified only by administrators, not by simple users. But discuss with JC: I am not a translator, if he thinks it is clear enough for simple users I will not say the contrary.

I just saw the reference to me today.

I think the idea of cache is not hard to understand and if properly documented should not be a problem at all. But I don't want my opinion to influence technical questions for which I have no real understanding. Although, I'd prefer a system where the translator does not have to worry about things.

t-cordonnier:
> 

So, if I give the possibility to select the number of cached translations, how would you recommend to a translator, without IT knowledge, what is the best value for him?
Maybe my formulation was not correct, I suppose a translator can understand what the cache is, but not sure he is able to know what is the best value for him.

brandelune:
I guess it depends on the "cost" of dropping possible matches and the benefits of caching.

PR #370, opened 2022-12-10
Goto x redone

kosivantsov:
## Which ticket is resolved?

- #1641 Navigate between segments inserted from /tm/auto and /tm/enforce
  * https://sourceforge.net/p/omegat/feature-requests/1641/
 

## What does this PR change?

- It introduses 4 new Goto menu items (grouped in a submenu): goto next/prev xauto/xenforced.

It substitutes #300


brandelune:
@briacp suggested that we could use the Ctrl/Cmd-Alt shortcuts for new functions. Would you mind considering that for your menu items?

Also, if you could put the menu item descriptions in the `Go to` menu list, that would be nice.

Considering the functions, they could benefit from a short description in `OmegaT5_ProjectFolder.xml`, especially the `project.folder.tm.auto` and the `project.folder.tm.enforce` subsections. With links from the menu descriptions to the subsections and reciprocally. I'm trying to stick to "Use such-and-such menu item to achieve this and that" kind of menu use descriptions (at least that's what I've been working on all day today ;-) ). So if you could keep your text close to that form, that would help, the readers, mostly.

Last, if you could add the changes.txt item, that would be great.

If there is anything you don't feel that you can do, just drop a draft here, and I'll handle it.

kosivantsov:
> @briacp suggested that we could use the Ctrl/Cmd-Alt shortcuts for new functions. Would you mind considering that for your menu items?

Not all of the actions are provided with default shortcuts (**Go to next/prev note**, for instance). I don't mind adding shortcuts, but I don't know what would be logical and easy to remember. If Ctrl+Alt shortcuts are to be used, I suggest to use **Ctrl+Alt+N/P** for segments with notes (I guess, these two actions will be used much more frequently than jumping between xauto and xenforced segments).

kosivantsov:
> Also, if you could put the menu item descriptions in the `Go to` menu list, that would be nice.

I'm not sure what you're talking about. What's `Go to` menu list?

brandelune:
This list : 

https://github.com/omegat-org/omegat/blob/master/doc_src/en/Menus_GoTo.xml

brandelune:
> Not all of the actions are provided with default shortcuts (**Go to next/prev note**, for instance). I don't mind adding shortcuts, but I don't know what would be logical and easy to remember. If Ctrl+Alt shortcuts are to be used, I suggest to use **Ctrl+Alt+N/P** for segments with notes (I guess, these two actions will be used much more frequently than jumping between xauto and xenforced segments).

C-A-(N/E) are already taken by Briac's "go to Notes pane" / "go to Editor pane".

Go to next unique is C-S-(K), K is not related to any letter in the item. And that's OK too.

kosivantsov:
> This list :
> 
> https://github.com/omegat-org/omegat/blob/master/doc_src/en/Menus_GoTo.xml

`Auto-Populated Segments` submenu items
These actions allow to navigate between segments that were automatically populated by exact matches from /tm/auto and /tm/enforce. See [/tm/auto]() and [/tm/enforce]() for details.

`Next Segment with xAUTO`
   Moves to the next segment that was auto-populated from /tm/auto

`Previous Segment with xAUTO`
   Moves to the previous segment that was auto-populated from /tm/auto

`Next Segment with xENFORCED`
   Moves to the next segment that was auto-populated from /tm/enforce

`Previous Segment with xENFORCED`
   Moves to the previous segment that was auto-populated from /tm/enforce

kosivantsov:
> Considering the functions, they could benefit from a short description in `OmegaT5_ProjectFolder.xml`, especially the `project.folder.tm.auto` and the `project.folder.tm.enforce` subsections. With links from the menu descriptions to the subsections and reciprocally. I'm trying to stick to "Use such-and-such menu item to achieve this and that" kind of menu use descriptions (at least that's what I've been working on all day today ;-) ). So if you could keep your text close to that form, that would help, the readers, mostly.

This info could be included in the sections on `project.folder.tm.auto` and `project.folder.tm.enforce`:
The auto-populated status is saved between project reloads if `Preferences → Editor → Save auto-populated status` is enabled.
Auto-populated status for is saves as `linked` segment property. For segments populated from /tm/auto its value is `xAUTO`. For segments populated from /tm/enforce its value is `xENFORCED`.
Auto-populated segments can be highlighted: `View → Mark Auto-populated Segments`. The highlighting color can be configured in `Preferences → Appearance → Colours → From auto memory/From enforced memory`

src/org/omegat/Bundle.properties
```diff
@@ -395,6 +395,12 @@ TF_MENU_GOTO_SELECTED_MATCH_SOURCE=Source of Selected &Match
 TF_MENU_GOTO_FORWARD_IN_HISTORY=&Forward in History
 TF_MENU_GOTO_BACK_IN_HISTORY=&Back in History
 
+TF_MENU_GOTO_X_SUBMENU=Auto-Populated Segments
+TF_MENU_GOTO_NEXT_XAUTO=Next Segment with xAUTO
+TF_MENU_GOTO_PREV_XAUTO=Previous Segment with xAUTO
+TF_MENU_GOTO_NEXT_XENFORCED=Next Segment with xENFORCED
+TF_MENU_GOTO_PREV_XENFORCED=Previous Segment with xENFORCED
```
miurahr, line 402: Are these have mnemonics to access from keyboard?
I have a question because these don't have shortcut key.

src/org/omegat/gui/editor/EditorController.java
```diff
@@ -1456,6 +1456,36 @@ public void nextTranslatedEntry() {
...
     }
 
+    private void linkedEntry(boolean forward, String linked) {
+        iterateToEntry(forward, ste -> {
+                TMXEntry info = Core.getProject().getTranslationInfo(ste);
+                if (String.valueOf(info.linked).equals(linked)) {
+                    return true;
+                }
+                return false;
+            });
```
miurahr, line 1466: This can be simplified because second argument is `@NotNull`

```suggestion
        iterateToEntry(forward, ste -> linked
            .equals(Core.getProject().getTranslationInfo(ste).linked));
```

miurahr (changes requested):
There is no way to access the feature with only keys.

brandelune:
> This info could be included in the sections on `project.folder.tm.auto` and `project.folder.tm.enforce`:

Ok. I'll do it. Thank you !

src/org/omegat/Bundle.properties
kosivantsov: Good point, I'll add mnemonics for sure. Thank you, Hiroshi!

kosivantsov: Done!

src/org/omegat/gui/editor/EditorController.java
miurahr:
```suggestion
        iterateToEntry(forward, ste -> {
                TMXEntry info = Core.getProject().getTranslationInfo(ste);
                return String.valueOf(info.linked).equals(linked);
            });
```

this will be.

Kazephil, line 1463: This is one of the things that makes programming so interesting.
My solution for line 1462 to 1464 was:

```java
if (linked.equals(String.valueOf(info.linked))) {
    return true;
}
```
It worked on a brief test, and passed the SpotBugs check mentioned by Hiroshi. I'll have to study his solution to learn why it's a better idiom.

kosivantsov: the direction of that comparison isn't too important, I guess, we need to make sure
that `info.linked` and `linked` are identical

kosivantsov: Hiroshi's is better because he uses one `return` statement which work for both cases

src/org/omegat/Bundle.properties
kosivantsov: Mnemonics added: [f8c4577](https://github.com/omegat-org/omegat/pull/370/commits/f8c457732e09c4aeb680be6a7077ea35a927ebdd)

kosivantsov: As to providing shortcuts, I'm not sure. We do have some menu items without shortcuts. Users can add them if they want to.

brandelune: I think it is better to have them. Adding shortcuts is not trivial and the functions are useful.

The functions that do not have shortcuts are functions that are generally used once in a while. In the Go to menu, we can filter on segments having notes so having a shortcut for "next note/previous note" is not really necessary.

(As far as I am concerned, the "search in dictionary" should have one too, for ex.)

miurahr (approved):

src/org/omegat/Bundle.properties
brandelune: It creates a UX hierarchy between "auto" only 2 keys (C+A) and "enforce" with 3 keys (C+A+S).

Sometimes, the Shift is used to create the "reverse" action. So that would be Forward=C+A / Backward = C+A+S.

What would you think of that ?

kosivantsov: Committed the shortcuts as suggested (with Shift to revert direction).
If everyone is happy, maybe this one could be merged?

brandelune:
@kosivantsov Sorry for the hassle, but as I'm writing the doc, I realized that "with xAUTO" is an implementation detail and that we should not mention that to the user.

We should rather use "from tm/auto/" or something like that. Users know that they put something in the respective subfolders but do not know that OmegaT assigned properties to those segments.

I'll change that along with the docs.

PR #373, opened 2022-12-14
MT connectors: cache with JCache

miurahr:
Replace MT connectors cache with JCache mechanism

## Pull request type

- Bug fix -> [bug]

## What does this PR change?

- Add Caffeine cache library and its jcache module
- Use jcache on MT cache
- Avoid OOM

## Other information

JCache is a standard API in Java JSR-107.

Successor of #369

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -194,6 +194,14 @@ dependencies {
         // JSON parser
         implementation "com.fasterxml.jackson.core:jackson-core:2.13.4"
         implementation "com.fasterxml.jackson.core:jackson-databind:2.13.4.2"
+
+        implementation("com.github.ben-manes.caffeine:caffeine:2.9.3") {
+            attributes {
+                attribute(Bundling.BUNDLING_ATTRIBUTE, project.objects.named(Bundling.class, Bundling.EXTERNAL))
+            }
+        }
+        implementation("com.github.ben-manes.caffeine:jcache:2.9.3")
+
     }
 
     // Test dependencies
--- a/docs_devel/docs/26.Caching.md
+++ b/docs_devel/docs/26.Caching.md
@@ -1,0 +1,51 @@
+# Caching
+
+## Caching 
+
+Java supports jcache layer for MT connectors and/or other drivers.
+OmegaT uses Caffeine cache library for the purpose.
+
+### How to create cache object
+
+Here is complete example how cache is used.
+
+
+    public class Exmaple {
+        private final Cache<String, String> cache;
+
+        public Example() {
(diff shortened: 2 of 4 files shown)


t-cordonnier (commented):
When I test it I receive javax.cache.CacheException: Cache MachineTranslatorCache already exists
The problem is that there will be one cache per MT engine but all share the same name because you defined it in BaseTranslate: either we use only one static MachineTranslatorCache, or, which I prefer, we include the instance name (can be the class name as long as MT engines are all singletons) in the cache name.

Another point: the example in the MD file is not correct, if the cache is defined in BaseTranslate class then it cannot be in one of its subclasses.

miurahr:
@t-cordonnier I don't know well about jcache, could you suggest changes?

t-cordonnier:
For the moment, only the changes I mentioned in previous message.
Except the exception, looks good, but I also discover JCache so maybe I will have something more to tell later

miurahr:
There is no maximum sizes configuration with jcache API. Is it acceptable?
 
Previous PR #369 removes eldest entry when cache reached a specified size.

miurahr:
@t-cordonnier any progress?

t-cordonnier:
Tested again, with the correction now the result appears correctly.

> There is no maximum sizes configuration with jcache API. Is it acceptable?

JCache is not incompatible with engine-specific configuration. For example with Caffeine you would have
com.github.benmanes.caffeine.jcache.configuration.CaffeineConfiguration<K,V>
which contains specification of size.
If really you want to make it configurable we could define a kind of XML file format which defines the engine to use and then, depending on the engine, specific parameters. If  the code is correctly written, migration to another engine could be done by modifying only the XML file and not the code, which contains references either to JCache classes or creation of configuration instances using reflection (i.e. via Class.forName("CaffeineConfiguration").newInstance())
Now think about the following: do you want to configure the cache in the code (not in Java, but a config file present in the JAR) or do you really expect the user to know what should be the ideal cache size for his usage?

Regarding the reason why this is not in the standard I found this:
https://stackoverflow.com/questions/37289599/limit-cache-size-using-jcache-configuration-api

> Previous PR https://github.com/omegat-org/omegat/pull/369 removes eldest entry when cache reached a specified size.

And what will your own abstraction layer do if you want to implement a new engine which does not have any kind of size parameter?

miurahr:
Thank you for pointing the method I can use.
I've updated to use CaffeineConfiguration to set expiry and maximum size of cache.

brandelune:
Do we set a milestone here ?

miurahr:
> Do we set a milestone here ?

You can do it.

PR #376, opened 2022-12-17
Add new start screen

brandelune:
This new start screen simplifies the onboarding process by:

- stating right away where the Help/user manual is
- giving only 6 basic steps to use OmegaT
- clearly stating the freedoms granted to users


Kazephil:
I'm starting to think there might be something wrong with my system...

I can follow the logic you're using, and I'm sure you tested it and made sure it worked, but…

…for some reason, when I build the documentation on this branch, the `First_Steps.xml` file doesn't build into the `first_steps.html` file on my system, and the pane remains blank.

All the checks pass, and the application builds fine, with the title of the Editor pane showing up as First Steps, but nothing appears in the pane itself.

I'll have to see if I can break down the documentation build process to find where the problem is.

brandelune:
All those PRs are based on master. Maybe that is related.

Kazephil:
Just figured it out.
I was building only the html5 documentation, which doesn't include the First Steps page.

I got it to work, and will review it later this evening.

doc_src/en/First_Steps.xml
```diff
@@ -0,0 +1,91 @@
...
+	  <listitem>
+		<para>If you want to know more, read the user manual.</para>
+	  </listitem>
+	</orderedlist>
+	</para>
+  </section>
+
+  <section>
+	<title>Freedoms</title>
+	<para>OmegaT is a free translation memory solution for professional
```
Kazephil, line 65: Replace "translation memory" with "computer-assisted translation (CAT)".
(The abbreviation isn't strictly necessary, but I prefer to include it. That's purely preference, though.)

doc_src/en/First_Steps.xml
```diff
@@ -0,0 +1,91 @@
...
+	<guimenu>> About</guimenu>. The license grants you the following freedoms:
+	
+	<itemizedlist>
+	  <listitem>
+		<para>The freedom to run OmegaT as you wish, for any purpose.</para>
+	  </listitem>
+
+	  <listitem>
+		<para>The freedom to study how OmegaT works, and change it so it does
+		your computing as you wish.</para>
```
Kazephil, line 77: I would change "and change it so it does your computing as you wish." to something closer to the GNU phrasing: "and customize it to suit your needs."

Kazephil (changes requested):

Kazephil:
I'm proposing minor tweaks, but it looks good overall.

The Readme on building the docs will also have to be updated to reflect the "instant start" --> "first steps" update.

miurahr:
@Kazephil  could you confirm?

Kazephil:
> @Kazephil could you confirm?

@miurahr 
I was waiting for @brandelune to confirm he was OK with the "one last tweak" above.

brandelune:
> > @Kazephil could you confirm?
> 
> @miurahr I was waiting for @brandelune to confirm he was OK with the "one last tweak" above.

Oh, I had missed that one.

Basically, that's a quote from the GPL site, so computing needs in board terms *but* specifically applied to OmegaT. So I think we're fine.

https://www.gnu.org/philosophy/free-sw.html#four-freedoms

brandelune:
Nice touch ;-)

Kazephil:
> Basically, that's a quote from the GPL site, so computing needs in board terms _but_ specifically applied to OmegaT. So I think we're fine.

OK. It still feels slightly off, so I just proposed a much more minor "specifically applied to OmegaT" variation in the same spirit as replacing "program" with "OmegaT".

Kazephil (approved):

Kazephil:
@miurahr 
We're good to go with this one!

miurahr:
> @miurahr We're good to go with this one!

Please also update `build.gradle`  that read `<languages>/InstantGuide.xml` @Kazephil

brandelune:
@Kazephil I'm also working on the build.gradle script to fix another issue...

Kazephil:
> @Kazephil I'm also working on the build.gradle script to fix another issue...

@brandelune 
OK. I can leave the `build.gradle` fix for `InstantGuide.xml` --> `FirstSteps.xml` up to you, then?

brandelune:
@miurahr The build process now works fine with https://github.com/omegat-org/omegat/pull/403. Please review.

brandelune:
I'll be merging all that with #403 to apply the fix and so that you can review and commit.

brandelune:
@miurahr Please use this branch to fix all the build issues that came from the new "First Steps" files.

brandelune:
This PR does this:

- add a new start screen with minimal process info and licensing info
- builds the document set to fit the current UI
- adapt the CSS to the old manuals so that they look reasonably similar to the new one
- create a multilingual index that includes languages that only have an old Instant Start tutorial

brandelune (commented):
I think we should stick to the original. The GPL even as a summary can't be changed by us.

Kazephil:
> I think we should stick to the original. The GPL even as a summary can't be changed by us.

In that case, we should also stick to "the program" rather than replacing it with "OmegaT".

As far as I can tell, though, the text of the four freedoms isn't part of the text of the license, so I think we can adapt it to mâle it OmegaT-specific. We can always add an "adapted to OmegaT" section to the paragraph that introduces the freedoms.

brandelune:
> > I think we should stick to the original. The GPL even as a summary can't be changed by us.
> 
> In that case, we should also stick to "the program" rather than replacing it with "OmegaT".

You're correct that the 4 freedoms are not part of the GPL. It's just a framework to check whether a license is a free license or not.

Here, the license applies to OmegaT (and associated files), so I don't think changing "the program" to "OmegaT" is a problem.

But limiting "computing needs" to "CAT computing needs" by using the "CAT" specifier is limiting the user freedom, so I don't think it is appropriate.

Kazephil:
> But limiting "computing needs" to "CAT computing needs" by using the "CAT" specifier is limiting the user freedom, so I don't think it is appropriate. 

That's an… interesting perspective.
I still find it the highly specific "Omegat" and very broad "computing" juxtaposition slightly jarring, but I guess someone out there might want to turn OmegaT into the next Emacs that does everything except make your coffee! :grin:

brandelune:
I guess that means that one could use parts of the code to do something totally unrelated to CATing. But Emacs has not yet managed to provide coffee so I guess we should not have too many expectations with "our" code.

Kazephil:
> But Emacs has not yet managed to provide coffee

Appliances keep getting needlessly "smarter", so it might just be a matter of time! :open_mouth:

Kazephil:
Merged into `master`.

PR #377, opened 2022-12-18
Reorganize the CSS stylesheet.

brandelune:
The CSS stylesheet has been reorganized to better identify parts that can be later modified when needed.

A number of small modifications that correspond to the recent manual review have also been added.


Kazephil:
Overall, I like this version better. It looks very good.

Two things:
- The CSS file in the `doc_src` folder is now `OmegaT_webhelp.css`, but the relevant `build.xml` section is still:
   ```xml
   <target name="css">
      <copy file="OmegaT.css"
            tofile="../docs/${language}/OmegaT_webhelp.css" />
    </target>
   ```
  The build fails because `OmegaT.css` can't be found.

- The font size for text marked with some form of literal tag (<filename>, <code>, and <literal>, for example) feels too big compared to the surrounding text. I think keeping it at the same size would be fine.

brandelune:
I think that's because there is a discrepancy with this file and the PRs that have not yet been merged.

brandelune:
Let's check the CSS when we have all the elements merged. We should be able to see the whole thing and make modifications if necessary.

brandelune:
I'm closing this one because I cherry-picked into #376.

PR #280, opened 2022-10-03
Backup project file to enable reuse in case of crash

t-cordonnier:
## Pull request type

Please check the type of change your PR introduces:

- [X] Bug fix
- [ ] Feature
- [ ] Documentation
- [ ] Build and release changes
- [ ] Other (describe below)

## Which ticket is resolved?

https://sourceforge.net/p/omegat/mailman/message/37715643/

This patch creates a backup of omegat.project; later, if OmegaT does not succeed to read the project file for any reason (usually because it was not correctly saved after a crash) it will re-use the backup file instead

PR #380, opened 2022-12-18
Reference to the omegat.project backup feature

brandelune:
The omegat.project backup feature was added by @t-cordonnier.

https://github.com/omegat-org/omegat/pull/280

I added reference to it in the documentation.

### DIFF
--- a/doc_src/en/HowTo_RestoreYourData.xml
+++ b/doc_src/en/HowTo_RestoreYourData.xml
@@ -15,7 +15,7 @@
 	backups</title>
 
 	<para>OmegaT creates backups of your project settings to ensure it can
-	retrieve them in cas of problem. See the <link
+	retrieve them in case of problem. See the <link
 	linkend="project.folder.omegat.project.file"
 	endterm="project.folder.omegat.project.file.title"/> section for
 	details.</para>
--- a/doc_src/en/HowTo_RestoreYourData.xml
+++ b/doc_src/en/HowTo_RestoreYourData.xml
@@ -13,7 +13,13 @@
   <section id="how.to.restore.your.data.automatic.backup">
 	<title id="how.to.restore.your.data.automatic.backup.title">Automatic
 	backups</title>
-	
+
+	<para>OmegaT creates backups of your project settings to ensure it can
+	retrieve them in cas of problem. See the <link
+	linkend="project.folder.omegat.project.file"
+	endterm="project.folder.omegat.project.file.title"/> section for
+	details.</para>
+	  
 	<para>OmegaT regularly and automatically stores all your progress in the
 	<link linkend="project.folder.project.save.tmx"
 	endterm="project.folder.project.save.tmx.title"/> file located in the
--- a/doc_src/en/OmegaT5_ProjectFolder.xml
+++ b/doc_src/en/OmegaT5_ProjectFolder.xml
@@ -150,9 +150,9 @@
 		  <para>It also acts as an <emphasis>address book</emphasis> that
 		  specifies where project resources are located.</para>
 
-		  <para>OmegaT also automatically creates an
-		  <filename>omegat.project.bak</filename> to restore your project
-		  settings in case of need.</para>
+		  <para>OmegaT also creates a backup of that file
+		  (<filename>omegat.project.bak</filename>) and will automatically use it to
+		  restore your project settings in case of need.</para>
 		</listitem>
 	  </varlistentry>
 	</variablelist>
--- a/doc_src/en/OmegaT5_ProjectFolder.xml
+++ b/doc_src/en/OmegaT5_ProjectFolder.xml
@@ -143,10 +143,10 @@
 		<term id="project.folder.omegat.project.file.title">omegat.project</term>
 		<listitem>
 		  <para>This file contains the project parameters defined in the <link
-		  linkend="dialogs.project.properties">project properties</link>: source and
-		  target languages, their respective tokenizers, the type of segmentation
-		  used, etc.</para>
-
+		  linkend="dialogs.project.properties">project properties</link>, such as
+		  the source and target languages, their respective tokenizers and the type of
+		  segmentation, etc.</para>
+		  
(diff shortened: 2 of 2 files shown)


doc_src/en/HowTo_RestoreYourData.xml
```diff
@@ -13,7 +13,13 @@
   <section id="how.to.restore.your.data.automatic.backup">
 	<title id="how.to.restore.your.data.automatic.backup.title">Automatic
 	backups</title>
-	
+
+	<para>OmegaT creates backups of your project settings to ensure it can
+	retrieve them in cas of problem. See the <link
```
miurahr, line 18:
```suggestion
	retrieve them in case of problem. See the <link
```

doc_src/en/OmegaT5_ProjectFolder.xml
```diff
@@ -4,143 +4,159 @@
...
+		  linkend="dialogs.project.properties">project properties</link>: source and
+		  target languages, their respective tokenizers, the type of segmentation
+		  used, etc.</para>
+
+		  <para>It also acts as an <emphasis>address book</emphasis> that
+		  specifies where project resources are located.</para>
+
+		  <para>OmegaT also automatically creates an
+		  <filename>omegat.project.bak</filename> to restore your project
+		  settings in case of need.</para>
```
miurahr, line 155: `in case of need` can be improved by adding what is a case.

miurahr (changes requested):
typo and ambiguous expression.

doc_src/en/OmegaT5_ProjectFolder.xml
t-cordonnier: The main use case is if OmegaT crashes while the file was locked, because in some OS (Windows, but maybe others) the original omegat.project may be empty if not properly closed after the end of your work. 
I let you decide whenever it is useful to explain this point in the documentation or not.

t-cordonnier: "to restore your project settings in case of need"
I would say instead "to be sure OmegaT can open your project even if the original file is corrupted". The use case I describe in the previous message shows us that this has nothing to do with the contents of the file itself (it is independent on the fact that you changed options or not) but with the locking mechanism which can potentially corrupt the file.
But maybe I think like a developer, I let you decide which formulation is the most clear for an user.

brandelune: I thought about that and since it is Omegat that does the thing (unlike the TM backups), I decided that it was not necessary to give more details. But I'll add something.

t-cordonnier: What do you call "the thing"? Creation of the backup file (and display a message if it is used) is made by OmegaT but it is in order to solve a potential problem with locking mechanism, and the locking mechanism is done by the OS, reason why its behaviour can change from one OS to another.
I simply meant that I never had the case where the file contains something wrong (such as an incomplete XML or a wrong character inside), the only use case was when the file looks completely empty because locking mechanism made it look like that.
And also I remember that yesterday somebody asks to explain the case, so did I. I don't know why the question is not here anymore, did you remove or rewrite one message?

brandelune: The "thing" is the operation on the backup file to restore the settings. In this case, OmegaT handles everything. In the TM case, it is the user who has to manipulate the backup data, rename it, etc.

I think the question (by @miurahr) has been removed because it was in the conversation that was solved by my correction, and GitHub does not display it anymore.

But I do understand what you wrote.

t-cordonnier: OK I see. I had not understood the relation with the backups of project_save.tmx
Maybe the possibility also to manage them in the UI may be considered for the future (the more: we have the SuperTmxMerge window to display the differences, if needed). But that is out of topic.

brandelune: We need to put that in a RFE (if that's not already there)...

doc_src/en/OmegaT5_ProjectFolder.xml
```diff
@@ -4,143 +4,159 @@
...
+		</listitem>
+	  </varlistentry>
+
+	  <varlistentry id="project.folder.omegat.project.file">
+		<term id="project.folder.omegat.project.file.title">omegat.project</term>
+		<listitem>
+		  <para>This file contains the project parameters defined in the <link
+		  linkend="dialogs.project.properties">project properties</link>: source and
+		  target languages, their respective tokenizers, the type of segmentation
+		  used, etc.</para>
```
Kazephil, line 148: I would change "...project properties: source and target languages, their respective tokenizers, the type of segmentation used, etc."

to

"... project properties, such as the source and target languages, their respective tokenizers and the type of segmentation."

Everything else looks good!

Kazephil (changes requested):

doc_src/en/OmegaT5_ProjectFolder.xml
```diff
@@ -4,143 +4,159 @@
...
+		</listitem>
+	  </varlistentry>
+
+	  <varlistentry id="project.folder.omegat.project.file">
+		<term id="project.folder.omegat.project.file.title">omegat.project</term>
+		<listitem>
+		  <para>This file contains the project parameters defined in the <link
+		  linkend="dialogs.project.properties">project properties</link>, such as
+		  the source and target languages, their respective tokenizers and the type of
+		  segmentation, etc.</para>
```
Kazephil, line 148: The ",etc." at the end of the sentence is redundant and needs to be removed.

Kazephil (changes requested):

brandelune:
I think you can make proposals directly here and I just need to click on a button to accept them.

Kazephil (approved):
I'll look into those direct proposals next time.

This one is good to go!

miurahr:
Rebased to `master`, squash commits and merged into `master`

PR #381, opened 2022-12-18
Document the Select Source Text feature

brandelune:
The feature added in https://github.com/omegat-org/omegat/pull/344 was not documented.


kosivantsov:
Thank you for writing up the explanation of the feature. I'm happy with it

miurahr:
merged into `master`.

PR #384, opened 2022-12-18
Document the new reload feature.

brandelune:
Document the new reload feature from https://github.com/omegat-org/omegat/pull/304.


miurahr:
merged into `master`.

PR #386, opened 2022-12-19
New default for OmegaT 5.8/6.0

brandelune:
Include:
- new layout taking into account bigger screens + adding new open panes
- bigger font size
- new default language pair, less confusing than EN-US/EN-GB
- all TM subfolders created
- all script subfolders created + some doc
- minimum TM threshold = 100
- \d+ as default custom tag
- a number of editor/view defaults set to true
- MT auto fetch set to false
- search history size set to 50

I need to make changes to the documentation (basically remove parts, remove screenshots since they were here to show potential modifications)


brandelune:
It looks like [1a5fa35](https://github.com/omegat-org/omegat/pull/386/commits/1a5fa3538d82c7adde446ecb75b97aa8abc8062b) is not properly set. The regular expression appears in the preference, but I need to click on "reset defaults" to enable it. It looks like it is not compiled or something. And I'm not sure what to check to accomplish that.

@t-cordonnier, could you help me understand the issue?

t-cordonnier (changes requested):
Line 116 should be

panel.customPatternRegExpTF.setText(Preferences.getPreferenceDefault(Preferences.CHECK_CUSTOM_PATTERN, Preferences.CHECK_CUSTOM_PATTERN_DEFAULT));

When you have a default value the method is Preferences.getPreferenceDefault with two parameters, while Preferences.getPreference has only one
Preferences.CHECK_CUSTOM_PATTERN is the name of the property
Preferences.CHECK_CUSTOM_PATTERN_DEFAULT is the default value

brandelune:
It looks like it is not sufficient to enable the regex as a regex, since it clearly appears in the field...

src/org/omegat/util/Preferences.java
```diff
@@ -215,7 +217,8 @@ public final class Preferences {
     public static final String MARK_TRANSLATED_SEGMENTS = "mark_translated_segments";
 
     public static final String MARK_AUTOPOPULATED = "mark_autopopulated";
-
+    public static final boolean MARK_AUTOPOPULATED_DEFAULT = true;
```
miurahr, line 220: @brandelune when you want to define a new default boolean in Preferences class, why not set ALL the defaults?
There are many direct value in `restoreDefaults` methods.

src/org/omegat/gui/preferences/view/EditingBehaviorController.java
```diff
@@ -72,10 +72,10 @@ private void initGui() {
 
     @Override
     protected void initFromPrefs() {
-        panel.defaultRadio.setSelected(!Preferences.isPreference(Preferences.DONT_INSERT_SOURCE_TEXT));
-        panel.leaveEmptyRadio.setSelected(Preferences.isPreference(Preferences.DONT_INSERT_SOURCE_TEXT));
+        panel.defaultRadio.setSelected(Preferences.isPreference(Preferences.DONT_INSERT_SOURCE_TEXT));
+        panel.leaveEmptyRadio.setSelected(!Preferences.isPreference(Preferences.DONT_INSERT_SOURCE_TEXT));
```
miurahr, line 76: It looks better to introduce a `Prefrences#isNotPreference` method other than calling `!Preferences.isPrefernce` for better recognition of logics.

miurahr:
Please don't merge master. I will handle the commits  when the topic work is finished. @brandelune

src/org/omegat/util/Preferences.java
brandelune: Because I wanted to modify only the minimum set.

We discussed with @t-cordonnier that the way preferences were set in OmegaT would require a major overhaul, but I don't have the skills to do that (I barely understood what I was doing this time...)

It's something I'd be interested in working on, but not at the moment.

src/org/omegat/gui/preferences/view/EditingBehaviorController.java
brandelune: I understand. See my comment above.

I have 2 concerns now:

1. The regex I set is not taken into account by default in OmegaT even though it is displayed, there is a need to "reset the defaults" to have OmegaT take it into account.
2. The whole set of preferences needs an overhaul, see comment above, but I'd rather work on that after the release, to not introduce more delays.

src/org/omegat/util/Preferences.java
miurahr: OK, I can understand it and will work on it in 6.1 versions.

src/org/omegat/gui/main/MainWindowUI.java
```diff
@@ -279,21 +279,19 @@ public static void initializeScreenLayout(MainWindow mainWindow) {
...
 
     /**
-     * Assume screen size is 800x600 if width less than 900, and 1024x768 if
-     * larger. Assume task bar at bottom of screen. If screen size saved,
-     * recover that and use instead (18may04).
+     * (20nov22) it looks like we don't have to worry too much about default
+     * size because if the screen is too small for the setting, OmegaT
+     * will just use the whole window size. So I guess we just have to
+     * consider a maximum useable width for big screens.
+     * Let's assume the "standard" is 1920×1080 for "pro" laptops
```
miurahr, line 286:
```suggestion
     * Let's assume the "standard" is 1920x1080 for "pro" laptops
```
Please don't use a non-Ascii character.

src/org/omegat/gui/main/MainWindowUI.java
```diff
@@ -279,21 +279,19 @@ public static void initializeScreenLayout(MainWindow mainWindow) {
     }
 
     /**
-     * Assume screen size is 800x600 if width less than 900, and 1024x768 if
-     * larger. Assume task bar at bottom of screen. If screen size saved,
-     * recover that and use instead (18may04).
+     * (20nov22) it looks like we don't have to worry too much about default
```
miurahr, line 282:
```suggestion
     * (20nov22) We don't  worry  about default
```

src/org/omegat/gui/main/MainWindowUI.java
```diff
@@ -279,21 +279,19 @@ public static void initializeScreenLayout(MainWindow mainWindow) {
...
+     * size because if the screen is too small for the setting, OmegaT
+     * will just use the whole window size. So I guess we just have to
+     * consider a maximum useable width for big screens.
+     * Let's assume the "standard" is 1920×1080 for "pro" laptops
+     * https://laptoping.com/laptop-display-specs-explained.html.
```
miurahr, line 287:
```suggestion
```

I think we don't need the reference. It is Ok to note our decision.

miurahr (changes requested):
Please update the comment

src/org/omegat/gui/main/MainWindowUI.java
```diff
@@ -279,21 +279,18 @@ public static void initializeScreenLayout(MainWindow mainWindow) {
...
         GraphicsEnvironment env = GraphicsEnvironment.getLocalGraphicsEnvironment();
         Rectangle scrSize = env.getMaximumWindowBounds();
-        if (scrSize.width < 900) {
-            // assume 800x600
-            return new Rectangle(0, 0, 580, 536);
-        } else {
-            // assume 1024x768 or larger
-            return new Rectangle(0, 0, 690, 700);
-        }
+        return new Rectangle(210, 0, 1500, 1080);
```
Kazephil, line 293: What's the point of assigning the `scrSize` variable since you're not using it?

Did you mean to assign `Rectangle(210, 0, 1500, 1080)` or the result of `env.getMaximumWindowBounds()` if the latter is smaller?

src/org/omegat/gui/preferences/view/EditingBehaviorController.java
Kazephil: I can't quite follow how the set of preferences works either, but based on my brief testing:
1. The above block sets the radio button in the Preferences dialog to "Leave the dialog empty", but it does not actually set the preference: the source segment is still inserted automatically when I enter a new segment. The `prefs` file generated by this version of the does not contain the "<wf_noSourceText>true</wf_noSourceText>" entry I get in the `prefs` file when I set the preference in a build made from master. (I couldn't figure out where to set it, though.)

2. If anyone is willing to explain to me how the current system works, I'd be happy to try to help with this after the release.

Kazephil (commented):
Final note: I can confirm that the custom tag regex is not applied immediately.

src/org/omegat/gui/main/MainWindowUI.java
brandelune: I just left the code there. Obviously it can be removed.

brandelune: Could you make a proposal in the GitHub interface as Hiroshi does ?

Kazephil: Did I understand your intent correctly?

If so, I'll try to propose appropriate code later this evening.

brandelune:
> Line 116 should be
> 
> panel.customPatternRegExpTF.setText(Preferences.getPreferenceDefault(Preferences.CHECK_CUSTOM_PATTERN, Preferences.CHECK_CUSTOM_PATTERN_DEFAULT));
> 
> When you have a default value the method is Preferences.getPreferenceDefault with two parameters, while Preferences.getPreference has only one Preferences.CHECK_CUSTOM_PATTERN is the name of the property Preferences.CHECK_CUSTOM_PATTERN_DEFAULT is the default value

(Not sure if you saw my reply, so I'm putting it here)

It looks like it is not sufficient to enable the regex as a regex, since it clearly appears in the field...

miurahr (approved):

brandelune:
@miurahr Please do not merge. There is still the issue of the regex to be solved.

src/org/omegat/gui/main/MainWindowUI.java
brandelune: There is no need to test the screen size, since OmegaT will use the whole screen if the size is too small.

The point is to set a size that is not too small for big screens. Smaller screens don't have this problem.

Kazephil: OK.

I've just proposed an alternative that eliminates the check and removes redundant code.
Please take a look.

Kazephil:
Since there's been a bit of a lull on this branch, I thought I'd summarize some of my observations and other issues I'm currently aware of regarding this topic.

_Quick notes/questions:_
1. As originally noted by Jean-Christophe, the "flagged text" regex is entered in the Preferences dialog, but the preference is not actually activated when OmegaT is run.
2. Similarly, on my system, the new default of not inserting the source text doesn't work. The "Leave the segment empty" option is checked, but not activated. Can anyone else confirm this?
3. Default font size: was 13 the consensus? I think 14 would be better, but that may just be due to my (not necessarily rational) preference for even-numbered font sizes.
4. Another personal preference, but I think it would look better if the initial size of the "Add Source Files" dialog should be about half that of the main window, and slightly offset from it. I'm willing to look into implementing that if it's not overly complicated and others also think it would be an improvement. (If it doesn't bother anyone else, it can remain as is, of course.)

Regarding 1 and 2 above, I tried to follow the logic and find where the preference is set, but the code is too advanced for my level of understanding. When I compared the preferences files generated by the build for these new defaults and the one from a build where the options are activated, I found that the:
-`<wf_noSourceText>true</wf_noSourceText>` and
- `<tagValidation_customPattern>\d+</tagValidation_customPattern>`
entries are not generated in the preferences file created by the new defaults build.

So far, all my efforts at trying to figure out where those the values are stored/produced in the code, and where those preference file entries are created, have been in vain.

I don't know if the current convoluted preferences code is the result of an accretion of changes, but I have to naively wonder if it wouldn't be simpler (after the 5.8/6.0 release is completed, of course) to create a `defaults.prefs` file and apply it if `omegat.prefs` does not exist or has been corrupted. Are there any technical reasons that make it desirable to handle setting the initial preferences within the code?

Just a few thoughts and observations. Suggestions and comments welcome!

brandelune:
Thank you Philippe.

> _Quick notes/questions:_
> 
> 1. As originally noted by Jean-Christophe, the "flagged text" regex is entered in the Preferences dialog, but the preference is not actually activated when OmegaT is run.

Thank you for confirming.

> 2. Similarly, on my system, the new default of not inserting the source text doesn't work. The "Leave the segment empty" option is checked, but not activated. Can anyone else confirm this?

I confirm.

> 3. Default font size: was 13 the consensus? I think 14 would be better, but that may just be due to my (not necessarily rational) preference for even-numbered font sizes.

:-) I'm fine with 14 :-)

> 4. Another personal preference, but I think it would look better if the initial size of the "Add Source Files" dialog should be about half that of the main window, and slightly offset from it. I'm willing to look into implementing that if it's not overly complicated and others also think it would be an improvement. (If it doesn't bother anyone else, it can remain as is, of course.)

Why not give that a try ?

> Regarding 1 and 2 above, I tried to follow the logic and find where the preference is set, but the code is too advanced for my level of understanding. When I compared the preferences files generated by the build for these new defaults and the one from a build where the options are activated, I found that the:
> 
> * `<wf_noSourceText>true</wf_noSourceText>` and
> * `<tagValidation_customPattern>\d+</tagValidation_customPattern>`
>   entries are not generated in the preferences file created by the new defaults build.
> 
> So far, all my efforts at trying to figure out where those the values are stored/produced in the code, and where those preference file entries are created, have been in vain.

... Same here.

Especially the "activation" part of the 2 above items.

> I don't know if the current convoluted preferences code is the result of an accretion of changes, but I have to naively wonder if it wouldn't be simpler (after the 5.8/6.0 release is completed, of course) to create a `defaults.prefs` file and apply it if `omegat.prefs` does not exist or has been corrupted. Are there any technical reasons that make it desirable to handle setting the initial preferences within the code?

That's definitely something we'll need to consider after the release.

Kazephil:
> Why not give that a try ?

I*ll see what I can do. :smile_cat:

brandelune:
I am going to split this PR in smaller parts to make it easier to integrate. Apologies for putting everything together here.

PR #387, opened 2022-12-20
Add PDFs to generated "User Manual" index.html - POC for documentation

damien-rembert:
## Pull request type
[documentation]

## Which ticket is resolved?

- SourceForge Documentation task 379 The localized site documentation page does not link to the correct PDF
  * https://sourceforge.net/p/omegat/documentation/379/

## What does this PR change?

This PR tries to address the lack of links for the most recent localised PDFs on the OmegaT website by adding a PDF link to the HTML page that can be generated using Gradle to get the latest version: 
[ SF Doc 379 The localized site documentation page does not link to the correct PDF](https://sourceforge.net/p/omegat/documentation/379/)

This is more of a proof of concept/work in progress than a real solution at the moment.
It adds a link to the PDF of the User Manual on [the page that links to its most recent HTML version](https://omegat.sourceforge.io/manual-standard/) which would then look like the attached screenshot. 
This implementation assumes that the most recent PDF version should be the same as the HTML, which does not seem to be the case.

## Other information

Screenshot 2022-12-20 at 18 44 57

### DIFF
--- a/doc_src/index_template.html
+++ b/doc_src/index_template.html
@@ -14,7 +14,11 @@
 <table>
 <% languages.each { lang -> %>
     <tr>
-        <td><a href="${lang.code}/index.html">${lang.name}</a></td>
+        <td>
+            <p>${lang.name}:</p>
+        </td>
+        <td><a href="${lang.code}/index.html">HTML/Web</a></td>
+        <td><a href="https://sourceforge.net/projects/omegat/files/OmegaT%20-%20Standard/OmegaT%20${lang.version}/OmegaT_documentation_${lang.code}.PDF/download">PDF</a></td>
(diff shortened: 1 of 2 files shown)


damien-rembert:
I don't think I can add the Documentation label as required in the template? If people with the necessary permissions could do it for me that would be great, thanks in advance.

miurahr:
Partially merged into `master`.
@kosivantsov  could you propose an update of `docs/` folder with the all doc_src changes?

kosivantsov:
> @kosivantsov could you propose an update of `docs/` folder with the all doc_src changes as another PR?

I don't think I have any changes to propose.

damien-rembert:
AFAIK the necessary bits have been merged and this PR has no reason to remain open.

PR #396, opened 2022-12-25
index.xml is a generated file, it should not be commited

brandelune:
index.xml is generated by the build process. It should not be commited to the repository.


damien-rembert:
If it should never be committed, it might be good to add it to the .gitignore file?

brandelune:
Does that work better this way?

miurahr:
You are recommended to create `doc_src/.gitignore` for the purpose.

miurahr:
rebase and merged into `master`

PR #397, opened 2022-12-26
Test: util.xml: improve coverage of test case

miurahr:
## Pull request type

- Other (describe below)

## Which ticket is resolved?
N.A.

## What does this PR change?

- Add a test case for XMLStreamReader


miurahr:
The class is to be deprecated, so no need to add test.

PR #401, opened 2022-12-29
Check and fine-tune the remaining chapters.

Kazephil:
Revision of the pane chapter, consisting mainly of phrasing changes to tighten up wording, and fixing typos.

There is one major change:
In the note giving examples of modifications translators can make to the editor to adapt it to their workflow, I changed the text of  item (starting on Line #430) about inserting source text to match the new proposed default.

### DIFF
--- a/doc_src/en/Dialogs_ProjectProperties.xml
+++ b/doc_src/en/Dialogs_ProjectProperties.xml
@@ -9,8 +9,8 @@
   linkend="menus.project.properties" endterm="menus.project.properties.title"/>
   to open the dialog.</para>
   
-  <para>The dialog is used to set the initial project properties when creating a
-  new project and also to modify them later, after the project has been
+  <para>This dialog is used to set the initial project properties when creating 
+  a new project or to modify them later, after the project has been
   created.</para>
 
   <para>See the <link linkend="introduction.create.and.open.new.project"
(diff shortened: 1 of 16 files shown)


Kazephil:
I'll try to do finish the _Menus_ chapter and one or two sections in the _Windows and Dialogs_ chapter later today.

brandelune:
Thank you. It looks good to me.

Kazephil:
@brandelune

I've added revisions for the _Menu_ chapter. There aren't any major differences in content, but given the descriptions in the manual for moving through segment history, I thought it would be better to put the "Back in History" command _*before*_ the "Forward in History" command, and I changed the manual, as well as the `bundle.properties` and `MainWindowMenu.java` files accordingly.

This update does not yet reflect potential changes to the "... from /tm/auto" or "...from /tm/enforce" names currently under discussion on the development list. I'll update those when we reach a final decision.

brandelune:
> @brandelune
> 
> I've added revisions for the _Menu_ chapter. There aren't any major differences in content, but given the descriptions in the manual for moving through segment history, I thought it would be better to put the "Back in History" command __before__ the "Forward in History" command, and I changed the manual, as well as the `bundle.properties` and `MainWindowMenu.java` files accordingly.

Thank you.
 
> This update does not yet reflect potential changes to the "... from /tm/auto" or "...from /tm/enforce" names currently under discussion on the development list. I'll update those when we reach a final decision.

Would you mind making a summary ? :-)

Kazephil:
@brandelune

One more batch. This one also includes the changes to the Scripting section from the new defaults branch.

brandelune (approved):
Thank you !

brandelune (approved):
Thank you !

Kazephil:
> > This update does not yet reflect potential changes to the "... from /tm/auto" or "...from /tm/enforce" names currently under discussion on the development list. I'll update those when we reach a final decision.
> 
> Would you mind making a summary ? :-)

Sure. Very briefly:

- Use "pre-translated" instead of "from /tm/auto".
- Use "enforced" or "locked" instead of "from /tm/enforce".
   - "Enforced" is better because it can imply locking (and overwriting), while the reverse isn't true.

brandelune:
> > > This update does not yet reflect potential changes to the "... from /tm/auto" or "...from /tm/enforce" names currently under discussion on the development list. I'll update those when we reach a final decision.
> > 
> > 
> > Would you mind making a summary ? :-)
> 
> Sure. Very briefly:
> 
> * Use "pre-translated" instead of "from /tm/auto".
> * Use "enforced" or "locked" instead of "from /tm/enforce".
>   
>   * "Enforced" is better because it can imply locking (and overwriting), while the reverse isn't true.

Ok, and we're talking about the doc, right, not (?) about the UI in `View`, are we?
(sorry if I sound lost, I'm focusing on something else at the moment)

Kazephil:
> Ok, and we're talking about the doc, right, not (?) about the UI in `View`, are we?

Both, actually.

PR #403, opened 2023-01-02
fix the documentation building process

brandelune:
# Issue:
The new manual removed the "Instant Start Guide" from the manual and integrated the information in the Introduction section.

The results were:
- the Instant start is not built as the default Editor contents
This is addressed in https://github.com/omegat-org/omegat/pull/376

- The `message.xml` that was used to create an Appendix extra header was removed
This is addressed in https://github.com/omegat-org/omegat/commit/6e51ff6d074e25249730953c41b64b41cbeb48b2

Removing message.xml and references to it in `whc-toc.xsl` created issues with the documentation building process since it was related to the declaration of a `translation` variable that was used in other parts of the build process.

# Proposed fix:

- I removed all references to `translation` in the documentation build process
- I removed references to the Instant Start Guide since it does not exist anymore as an independent guide
- I removed an unused gendoc.bat file that was used in previous `ant` based documentation generation processes
- I removed the l10n packages creation tasks since they referred to the Instant Start Guide and are irrelevant now that all the l10n process in hosted on Github
- I fixed the resulting index layout

# Issues with the proposed fix:
## Independent `Instant Start Guides` are not build
Older versions of the manual used to have an independent "Instant Start Guide" for localization teams that could not translate the full manual. This file does not exist anymore and the build process does not have any replacement.

→ The language teams that only had an Instant Start Guide available are not considered by the build process and that file is thus removed from the multilingual documentation set.

## The new CSS is not adapted to fit the old manuals
The new CSS was created with the 5.8 manual anchors in mind and it will need to be adapted to the old manuals.

# Discussion:
## What do we do with the independent Instant Start Guides?
- Do we accept to remove them?
- Do we manually add them to the doc set?
- Do we modify the build system to check for their presence and add them when they are the only doc?
- Do we create a file that would play that role with the contents of the "Basic Steps" above (https://github.com/omegat-org/omegat/pull/376)?
- Do we keep the Instant Start in the documentation package so that languages that have only that can still use it?


doc_src/index_template.html
```diff
@@ -11,15 +11,14 @@
...
 <% languages.each { lang -> %>
     <tr>
         <td>
-            <p>${lang.name}:</p>
+            <p><a href="${lang.code}/index.html">${lang.name}</a></p>
         </td>
-        <td><a href="${lang.code}/index.html">HTML/Web</a></td>
-        <td><a href="https://sourceforge.net/projects/omegat/files/OmegaT%20-%20Standard/OmegaT%20${lang.version}/OmegaT_documentation_${lang.code}.PDF/download">PDF</a></td>
         <td>(<span class="${lang.status}">${lang.version}</span>)</td>
+        <td><a target="blank" href="https://sourceforge.net/projects/omegat/files/OmegaT%20-%20Standard/OmegaT%20${lang.version}/OmegaT_documentation_${lang.code}.PDF/download">PDF</a></td>
```
damien-rembert, line 21: Not sure if I'm missing something but just to be sure we're on the same page, my understanding is that this file builds [this page](https://omegat.sourceforge.io/manual-standard/) and the page that is currently online is not the version we're working on here as it has been edited recently and not built and uploaded since.

The version in the repo turned it into four cells that looked like that: 
-The language name (in plain text) |
-"HTML/Web" as a link to the HTML manual on omegat.sourceforge.io |
-"PDF" as a link to the pdf in the downloads on sourceforge |
-a version and status indicator

eg: French: | HTML/Web | PDF | (3.6.0)

This change would instead display:
-The language as a link to the HTML manual on omegat.sourceforge.io |
-a version and status indicator |
-"PDF" as a link to the pdf in the downloads on sourceforge

eg: French | (3.6.0) | PDF

If so, I think merging the language name and the link to the web version back into one cell makes things confusing now that there are two links:
Is the PDF not translated? And if it is, then where does the link that says French go? Why is there a version number in the middle?

I'm not happy with "Web/HTML" either (and to be honest, it was more of a question, offering two options) but I think we need to somehow indicate "this is the HTML version (hosted on this website)" and "this is the PDF version as a download". 
We could maybe add a header before line 15, indicating what each column contains?

damien-rembert:
I think offering Instant Start pages is better than nothing at all, if the instructions they contain are still correct. 

> Do we create a file that would play that role with the contents of the "Basic Steps" above ?

I think so but (somewhat going back to the idea of first run wizard/tour of OmegaT) maybe it would be nice to be able to dismiss it? With a tickbox at the bottom "display this screen next time" (that would be matched in the settings by "display the 'first steps' screen on next start"). And when this is not displayed you could have useful options instead (eg "Open recent...", "Open...", "Open Team Project", "Settings")

I will have a look at the documentation build and translation process but at the moment that's pretty much all I can say :)

doc_src/index_template.html
brandelune: We have an old problem with the various OmegaT versions and the way the manual is handled.

- The "standard" version comes with a very outdated "updated" manual.
   https://omegat.sourceforge.io/manual-standard/
- The "latest" version is recent code, but not documented.
   https://omegat.sourceforge.io/manual-latest/

The standard version manuals have not been built in a long time, which explains the discrepancy that you see online. That is ugly.

What we are doing now is **stop** having two versions. We'll always have a version that is properly documented and will organize so that the documentation is produced on time.

That's the reason why it's taking so much time to release this 5.8/6.0 (if you're not sure about the difference: they are the same but 5.8 will be based on Java 8 and 6.0 will be based on Java 11, only 6.0 will be advertised and 5.8 will only be mentioned for people who have _really old machines/OSs_).

That being said, the build process currently builds the "latest" version:
https://omegat.sourceforge.io/manual-latest/

The "Web/HTML" part is kind of unnecessary because this file is also the one that is included in the distribution and is displayed in your browser when you open the help menu in OmegaT. So, people kind of know that the underlined part is a link and will click on their language to have access to their localized manual.

Also, I'm not sure that we need to link to the PDF there. It would be better to keep the links to the PDFs for the omegat.org site, especially since we don't have yet a satisfactory PDF output due to plenty of reasons (check https://sourceforge.net/p/omegat/mailman/omegat-development/thread/4DEF6BE8-6B5E-4938-BCC1-8B16AC7120A8%40traduction-libre.org/#msg37751519).

The manual web hosting is a convenience and people who use it will (most probably) go/refer straight to the localized version that they need and not to the index. So that online index is not as important as its local clone, which is the reason why I prefer to keep its layout and the info it displays super simple.

brandelune:
> I think offering Instant Start pages is better than nothing at all, if the instructions they contain are still correct.

I've just changed the build process to:
- retain the Instant Start Guide (for the languages that only had that).
- add a condition in the build process to keep the relevant CSS files

So, basically, (except for the fact that the `whc` tasks now apply to all the manuals, but it works fine so we should not bother too much), we now have a working build process. **Sincere apologies for the hassle** (@miurahr).

What needs to be done, if you could help @damien-rembert, is add the languages that only have the ISG (and only those) to that index.html file.

I can see 2 ways to go about that:

- Split the sections
    - Keep the OmegaT Manual section on top, with the version numbers
    - Put a "OmegaT Instant Start Guide" below with only the languages that *only have that* (since it is included in the old manual otherwise and superseded by the Introduction in the new manual...)

- Mix the languages
    - the languages that have a manual keep the version number
    - the languages that only have the instant start do not have a version number

@kosivantsov, we need your opinion on that!

miurahr (approved):
LGTM now.

brandelune:
Don't merge. I'm moving all that to #376 and I'll close this PR.

miurahr:
I've rebased and merged into master.

brandelune:
> Don't merge. I'm moving all that to #376 and I'll close this PR.

Please remove the merge into master. I'm fixing all that into #376.

brandelune:
> I've just changed the build process to:
> 
> * retain the Instant Start Guide (for the languages that only had that).
> * add a condition in the build process to keep the relevant CSS files

> What needs to be done, if you could help @damien-rembert, is add the languages that only have the ISG (and only those) to that index.html file.

I managed to update the index in #376 so we should be good.

> * Mix the languages
>   
>   * the languages that have a manual keep the version number
>   * the languages that only have the instant start do not have a version number

I set a version property to 2.3.0 and added a new non-colored style for indexes that have that value.

PR #404, opened 2023-01-07
Clarify the Bidi related information

brandelune:
## Which ticket is resolved?

Documentation: https://sourceforge.net/p/omegat/documentation/400/

## What does this PR change?

- This PR reverts the use of `Bidi` instead of `Unicode` in the Bidi related features.
- It adds a small appendix to clarify the use of the various features.
- It adds better cross-reference between the manual parts.

### DIFF
--- a/doc_src/en/App_Bidi.xml
+++ b/doc_src/en/App_Bidi.xml
@@ -4,36 +4,35 @@
 <section id="app.bidi">
   <title id="app.bidi.title">Directional Formatting Characters</title>
 
-  <para>Bidi control characters can be found in the menu <link
+  <para>Bidi control characters are available from <link
   linkend="menus.edit" endterm="menus.edit.title"/><link
   linkend="menus.edit.insert.unicode.control.character"
-  endterm="menus.edit.insert.unicode.control.character.title"/> and may be used
+  endterm="menus.edit.insert.unicode.control.character.title"/>. They can be used
   to:</para>
 
   <itemizedlist>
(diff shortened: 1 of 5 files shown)


Kazephil (commented):
@brandelune 
Everything other than the proposed changes to the `App_Bidi.xml` file looks good to me.

brandelune (commented):
Thank you.

Kazephil:
I just noticed that all the bidi information is repeated in PRs #405 and #406. Can we either merge this one first or update and combine them into a single PR?

brandelune:
There are no problems with the merge. I just branched off the branches to base the new work on the old. It will not impact the merge (according to what I read on stackoverflow ;-).

Kazephil:
> There are no problems with the merge. I just branched off the branches to base the new work on the old. It will not impact the merge (according to what I read on stackoverflow ;-).

As long as we're careful about not making changes to the bidi content in the other branches before this one is merged, I guess there shouldn't be any problems. I'm just worried about my brain becoming more confused than Git... :open_mouth:

brandelune:
> As long as we're careful about not making changes to the bidi content in the other branches before this one is merged, I guess there shouldn't be any problems. I'm just worried about my brain becoming more confused than Git... 😮

The bidi contents in the other branches is linked to the various commits. If a commit is modified then it is modified for all the locations that refer to it.

kosivantsov (approved):
Looks good to me

PR #405, opened 2023-01-07
Use appropriate verbs in the View menu

brandelune:
A review of the View menu reached the conclusion that more appropriate verbs should be used.

## Which ticket is resolved?

https://sourceforge.net/p/omegat/documentation/401/

## What does this PR change?

- This PR modifies the items in the View menu with verbs that better describes the action
- It modifies the descriptions in the manual to reflect the changes
- It adds a reference to the Color preference so that the user knows where to change the colors used.


Kazephil (commented):
@brandelune 
I made very minor wording tweaks. If there are no problems with those, I think this PR is ready to get pushed to `master`.

Kazephil:
Merged into master.

PR #406, opened 2023-01-07
Replace "non-unique" with "repeated"

brandelune:
## Which ticket is resolved?

https://sourceforge.net/p/omegat/documentation/402/

## What does this PR change?

- This PR changes the "non-unique" strings into "repeated", in the manual and in the UI
- It also clarifies one wording in the Preferences, for "Mark all repeated segments"


miurahr:
why has this set of changes  BiDi things?

brandelune:
I branched off the bidi branch because it was more convenient to have the bidi modifications as a base.

I checked on stackoverflow and found that it did not impact the final merges and does not even require to merge in a particular order.

If you could give me instructions as to how to accomplish the same thing so that it looks the way you prefer, I'll do that.

Kazephil:
@brandelune
I've just sent a pull request that resolves the conflicts to your fork. Please check and update this PR.

Kazephil:
Merged into `master`.

PR #409, opened 2023-01-08
Add links to first steps guide and headers to index

damien-rembert:
## Pull request type

## Which ticket is resolved?

- Add the languages that only have the Instant Start Guide
  * Following discussion in #403 

## What does this PR change?

- If a language folder has an instant start guide file and does not have a manual, running `./gradlew geDocIndex` will add it to the index.html generated.

- Add a few HTML and CSS options to help make the table look nicer.


brandelune:
Excellent! Thank you!

Regarding the addition of a new property, I think it's better to check whether the instant start xml file is present since new translations will be coming _without_ the instant start (it's been removed from the current manual) and that would force us to manually remove the property.

Regarding the HTML header :-) I really think it is redundant if we have the link applied to the language. People are used to clicking on links and move to HTML pages.

damien-rembert:
I'm sorry to insist, but while I also think people know how to navigate html links, I still think that we're not providing them with clear information here.
There are two links by language so people have to pick one. Only it's not instantly clear what each link stands for, or their relationship: one is a language, one is a file format.

You could (if only for a second) wonder "are all the PDFs the same file?" Or even misread this as "English: v4 PDF" and assume that the only option is the PDF.

In a similar situation, my bank offers to "view" or "download" bank statements.

Otherwise I think either both formats need to be indicated or their relationship needs to be made explicit eg "English" for the html and "download as a PDF", "PDF alternative" or "PDF version". 

It's not a big thing and I'm not really happy with any of the options I'm offering but it just causes me to do a double take every time I see it so I'm trying to find a way that feels a bit smoother. But I'll forget all about it as soon as I'm not working on that file so no worries if you feel that the current format is the best option :)

brandelune:
Thank you for the explanations. If you think it's important, I'm fine with your decision.

My issue with the current proposal is that the PDF was never conceived on an equal stand with the html output, and that's the reason why we never linked to it in the index file.

We moved from HTML to DocBook as a documentation source format because of its ability to provide PDF output, but PDF is still another tool chain, with plenty of issues and a much higher learning curve to obtain a nice looking document. And I don't think anybody actually took the time to study how to improve the current output (check it for yourself, you'll see what I mean).

We still have an increasing number of important issues with the 5.8 (currently only EN) output.
https://sourceforge.net/p/omegat/documentation/399/

We have a number of options that are currently discussed, but have not reached a conclusion, to the point where I was even considering to _not_ provide a PDF output for the current version (what we have at the moment is really ugly).
http://doublet.jp/_omegat/en/OmegaT_documentation_en.PDF

So, my idea is that we make sure that the translated PDFs are properly linked to from the omegat.org site, for people who want them as a secondary option, but having them right in the online documentation that's facing the new user puts them on an equal stand with the HTML, which is going to greatly disappoint the people who download it. And I'm not really in favor of that, not in the absolute, but as long as we don't have a way to output a PDF that is as good as the HTML.

I hope you understand my position better now.

damien-rembert:
That makes a lot of sense, thanks for taking the time to explain.

brandelune:
And just like for the video, I'm willing to really investigate the issue after the current release :-) I think PDF as a format has some value and if properly created can look really nice.

damien-rembert:
The Instant-Start-Guide only languages do not depend on properties files now.
At the moment there are lots of duplicates as it doesn't try to make languages unique. 
It would be possible to do it but should not be required since only the languages without user manuals will have Instant Start Guides.
It does not look good at the moment but it might look better if it went like this (let's imagine we do not have a French User Manual):

```
español                                  | (3.0.2)    | PDF
français (First Steps only)    |(N/A)       | PDF
galego                                    | (3.3.2)    | PDF
```

In that case

brandelune:
It's ok to sort by language code. Wikipedia does that too.

brandelune:
It should be "instant start only" since the "first steps" file belongs to the new manual.

damien-rembert:
Here is what it looks like now. Again duplicates should not be a problem when languages that are not "Instant Guide Only" do not have it anymore. 
Screenshot 2023-01-10 at 13 32 33

brandelune:
Sorry for the hassle, but to avoid duplicates you should be able to test on the presence of the main manual xml file (OmegaTUsersManual_xinclude full.xml). If both the instant start and that file are present, there is no need to have a separate instant start.

damien-rembert:
I went about it in a slightly different way but Groovy is pretty new to me. But it works :)

brandelune:
Also, adding the "Instant Start only" string means that we have to manage the localization of that file... If we go that way, I'd add a "please help with the translation of the manual in your language", or something similar. But I don't want that to become a development rabbit hole... :-)

damien-rembert:
Since this file is built programmatically, that string could be anything we want, we just need to know what we want to display. We could display the title of each isg (which I imagine would be the localised version of instant start guide). We could also display a call for help to translators. We could also not display anything and consider that the lack of version number says enough.

brandelune:
> We could also not display anything and consider that the lack of version number says enough.

+1

damien-rembert:
Changed accordingly.
That's a lot of white space but I think that's for another time.
Screenshot 2023-01-11 at 14 12 25

Kazephil:
@damien-rembert
Since JC is happy with the contents of this PR, can you rebase your branch on the latest master to resolve the `build.gradle` conflict?

Once that's done, I can merge it into master.

damien-rembert:
Done!
Only the changes to `build.gradle` and `doc_src/index_template.html` are to be kept, the other files were only modified to illustrate how the change worked.

Kazephil:
> Done! Only the changes to `build.gradle` and `doc_src/index_template.html` are to be kept, the other files were only modified to illustrate how the change worked.

Thank you!
I've just pushed it to `master`.

Closing this PR.

PR #413, opened 2023-01-13
docs: App_Bidi: add missing period

miurahr:


Kazephil:
Nice catch!

Kazephil:
Merged into `master`.

PR #414, opened 2023-01-14
Filters: fix to catch exception when loading file and ignore when failure

miurahr:
When XML filters try to load invalid XML files when loading project, it raises fatal error and abort loading project.
It is not useful for ordinal translators who just accept source file.

This changes to show error message and continue to load files when the failure occurred. 

## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?

-  XML/DocBook filter breaks project load when XML has error  
  * https://sourceforge.net/p/omegat/bugs/1138/

## What does this PR change?

- Catch `TranlationException` in `RealProject#loadSourceFiles` and show message and ignore a file.
- Add test case that raise `TranslationExeption` when  visits invalid docbook source file.
- Spotless code formatting

### DIFF
--- a/src/org/omegat/core/data/RealProject.java
+++ b/src/org/omegat/core/data/RealProject.java
@@ -110,14 +110,15 @@ import org.omegat.util.gui.UIThreadsUtil;
 import gen.core.filters.Filters;
 
 /**
- * Loaded project implementation. Only translation could be changed after project will be loaded and set by
- * Core.setProject.
+ * Loaded project implementation. Only translation could be changed after
+ * project will be loaded and set by Core.setProject.
  *
- * All components can read all data directly without synchronization. All synchronization implemented inside
- * RealProject.
+ * All components can read all data directly without synchronization. All
+ * synchronization implemented inside RealProject.
  *
- * Since team sync is long operation, auto-saving was split into 3 phrases: get remote data in background, then rebase
- * during segment deactivation, then commit in background.
+ * Since team sync is long operation, auto-saving was split into 3 phrases: get
+ * remote data in background, then rebase during segment deactivation, then
+ * commit in background.
  *
  * @author Keith Godfrey
  * @author Henry Pijffers (henry.pijffers@saxnot.com)
@@ -168,18 +169,18 @@ public class RealProject implements IProject {
     private DirectoryMonitor tmOtherLanguagesMonitor;
 
     /**
-     * Indicates when there is an ongoing save event. Saving might take a while during
-     * team sync: if a merge is required the save might be postponed indefinitely while we
-     * wait for the user to confirm the current segment.
+     * Indicates when there is an ongoing save event. Saving might take a while
+     * during team sync: if a merge is required the save might be postponed
+     * indefinitely while we wait for the user to confirm the current segment.
      */
     private boolean isSaving = false;
 
     /**
-     * Storage for all translation memories, which shouldn't be changed and saved, i.e. for /tm/*.tmx files,
-     * aligned data from source files.
+     * Storage for all translation memories, which shouldn't be changed and
+     * saved, i.e. for /tm/*.tmx files, aligned data from source files.
      *
-     * This map recreated each time when files changed. So, you can free use it without thinking about
-     * synchronization.
+     * This map recreated each time when files changed. So, you can free use it
+     * without thinking about synchronization.
      */
     private Map<String, ExternalTMX> transMemories = new TreeMap<>();
 
@@ -210,11 +211,13 @@ public class RealProject implements IProject {
         EMPTY_TRANSLATION = new TMXEntry(empty, true, null);
     }
 
-    private final boolean allowTranslationEqualToSource = Preferences.isPreference(Preferences.ALLOW_TRANS_EQUAL_TO_SRC);
+    private final boolean allowTranslationEqualToSource = Preferences
+            .isPreference(Preferences.ALLOW_TRANS_EQUAL_TO_SRC);
 
     /**
-     * A list of external processes. Allows previously-started, hung or long-running processes to be
-     * forcibly terminated when compiling the project anew or when closing the project.
+     * A list of external processes. Allows previously-started, hung or
+     * long-running processes to be forcibly terminated when compiling the
+     * project anew or when closing the project.
      */
     private final Stack<Process> processCache = new Stack<>();
 
@@ -230,7 +233,8 @@ public class RealProject implements IProject {
         config = props;
         if (config.getRepositories() != null && !Core.getParams().containsKey(CLIParameters.NO_TEAM)) {
             try {
-                remoteRepositoryProvider = new RemoteRepositoryProvider(config.getProjectRootDir(), config.getRepositories(), config);
+                remoteRepositoryProvider = new RemoteRepositoryProvider(config.getProjectRootDir(),
+                        config.getRepositories(), config);
             } catch (Exception ex) {
                 // TODO
                 throw new RuntimeException(ex);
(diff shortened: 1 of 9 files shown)


src/org/omegat/filters2/master/FilterMaster.java
```diff
@@ -202,6 +203,8 @@ public IFilter loadFile(String filename, FilterContext fc, IParseCallback parseC
             filterObject = lookup.filterObject;
 
             filterObject.parseFile(inFile, lookup.config, fc, parseCallback);
+        } catch (TranslationException e) {
+            throw e;
```
miurahr, line 207: Here is a core part of the change.
When filter got error, TranslationException will deliver to parent caller.

src/org/omegat/core/data/RealProject.java
```diff
@@ -1141,35 +1179,50 @@ private void loadSourceFiles() throws Exception {
...
+                        fi.fileEncoding = "";
+                    }
+                    projectFilesList.add(fi);
                 }
-                projectFilesList.add(fi);
+            } catch (TranslationException e) {
+                // ignore failed file and continue loading next.
+                Log.logErrorRB("TF_SOURCE_LOAD_ERROR", e.getLocalizedMessage());
+                Core.getMainWindow().displayErrorRB(e, "TF_SOURCE_LOAD_ERROR", filepath);
             }
```
miurahr, line 1226: When we got `TranslationException` from `FilterMaster` class, OmegaT notify the error to user in error dialog with problematic filename, then continue the process.

miurahr: It would be better to insert explicit call of `continue` to express we don't abort here.

brandelune:
Screen Shot 2023-01-30 at 20 25 40

I think the order of messages should be different (@Kazephil)

1. Announce the error: Error in file ..., line 318
2. Describe the error: The element "para" must be terminated ...
3. Announce the resulting action: The file will not be loaded.

We don't need the org.omegat.filters2... exception message (although it could be logged)

Kazephil:
> I think the order of messages should be different (@Kazephil)
> 
> 1. Announce the error: Error in file ..., line 318
> 2. Describe the error: The element "para" must be terminated ...
> 3. Announce the resulting action: The file will not be loaded.
> 
> We don't need the org.omegat.filters2... exception message (although it could be logged)

I agree that the `org.omegat.filters2`... exception should be stored in a log rather than displayed in the error dialog.

For the order, though, I think the "Failed to load..." message is actually the most important piece of information for the user, and should come first. Only more technically-inclined users will follow up on the additional--and more specific--information about the error.

I would use:
1. What happened:
   "Failed to load source file `filename`."
  (I don't think "in specific project" is necessary, incidentally).
2. Where the error occurred
    "Error in file ..., line 318."
3. What the error is:    
    "The element "para" must be terminated"...

This immediately informs the user that a file cannot be loaded, and follows it up with additional information that either lets the user fix the problem him- or herself, or ask for assistance.

brandelune:
Ok, but with this fix OmegaT "deliberately" chooses to not load the file and load the rest of the project. So I think we should use the active voice here, or at least not "failed".

Kazephil:
> Ok, but with this fix OmegaT "deliberately" chooses to not load the file and load the rest of the project. So I think we should use the active voice here, or at least not "failed".

Ah! In that case, maybe the first error message should be something along the lines of:

"Skipped source file `filename` because"... (If the message is shown after the rest of the project has loaded.)

or

"Skipping source file `filename` because"... (If the message is shown while the project is loading.)

Options for the part after "because" include:
- ... "it could not be loaded."
- ... "it contains an error."/"it contains errors." 
  (This one would need to change depending on whether there are several errors or just one.)
- ... "the file contains invalid XML."/ "the file is not valid XML." (If this only applies to filters for XML files.)

I would probably stick with "it could not be loaded." since it immediately informs the uses that the `filename` was not loaded. The next two line provide specific information about why the file didn't load, so there is probably no need to make the first one overly specific.

brandelune:
> > Ok, but with this fix OmegaT "deliberately" chooses to not load the file and load the rest of the project. So I think we should use the active voice here, or at least not "failed".
> 
> Ah! In that case, maybe the first error message should be something along the lines of:
> 
> "Skipped source file `filename` because"... (If the message is shown after the rest of the project has loaded.)
> 
> or
> 
> "Skipping source file `filename` because"... (If the message is shown while the project is loading.)
> 
> Options for the part after "because" include:
> 
> * ... "it could not be loaded."
> * ... "it contains an error."/"it contains errors."
>   (This one would need to change depending on whether there are several errors or just one.)
> * ... "the file contains invalid XML."/ "the file is not valid XML." (If this only applies to filters for XML files.)
> 
> I would probably stick with "it could not be loaded." since it immediately informs the uses that the `filename` was not loaded. The next two lines provide specific information about why the file didn't load, so there is probably no need to make the first one overly specific.

OK. I leave it to the two of you to decide what's best.

miurahr:
Little tweaks error message when getting `SAXException` in filters in  filters3 categories.

miurahr:
> OK. I leave it to the two of you to decide what's best.

@brandelune 
Could you propose how to change

 `TF_SOURCE_LOAD_ERROR=Failed to load source file {0} in specified project!`

?
Following message is produced by JDK parser library.

Kazephil:
> Could you propose how to change
> 
> `TF_SOURCE_LOAD_ERROR=Failed to load source file {0} in specified project!`

I would use:

`TF_SOURCE_LOAD_ERROR=Skipped source file {0} because it could not be loaded.`

miurahr:
> TF_SOURCE_LOAD_ERROR=Skipped source file {0} because it could not be loaded.

Thank you for the clear feedback. I think I've updated all things.

Kazephil:
> Following message is produced by JDK parser library.

Thanks. I tried it yesterday, and it seems to work as intended.

Two questions, though:
1. The order of the error message after "Skipped source file..." hasn't changed. Is that what you meant by:
   > Following message is produced by JDK parser library.

   in an earlier comment?

2. I get two failed tests when I run Gradle check:
    > org.omegat.filters.XLIFFFilterTest > testInvalidXMLOnWeirdPath FAILED
    java.lang.AssertionError at XLIFFFilterTest.java:264

    > org.omegat.filters.XLIFFFilterTest > testInvalidXML FAILED
    java.lang.AssertionError at XLIFFFilterTest.java:233

   Are those file supposed to fail?

miurahr:
> > Following message is produced by JDK parser library.
> 
> Thanks. I tried it yesterday, and it seems to work as intended.
> 
> Two questions, though:
> 
> 1. The order of the error message after "Skipped source file..." hasn't changed. Is that what you meant by:
>    > Following message is produced by JDK parser library.
>    
>    
>    in an earlier comment?

Yes, when you want to change it, you need to change OpenJDK project.

miurahr:
> 2. I get two failed tests when I run Gradle check:
>    > org.omegat.filters.XLIFFFilterTest > testInvalidXMLOnWeirdPath FAILED
>    > java.lang.AssertionError at XLIFFFilterTest.java:264
>    
>    
>    > org.omegat.filters.XLIFFFilterTest > testInvalidXML FAILED
>    > java.lang.AssertionError at XLIFFFilterTest.java:233
>    
>    
>    Are those file supposed to fail?

You suggest to hide root cause of error of `SAXException`, then I changes the behavior.

These test expect  a line `SAXException` which you suggested to be removed.

miurahr:
Now recreate commit history  as good enough to merge.

miurahr:
merged into master.

PR #255, opened 2022-09-10
Team: Improve a way to open team project and backup file creation

miurahr:
Improve a way to open team project

## Pull request type

Please check the type of change your PR introduces:

- [x] Bug fix
- [ ] Feature
- [ ] Documentation
- [ ] Build and release changes
- [ ] Other (describe below)

## Which ticket is resolved?

- Link: https://sourceforge.net/p/omegat/bugs/1110/
- Title:  Team project pile upback up files of omegat.project

## What does this PR change?

- Update `ProjectUICommands#projectOpen` method
    - Simplify detection of condition whether override local or not.
    - add missing message entries in `Bundle.properties`
   - Detect remote omegat.project file is identical with local one and if identical it does not make backup

PR #415, opened 2023-01-15
Improve reuse of a backup project files

miurahr:
This change intend to improve an error handler and retry backup file when loading `omegat.project`

A root cause of the problem is a change #280 and #255 is not well integrated.

## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?

- Error displayed when opening a project on Windows 10
  * https://sourceforge.net/p/omegat/bugs/1136/

## What does this PR change?

- Reuse backup files.
- Pick a latest backup file
- When project file are broken, reuse backup file
- Don't copy backup file to project file
- Don't produce `omegat.project.bak` file.


briacp (commented):
Thanks for the fix, this looks good.

miurahr:
Thanks @briacp for review. merged into master now.

miurahr:
Original issuer also confirmed the fix.

miurahr:
The issue was enrolled after v5.7.1, so we don't need to add it to changelog

miurahr:
#280 make backup file every time when successfully loading project file, but #255 make it when project file is changed.
If user just generated a project and doesn't change it, or just downloading team project, there is not a backup file.

PR #416, opened 2023-01-15
Add regression test for RFE#1506: handling XLIFF state attributes

miurahr:
[RFE#1506 XLIFF: update target tag's state attribute according to translation status](https://sourceforge.net/p/omegat/feature-requests/1506/)  requests OmegaT XLIFF filter to handle state attribute of target tag.

A proposal here add a regression test and fix a bug that detected by a regression test.

## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?

- XLIFF: update target tag's state attribute according to translation status
  * https://sourceforge.net/p/omegat/feature-requests/1506/

## What does this PR change?

- Add a case file: file-xliff-RFE1506.xliff
- There also is expectation file-xliff-RFE1506-translated-expectation.xliff
- Check output expectation with XMLStreamReader
- Fix a bug detected by the test
  - when source has "needs-review-translation" state, then it should become "translated" in the translation.

## Other information

Related PR  #356

### DIFF
--- a/src/org/omegat/filters3/xml/xliff/XLIFFDialect.java
+++ b/src/org/omegat/filters3/xml/xliff/XLIFFDialect.java
@@ -63,8 +63,9 @@ public class XLIFFDialect extends DefaultXMLDialect {
     private boolean ignoreTypeForBptTags;
     private boolean changeStateToNeedsReviewTranslation;
     /**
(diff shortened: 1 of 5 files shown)


miurahr:
merged into master

PR #418, opened 2023-01-16
docs_build: Don't override index.html if manual exists by instant-start task

miurahr:
PR #376 force override manual's `index.html` by `first_step.html` by `instant-start` task

This fix the bug.

## Pull request type

- Bug fix -> [bug]

## What does this PR change?

- Check manual source is exist or not
- when manual does not exist, copy first_step.html as index.html

### DIFF
--- a/doc_src/build.xml
+++ b/doc_src/build.xml
@@ -159,7 +159,17 @@
 	  <available property="FirstSteps.exists" file="../doc_src/${language}/First_Steps.xml"/>
 	</target>
 
-	<target name="old-css" depends="check-first-steps" unless="${FirstSteps.exists}">
+    <target name="check-manual-exist">
+      <condition property="ManualContents.exists">
+        <or>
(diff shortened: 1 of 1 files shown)


brandelune:
I don't understand the problem you are trying to fix. That is not a bug. That is the intended behavior. The index is later replaced by the existing manual index *if* that index exists.

miurahr:
> I don't understand the problem you are trying to fix. That is not a bug. That is the intended behavior. The index is later replaced by the existing manual index _if_ that index exists.

It is right only when you run `updateManuals` task. It is a bug for task  `instantStartGuides` et al.

brandelune:
Whatever works is fine. Sorry for having broken the local tasks. I was not aware that they needed to be run independently.

brandelune (approved):

miurahr:
merged into master.

PR #423, opened 2023-01-17
GUI: filter sort order of filter list

miurahr:
Put sort order for local / global file filter list in preferences.
 
Related #356


brandelune (approved):
Excellent! Thank you!

Although, it looks like the sorting order is not saved after quit. Would you mind checking whether that is possible?

miurahr:
> Although, it looks like the sorting order is not saved after quit. Would you mind checking whether that is possible?

I don't think it valuable.

kosivantsov:
The sorting is nice, but the list itself became almost unusable:
if you open filter options and settings, you get them for the filter that was sitting in the selected place before the sorting was introduced:
image

kosivantsov:
And this is [RFE 1667](https://sourceforge.net/p/omegat/feature-requests/1667/) (just a reminder)

PR #424, opened 2023-01-18
MT: Fix cache clear logic

miurahr:
Fix the bug that when project save  cache cleared.

## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?
- MT cache cleared on saving or committing translation to memory
  * https://sourceforge.net/p/omegat/bugs/1139/
 
## What does this PR change?

- Introduce internal method `getCacheLayer` and `setCacheClearPolicy` method that can be override by MT connectors.
- Clear cache only when close event issued
- Remove unused constant variable


t-cordonnier (commented):
LGTM

kosivantsov:
Seems to be working as expected, thank you for your quick fix

PR #425, opened 2023-01-18
team: git: suppress debug log for git connections

miurahr:
## Pull request type

- Other (describe below)

refactoring

## Which ticket is resolved?
N.A.

## What does this PR change?

Current GitCredential SSH access code registers SLF4J logger. Now introduces JSchCompatLogger class that is proxy from Jsch logging request pass to org.omegat.util.Log.

Logging level for team credentials is set to WARNING.


src/org/omegat/core/team2/impl/JSchCompatLogger.java
```diff
@@ -0,0 +1,68 @@
...
+ */
+
+package org.omegat.core.team2.impl;
+
+
+import java.util.logging.Logger;
+
+import org.omegat.util.Log;
+
+public class JSchCompatLogger implements com.jcraft.jsch.Logger {
```
miurahr, line 33: This can be internal class or unnamed class in GITCredentialsProvider class.

PR #427, opened 2023-01-18
Quit confirmation default to be cancel

miurahr:
Change quit confirmation to be "cancel" in default for safe.

## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?

- #1130 Default of quit confirmation
  * https://sourceforge.net/p/omegat/bugs/1130
 
## What does this PR change?

- Replace `JOptionPane.showConfirmDialog` with `JOptionPane.showOptionDialog` with a custom initial value.
-
-

### DIFF
--- a/src/org/omegat/gui/main/MainWindowMenuHandler.java
+++ b/src/org/omegat/gui/main/MainWindowMenuHandler.java
@@ -44,6 +44,7 @@ import java.util.regex.Pattern;
 
 import javax.swing.JOptionPane;
 import javax.swing.SwingWorker;
+import javax.swing.UIManager;
 import javax.swing.text.JTextComponent;
 
 import org.omegat.Main;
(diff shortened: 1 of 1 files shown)


brandelune:
I don't think there is a safety issue in that dialog. See my comment on SF.

brandelune:
I just tested. Thank you.

I don't think there is a safety issue here or that we have something broken in terms of UX that needs to be improved. Moving from a button to another with TAB is not an obvious action for a lot of computer users. So we are adding an extra step for a very safe and very common action.

Regardless, as Kos said, TAB followed by SPACE works on macOS too. The problem I see is that TAB is not fully highlighted. CANCEL is still in blue and there is a thin blue border around OK that appears but it is not very visible and thus not great in terms of accessibility.

miurahr:
There is no problem on Linux and windows. Are there some UX problem in macOS?

The change here does not use any special UI parts. I just use a java standard dialog.

If  a java standard dialog has UX problem on macOS. There are potentially many problems on macOS. 
It is very strange feedback for me.

brandelune:
I just checked LibreOffice UI for quitting an unsaved document. The default is "Save" it is not "Cancel". People intend to quit. We just need a 1 step confirmation.

miurahr:
> I just checked LibreOffice UI for quitting an unsaved document. The default is "Save" it is not "Cancel". People intend to quit. We just need a 1 step confirmation.

L.O. ask whether "save" or "not save",  OmegaT ask whether "save and quit" or "back to editor". Meaning is different.

brandelune:
> L.O. ask whether "save" or "not save", OmegaT ask whether "save and quit" or "back to editor". Meaning is different.

I know that the meaning is different. Because LO does not save automatically. OmegaT cannot not save on quit.

In LO we have 3 possibilities:

- C-Q / Enter -> Save + Quit (1 step confirmation)
- C-Q / Esc -> Back to editor (1 step confirmation)
- C-Q / TAB / Enter -> Don't save + Quit (2 steps confirmation)

In OmegaT we have 2 possibilities:
- C-Q / Enter -> (Save +) Quit (1 step confirmation)
- C-Q / Esc -> Back to editor (1 step confirmation)
(- We cannot "Quit and not save" unless we kill OmegaT from an external process: command line, etc.)

Which is exactly the same as LO.

There is no UX bug since it is a user choice that can be disabled, there is no theoretical safety issue since OmegaT never looses data, the behavior is exactly as intended.

miurahr:
OK, I'd like to take another approach when user select quit command and protecting from continue to enter something in editor pane.

brandelune:
I think we have one problem with Quit now, and maybe that is what you want to were considering:
When we have "auto-insert best match" and the segment is empty *but* must be filled by an auto-insert, if we quit, the segment is filled without the user's intention and the unintentional segment is registered.

This is probably outside the scope of your suggestion, but I think that's the one issue related to Quit that needs to be solved.

PR #428, opened 2023-01-19
Fix search window field font

miurahr:
## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?
- Font Rendering issue in OmegaT (Santali)
  * https://sourceforge.net/p/omegat/bugs/1133/

## What does this PR change?

-  set configured font to search / replace field on search dialog 
-  register font change event listener that update search dialog fonts

### DIFF
--- a/src/org/omegat/gui/search/SearchWindowController.java
+++ b/src/org/omegat/gui/search/SearchWindowController.java
@@ -154,8 +154,8 @@ public class SearchWindowController {
         form.m_dateToSpinner.setModel(dateToModel);
 
         // Box Number of results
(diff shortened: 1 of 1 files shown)


src/org/omegat/gui/search/SearchWindowController.java
```diff
@@ -1131,7 +1168,8 @@ private void setEnabled(Container component, boolean enabled) {
     }
 
     /*
-     * TODO: This should be a method on EditorController exposed via IEditor, NOT here.
+     * TODO: This should be a method on EditorController exposed via IEditor,
+     * NOT here.
```
miurahr, line 1172: Here is a TODO comment that should be handled by another topic. @yu-tang

PR #431, opened 2023-01-20
New default layout

brandelune:
This PR keeps the layout related commits from https://github.com/omegat-org/omegat/pull/386

### DIFF
--- a/src/org/omegat/gui/main/MainWindowUI.java
+++ b/src/org/omegat/gui/main/MainWindowUI.java
@@ -9,6 +9,7 @@
                2008 Andrzej Sawula, Alex Buloichik
                2014 Piotr Kulik
                2015 Aaron Madlon-Kay
+               2023 Jean-Christophe Helary
                Home page: http://www.omegat.org/
                Support center: https://omegat.org/support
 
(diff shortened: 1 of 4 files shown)


Kazephil:
This one looks good here!

brandelune:
@Kazephil I just added some code to check the available screenspace so that OmegaT stays within the empty available space and does not hide behind toolbars or whatnot.

@miurahr please, check the validity of the code. It works, but it might not be the best way to put it.

Kazephil:
> I just added some code to check the available screenspace so that OmegaT stays within the empty available space and does not hide behind toolbars or whatnot.

Nice!

brandelune:
It works on macOS with the Dock on the left/right at the bottom. There is only a small glitch when the Dock is on the left, but I don't want to bother about that. The code is generic enough. Could you test that on Windows ? You'd need to remove your default OmegaT preference folder first.

src/org/omegat/gui/main/MainWindowUI.java
```diff
@@ -267,7 +269,34 @@ public void mouseClicked(MouseEvent e) {
...
+		
+		if (omegatHeight > screenHeight) {
+			omegatHeight = screenHeight;
+		}
+		
+		// Attempt to center the OmegaT main window on the screen
+		int omegatLeftPosition = (screenWidth - omegatWidth)/2;
+
+		Rectangle defaultWindowSize = new Rectangle(omegatLeftPosition, 0, omegatWidth, omegatHeight);
+
```
miurahr, line 298: Please format indent.  It may be good idea to run spotlessChangedApply task

miurahr (changes requested):

miurahr:
Now rebased and merged into master. Thank you for hard work!

PR #434, opened 2023-01-20
New view options

brandelune:
This PR keeps the view options related commits from https://github.com/omegat-org/omegat/pull/386

### DIFF
--- a/src/org/omegat/gui/editor/EditorSettings.java
+++ b/src/org/omegat/gui/editor/EditorSettings.java
@@ -92,7 +92,7 @@ public class EditorSettings implements IEditorSettings {
         markWhitespace  = Preferences.isPreference(Preferences.MARK_WHITESPACE);
         markBidi  = Preferences.isPreference(Preferences.MARK_BIDI);
         displayModificationInfo = Preferences.getPreferenceDefault(Preferences.DISPLAY_MODIFICATION_INFO,
-                DISPLAY_MODIFICATION_INFO_NONE);
+                DISPLAY_MODIFICATION_INFO_SELECTED);
         autoSpellChecking = Preferences.isPreference(Preferences.ALLOW_AUTO_SPELLCHECKING);
         markAutoPopulated = Preferences.isPreferenceDefault(Preferences.MARK_AUTOPOPULATED,
                 Preferences.MARK_AUTOPOPULATED_DEFAULT);
--- a/src/org/omegat/gui/editor/EditorSettings.java
+++ b/src/org/omegat/gui/editor/EditorSettings.java
@@ -102,7 +102,8 @@ public class EditorSettings implements IEditorSettings {
         viewActiveSourceBold = Preferences.isPreferenceDefault(Preferences.VIEW_OPTION_SOURCE_ACTIVE_BOLD,
                 Preferences.VIEW_OPTION_SOURCE_ACTIVE_BOLD_DEFAULT);
         markFirstNonUnique = Preferences.isPreference(Preferences.VIEW_OPTION_UNIQUE_FIRST);
-        markGlossaryMatches = Preferences.isPreference(Preferences.MARK_GLOSSARY_MATCHES);
+        markGlossaryMatches = Preferences.isPreferenceDefault(Preferences.MARK_GLOSSARY_MATCHES,
+                Preferences.MARK_GLOSSARY_MATCHES_DEFAULT);
         markLanguageChecker = !Preferences.isPreferenceDefault(Preferences.LT_DISABLED,
                 Preferences.LT_DISABLED_DEFAULT);
         doFontFallback = Preferences.isPreference(Preferences.FONT_FALLBACK);
--- a/src/org/omegat/gui/editor/EditorSettings.java
+++ b/src/org/omegat/gui/editor/EditorSettings.java
@@ -87,7 +87,8 @@ public class EditorSettings implements IEditorSettings {
                 Preferences.MARK_NON_UNIQUE_SEGMENTS_DEFAULT);
         markNoted = Preferences.isPreference(Preferences.MARK_NOTED_SEGMENTS);
         markNBSP  = Preferences.isPreference(Preferences.MARK_NBSP);
-        markParagraphDelimitations = Preferences.isPreference(Preferences.MARK_PARA_DELIMITATIONS);
+        markParagraphDelimitations = Preferences.isPreferenceDefault(Preferences.MARK_PARA_DELIMITATIONS,
+                Preferences.MARK_PARA_DELIMITATIONS_DEFAULT);
         markWhitespace  = Preferences.isPreference(Preferences.MARK_WHITESPACE);
         markBidi  = Preferences.isPreference(Preferences.MARK_BIDI);
         displayModificationInfo = Preferences.getPreferenceDefault(Preferences.DISPLAY_MODIFICATION_INFO,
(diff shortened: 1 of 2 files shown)


miurahr (changes requested):
It should place default constants for editor in `EditorSettings` class as `private static final boolean`, not in `Preferences` class.
It is also a good habit to put implicit defaults for other settings.

brandelune:
I don't understand what you are asking me to change.

src/org/omegat/gui/editor/EditorSettings.java
```diff
@@ -83,24 +83,28 @@ protected EditorSettings(final EditorController parent) {
         markTranslated = Preferences.isPreference(Preferences.MARK_TRANSLATED_SEGMENTS);
         markUntranslated = Preferences.isPreference(Preferences.MARK_UNTRANSLATED_SEGMENTS);
         displaySegmentSources = Preferences.isPreference(Preferences.DISPLAY_SEGMENT_SOURCES);
-        markNonUniqueSegments = Preferences.isPreference(Preferences.MARK_NON_UNIQUE_SEGMENTS);
+        markNonUniqueSegments = Preferences.isPreferenceDefault(Preferences.MARK_NON_UNIQUE_SEGMENTS,
+                Preferences.MARK_NON_UNIQUE_SEGMENTS_DEFAULT);
```
miurahr, line 87:
```suggestion
               MARK_NON_UNIQUE_SEGMENTS_DEFAULT);
```

src/org/omegat/gui/editor/EditorSettings.java
```diff
@@ -83,24 +83,28 @@ protected EditorSettings(final EditorController parent) {
...
 
         // options from preferences 'view' pane
         viewSourceBold = Preferences.isPreferenceDefault(Preferences.VIEW_OPTION_SOURCE_ALL_BOLD,
                 Preferences.VIEW_OPTION_SOURCE_ALL_BOLD_DEFAULT);
         viewActiveSourceBold = Preferences.isPreferenceDefault(Preferences.VIEW_OPTION_SOURCE_ACTIVE_BOLD,
                 Preferences.VIEW_OPTION_SOURCE_ACTIVE_BOLD_DEFAULT);
         markFirstNonUnique = Preferences.isPreference(Preferences.VIEW_OPTION_UNIQUE_FIRST);
-        markGlossaryMatches = Preferences.isPreference(Preferences.MARK_GLOSSARY_MATCHES);
+        markGlossaryMatches = Preferences.isPreferenceDefault(Preferences.MARK_GLOSSARY_MATCHES,
+                Preferences.MARK_GLOSSARY_MATCHES_DEFAULT);
```
miurahr, line 107:
```suggestion
                MARK_GLOSSARY_MATCHES_DEFAULT);
```

src/org/omegat/util/Preferences.java
```diff
@@ -192,6 +192,7 @@ public final class Preferences {
      * backwards-compatibility.
      */
     public static final String MARK_GLOSSARY_MATCHES = "transtips";
+    public static final boolean MARK_GLOSSARY_MATCHES_DEFAULT = true;
```
miurahr, line 195:
```suggestion

```

src/org/omegat/util/Preferences.java
```diff
@@ -215,7 +217,8 @@ public final class Preferences {
     public static final String MARK_TRANSLATED_SEGMENTS = "mark_translated_segments";
 
     public static final String MARK_AUTOPOPULATED = "mark_autopopulated";
-
+    public static final boolean MARK_AUTOPOPULATED_DEFAULT = true;
```
miurahr, line 220:
```suggestion

```

src/org/omegat/util/Preferences.java
```diff
@@ -342,7 +345,8 @@ public final class Preferences {
      * mark unique segments
      */
     public static final String MARK_NON_UNIQUE_SEGMENTS = "mark_non_unique_segments";
-
+    public static final boolean MARK_NON_UNIQUE_SEGMENTS_DEFAULT = true;
```
miurahr, line 348:
```suggestion

```

miurahr (changes requested):

brandelune:
I still don't understand, but I did it.

miurahr:
Reviewer cannot add a new code. You should add a constant variable definition in the class.

brandelune:
> It should place default constants for editor in `EditorSettings` class as `private static final boolean`, not in `Preferences` class.
> It is also a good habit to put implicit defaults for other settings.

If we split preference settings in all the related files, don't you think it makes it more difficult to find them ?

miurahr:
> > It should place default constants for editor in `EditorSettings` class as `private static final boolean`, not in `Preferences` class.
> > It is also a good habit to put implicit defaults for other settings.
> 
> If we split preference settings in all the related files, don't you think it makes it more difficult to find them ?

I don't understand what you want to do.

There is a golden rule for object-oriented programming.
All the related properties, methods are going to be inside the class.

I think as follows based on the OOP rule.
Keys of preferences are properties (constant) of Preferences class. 
The default values (constant ) of  Editor behavior  are properties of  Editor GUI parts, so it should  belong to `EditorSettings` class.

I think that finding variable is a task of IDE, cross reference tools, or other computer-assisted programming environment, especially in large project in Java ecosystem.

brandelune:
> I don't understand what you want to do.
> 
> There is a golden rule for object-oriented programming. All the related properties, methods are going to be inside the class.

I'm fine with golden rules, but I'm just changing 5 default preferences in the _current_ version of the code.

> I think as follows based on the OOP rule. Keys of preferences are properties (constant) of Preferences class. The default values (constant ) of Editor behavior are properties of Editor GUI parts, so it should belong to `EditorSettings` class.

If we want to reorganize the whole preference code to be closer to the traditional OOP rules, we can do that in 6.1.

miurahr:
> > I don't understand what you want to do.
> > There is a golden rule for object-oriented programming. All the related properties, methods are going to be inside the class.
> 
> I'm fine with golden rules, but I'm just changing 5 default preferences in the _current_ version of the code.
> 
> > I think as follows based on the OOP rule. Keys of preferences are properties (constant) of Preferences class. The default values (constant ) of Editor behavior are properties of Editor GUI parts, so it should belong to `EditorSettings` class.
> 
> If we want to reorganize the whole preference code to be closer to the traditional OOP rules, we can do that in 6.1.

I just comment about your 5 default preferences. I don't understand why you tell whole preference code.

brandelune:
Sorry, maybe there is a confusion.

Is there any problem with the current proposals ?

miurahr:
There is no big problem in current code; except for failure of build.


I want to understand a reason why default value should be in Preferences class.

In many cases,  OmegaT code use direct value for defaults. Some put its default as  class constant,  and very rare who put it in Preferences class.

It may be a good change to improve Preferences system when we find something not optimal in the system.

There can be an idea to put defaults values in external file such as `PreferenesDefaults.properties` and developer and manual editor can share its values. This may help document writer to understand  behaviors.
Is it your concerned?

brandelune:
> There is no big problem in current code; except for failure of build.

When I merge master on this branch it builds fine.
 
> I want to understand a reason why default value should be in Preferences class.

I see.
 
> In many cases, OmegaT code use direct value for defaults. Some put its default as class constant, and very rare who put it in Preferences class.

I see.
 
> It may be a good change to improve Preferences system when we find something not optimal in the system.

Indeed.

> There can be an idea to put defaults values in external file such as `PreferenesDefaults.properties` and developer and manual editor can share its values. This may help document writer to understand behaviors. Is it your concerned?

Yes.

But what matters more than may concern is that people who are involved in development or who want to be involved in development find that modifying preferences is easy.

So, if there is a better (more systematic) way to implement preferences than what exists now we should consider it.

miurahr (approved):

miurahr:
I've squashed and merged.

PR #435, opened 2023-01-20
Remove TaaS references

brandelune:
This PR keeps the TAAS related commits from https://github.com/omegat-org/omegat/pull/386


miurahr:
merged into master.

PR #440, opened 2023-01-21
Standardize use of "repeated/repetition" for segments

Kazephil:
This is a follow-up to the change from "non-unique" to "repeated" that also changes "duplicate" makes consistent use of "repeated" or "repetition" when talking about segments.

I have not changed the names of the keys in the program, because it does not seem necessary to go that far. I also haven't changed any references to duplicates that do not refer to segments.


brandelune (approved):

miurahr:

PR #441, opened 2023-01-21
Change "convert numbers" to "replace numbers" for fuzzy matches

Kazephil:
This PR addresses the concern raised [here](https://sourceforge.net/p/omegat/mailman/message/37764920/) by Manuel.


src/org/omegat/gui/preferences/view/EditingBehaviorPanel.java
```diff
@@ -184,7 +184,7 @@ public void actionPerformed(java.awt.event.ActionEvent evt) {
         gridBagConstraints.anchor = java.awt.GridBagConstraints.WEST;
         add(prefixText, gridBagConstraints);
 
-        org.openide.awt.Mnemonics.setLocalizedText(convertNumbers, OStrings.getString("WF_OPTION_CONVERT_NUMBERS")); // NOI18N
+        org.openide.awt.Mnemonics.setLocalizedText(convertNumbers, OStrings.getString("WF_OPTION_REPLACE_NUMBERS")); // NOI18N
```
brandelune, line 187: /**
     * This method is called from within the constructor to initialize the form.
     * WARNING: Do NOT modify this code. The content of this method is always
     * regenerated by the Form Editor.
     */

This file should not be modified.

Kazephil: Oh! The dangers of search & replace!
Thanks you! I'll remember to exclude `.form` files from now on.

I've just redone the commit without that file.

brandelune: It's not the `.form` file that should not be touched. It's the resulting `.java` file. The one that I pointed at above.

brandelune: It's all written in the header file.

Kazephil: Oops!
If at first you don't succeed... fail more spectacularly!

I read too quickly. Hopefully the third time will be the charm...

PR #443, opened 2023-01-21
fix: context-bound matches were ignored in /tm/auto

briacp:
## Pull request type

Bug

## Which ticket is resolved?

- Fix context-bound matches in TMX that were ignored in /tm/auto
  * https://sourceforge.net/p/omegat/bugs/1140/

### DIFF
--- a/test/src/org/omegat/core/data/AutoTmxTest.java
+++ b/test/src/org/omegat/core/data/AutoTmxTest.java
@@ -45,103 +45,166 @@ import org.omegat.util.TestPreferencesInitializer;
  * @author Alex Buloichik (alex73mail@gmail.com)
  */
 public class AutoTmxTest {
-    RealProject p;
+	RealProject p;
 
-    @Before
-    public final void setUp() throws Exception {
-        TestPreferencesInitializer.init();
-        Core.setSegmenter(new Segmenter(SRX.getDefault()));
-    }
+	@Before
+	public final void setUp() throws Exception {
+		TestPreferencesInitializer.init();
+		Core.setSegmenter(new Segmenter(SRX.getDefault()));
+	}
 
-    @Test
-    public void test1() throws Exception {
-        ProjectProperties props = new ProjectProperties();
-        props.setSourceLanguage("en");
-        props.setTargetLanguage("fr");
-        props.setTargetTokenizer(LuceneFrenchTokenizer.class);
-        File file = new File("test/data/autotmx/auto1.tmx");
-        ExternalTMX autoTMX = new ExternalTMFactory.TMXLoader(file)
-                .setDoSegmenting(props.isSentenceSegmentingEnabled())
-                .load(props.getSourceLanguage(), props.getTargetLanguage());
+	@Test
+	public void test1() throws Exception {
+		ProjectProperties props = new ProjectProperties();
+		props.setSourceLanguage("en");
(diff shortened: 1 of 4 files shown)


miurahr:
Could you apply `gradlew spotlessChangedApply` for chages?

damien-rembert:
Sorry not really a comment about this PR but I'm a bit confused: I thought we had run Spotless on the whole codebase but if we have, then why is it that running Spotless is reformatting that were not touched in this PR (eg line 46 of src/org/omegat/core/data/ImportFromAutoTMX.java)?
If Spotless has not been run against the whole codebase, I think it would be better to do that before we ask people to use spotlessChangedApply. That's not a problem here as the commit message indicates clearly the reason for the change but I think that the reason for the changes in the comment would have been less clear if the formatting had been included in one of the previous commits.

test/src/org/omegat/core/data/AutoTmxTest.java
```diff
@@ -128,6 +132,66 @@ public boolean existEntryInProject(EntryKey key) {
...
+        checkTranslation(ste, "alternative", TMXEntry.ExternalLinked.xENFORCED);
+    }
+
+    /**
+     * Test that an alternate translation matching an id is correctly applied
+     * when the memory is in tm/auto.
+     */
+    @Test
+    public void test_alternateOnAuto() throws Exception {
+        ExternalTMX enforceTMX = prepareTestTM();
```
damien-rembert, line 157: Sorry, nit, but I think this should be named autoTMX

test/src/org/omegat/core/data/AutoTmxTest.java
```diff
@@ -128,6 +132,66 @@ public boolean existEntryInProject(EntryKey key) {
...
+
+        SourceTextEntry ste = createSTE("1_0", "Edit");
+        p.allProjectEntries.add(ste);
+        checkTranslation(ste, "foobar", null);
+        p.importHandler = new ImportFromAutoTMX(p, p.allProjectEntries);
+        p.appendFromAutoTMX(enforceTMX, false);
+        checkTranslation(ste, "alternative", TMXEntry.ExternalLinked.xAUTO);
+    }
```
damien-rembert, line 165: Just a suggestion and it's probably overkill but since you have set up everything already maybe add a test_alternateInDefaultTmFolder to check that nothing happens when the tmx is added to the default folder?
I don't think test1 and test2 do that...?
And while I'm suggesting things if you feel like renaming them, I think that would be great :)

damien-rembert (commented):
Looking good!

test/src/org/omegat/core/data/AutoTmxTest.java
briacp: >  to check that nothing happens when the tmx is added to the default folder

Unfortunately, this doesn't seem to be easy to test as the field `org.omegat.core.data.RealProject.transMemories` and the method `org.omegat.core.data.RealProject.loadTM()` are private. I'm not really sure this PR is right for refactoring this class.

> And while I'm suggesting things if you feel like renaming them, I think that would be great 

yes, thanks for the suggestion, it's done in e21efec67d0f03c496be327e327aa8840d19d27b.

miurahr (approved):
LGTM

miurahr:

Issue #421, opened 2023-01-17
Add documentation to the Stax filters feature


brandelune:
https://github.com/omegat-org/omegat/pull/356

The feature relates to the file filters and to file formats.
The following files need to be modified:

https://github.com/omegat-org/omegat/blob/master/doc_src/en/HowTo_TranslateOtherFiles.xml
https://github.com/omegat-org/omegat/blob/master/doc_src/en/App_FileFilters.xml

brandelune:
https://github.com/omegat-org/omegat/pull/356#issuecomment-1382692284 @t-cordonnier one of the filter seems to be for SDL packages. Does that mean that the plugin creates a valid return package ? I could not find that information here: http://185.13.37.79/?q=/node/88

brandelune:
The MSO filter has the exact same options as OmegaT's default with a very similar name.

Now that the Stax filter is integrated at the same level as OmegaT's (or is it?) we need to clarify the naming and give an explanation about why we keep the two.

Are we slowly phasing out the original one ? Do we suggest new users to use yours instead ? And Why ?

That's the kind of info I'd like to summarize in the Filter explanation.

t-cordonnier:
> Does that mean that the plugin creates a valid return package ? 

This filter reads and updates the SDLXLIFF files contained in the package, for that reason it returns a valid _input_ package (sdlppx) which you can use in Trados to continue the translation.
To create a return package (sdlrpx) I think it should also do some changes in the project file (sdlproj) which actually it does not. This is something I could consider as an option for the future.

That said this filter was mostly an experiment to check feasibility of zip-based formats, and build the ZIP architecture which I finally used for OpenXML. Now you should test it to say whenever it is really useful or not: maybe other approach similar to MED and OMT packages may be preferable in the future.

In particular, this filter does not yet make anything from the translation memories (SDLTM) which are in the package. A real package approach, instead of seeing it as a file format, may be better.

Test and tell me what you think.

t-cordonnier:
> Are we slowly phasing out the original one ? Do we suggest new users to use yours instead ? And Why ?

Personally I may say yes and consider that since the new filter has more features than the old one, the only reason why to keep the old one is to preserve compatibility, as a translation made with old filter cannot be fully compatible with the new one (first of all because the tags are different)
But it is up to you to decide.

brandelune:
> Test and tell me what you think.

[...]

PR #444, opened 2023-01-22
Add documentation to the Stax filters feature

brandelune:
This PR corresponds to the https://github.com/omegat-org/omegat/issues/421 documentation issue.

### DIFF
--- a/src/org/omegat/Bundle.properties
+++ b/src/org/omegat/Bundle.properties
@@ -1346,7 +1346,7 @@ RB_FILTER_SUPPORT_JAVA8_ENCODING=Force Unicode literals compatibility with Java 
 RB_FILTER_EXCEPTION=The Java Resource Bundles filter threw an exception:
 
 # XLIFFFilter.java
-XLIFF_FILTER_NAME=Monolingual XLIFF
+XLIFF_FILTER_NAME=XLIFF (legacy filter)
 
 # RESXFilter.java
 RESX_FILTER_NAME=ResX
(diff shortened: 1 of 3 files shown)


miurahr (changes requested):
You describe the commit is updating document, but it modified Bundle.properties.
Please don't change that is not related to the purpose.

brandelune:
I think we're good now.

miurahr (approved):

src/org/omegat/Bundle.properties
```diff
@@ -1507,10 +1507,10 @@ SDLPROJECT_FILTER_NAME=SDL project
 SDLXLIFF_FILTER_NAME=SDL XLIFF
 
 # org.omegat.filters4.xml.xliff.Xliff1Filter.java
-XLIFF1FILTER_FILTER_NAME=XLIFF 1.x Bilingual
+XLIFF1FILTER_FILTER_NAME=XLIFF
```
Kazephil, line 1511: I thought you wanted to name them:
- XLIFF 1.x
- XLIFF 2.x

It's probably fine with XLIFF 2 right below, but just XLIFF might be misinterpreted as applying to all versions of the format.

Kazephil: OK. In that case this is probably ready to be merged if you're happy with the wording tweaks I made in `App_filter.xml`.

Kazephil:
Merged into `master`.

PR #445, opened 2023-01-23
Add structure to the file filters section in the Bundle

brandelune:
## What does this PR change?
- This PR add structure to the file filters section of the Bundle so that further additions/modifications in that section are easier to locate.
- The commit adds sections by type of filter (document format, l10n resource, subtitles, etc.) and adds a clear title for each filter.

## Other information
The whole filter list should be documented in the manual and describe the ways to use them, especially for "exotic" formats like Illustrator or WP XML export, for which the file types require some kind of manipulation on the software side to access (either explicit "export" or more complex tasks).


miurahr (approved):

miurahr:
Looks Good.

@brandelune @kosivantsov  could you mind to regenerate all the L10N bundle from this up-to-date changes and update all the languages in another commit in the topic branch?

PR #446, opened 2023-01-23
remove references to the top message area

brandelune:
I introduced a bug in
https://github.com/omegat-org/omegat/commit/d2cc60ad5c921025251e9368545d8293afeb0b08

I did not fully remove references to the top message area that is now empty.


miurahr:
merged into master.

PR #449, opened 2023-01-26
Feature: add shortcuts to navigate notes

brandelune:
Now that we have the Notes pane displayed by default, shortcuts to enter the Notes pane and go back to the Editor pane, I think it would be very useful to add shortcuts to navigate the annotated segments.

I'm proposing C-A-UP/DOWN

### DIFF
--- a/doc_src/en/App_ShortCuts.xml
+++ b/doc_src/en/App_ShortCuts.xml
@@ -853,15 +853,15 @@
             <row>
               <entry>Next Note</entry>
               <entry>gotoNextNoteMenuItem</entry>
-              <entry>ctrl alt DOWN</entry>
-              <entry>meta alt DOWN</entry>
+              <entry>ctrl alt N</entry>
+              <entry>meta alt N</entry>
             </row>
 
             <row>
               <entry>Previous Note</entry>
               <entry>gotoPreviousNoteMenuItem</entry>
-              <entry>ctrl alt UP</entry>
-              <entry>meta alt UP</entry>
+              <entry>ctrl alt P</entry>
+              <entry>meta alt P</entry>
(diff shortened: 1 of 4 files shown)


kosivantsov:
I would be more in favor to have something based on N/P (next/previous), like C-A-N/P.
Jumping to various panes could be assigned to C-A-1, C-A-2 and so on.

brandelune:
> I would be more in favor to have something based on N/P (next/previous), like C-A-N/P.
> Jumping to various panes could be assigned to C-A-1, C-A-2 and so on.

Ok, so you propose that
- we use `C-A-number` for the current _Go to Notes_ / _Go to Editor_ with future extensions to other panes
- and we use the `C-A-N` (currently assigned to _Go to Notes_) and `C-A-P` to navigating notes.

I like the `C-A-N/P` pair because it allows me to use the 2 hands to reach different parts of the keyboard. For numbers on the left side of the kb `C-A-number` is not as convenient (I don't have an ALT key on the right and maybe the fact that I am left-handed plays.)

With all those caveats, what do you think about `C-A-0` for the current _Got to Editor_ and `C-A-9` for the current _Go to Notes_?

---
→ We could also have a _Go to Pane_ dialog that uses the same mechanism as the _Go to segment number_ in the future. That would probably be better than assigning shortcuts to all the panes.

brandelune:
@kosivantsov I've implemented your request. We should be good now.

kosivantsov:
> @kosivantsov I've implemented your request. We should be good now.

This is great! 0 for the Editor feels very natural, very Unix-like =)

miurahr:
Squashed and merged into master.

PR #451, opened 2023-01-27
Unify to "-our" spelling in user facing material

Kazephil:
I'm putting this up as a draft to get opinions first before finalizing this proposed cosmetic change.

As far as I can see, the current user facing material (manual, UI, `changes.txt`) primarily uses American spelling, with the occasional instance of British spelling sneaking up in words like colour (color).

I'd like to reflect the more international composiiton of both the OmegaT team and its community of users, and since I'm Canadian, I'd like to update all user facing material to Canadian spelling, which is most often a compromise between American and British spelling.

In practice, that means sticking to "-ize" in words like "recognize" or "realize", but using "-our" in words like "colour".

Thoughts, objections, comments?

### DIFF
--- a/doc_src/en/HowTo_UseTeamProject.xml
+++ b/doc_src/en/HowTo_UseTeamProject.xml
@@ -2,7 +2,7 @@
 <!DOCTYPE section PUBLIC "-//OASIS//DTD DocBook XML V4.5//EN"
 "../../../docbook-xml-4.5/docbookx.dtd">
 <section id="how.to.use.team.project">
-  <title id="how.to.use.team.project.title">Use a Team Project</title>
+  <title id="how.to.use.team.project.title">Use a team project</title>
 
   <para>Team projects use synchronization mechanisms between project members.</para>
 
--- a/doc_src/en/HowTo_Run.xml
+++ b/doc_src/en/HowTo_Run.xml
@@ -69,7 +69,7 @@
 	endterm="launch.with.command.line.title"
 	linkend="launch.with.command.line"/> section for details.</para>
 
-	<para>You can modify the behavior of OmegaT.app by editing the
+	<para>You can modify the behaviour of OmegaT.app by editing the
 	<filename>Configuration.properties</filename> (OmegaT configuration) as well
 	as the <filename>Info.plist</filename> (Java configuration) files located
 	inside the OmegaT.app package.</para>
@@ -150,7 +150,7 @@
 	<title id="launch.with.command.line.title">Command line launch</title>
 
 	<para>Using the command line allows you to set various options that control
-	or modify the behavior of the application. You can also define and save sets
+	or modify the behaviour of the application. You can also define and save sets
 	of options in scripts that you can then use to launch OmegaT for a
 	particular purpose.</para>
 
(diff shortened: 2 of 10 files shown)


brandelune (approved):

kosivantsov:
Totally agree. "Localisation" in `changes.txt` should be then changed to "Localization", right?

doc_src/en/HowTo_UseTeamProject.xml
```diff
@@ -2,7 +2,7 @@
 <!DOCTYPE section PUBLIC "-//OASIS//DTD DocBook XML V4.5//EN"
 "../../../docbook-xml-4.5/docbookx.dtd">
 <section id="how.to.use.team.project">
-  <title id="how.to.use.team.project.title">Use a Team Project</title>
+  <title id="how.to.use.team.project.title">Use a team project</title>
```
damien-rembert, line 5: I would have thought the case for titles would remain the same?

Kazephil: Lack of explanation on my part.

We've been using title case for chapter headings, and sentence case for everything below that.

This one happens to be an inconsistency I noticed at the same time, and I added it before I forgot to fix it.

Kazephil:
> Totally agree. "Localisation" in `changes.txt` should be then changed to "Localization", right?

Thanks Kos and Jean-Christophe!

Yes, "localisation" should be "localization". I'll also need to check for things like  "center" vs "centre", "controled" vs "controlled", and the distinction between "licence" (noun) and "license" (verb)—except, of course, in the GPL itself which is quoted verbatim in ira original spelling.

I just wanted to get other people's opinions before making sweeping changes. :smiley:

I should also add a `Bundle_us.properties` file for those who want their UI un American spelling.

I'm currently on mobile, so I'll do all that later today.

brandelune:
> > Totally agree. "Localisation" in `changes.txt` should be then changed to "Localization", right?
> 
> Thanks Kos and Jean-Christophe!

Damien ? :-)

> I just wanted to get other people's opinions before making sweeping changes. 😃

Nice idea.

> I should also add a `Bundle_us.properties` file for those who want their UI un American spelling.

Nice too.

damien-rembert:
Hey, sorry, I don't really have anything further to contribute here. Just the thumbs up! I think the idea of not going for GB or USA English is brilliant and considering @Kazephil seems to be one of the few very active contributors (or contributours ;) maybe?) who is a native speaker, it also makes sense.

Kazephil:
> I think the idea of not going for GB or USA English is brilliant

Thanks for the kind words!

I just realized that I should probably do the same thing with the text on the website, assuming @briacp has no objection.

Next, maybe we can coin "_contributours_" as a new term that encompasses people making any sort of contribution, whether it's documentation, code, or something else! :laughing:

briacp:
That's great à great idea, thanks @Kazephil

Kazephil:
OK, I think this one is ready to go in terms of applying Canadian spelling conventions to the manual, `Bundle.properties` and `changes.txt`.

Please let me know if you see anything suspicious. :-)

I've  also added a `Bundle_us.properties` file for people who want American spelling in their UI.
Let me know if I also have to make changes somewhere in the code to make sure it's recognized.

PR #402, opened 2023-01-01
Html parsing error on repeated tags sf205

damien-rembert:
## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?

- script + style HTML elements in source, missing stylesheet in result.
  * URL: [SourceForge bug 609](https://sourceforge.net/p/omegat/bugs/609/) 

- HTML parsing error
  * URL: [SourceForge bug 205](https://sourceforge.net/p/omegat/bugs/205/)

## What does this PR change?

-Split the recurse boolean
-Add missing HTML elements that can contain text and act as start of paragraph
-Rename variables to make the process of FilterVisitor clearer
-Correct a typo in the PR template
-Add annoying MacOS system files to gitignore

PR #452, opened 2023-01-28
Bugfix SF613 SF873 HTML drops tags

damien-rembert:
## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?

- SF613 HTML: some content not available
  * URL: [SourceForge bug 613](https://sourceforge.net/p/omegat/bugs/613/)

- SF873 HTML segmentation dropping whole tags
  * URL: [SourceForge bug 873](https://sourceforge.net/p/omegat/bugs/873/)

## What does this PR change?

- make recurseChildren true for paragraph tags as they can contain tags that need to be translated
- add test files for each bug (not sure we really need one for each but hey)

## Other information

Sorry if this is messy, it is building on #402 and using the split "recurse" variables added there.
I didn't want to push to the other PR as it's been accepted already...
The only difference here is the added commit [0aed648](https://github.com/omegat-org/omegat/pull/452/commits/0aed648613e107d005f869660cc0e5a4323b5420)


src/org/omegat/filters2/html2/FilterVisitor.java
```diff
@@ -153,38 +162,22 @@ public boolean shouldRecurseChildren() {
...
+            if (hasAnEndTag(tag)) {
+                recurseSelf = true;
+                recurseChildren = false;
             }
         } else {
-            // recurse = true;
-            if (isParagraphTag(tag) && text) {
-                endup();
+            if (isParagraphTag(tag)) {
+                handleParagraphTag();
```
damien-rembert, line 180: Here is the difference when compared to the other PR (#402): `handleParagraphTag` handles the check that I removed here (`&& text`) and sets recurseChildren to true.

miurahr (approved):
Looks good.
Is proposal ready for merge?

damien-rembert:
> Is proposal ready for merge?
Yep! I've pulled in the master branch recently and the change when compared to the other PR is only minimal so it's good to go :)

miurahr:
rebase to master,  reword title to add target (filter/html), and merge into master.

PR #455, opened 2023-01-29
fix(gui/filter): wrong target selection on filter customizer table

miurahr:
## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?

ref: https://sourceforge.net/p/omegat/mailman/message/37770101/

## What does this PR change?

- We add a sorter at commit bb8f181998e73a1b023e9e3942d84830e18cacd8  in #423
- When the table is sorted, the indices of the rows in the view don't match with the indices in the model anymore.
- This add `convertRowIndexToModel` method to convert from a sorter index to model index.


kosivantsov:
Looks good. Thank you for the quick fix!

kosivantsov (approved):
Tested in several projects, seems to be working as expected

PR #458, opened 2023-02-02
MT:  http json tests, fixes and improves error case behavior

miurahr:
## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?

- #1141 MT engines show JSON parser error as translated text
  * https://sourceforge.net/p/omegat/bugs/1141
 

## What does this PR change?

- Add tests to make http connection,  check request and send back mock response. 
  - Apertium
  - DeepL
  - IBMWatson
  - Google2
  - MyMemory (Human, Machine)
- Fix bugs that produce NPE,  send error message to translate pane.
- Throw MachineTranslateError instead of generic java.lang.Exception when http connection error

### DIFF
--- a/src/org/omegat/core/machinetranslators/DeepLTranslate.java
+++ b/src/org/omegat/core/machinetranslators/DeepLTranslate.java
@@ -108,7 +108,7 @@ public class DeepLTranslate extends BaseCachedTranslate {
         String apiKey = getCredential(PROPERTY_API_KEY);
         if (apiKey == null || apiKey.isEmpty()) {
             if (temporaryKey == null) {
-                throw new Exception(OStrings.getString("DEEPL_API_KEY_NOTFOUND"));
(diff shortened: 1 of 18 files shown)


src/org/omegat/core/machinetranslators/DeepLTranslate.java
```diff
@@ -57,16 +59,33 @@
...
     //
     // See https://www.deepl.com/docs-api/accessing-the-api/api-versions/
-    protected static final String DEEPL_URL = "https://api.deepl.com/v1/translate";
+    protected static final String DEEPL_PATH = "/v1/translate";
+    protected final String deepLUrl;
+
     private final static int MAX_TEXT_LENGTH = 5000;
 
+    public DeepLTranslate() {
+        deepLUrl = "https://api.deepl.com/v1/translate";
```
miurahr, line 76: URL Constant should be static final string such as `DEEPL_V1_BASE_URL`

miurahr:
Test failed with known issue in master branch that #539 is for fixing it.

PR #460, opened 2023-02-02
Docs/UI: change local to private and global to general

brandelune:
Create a better alternative than local/global for preferences:

> The difference is between "stored in the configuration folder" or "stored in the project folder".
> And we also have internal/external (searches) and local/remote (projects)

> Would "private" vs "general" work?

> - private/general settings,
> - private post-processing commands / general post processing commands

> - internal/external searches,
> - private external searches / general external searches

> - local/remote projects


brandelune:
Waiting for input on the mailing list to remove Draft status.

PR #464, opened 2023-02-06
Sf610 do not split segment on removed comment

damien-rembert:
## Pull request type

- Bug fix

## Which ticket is resolved?

- SF610 HTML comment tag segments
  * https://sourceforge.net/p/omegat/bugs/610/
 
## What does this PR change?

- HTML comments do not split segments when they are set to be removed
- Behaviour unchanged if comments are kept

## Other information

Content of HTML file:
HTML file
Now, when set to remove HTML comments:
Now, when set to remove HTML comments
Before the change - Now when set to keep HTML comments:
Before the change - Now when set to keep HTML comments


damien-rembert:
The failed build doesn't seem to be related with the changes on that branch

brandelune:
It looks like the documentation also should be changed.

Specifically that part:
https://github.com/damien-rembert/omegat/blob/8ad9b074ae552807e3a8251a741e43043766330f/doc_src/en/App_FileFilters.xml#L633

"will not be displayed in ths source segment and will not be copied..."

@Kazephil

brandelune:
Althought the fix works as planned, it seems to be that comments are not ignored, they should not be splitting parts and should appear in the segment as tags.

When the project properties have "Remove tags" the segment should be the same as when this preference is set.

Kazephil:
> It looks like the documentation also should be changed.

I'll take a look at it sometime after dinner.

I have a few follow-up questions though:
- If HTML comments are not removed, does the text always split at inline comments?
- Is splitting the text at inline comments really desirable behaviour?
  - What's the use case for not removing the comments and having the text split into separate segments?
     I thought it might have been to translate them, but based on the screenshots above, the content of the comments is not shown  regardless of whether the comments are removed or not.

damien-rembert:
Thanks for all the great feedback!
I'll set this as draft and try to display the comments as inline protected tags.

damien-rembert:
Updated the change following the feedback it got, now if comments are set to be kept, they are treated as tags (see screenshot). And when comments are to be removed, they do not break the segment.

Screenshot 2023-02-08 at 22 42 43

brandelune:
It works very nicely. Thank you.

The UI for the filter preference should be "Remove the comments" (not limited to the translated file since they are not displayed in the source either).

And the manual should be modified (see my first comment)

Kazephil:
> The UI for the filter preference should be "Remove the comments"

"Remove comments" (no "the"). :grinning: 

This implementation looks good to me. If this behaviour is finalized, I can look into updating the UI string and documentation tonight while Damien works on solving the failing tests.

brandelune:
> > The UI for the filter preference should be "Remove the comments"
> 
> "Remove comments" (no "the"). 😀
> 
> This implementation looks good to me. If this behaviour is finalized, I can look into updating the UI string and documentation tonight while Damien works on solving the failing tests.

You're in charge :-)

damien-rembert:
Sadly testing it with the omegat.org front page has revealed something broken in the logic. 
Sorry about that, I will make sure to test it fully *before* reopening the PR.

Kazephil:
> Sadly testing it with the omegat.org front page has revealed something broken in the logic.

That's unfortunate. :crying_cat_face: 

If you want, I could still send a PR with the UI and documentation updates directly to your branch instead of here so they're ready to go once you've solved the coding part.

brandelune:
Yes, let's do that !

src/org/omegat/filters2/html2/FilterVisitor.java
```diff
@@ -118,13 +118,13 @@ public FilterVisitor(HTMLFilter2 htmlfilter, BufferedWriter bufwriter, HTMLOptio
...
      * translatable paragraph,
-     * <li>Otherwise (paragraph tag follows), they are written out directly.
+     * <li>Otherwise (eg if a paragraph tag follows), they are written out directly.
      * </ul>
      */
     protected List<Node> followingNodes;
 
     /** The tags behind the shortcuts */
-    protected List<Tag> sTags;
+    protected List<Node> sTags;
```
damien-rembert, line 127: Allowing comments (Remark nodes) to be handled as shortcut the way tags are

src/org/omegat/filters2/html2/FilterVisitor.java
```diff
@@ -318,18 +317,25 @@ public void visitStringNode(Text string) {
...
+            recurseChildren = true;
+            if (betweenPreformattingTags) {
+                isTextUpForCollection = true;
+            }
+
+            if (isTextUpForCollection) {
+                queueTranslatable(remark);
+            } else {
+                queuePrefix(remark);
+            }
```
damien-rembert, line 331: This is mostly the behaviour used for Text node

damien-rembert:
All tests passing as it should have been from the start 🙃 
Thanks for the docs and your patience!

Kazephil:
That was quick!
I'll try to get the UI and documentation updates to you later today.

Kazephil:
> I'll try to get the UI and documentation updates to you later today.

Done! (Even if I'm talking to myself here :laughing: )

damien-rembert:
Thank you

brandelune (approved):

Kazephil:
This one will be ready to merge as soon as Damien gets a change to merge the UI and documentation on his fork and push them here.

brandelune:
> This one will be ready to merge as soon as Damien gets a change to merge the UI and documentation on his fork and push them here.

Feel free to correct/add/subtract things from the docs/UI as you see fit.

damien-rembert:
Sorry, I had not realised we were waiting for me! I'll do that later today :)

damien-rembert:
Updated with the docs @Kazephil kindly provided 🥳

PR #466, opened 2023-02-08
fix(mt): avoid NPE in connector tests

miurahr:
## Pull request type

refactoring

## What does this PR change?

- Avoid NPE when try running connection test with mt connectors

## Other information

Related #458


miurahr:
Similar workaround in 3rd party plugin.

https://github.com/omegat-org/azure-translate-plugin/blob/main/src/main/java/org/omegat/connectors/machinetranslators/azure/MicrosoftTranslatorAzure.java#L82-L96

PR #467, opened 2023-02-08
fix: always try to calculate the third TM match score (untokenized)

briacp:
## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?

- Title Matches containing non-alphabetical text are ignored in TM leverage
  * https://sourceforge.net/p/omegat/bugs/1003/
 
## What does this PR change?

- Removes the test that prevents the third percentage to be calculated if the first two are below the threshold.

### DIFF
--- a/src/org/omegat/core/statistics/FindMatches.java
+++ b/src/org/omegat/core/statistics/FindMatches.java
@@ -419,9 +419,6 @@ public class FindMatches {
      * @return true if we have chance
      */
     protected boolean haveChanceToAdd(final int simStem, final int simNoStem, final int simExactly) {
-        if (simStem < fuzzyMatchThreshold && simNoStem < fuzzyMatchThreshold) {
-            return false;
-        }
         if (result.size() < maxCount) {
             return true;
         }


brandelune:
@briacp for some reason it seems like I can't fetch your repository:

`git remote add briac git@github.com:briac/omegat`

works but when I do `git fetch briac` my ssh key passphrase is refused and I get

```
Enter passphrase for key '/Users/suzume/.ssh/id_ed25519': 
ERROR: Repository not found.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

miurahr (approved):
LGTM.

brandelune:
Do we need to document that "new" behavior ?

miurahr:
> Do we need to document that "new" behavior ?

Yes.
Especially for the languages that does not use space for word separator like Japanese.

brandelune:
I'm not seeing any difference for Japanese source on a small project I have here. Do you have a sample where the new behavior changes the percentages for Japanese source?

miurahr:
> I'm not seeing any difference for Japanese source on a small project I have here. Do you have a sample where the new behavior changes the percentages for Japanese source?

I can observed change at 
OmegaT L10N project, doc_src/Dialogs_ProjectProperties.xml,  segment 3450
Previous 2 candidate, now 5 candidate and 3 is under  30% is first and second match percent.

brandelune:
Ok. But that is not related to Japanese. I have the same thing in EN-FR.

So, what we should document is the fact that the threshold is taken into account for **all** the matches not the first 2, is that correct?

briacp:
> So, what we should document is the fact that the threshold is taken into account for all the matches not the first 2, is that correct

Correct, the previous behavior did not calculate the 3rd score if the first and second were below the threshold. Now, all the scores are calculated and the TM match if rejected if all the three scores are below the threshold.

PR #470, opened 2023-02-09
refactor(mt): putToCache in getTranslation

miurahr:
BaseTranslate abstract method to update cache by result of translate method.

This help hiding cache layer from connectors plugins and remove duplication of codes.

## Pull request type
- Other (describe below)

refactor

## Which ticket is resolved?

- dev ML: https://sourceforge.net/p/omegat/mailman/message/37776353/

## What does this PR change?

-  call `putToCache` in `getTranslation` method

### DIFF
--- a/src/org/omegat/core/machinetranslators/BaseTranslate.java
+++ b/src/org/omegat/core/machinetranslators/BaseTranslate.java
@@ -157,7 +157,7 @@ public abstract class BaseTranslate implements IMachineTranslation {
     @Override
     public String getTranslation(Language sLang, Language tLang, String text) throws Exception {
         if (enabled) {
-            return translate(sLang, tLang, text);
+            return putToCache(sLang, tLang, text, translate(sLang, tLang, text));
         } else {
             return null;
         }
@@ -220,8 +220,10 @@ public abstract class BaseTranslate implements IMachineTranslation {
     }
 
     protected String putToCache(Language sLang, Language tLang, String text, String result) {
-        cache.put(sLang.toString() + "/" + tLang.toString() + "/" + text, result);
-        return null; // always return null, just for method compatibility
+        if (result != null) {
+            cache.put(sLang.toString() + "/" + tLang.toString() + "/" + text, result);
+        }
+        return result;
     }
 
     /**


miurahr:
Each driver already handle cache so there are no feature and behavior difference. This purpose is for stabilizing  API for MT connector plugins.
Without the change, MT plugins need to access OmegaT internal cache layer details.

miurahr:
There are not enough unit test for MT connectors.  Improvement of test cases are in another PR and planed for v6.1.

miurahr:
I just wonna you to know an activity type of plugins programming interface(API). @brandelune

brandelune (approved):

t-cordonnier:
I have an engine (http://185.13.37.79/?q=/node/5) for which using a cache is a non-sense, as it works by having everything in memory.

"Without the change, MT plugins need to access OmegaT internal cache layer details."
well, they only need to acceess one method with usual parameters. They don't have to know what this method does.
Actually all call putToCache without any difficulty. My engine intentionally does not.

I would be in favour of removing cache from BaseTranslate and create an intermediate class CacheTranslate instead: then only engines which really need a cache (because they are slow or network access can have a cost) implement it.

miurahr:
> I have an engine (http://185.13.37.79/?q=/node/5) for which using a cache is a non-sense, as it works by having everything in memory.
> 
> "Without the change, MT plugins need to access OmegaT internal cache layer details." well, they only need to acceess one method with usual parameters. They don't have to know what this method does. Actually all call putToCache without any difficulty. My engine intentionally does not.

Good catch. But we should see `MachineTranslteTextArea` class.
When user set preference to be "not always search MT" and "invoke C-M , or select   editOverwriteTranslationMenuItem" then OmegaT calls to `IMachineTranslation#getCachedTranslation`.  when it returns null, then calls `IMachineTranslation#getTranslation`. 

A driver writer who want to skip a cache layer like you,  can override these two methods to be  `getCachedTranslation` as `return null` and orerrideing `getTranslation` which does NOT use a `getFromCache` method, then you can implement in-memory only connector.

`BaseTranslate` designed for the cases of many connectors to reduce duplication. I noticed that are many duplicated code because of the fact above.

miurahr:
There is also a duplicated logic; `String trText = text.length() > MAX_TEXT_LENGTH ? text.substring(0, MAX_TEXT_LENGTH - 3) + "..." : text;`

There are no engine that can accept infinite length of text, so it is also a target to be handle in abstract class.

@t-cordonnier  your proposal help a MT connector that don't want to create cache object, so I'd like to implement based on your way.

miurahr:
@t-cordonnier  please look at #474

t-cordonnier:
> Good catch. But we should see `MachineTranslteTextArea` class. When user set preference to be "not always search MT" and "invoke C-M , or select editOverwriteTranslationMenuItem" then OmegaT calls to `IMachineTranslation#getCachedTranslation`. when it returns null, then calls `IMachineTranslation#getTranslation`.
> 

Yes, this has been my nightmare a few years ago.

When I started DGT-OmegaT, then standard OmegaT did not use cache at all, so I introduced my own mechanism based on WeakHashMap. Then the interface IMachineTranslation did not have cache methods at all: author of an engine could use CacheTranslate class to declare that he wants a cache, but then it was the entire responsability of the CacheTranslate class to manage this cache.
But this model falled down when OmegaT introduced these two options, because then MachineTranslateTextArea explicitly makes the distinction between a cached and a non-cached entry.

> A driver writer who want to skip a cache layer like you, can override these two methods to be getCachedTranslation as return null

Or it can return the same as getTranslation: my engine does not need a cache because search of the translation is fast and implies neither network, nor disk, nor consumption of something you have paid for.

But for me it remains somehow problematic that the author of a MT engine still has to care about this distinction. Honestly it took me some time to understand what these two options really did, but it is essential for the author or an engine to decide what to do with getCachedTranslation.
So, I see two solutions.
1. make getCachedTranslation and getTranslation final in CacheTranslate (or the name you give to it): if, as the author of a new MT engine, I decide to use this class CacheTranslate, then I only take care of method `translate` and I leave the management of the cache to the base class - getFromCache and putToCache can even be private in this case.
2. Have BaseTranslate and CachedTranslate _contain_ (and not be a base class of) IMachineTranslation, which requires only the ` translate` method.  Engines extend an interface which does not have caching methods, except a boolean "needsCache" to declare whenever it wants to be cached or not. 
First method is simpler but it requires more explanation of the API for a developer which would like to implement a new engine.

t-cordonnier:
> There is also a duplicated logic; `String trText = text.length() > MAX_TEXT_LENGTH ? text.substring(0, MAX_TEXT_LENGTH - 3) + "..." : text;`
> 
> There are no engine that can accept infinite length of text, so it is also a target to be handle in abstract class.

An engine which is not network-based would probably not have an explicit limitation. And you probably noticed that the value differs from one engine  to another.

We could add a method like this in IMachineTranslate:
default int maxTextLength() { return Integer.MAX_VALUE; }
then only engines which have such a limit will need to implement it, and we don't break compatibility

miurahr:
> First method is simpler and does not break the code of existing engines... - getFromCache and putToCache can even be private in this case.

No.
https://wiki.eclipse.org/Evolving_Java-based_APIs_2

> Decrease access; that is, from protected access to default or private access; or from public access to protected, default, or private access | - | Breaks compatibility

t-cordonnier:
> No. https://wiki.eclipse.org/Evolving_Java-based_APIs_2

When I wrote this I remembered that you just decided to move all calls to putToCache from existing engines into the main class. But you are right, I did not think about external engines which are not part of the core.

miurahr:
>An engine which is not network-based would probably not have an explicit limitation. And you probably noticed that the value differs from one engine to another.
>
>We could add a method like this in IMachineTranslate:
>default int maxTextLength() { return Integer.MAX_VALUE; }
>then only engines which have such a limit will need to implement it, and we don't break compatibility

Thank you for suggestion. I'll take it.

miurahr:
I'd like to close here and discussion on #473 for merge into 5.8/6.0.

PR #473, opened 2023-02-11
Refactor(MT):  handle maxTextLength in BaseTranslate

miurahr:
## Pull request type

- Other (describe below)

refactor

## Which ticket is resolved?

Suggested in PR #470

## What does this PR change?

- Introduce `BaseCachedTranslate` class that has Cache layer.
- Change `BaseTranslate` not to handle cache
- Introduce `BaseTranslate#getTruncateText`  and handle source text max length in common
- Add constructor with maxTextLength

PR #474, opened 2023-02-11
refactor(mt): introduce BaseCachedTranslate class

miurahr:
## Pull request type

- Other (describe below)

refactor

## Which ticket is resolved?

Suggested in PR #470

## What does this PR change?

- Introduce `BaseCachedTranslate` class that has Cache layer.
- Change `BaseTranslate` not to handle cache

### DIFF
--- a/src/org/omegat/core/machinetranslators/BaseCachedTranslate.java
+++ b/src/org/omegat/core/machinetranslators/BaseCachedTranslate.java
@@ -1,0 +1,177 @@
+/**************************************************************************
+ OmegaT - Computer Assisted Translation (CAT) tool
+          with fuzzy matching, translation memory, keyword search,
+          glossaries, and translation leveraging into updated projects.
+
+ Copyright (C) 2010-2015 Alex Buloichik
+               2013 Didier Briel
+               2022,2023 Hiroshi Miura
+               Home page: http://www.omegat.org/
+               Support center: https://omegat.org/support
+
+ This file is part of OmegaT.
+
+ OmegaT is free software: you can redistribute it and/or modify
+ it under the terms of the GNU General Public License as published by
+ the Free Software Foundation, either version 3 of the License, or
+ (at your option) any later version.
+
+ OmegaT is distributed in the hope that it will be useful,
+ but WITHOUT ANY WARRANTY; without even the implied warranty of
+ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ GNU General Public License for more details.
+
+ You should have received a copy of the GNU General Public License
+ along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ **************************************************************************/
+
+package org.omegat.core.machinetranslators;
+
+import java.util.OptionalLong;
+
+import javax.cache.Cache;
+import javax.cache.CacheManager;
+import javax.cache.Caching;
+import javax.cache.expiry.CreatedExpiryPolicy;
+import javax.cache.expiry.Duration;
+import javax.cache.spi.CachingProvider;
+
+import com.github.benmanes.caffeine.jcache.configuration.CaffeineConfiguration;
+
+import org.omegat.core.CoreEvents;
+import org.omegat.core.events.IProjectEventListener;
+import org.omegat.gui.exttrans.IMachineTranslation;
+import org.omegat.util.Language;
+
+/**
+ * Base class for machine translation.
+ *
+ * @author Alex Buloichik (alex73mail@gmail.com)
+ * @author Didier Briel
+ * @author Hiroshi Miura
+ */
+@SuppressWarnings("deprecation")
+public abstract class BaseCachedTranslate extends BaseTranslate implements IMachineTranslation {
+
+    private static final int SIZE_OF_CACHE = 1_000;
+
+    /**
+     * Machine translation implementation can use this cache for skip requests
+     * twice. Cache will be cleared when project change.
+     */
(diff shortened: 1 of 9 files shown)


src/org/omegat/core/machinetranslators/BaseTranslate.java
```diff
@@ -86,56 +73,10 @@ public BaseTranslate() {
...
-     * Creat cache object.
-     * <p>
-     * MT connectors can override cache size and invalidate policy.
-     * 
-     * @param name
-     *            name of cache which should be unique among MT connectors.
-     * @return Cache object
-     */
-    protected Cache<String, String> getCacheLayer(String name) {
-        return getCaffeineCache(name, 1_000, Duration.ONE_DAY);
```
miurahr, line 103: Removed cache handling code from `BaseTranslate` class.

src/org/omegat/core/machinetranslators/BaseTranslate.java
```diff
@@ -157,25 +98,77 @@ public void setGlossarySupplier(IMTGlossarySupplier glossarySupplier) {
...
     }
 
     @Override
     public String getCachedTranslation(Language sLang, Language tLang, String text) {
-        if (enabled) {
-            return getFromCache(sLang, tLang, text);
-        } else {
-            return null;
-        }
+        return null;
```
miurahr, line 109: Always return null from `BaseTranslate`

src/org/omegat/core/machinetranslators/BaseTranslate.java
```diff
@@ -157,25 +98,77 @@ public void setGlossarySupplier(IMTGlossarySupplier glossarySupplier) {
...
+     *            Source langauge.
+     * @param tLang
+     *            Target language.
+     * @param text
+     *            source text.
+     * @return translated text if exists in cache, otherwise null.
+     */
+    protected String getFromCache(Language sLang, Language tLang, String text) {
+        return null;
+    }
```
miurahr, line 135: Keep method for backward binary compatibility.

src/org/omegat/core/machinetranslators/BaseTranslate.java
```diff
@@ -285,6 +260,20 @@ protected boolean isCredentialStoredTemporarily(String id) {
...
+     *            original source text.
+     * @return truncated text.
+     */
+    protected String getTruncateText(String text) {
+        if (maxTextLength <= 0) {
+            return text;
+        }
+        return text.length() > maxTextLength ? text.substring(0, maxTextLength - 3) + "..." : text;
+    }
+
```
miurahr, line 276: For future API stability.  see #473

miurahr (commented):
These are the changes other than spotless coding style change.

src/org/omegat/core/machinetranslators/BaseTranslate.java
miurahr: It seems to be able to be private. I'd like to remove it.

t-cordonnier (changes requested):
As said in another comment, I would recommend that in BaseCachedTranslate, methods `getTranslation` and `getCachedTranslation` become final, and all cache-related methods become private. This prevents the case where a user overrides getCachedTranslation in a way where the variable `cache` is still here but not used.
Users which don't agree with your cache policy can simply propose an alternative to BaseCachedTranslation class.

Except of that I approve this code.

miurahr:
> all cache-related methods become private. 

@t-cordonnier  You cannot change `BaseCachedTranslate#getFromCache` and `BaseCachedTranslate#setToCache` method to be private without changing `BaseTranslate#putFromCache` and `BaseTranslate#putToCache`  that breaks binary compatibility against 3rd-party MT plugins that support OmegaT  5.7.1 and before.

Could you suggest a better way?

miurahr:
@t-cordonnier  does #473 make you satisfied?

t-cordonnier:
> @t-cordonnier You cannot change `BaseCachedTranslate#getFromCache` and `BaseCachedTranslate#setToCache` method to be private without changing `BaseTranslate#putFromCache` and `BaseTranslate#putToCache` that breaks binary compatibility against 3rd-party MT plugins that support OmegaT 5.7.1 and before.
> 
> Could you suggest a better way?

You can depreciate the method putToCache in BaseTranslate and use a new name in BaseCachedTranslate
Or even depreciate class BaseTranslate and start a new hierarchy

> @t-cordonnier does https://github.com/omegat-org/omegat/pull/473 make you satisfied?

Sorry when I reviewed it I thought it was identical except for adding maxTextLength handling
I will have a new look and tell you later

miurahr:
A change #473 will be future figure and I think we reach a result;

- make putToCache and getFromCache deprecated  (in 6.1)
- make getCachedTranslate and getTranslate final (in 6.1)
- add getMaxTextLength()
- introduce BaseCachedTranslate class

I've changed last two here,  and plan to first two.

src/org/omegat/core/machinetranslators/BaseCachedTranslate.java
```diff
@@ -0,0 +1,177 @@
...
+        } else {
+            return null;
+        }
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public final String getCachedTranslation(Language sLang, Language tLang, String text) {
```
miurahr, line 128: Now the method is final. Adjusted from review comment.

t-cordonnier (approved):
Looks ok for me now

Issue #475, opened 2023-02-12
Document 3rd match percentage calculation


brandelune:
https://github.com/omegat-org/omegat/pull/467

Files to modify:
OmegaT5_EditorsPanes.xml
OmegaT_new.css

brandelune:
Created PR.

PR #476, opened 2023-02-12
Add new ex. to fuzzy matches and new explanation to 3rd percentage

brandelune:
Documentation issue:
https://github.com/omegat-org/omegat/issues/475

Related to PR:
https://github.com/omegat-org/omegat/pull/467

### DIFF
--- a/doc_src/en/OmegaT5_EditorsPanes.xml
+++ b/doc_src/en/OmegaT5_EditorsPanes.xml
@@ -764,8 +764,12 @@ Dossier de configuration
       </listitem>
 
       <listitem>
-        <para>Percentage calculated on the entire text, including tags and
-        numbers (generally the lowest).</para>
+        <para>Percentage calculated on <emphasis role="bold">the entire text,
+        including tags and numbers</emphasis> (generally the lowest).</para>
+		<note>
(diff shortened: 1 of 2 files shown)


brandelune:
I've changed the line according to Briac's description. Although not directly related to this change (but related to the fuzzy pane) I added a second fuzzy match to the example and modified the CSS accordingly.

Kazephil:
@brandelune 
Looks good to me. I think we also need a note in the description of the minimum threshold to add the [information provided by Briac](https://github.com/omegat-org/omegat/pull/467#issuecomment-1427039197).

I just sent a PR for that to your fork.

brandelune:
> @brandelune Looks good to me. I think we also need a note in the description of the minimum threshold to add the [information provided by Briac](https://github.com/omegat-org/omegat/pull/467#issuecomment-1427039197).
> 
> I just sent a PR for that to your fork.

I don't think the extra info is required since it should have been like this from the beginning and people did not expect it to not work that way (hence the bug report). I don't think we should add info related to bug fixes in the manual :-)

miurahr (approved):

Kazephil (approved):

miurahr:

PR #477, opened 2023-02-13
Docs: emphasize info about fuzzies and add CSS

brandelune:
This was included in the 3rd TM match % commit but was unrelated so I separated it.


miurahr:
merged into master

PR #478, opened 2023-02-13
fix(filters4): get string data from character event

miurahr:

## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?

https://github.com/omegat-org/omegat/pull/239#issuecomment-1407349578

## What does this PR change?

- Character type of XMLEvent object(ev) of OpenJDK default StAX parser returns ev.asCharacters().getData() for toString() It is not guarantee toString() returns getData()
- Replace  ev.toString()  into  ev.asCharacters().getData()
- Add OpenXML filter test case to improve coverage


src/org/omegat/filters4/xml/openxml/OpenXmlFilter.java
```diff
@@ -510,8 +510,10 @@ private void browseRunContents(List<XMLEvent> run, ListIterator<XMLEvent> runIte
                 nList.add(eFactory.createEndElement(qR, null));
                 res.append("<" + prefixInt + tcInt + "/>");
                 tagsMap.put("" + prefixInt + tcInt, nList);
+            } else if (next.isCharacters()){
+                res.append(next.asCharacters().getData());
             } else {
-                res.append(next); // character data
+                // XXX: skip
```
miurahr, line 516: Is it really not come here?

PR #479, opened 2023-02-13
Add left thick border to glossary pane mock-up

brandelune:
This is intended to create the same L&F among various pane mock-ups (cf. Editor pane, fuzzy match pane).


miurahr:
Why not also change fuzzy match mock-up?

brandelune:
> Why not also change fuzzy match mock-up?

That's in #477 .

miurahr:
got it.   merged into master.

PR #482, opened 2023-02-14
fix(mt/google): throw MachineTranslateError exception when got error

miurahr:
## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?

N.A.

## What does this PR change?

- Google2MT  raise MachineTranslationError when got http/json parser error


miurahr:
merged into master.

PR #485, opened 2023-02-15
feat(pref/plugins): plugin table sorter

miurahr:
## Pull request type

## Which ticket is resolved?

- Title
  * URL

## What does this PR change?

-  sort plugin tables by its name and class name


miurahr:
merged into master.

PR #487, opened 2023-02-15
feat(pref/appearance): improve theme list

miurahr:
Enhance preference settings for theme.
Current implementation use ComboBox. It is good when number of selections is small.

When add  flat-theme plugin, there are so many themes, then UI is bad.

This proposal changes combobox to be  a single selection list that help users to see
all theme candidates.

## Pull request type

## Which ticket is resolved?

- Title ex. fix...
  * URL ex. https://...
 
- Title
  * URL

## What does this PR change?

-
-
-


miurahr:
I give up it.

PR #491, opened 2023-02-18
refactor(filters4): avoid catching NPE

miurahr:
Xliff2Filter use  `try { ... } catch (NPE e) { ..} `. This type of block is not recommended,
 and update as `if (foo !=null) { ...} else {... }`.   

## Pull request type

refactor

## What does this PR change?

- Refactor Xliff2Filter to avoid catch NPE to process.
- Update javadoc/comments


t-cordonnier (approved):
ok for me

PR #493, opened 2023-02-18
Remove change orientation shortcut from appendix

brandelune:
The appendix still had info about the shortcut.


miurahr:
merged into master.

PR #496, opened 2023-02-20
fix: Color the fuzzy matches from /tm/mt

briacp:
## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?

- Color the fuzzy matches from /tm/mt
  * https://sourceforge.net/p/omegat/bugs/1144

## What does this PR change?

- check if a fuzzy match comes from the /tm/mt folder before being inserted
- refactor the code that checks if a NearString comes from /tm/mt (the clunky test was also used in MainWindow)
- spotlessChangedApply on the 3 modified files


miurahr (approved):

brandelune:
Please, add a reference to the SF ticket in the changes.txt file.

PR #499, opened 2023-02-21
[1135] Avoid mnemonics in context menus

t-cordonnier:
## Pull request type

[bug]

## Which ticket is resolved?

https://sourceforge.net/p/omegat/bugs/1135

## What does this PR change?

- Remove & in contextual menus, since they do not support mnemonics

### DIFF
--- a/src/org/omegat/gui/editor/EditorPopups.java
+++ b/src/org/omegat/gui/editor/EditorPopups.java
@@ -7,6 +7,7 @@
                2010 Wildrich Fourie
                2011 Alex Buloichik, Didier Briel
                2015 Aaron Madlon-Kay
+               2023 Thomas Cordonnier
                Home page: http://www.omegat.org/
                Support center: https://omegat.org/support
 
@@ -62,6 +63,7 @@ import org.omegat.util.gui.UIThreadsUtil;
  * @author Wildrich Fourie
  * @author Didier Briel
  * @author Aaron Madlon-Kay
+ * @author Thomas Cordonnier
  */
 public final class EditorPopups {
     public static void init(EditorController ec) {
@@ -75,6 +77,14 @@ public final class EditorPopups {
     }
 
     private EditorPopups() {
+    }
+    
+    /** 
+     * Translate menu item but ensuring it does not contain mnemonics,
+     * because they do not work with context menus
+     **/
+    public static String translateMenu(String id) {
+        return OStrings.getString(id).replace("&","");
     }
 
     /**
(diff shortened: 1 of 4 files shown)


src/org/omegat/gui/editor/EditorPopups.java
```diff
@@ -76,6 +78,14 @@ public static void init(EditorController ec) {
 
     private EditorPopups() {
     }
+    
+    /** 
+     * Translate menu item but ensuring it does not contain mnemonics,
+     * because they do not work with context menus
+     **/
+    public static String translateMenu(String id) {
+        return OStrings.getString(id).replace("&","");
```
miurahr, line 87: This is not enough.
Please look
https://github.com/omegat-org/lib-mnemonics/blob/master/src/org/openide/awt/Mnemonics.java#L104-L108

t-cordonnier: OK, have a look to the PR in lib-mnemonics, if you aprove it it would make possible to have all about mnemonics in same location (else I must copy the regex here)
Tell me what you think

miurahr: You can find a lib-mnemonics version 1.1 now.
https://central.sonatype.com/artifact/org.omegat/lib-mnemonics/1.1

t-cordonnier: Corrected

src/org/omegat/gui/editor/EditorPopups.java
```diff
@@ -76,6 +80,14 @@ public static void init(EditorController ec) {
 
     private EditorPopups() {
     }
+    
+    /** 
+     * Translate menu item but ensuring it does not contain mnemonics,
+     * because they do not work with context menus
+     **/
+    public static String translateMenu(String id) {
```
damien-rembert, line 88: Could we name this method something else?
It seems to me like if we can avoid naming things "translate", we should try to do that, considering lots of things in OmegaT already "translate" in one way or another.

miurahr: Could you consider it is as private method?
I think we can call `Mnemonics.removeMnemonics` from `DictionaryPopup` class and `MultipleTranslationPopup` class.

t-cordonnier: The goal of the method was to ensure that all popups which come to the same editor receive the same treatments, in case we later need to add something more than only removing the mnemonics.
But considering the polemics it created, I prefer to remove it completely.
Corrected now.

miurahr:
squashed and merged.

PR #501, opened 2023-02-23
Fix error: not "ten days" but "ten files"

brandelune:
https://sourceforge.net/p/omegat/mailman/message/37781955/

by @miurahr.


miurahr:
merged into master.

PR #503, opened 2023-02-24
Hide tag only

t-cordonnier:
## Pull request type

[enhancement]

## Which ticket is resolved?

https://sourceforge.net/p/omegat/feature-requests/1672/

## What does this PR change?

- Adds a checkbox menu  item "Hide segments with only tags"  in "View" menu

When active, a new filter hides all segments which contain only tags. They are only hidden, still in memory. It works using the interface IEditorFilter already used when you click on "Set filter" in one of the searches window
In case you set a filter using search window, this new filter will temporarily replace the filter to hide segments. When you click later on "remove filter", it will restore the previous state, meaning that the "hide segments" filter comes back if the checkbox item is still active.
Segments remain in memory, they are not lost. If you don't translate them, they will be translated as identical to the source, as it would have been the case if you forgot them during translation without the menu active.


brandelune:
@t-cordonnier 

You may have missed the discussion on SF:
https://sourceforge.net/p/omegat/feature-requests/1672/#01ee

Also, regardless of the state of advancement of this feature, let's agree to not merge it before 6.1.

t-cordonnier:
Not sure I understand why you see this as a generalization of an option which is in the filter. The only common part I see is that it uses a regular expression.
But to be clear: my work modifies _display_, not _loading_, contrarily to file filters. Oh, maybe you are confused because I spoke about editor filters? But no, editor filters (contrarily to file filters) work by hiding segments in the editor, without any consequence on what is in memory.
Hope it is more conform to what you would expect.

miurahr (changes requested):
Please add GUI  parts in `src/org/omegat/gui/preferences/view/EditingBehavirPanel.form(.java)` using NetBeans GUI editor.

t-cordonnier:
Actually JC refuses the pull request with other change requests which are much more complicated, so I will not work on it right now. 
You can decide to close or not, as you prefer

miurahr:
OK, I close here.

PR #506, opened 2023-02-28
Fix aligner "hide tags" string

brandelune:
In the aligner, tags are not hidden, they're removed.

https://sourceforge.net/p/omegat/feature-requests/1677/#2996/0520/fd4b/f0b6


kosivantsov (approved):
Looks good

miurahr:
squashed and merged into master.

PR #511, opened 2023-03-03
fix: work around for JRE 8 l10n bug

miurahr:
- JRE 8 has a localization bug in Chinese
  - Locale class has wrong display name for ZA, SD, SD_IN and SD_PK
- The problem is reported at https://github.com/OmegaT-L10N/zh_CN/issues/4 and https://github.com/OmegaT-L10N/zh_CN/issues/5

## Pull request type

- Bug fix -> [bug]
- 
## Which ticket is resolved?

- Title ex. fix...
  * URL ex. https://...
 
- SD-Sindhi language 
  https://github.com/OmegaT-L10N/zh_CN/issues/4

- ZA-Shuang language
   https://github.com/OmegaT-L10N/zh_CN/issues/5

## What does this PR change?

-
-
-

### DIFF
--- a/src/org/omegat/util/Language.java
+++ b/src/org/omegat/util/Language.java
@@ -108,11 +108,31 @@ public class Language implements Comparable<Object> {
         }
     }
 
+    private final static String ZA_IN_ZH= "\u58EE\u8BED";
+    private final static String SD_IN_ZH= "\u4FE1\u5FB7\u6587";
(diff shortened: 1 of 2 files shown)


miurahr:
You may have an opinion that OmegaT 6.0 will ship with JRE 11 which don't have the bug.

brandelune:
> You may have an opinion that OmegaT 6.0 will ship with JRE 11 which don't have the bug.

If the fix is not too hard, let's do it for 5.8.

miurahr:
merged into master.

PR #512, opened 2023-03-04
fix: catch plugin load error

miurahr:
When starting OmegaT on JRE 8 with okapi plugin which requires Java 11, OmegaT fails to start.
This catch load error with Java version mismatch, and continue starting with leaving log.

## Pull request type

- Other (describe below)

improve error handling

## What does this PR change?

- catch `UnsupportedClassVersionError` and leave log


miurahr:
merged into master.

PR #516, opened 2023-03-10
Topic/jc/dev list fixes

brandelune:
Fixes a number of small UI recently identified and discussed on the dev list:

- Advanced search functions labels
  If we don’t have a fix for the code, we can have one line in the manual to describe the feature scope, along with a change in the labels
  https://sourceforge.net/p/omegat/mailman/message/37787435/

- External TM creation
  That’s a bug, but the easy fix is to document it
  https://sourceforge.net/p/omegat/mailman/message/37785525/

- Clarify which bilingual formats can be used as TMs
  That’s a documentation issue
  https://sourceforge.net/p/omegat/mailman/message/37785523/

Also fixes "MediaWiki" related strings.


miurahr:
merged into master.

PR #521, opened 2023-03-20
Add information that the credentials may be wrong.

brandelune:
Fixes bug [1153](https://sourceforge.net/p/omegat/bugs/1153/) in a soft way, since it only changes the error message to help the user debug the issue.

It is a temporary fix and a real fix should address the fact that OmegaT properly detects when credentials are wrong and should provide that information right away.


src/org/omegat/Bundle.properties
```diff
@@ -2324,7 +2324,7 @@ GIT_ERROR=Git '{0}' error: {1}
 GIT_CONFLICT=Push failed. Will be synchronized next time.
 GIT_SAVE_PLAINPASSWORD=Allow to save password as plain text
 GIT_CREDENTIAL_MESSAGE=Git notification: {0}
-VCS_OFFLINE=No connection available, the project is in offline mode.
+VCS_OFFLINE=OmegaT does not have access to the remote repository: verify your connection or your credentials. The project is in offline mode.
```
miurahr, line 2327: Please change key when meaning is changed.
It does not fix in localized message.

miurahr (changes requested):

src/org/omegat/Bundle.properties
```diff
@@ -2324,7 +2324,7 @@ GIT_ERROR=Git '{0}' error: {1}
 GIT_CONFLICT=Push failed. Will be synchronized next time.
 GIT_SAVE_PLAINPASSWORD=Allow to save password as plain text
 GIT_CREDENTIAL_MESSAGE=Git notification: {0}
-VCS_OFFLINE=No connection available, the project is in offline mode.
+VCS_ACCESS_PROBLEM=OmegaT does not have access to the remote repository: verify your connection or your credentials. The project is in offline mode.
```
Kazephil, line 2327: Slightly shorter alternative:
"OmegaT cannot access the remote repository. Check your connection or credentials. The project is in offline mode."

It's a matter of preference more than anything else. Pick the version you prefer! :smile:

Kazephil (approved):

miurahr (approved):

PR #523, opened 2023-03-28
[6.1] newUI

miurahr:
Propoosal of `newUI` that can be added as optional feature.

The newUI allow users to select project, select files without modal dialogs.
Hide all traditional menus in burger menu.

## Pull request type

## Which ticket is resolved?

- [Dev ML post](https://sourceforge.net/p/omegat/mailman/message/37796505/)

## What does this PR change?

- Introduce BaseMainWindowMenu class that has most of previous MainWindowMenu class.
- MainWindowMenu class only has its specific part, that create classic menuBar object.
- Add NewMainMenu class that realize  Hamburger menu, selectors, and search/settings buttons..

## Other information

image


brandelune:
Restarting does not change the UI state.

src/org/omegat/util/Preferences.java
```diff
@@ -500,6 +500,7 @@ public final class Preferences {
     public static final String THEME_CLASS_NAME = "theme_class_name";
     public static final String THEME_CLASS_NAME_DEFAULT = "org.omegat.gui.theme.DefaultFlatTheme";
     public static final String GLOSSARY_SORT_BY_LENGTH = "glossary_sort_by_length";
+    public static final String APPLY_NEW_UI = "apply_new_ui";
```
miurahr, line 503: Preference key is here.

miurahr: Class name and property name that have prefixed with "new" is bad idea.

miurahr:
> > > Restarting does not change the UI state.
> > 
> > 
> > It works on Linux and Windows. I don't know why it does not work. Please share your insight.
> 
> Something happens: the main menu is disabled, but nothing comes in its stead. Screen Shot 2023-03-30 at 21 36 06 Screen Shot 2023-03-30 at 21 35 43
> 
> In any case, I don't think that state should be triggered by an item in the preference settings that requires a restart. This state should be available as a "View" item, like a Pane that we minimize or maximize: a part of the UI is available, or it is not.

I don't know macOS and I don't and wont have it,  I'd like to disable newUI for mac. When someone can fix it, he/she can enable it.

brandelune:
> I don't know macOS and I don't and wont have it, I'd like to disable newUI for mac. When someone can fix it, he/she can enable it.

I don't agree with that. Either we have features that work reasonably well on the 3 platforms or we don't have the feature.

Let's just try things the way it was in your first implementation. We have the "tool bar" on by default for now.

We can't make global design decisions based on some arbitrary preference setting code that doesn't work.

miurahr:
> > I don't know macOS and I don't and wont have it, I'd like to disable newUI for mac. When someone can fix it, he/she can enable it.
> 
> I don't agree with that. Either we have features that work reasonably well on the 3 platforms or we don't have the feature.
> 
> Let's just try things the way it was in your first implementation. We have the "tool bar" on by default for now.
> 
> We can't make global design decisions based on some arbitrary preference setting code that doesn't work.

Please tell me how to fix this. What is wrong?

miurahr:
> Also, as I wrote above, I think that this "Quick access tool bar" should be like a View item, or like a Pane that we can open or close at will. It should not require restart.

This is not a "tool bar" but a "menu bar" with selector menu items and icon only menus.
image

So it is on same place as menu bar.  There is not necessary new pane to place of quick access tool items. 

Do you like a "tool bar" for mac?

miurahr:
Now support dark theme.
image

When 3rd party theme provide  no-decoration style,  newUI put  burger menu ,  selector and access buttons on top of window.

brandelune:
> > Also, as I wrote above, I think that this "Quick access tool bar" should be like a View item, or like a Pane that we can open or close at will. It should not require restart.
> 
> This is not a "tool bar" but a "menu bar" with selector menu items and icon only menus. image
> 
> So it is on same place as menu bar. There is not necessary new pane to place of quick access tool items.

On the mac, the menu bar is at the top of the screen and is not associated to the application window.

Screen Shot 2023-04-01 at 11 58 05

In your first implementation (before the preference setting implementation), the new UI was in place of a "tool bar" as you'd see in LibreOffice for ex. Except that the standard menu bar was gone.

That's the reason why I call that a "toolbar".

> Do you like a "tool bar" for mac?

In LibreOffice, toolbars offer quick access to functions that are otherwise available in the menus, but do not remove menu access.

So, in what you are trying to implement, I think there are 2 elements that we need to separate:
- The quick access part
   The quick access part should be available as easily as a View menu item.

- The "hamburger" menu part
   The "hamburger" menu should/could be available as it is in Emacs, by hitting a key

Which means that this new UI cannot be a full replacement for the existing menu bar. The menu bar exists on its own and should not be removed from OmegaT.

I'd like to show you what it looked like before you implemented the preference setting, but since you force-pushed I can't go back to the original commits.

PR #526, opened 2023-04-03
fix: don't display preamble line in search results if no filename set

briacp:
## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?

- Fixed "null" file names in search results when "Display: file names" checkbox was unchecked
  * https://sourceforge.net/p/omegat/bugs/1154/
 

## What does this PR change?

- only display the preamble line (e.g. `filename.xxx +3 more`) when the "Display: file names" option is checked

## Other information

See https://sourceforge.net/p/omegat/bugs/1154/#9b36 for screenshots of the different search results

### DIFF
--- a/src/org/omegat/Bundle.properties
+++ b/src/org/omegat/Bundle.properties
@@ -1087,8 +1087,8 @@ ST_NOTHING_FOUND=No matches returned by search.
 SW_NR_OF_RESULTS=Matching segments: {0}
 
 SW_GLOSSARY_RESULT=Glossary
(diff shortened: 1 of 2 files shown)


brandelune:
I think the "+ n more" display should not be linked to the "display file name" option, but only to the "display all matches" option.

briacp:
Good idea, I updated the branch to have the following behavior:

## All off
image

## All on
image

## Only "File names"
image

## Only "All matches"
image

brandelune (approved):

PR #538, opened 2023-04-15
Support for SRX format

t-cordonnier:
For the history refer to #24


miurahr (approved):

miurahr:
I merged it but I found several bugs in the code.
Please fix it.
I will push several test cases for SRX reader/writer  in master branch. Please make it passed.

miurahr:
A change here generate broken `segmentaion.srx` file. Its "LanguageCode" field becomes localized language name other than SRX standard language code.

This cause loader failed to load the file which is generated by new SRX.java code. @t-cordonnier

PR #539, opened 2023-04-16
Enhance/Fix bug: SRX reader/writer, support buggy CONF file to migrate correct SRX file 

miurahr:
## Pull request type

- BUG fix

## Which ticket is resolved?
 
-  SRX conf file is produced with localized name of language
   -  https://sourceforge.net/p/omegat/bugs/1160/
   
- Use Jackson library as XML parser
   - https://sourceforge.net/p/omegat/feature-requests/1582/
   
## What does this PR change?

- Use Jackson XML mapper to write SRX file.
- Use XMLStreamInput to read SRX and conf file.
- Fix MapRule#setLanguage to store Language Code instead of Language name
- Add org.omegat.core.segmentation.LanguageCodes.getLanguageCodeByName static method to get the code
- Rename MapRule#getLanguage to MapRule#getLanguageName
- Rename MapRUle#getLanguageCode to MapRule#getLanguage
- Fix SRX#copy method to copy also a version, or other attributes
- Set  segmentSubFlows and cascade is true in default
- Drop JAXB dependency from runtime
- Fix SRX parser to store src->header->version attribute into SRX#version properly.
- When CONF reader encounter a language name with UNKOWN name value, OmegaT leave a log message with it.

### DIFF
--- a/src/org/omegat/core/segmentation/SRX.java
+++ b/src/org/omegat/core/segmentation/SRX.java
@@ -28,7 +28,10 @@
 package org.omegat.core.segmentation;
 
 import java.beans.ExceptionListener;
+import java.beans.XMLDecoder;
 import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileNotFoundException;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
@@ -176,7 +179,8 @@ public class SRX implements Serializable {
     }
 
     /**
-     * Loads segmentation rules from an XML file.
+     * Loads segmentation rules from an XML file. If there's an error loading a
+     * file, it calls <code>getDefault</code>.
      * <p>
      * Since 1.6.0 RC8 it also checks if the version of segmentation rules saved
      * is older than that of the current OmegaT, and tries to merge the two sets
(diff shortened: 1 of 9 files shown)


miurahr:
Attached file is generated by proposed  `writeSrx` with Jackson xml mapper.
[segmentation.srx.zip](https://github.com/omegat-org/omegat/files/11241278/segmentation.srx.zip)

t-cordonnier:
> Attached file is generated by proposed `writeSrx` with Jackson xml mapper. [segmentation.srx.zip](https://github.com/omegat-org/omegat/files/11241278/segmentation.srx.zip)

Hi Hiroshi, thanks for all of this
Good that you provided a sample, I am a little bit surprised to see the rule names in japanese. I try to remember that, almost for the default file which is inside OmegaT.jar, the UI displays the name in the language of the user but stores them in english in the CONF file. I remember that I had to do a correction about this in DGT. Give me some time I will investigate if I can find again the correction I did.

src/org/omegat/core/segmentation/datamodels/MappingRulesModel.java
```diff
@@ -89,7 +90,13 @@ public void setValueAt(Object aValue, int rowIndex, int columnIndex) {
...
         switch (columnIndex) {
         case 0:
-            maprule.setLanguage((String) aValue);
+            String target = (String) aValue;
+            String code = LanguageCodes.getLanguageCodeByName(target);
+            if (code != null) {
+                maprule.setLanguage(code);
+            } else {
+                maprule.setLanguage(target);
+            }
```
miurahr, line 99: A code here is for fixing the issue which  OmegaT produce a CONF/SRX file with localized language name.

test/src/org/omegat/core/segmentation/SRXTest.java
```diff
@@ -93,8 +91,8 @@ public void testSrxReaderDefault() {
         assertEquals(18, mapRuleList.size());
         for (MapRule mapRule : mapRuleList) {
             if (mapRule.getPattern().equals("JA.*")) {
-                assertEquals(LanguageCodes.JAPANESE_CODE, mapRule.getLanguageCode());
-                assertEquals(OStrings.getString(LanguageCodes.JAPANESE_KEY), mapRule.getLanguage());
+                assertEquals(LanguageCodes.JAPANESE_CODE, mapRule.getLanguage());
+                assertEquals(OStrings.getString(LanguageCodes.JAPANESE_KEY), mapRule.getLanguageName());
             }
```
miurahr, line 96: Check here, `MapRule#getLanguage` returns CODE, and `MapRule#getLanguageName` returns localized name.

All previous versions failed here in environment other than English.

src/org/omegat/core/segmentation/SRX.java
```diff
@@ -418,24 +397,29 @@ public List<Exception> getExceptionsList() {
...
-        SRX srx = loadSrxFile(SRX.class.getResource("defaultRules.srx"));
-        srx.includeEndingTags = true;
-        srx.segmentSubflows = true;
-        return srx;
+        try {
+            SRX srx = loadSrxInputStream(SRX.class.getResourceAsStream("defaultRules.srx"));
+            srx.includeEndingTags = true;
+            srx.segmentSubflows = true;
+            return srx;
+        } catch (IOException e) {
```
miurahr, line 405: Is it OK for default values?

src/org/omegat/core/segmentation/datamodels/MappingRulesModel.java
t-cordonnier: Works better but now if I close OmegaT and open it again, names are in english.
I will search whenever I can find a method to keep localization in the UI without having it in the file

src/org/omegat/core/segmentation/SRX.java
t-cordonnier: Yes, for the default values you can refer to https://www.unicode.org/uli/pas/srx/srx20.html#formathandle

src/org/omegat/core/segmentation/datamodels/MappingRulesModel.java
miurahr: `getValueAt` is what you look for

t-cordonnier: Yes, just pushed a correction to this branch 
Seems to work, now if I remove a rule set and create a new one still with name in french, setValueAt makes the correct correspondance between what I type and the code

PR #548, opened 2023-04-24
docs(first step): remove version string from First_Step

miurahr:
When show on OmegaT version number in First step article, it should be as same as OmegaT release version.
Current version number is set from manual version, that is incorrect.

## Pull request type

- Bug fix -> [bug]
- Other (describe below)

## Which ticket is resolved?

-  First step document show wrong OmegaT version
  * https://sourceforge.net/p/omegat/documentation/409/

## What does this PR change?

- Remove referece to manual version.
-
-

### DIFF
--- a/doc_src/ca/First_Steps.xml
+++ b/doc_src/ca/First_Steps.xml
@@ -5,7 +5,7 @@
 %manualvariables;
 ]>
 <article lang="ca" id="first.steps">
-  <title>OmegaT &vernb;</title>
+  <title>OmegaT</title>
   <section>
 	<title id="first-steps">Primers passos</title>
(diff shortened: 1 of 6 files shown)


brandelune (changes requested):
I'm not sure I understand the problem. The version number is automatically corrected during the build process.

Plus, I don't think that we should add strings, even in the title. People know what OmegaT is.

miurahr:
> The version number is automatically corrected during the build process.

No. it is replaced with wrong document version, not a omegat software version.

miurahr:
> > The CSS files should not be changed. Only the file here: https://github.com/omegat-org/omegat/blob/master/doc_src/OmegaT_new.css
> > should be changed.
> 
> I don't change a source file, just run build process. If it should not be chaged, there may be a bug in document build script.

https://github.com/omegat-org/omegat/commit/c8b574bd1fbdb6232b23d1ca5e58b88e5a034935

Then souce css was changed by @Kazephil  last week and only update an english build results, but not against other languages.

miurahr:
I remake PR only with a removal of version number.

brandelune (approved):

PR #551, opened 2023-04-26
chore(gradle/modular): introduce test-fixtures

miurahr:
Add test fixtures to share common test functions among modules.

## Pull request type

- Other (describe below)

refuctoring build.gradle and test structure for modular source

## What does this PR change?

- When adding new module with test code, new test will be depends on `TestCore` class.
- `Test Fixtures` is a standard feature of gradle, that share common feature and resource among sub projects. 
- This move `TestCore`, `TestPreferenceInitializer`  and `TestCoreInitializer` into test-fixture.

## Other information

- It is mandatory when we move MT connectors into its module


t-cordonnier:
Interesting, but could you do it in a branch which already contains a module? I would prefer to test the feature with a concrete case rather than evaluating only the changes in the code

miurahr:
> Interesting, but could you do it in a branch which already contains a module? I would prefer to test the feature with a concrete case rather than evaluating only the changes in the code

Ok, I'll send it with another PR.

briacp (commented):
great idea to use `java-test-fixtures`, thanks Hiroshi

PR #553, opened 2023-04-27
Exclude `xml-apis` from tipoftheday dependencies in build.gradle

briacp:
Otherwise, Eclipse is confused as the javax.xml.* modules are duplicated: `The package [...] is accessible from more than one module: <unnamed>, java.xml`

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -257,7 +257,9 @@ dependencies {
         }
 
         // dependencies in submodules, you should put it as compileOnly in subprojects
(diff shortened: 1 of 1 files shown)


miurahr:
@briacp Thank you for heads me up. this is an issue in transitive dependency [CSSBox4](https://codeberg.org/miurahr/cssbox4) and I just fixed it and released new.

briacp:
Fixed by #555

PR #554, opened 2023-04-28
Allow DTD parsing in order to load internal entities

briacp:
## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?

- Project not loaded if it contains DTD internal entities
  * https://sourceforge.net/p/omegat/bugs/1163/

## What does this PR change?

- Essentially, it sets the `XMLInputFactory.SUPPORT_DTD` flag to true when creating the XmlMapper

## Other information

See also: https://sourceforge.net/p/omegat/mailman/message/37805777/

### DIFF
--- a/release/changes.txt
+++ b/release/changes.txt
@@ -31,6 +31,9 @@
 
   Bug fixes:
 
+  - Project not loaded if it contains DTD internal entities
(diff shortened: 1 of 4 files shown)


miurahr (approved):
@briacp it is a good catch!

PR #559, opened 2023-04-30
[modular, 6.1] Migrate machinetranslation connectors as module

miurahr:
## Pull request type

- Other (describe below)

## What does this PR change?

- Add test fixtures to share common test functions among modules.
- When adding new module with test code, new test will be depends on TestCore class.
- Test Fixtures is a standard feature of gradle, that share common feature and resource among sub projects.
- This move TestCore, TestPreferenceInitializer and TestCoreInitializer into test-fixture.
- Migrate machine translator connectors as module of omegat project.

## Other information

Related PR #551


t-cordonnier (commented):
I tested the new release, first with gradlew run, then building package => OK
Launched the tests => needs test fixtures, when it is here test run OK

So, LGTM.

t-cordonnier (approved):
OK for me after the tests mentioned in previous comment

miurahr:
> OK for me after the tests mentioned in previous comment

Thank you for feedback. 
I think we need to decide how modular OmegaT project is.
It is not discussed well which features should be subproject.

I believe MT connectors obviously are not "core API" of OmegaT, so it is valuable to modularize.
Because I don't change package path, there are classes in  "org.omegat.core.machinetranslators", where something is "core".

It is beause I put the PR as draft.

miurahr:
I've modularized each connectors individually for future plugin management feature.

miurahr:
It is a chance we change the package path of MT connectors. I'd like to propose "org.omegat.connectors.machinetranslators"

t-cordonnier:
> It is a chance we change the package path of MT connectors. I'd like to propose "org.omegat.connectors.machinetranslators"

OK, what do you propose for global classes like BaseTranslate and BaseCacheTranslate?
Should we or not have one sub-package per connector?

miurahr:
waiting #586 to be merged.

miurahr:
Let's merge and ask feedback from community.

kosivantsov:
What about the MT connectors' localizeble strings? Will they be moved to each individual plugin, or remain in OmegaT's `Bundle.properties`?

And how do I build each plugin? `../gradlew jar` in plugins' directories doesn't produce jar files.

miurahr:
> What about the MT connectors' localizeble strings? Will they be moved to each individual plugin, or remain in OmegaT's `Bundle.properties`?

They will be moved to each individual plugin. 
Thank you for feedback!

miurahr:
There is a bug that module jar does not installed in distributions.
I've pushed the fix as #656

PR #560, opened 2023-05-01
refactor: improve LanguageToolTest

miurahr:
- exclude duplicated slf4j logging module
- use Assert.assertThrows for error check test
- improve a test message because an old XML API was removed in LanguageTool 3.7 and later.

## Pull request type
- Other (describe below)

refactoring

### DIFF
--- a/test/src/org/omegat/languagetools/LanguageToolTest.java
+++ b/test/src/org/omegat/languagetools/LanguageToolTest.java
@@ -28,8 +28,8 @@ package org.omegat.languagetools;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertThrows;
 import static org.junit.Assert.assertTrue;
-import static org.junit.Assert.fail;
 
 import java.util.Collections;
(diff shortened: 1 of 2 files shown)


test/src/org/omegat/languagetools/LanguageToolTest.java
```diff
@@ -99,14 +99,10 @@ public void testRemoteServer() throws Exception {
...
-            try {
-                new LanguageToolNetworkBridge(SOURCE_LANG, TARGET_LANG, "http://localhost:8081");
-                fail("URL not specifying API v2 should fail due to XML response instead of JSON");
-                // TODO: LanguageTool will drop XML entirely in version 3.6; this
-                // test might need to be adjusted then.
-            } catch (Exception e) {
-                // OK
-            }
+            assertThrows("URL not specifying API actions should fail due to missing argument.",
+                    java.lang.Exception.class,
```
miurahr, line 103: Is it really good when LanguageToolNetworkBridge class throws generic `Exception` class other than specific exception that has a reason?

miurahr: I will raise another refactoring PR for it.

PR #569, opened 2023-05-07
[6.1][RFE#1169] feat: use Apache MINA SSHD for SVN transport

miurahr:
SVNKit 1.10.5 and later support apache sshd and recommend to switch to it.

SVNKIT-759: Apache sshd based implementation for svn+ssh:// protocol as an alternative to trilead.

## Pull request type

## Ticket

- Team: SVN support svn+ssh protocol with key authentication
    - https://sourceforge.net/p/omegat/feature-requests/1700/
    
-  Team: SVN: Replacing trilead-ssh2 with Apache sshd
    - https://sourceforge.net/p/omegat/feature-requests/1699/

## What does this PR change?

- Set System property to specify Apache MINA-SSHD for SVNKit for team subversion connections
- It supports EC cipher
- SVN/Git use a common apache ssh library for team connections

## Other information

- [x] Integration test with SVN+SSH

### DIFF
--- a/src/org/omegat/core/team2/impl/SVNAuthenticationManager.java
+++ b/src/org/omegat/core/team2/impl/SVNAuthenticationManager.java
@@ -25,16 +25,12 @@
 
 package org.omegat.core.team2.impl;
 
+import java.io.Console;
+import java.util.Arrays;
 import java.util.logging.Logger;
 
 import javax.net.ssl.TrustManager;
 
-import org.omegat.core.Core;
-import org.omegat.core.KnownException;
-import org.omegat.core.team2.ProjectTeamSettings;
-import org.omegat.core.team2.TeamSettings;
-import org.omegat.util.Log;
-import org.omegat.util.OStrings;
 import org.tmatesoft.svn.core.SVNErrorCode;
 import org.tmatesoft.svn.core.SVNErrorMessage;
 import org.tmatesoft.svn.core.SVNException;
@@ -48,15 +44,23 @@ import org.tmatesoft.svn.core.auth.SVNSSHAuthentication;
 import org.tmatesoft.svn.core.auth.SVNUserNameAuthentication;
 import org.tmatesoft.svn.core.io.SVNRepository;
 
-import com.jcraft.jsch.agentproxy.AgentProxyException;
-import com.jcraft.jsch.agentproxy.ConnectorFactory;
-import com.jcraft.jsch.agentproxy.TrileadAgentProxy;
+import org.omegat.core.Core;
+import org.omegat.core.KnownException;
+import org.omegat.core.team2.ProjectTeamSettings;
+import org.omegat.core.team2.TeamSettings;
+import org.omegat.gui.main.ConsoleWindow;
+import org.omegat.util.Log;
+import org.omegat.util.OStrings;
+
+import gen.core.project.RepositoryDefinition;
 
 /**
  * Authentication manager for SVN. See details about authentication at the
- * http://wiki.svnkit.com/Authentication. Authentication manager created for each repository instance.
+ * http://wiki.svnkit.com/Authentication. Authentication manager created for
+ * each repository instance.
  *
- * Only username+password authentication supported. Proxy not supported for https:// repositories.
+ * Only username+password authentication supported. Proxy isn't supported for
+ * https:// repositories.
  *
  * @author Alex Buloichik (alex73mail@gmail.com)
  */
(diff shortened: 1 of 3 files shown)


miurahr:
Now ready for review.
This enhancement help us to put the integration test in CI process.

miurahr:
Improvements for Integration tests are now become an individual PR. After merge #633 I'd like to repush the change here.

src/org/omegat/core/team2/impl/SVNAuthenticationManager.java
```diff
@@ -130,66 +119,84 @@ protected SVNAuthentication ask(String kind, SVNURL url, String message) throws
...
 
     @Override
     public SVNAuthentication getFirstAuthentication(String kind, String realm, SVNURL url)
             throws SVNException {
-        if (predefinedUser != null && predefinedPass != null) {
-            if (ISVNAuthenticationManager.PASSWORD.equals(kind)) {
+        // Check supported kind and return when there are predefined credentials
+        switch (kind) {
+        case SSL:
+            return SVNSSLAuthentication.newInstance(kind, null, false, url, false);
```
miurahr, line 154: We don't support SSL authentication and also there is no test case for that.

src/org/omegat/core/team2/impl/SVNAuthenticationManager.java
```diff
@@ -232,4 +239,46 @@ public String getProxyUserName() {
         public void acknowledgeProxyContext(boolean accepted, SVNErrorMessage errorMessage) {
         }
     };
+
+    private SVNAuthentication getResult(String kind, SVNURL url, String user, char[] pass)
```
miurahr, line 243: A method name `getResult` has not enough meaning. It should be like `getAuthenticatorInstance`.

src/org/omegat/core/team2/impl/SVNRemoteRepository2.java
```diff
@@ -58,12 +70,21 @@
...
     RepositoryDefinition config;
     File baseDirectory;
     SVNClientManager ourClientManager;
     ProjectTeamSettings projectTeamSettings;
 
+    /**
+     * Constructor of SVN remote repository.
+     */
+    public SVNRemoteRepository2() {
+        // Specify MINA-SSHD for transport.
```
miurahr, line 84: It is better to point reference document that explain it.

src/org/omegat/core/team2/impl/SVNAuthenticationManager.java
```diff
@@ -68,17 +73,17 @@ public class SVNAuthenticationManager implements ISVNAuthenticationManager {
...
+    private final RepositoryDefinition repoDef;
     private final String predefinedUser;
     private final String predefinedPass;
-    //private final ProjectTeamSettings teamSettings;
+    private final ProjectTeamSettings teamSettings;
 
-    public SVNAuthenticationManager(String repoUrl, String predefinedUser, String predefinedPass,
-            ProjectTeamSettings teamSettings) {
-        this.repoUrl = repoUrl;
+    public SVNAuthenticationManager(RepositoryDefinition repoDef, String predefinedUser,
```
miurahr, line 81: A change of constructor arguments, it might break compatibility. It may be better to keep a constructor with old signature.

src/org/omegat/core/team2/impl/SVNAuthenticationManager.java
```diff
@@ -232,4 +239,46 @@ public String getProxyUserName() {
...
+            return SVNUserNameAuthentication.newInstance(user, false, url, false);
+        } else {
+            throw new SVNException(SVNErrorMessage.create(SVNErrorCode.RA_UNKNOWN_AUTH));
+        }
+    }
+
+    private Credentials loadCredentials(SVNURL url) {
+        Credentials credentials = new Credentials();
+        // check stored credential
+        credentials.username = TeamSettings.get(repoDef.getUrl() + "!" + KEY_USERNAME_SUFFIX);
```
miurahr, line 259: We use team setting here but we drop teamSetting argument in SVNAuthenticationManager constructor.
It would be better to keep ProjectTeamSettings and checks it.

src/org/omegat/core/team2/impl/SVNAuthenticationManager.java
miurahr: It would be better to raise exception as same as before.
```
throw new SVNException(SVNErrorMessage.create(SVNErrorCode.RA_UNKNOWN_AUTH));
```

src/org/omegat/core/team2/impl/SVNAuthenticationManager.java
miurahr: projectTeamSettings don't have a credential record.

miurahr:
Rebased on master.

PR #575, opened 2023-05-11
fix: write the project properties in the backup instead of copying file (bug #1166)

briacp:
## Which ticket is resolved?

- Error when trying to backup the project file on Windows
  * https://sourceforge.net/p/omegat/bugs/1166/
 

## What does this PR change?

- Instead of creating a backup by copying the `omegat.project` file, it serialize the project properties in a new backup file, thus circumventing the file lock.

### DIFF
--- a/test/src/org/omegat/util/FileUtilTest.java
+++ b/test/src/org/omegat/util/FileUtilTest.java
@@ -62,14 +62,20 @@ public class FileUtilTest {
 
     private File base;
 
+    private String oldTz;
+
     @Before
     public final void setUp() throws Exception {
+        oldTz = System.setProperty("user.timezone", "UTC");
         base = Files.createTempDirectory("omegat").toFile();
     }
 
     @After
     public final void tearDown() throws Exception {
         safeDeleteDirectory(base);
+        if (oldTz != null) {
+            System.setProperty("user.timezone", oldTz);
+        }
     }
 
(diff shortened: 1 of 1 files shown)


src/org/omegat/gui/main/ProjectUICommands.java
```diff
@@ -649,7 +650,11 @@ private static void projectOpenImpl(File projectRootFolder) {
                             backup.getName());
                     Core.getProject().saveProjectProperties();
                 } else if (FileUtil.getRecentBackup(projectFile) == null) {
-                    FileUtil.backupFile(projectFile);
+                    File backup = new File(projectRootFolder,
+                            String.format("%s.%s%s", OConsts.FILE_PROJECT,
+                                    new SimpleDateFormat("yyyyMMddHHmm").format(new Date(projectFile.lastModified())),
+                                    OConsts.BACKUP_EXTENSION));
```
miurahr, line 656: Here is copy of a similar lines of code with FileUtil class.
Are any ways to update `FileUtil.backupFile`  method? 
My idea is adding `FileUtil.getBackupFilename` utility method and `FileUtil.backupFile` and here share a same logic.

briacp: That's a good idea, thanks for the suggestion.

miurahr (approved):

PR #576, opened 2023-05-11
fix: avoid using filter option with `null` name

briacp:
## Which ticket is resolved?

- Default filters settings don't initialize options
  * https://sourceforge.net/p/omegat/bugs/1167/

### DIFF
--- a/test/src/org/omegat/filters2/FilterMasterTest.java
+++ b/test/src/org/omegat/filters2/FilterMasterTest.java
@@ -60,7 +60,8 @@ public class FilterMasterTest {
     public final void setUpCore() throws Exception {
         tempFilter = Files.createTempFile(FilterMaster.FILE_FILTERS, null).toFile();
 
-        FilterMaster.setFilterClasses(Arrays.asList(new Class[] { org.omegat.filters3.xml.xhtml.XHTMLFilter.class }));
+        FilterMaster.setFilterClasses(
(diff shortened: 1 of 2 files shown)


miurahr (approved):

test/src/org/omegat/filters2/FilterMasterTest.java
```diff
@@ -0,0 +1,124 @@
+package org.omegat.filters2;
```
miurahr, line 1: We need a copyright header here.

miurahr:
New test causes  a compile error of test code in master branch. @briacp  could you check it?

miurahr:
```
@@ -61,7 +61,7 @@ public class FilterMasterTest {
         tempFilter = Files.createTempFile(FilterMaster.FILE_FILTERS, null).toFile();
 
         FilterMaster.setFilterClasses(
-                Arrays.asList(new Class[] { org.omegat.filters3.xml.xhtml.XHTMLFilter.class }));
+                Arrays.asList(new Class<?>[] { org.omegat.filters3.xml.xhtml.XHTMLFilter.class }));
     }

```

kosivantsov:
Could this one also be backported to `releases/5.8` and `releases/6.0`? Or [this comment](https://sourceforge.net/p/omegat/bugs/1167/#d162/b924) still applies?

PR #577, opened 2023-05-13
[6.1][BUGS#1168] fix: project loading error on source file system  link  loop

miurahr:
When there is symbolic link loop in source folder, OmegaT give up loading project
and raise error dialog.
This change to skip source loading when detect link error, and continue loading the project.

## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?

-  OmegaT cannot load a project and error dialog with FileSystemLoopException
   * https://sourceforge.net/p/omegat/bugs/1168

## What does this PR change?

- Raise better error message when giving up loading project.(can backport to 5.8)
- Catch link loop error and stop loading source files (java 9 required)


miurahr:
Improvement for error message can be port to 6.0.

Kazephil:
> Improvement for error message can be port to 6.0 when requested.

Brought back into this branch, actually. I rephrased the error messages in your `releases/6.0` branch (see PR #589).

Kazephil (changes requested):
Looks good to me.
See detailed comment here: https://github.com/omegat-org/omegat/pull/589#issuecomment-1550770263

You'll have to bring the commits that improve the error messages from #589 into this branch before merging.

miurahr:
> You'll have to bring the commits that improve the error messages from #589 into this branch before merging.

I've picked message enhancements.

Kazephil:
> > You'll have to bring the commits that improve the error messages from #589 into this branch before merging.
> 
> I've picked message enhancements.

Thank you. Unfortunately (as I noted in #600), it looks like I'll have to change it again. Only one line seems to fit in the error dialog, even though the pane showing the link itself is quite large.

(Maybe it's just my system? I'm on Arch Linux.)

I found a bigger problem, though: using Java 11 on Arch Linux, when I build OmegaT for this branch and try to load a project with a looped link in the source folder, I get the (first line of the) new error message, but the project does not load.

I tried with both the Gradle `run ` and `InstallDist` tasks just in case compiling first makes a difference and on my system, for both the 5.8 loop exception branch and the 6.1 loop exception branch:
- With the 5.8 loop exception branch (#577), the error message is shown, and the project fails to load.
- With the 6.1 loop exception branch (this branch), the error message is shown and the project fails to load.

So, exactly the same behaviour for all branches. (It was the same for 6.0 when I tried, but I want to wait until the build against Java 11 PR has been merged into 6.0.)

Kazephil (approved):

miurahr:
> I found a bigger problem, though: using Java 11 on Arch Linux, when I build OmegaT for this branch and try to load a project with a looped link in the source folder, I get the (first line of the) new error message, but the project does not load.

I found the exception is thrown by java stream itself.
I try to catch it and return file lists that is walked before link loop detected.

A new  code  is able to run on Java 8, too.
It can be ported to 5.8/6.0.

src/org/omegat/util/FileUtil.java
```diff
@@ -472,9 +475,24 @@ public static List<String> buildRelativeFilesList(File rootDir, List<String> inc
...
+                stream.forEach(p -> result.add(root.relativize(p).toString().replace('\\', '/')));
+            } catch (UncheckedIOException e) {
+                if (e.getCause() != null && e.getCause() instanceof FileSystemLoopException) {
+                    // source file folder has a looped symbolic links
+                    String p = ((FileSystemLoopException) e.getCause()).getFile();
+                    Core.getMainWindow().displayErrorRB(e.getCause(), "TF_LOAD_ERROR_SOURCE_LOOP_EXCEPTION", p);
+                    Log.logErrorRB(e, "TF_LOAD_WARN_SOURCE_LOOP_EXCEPTION", p);
+                    // stopping walking source folder and continue
+                } else {
+                    throw e; // propagate to caller
```
miurahr, line 491: It may be better we check an `UncheckedIOException` cause and throw a proper error  class.

miurahr: Bing search chat teach me... 
>According to the web results, UncheckedIOException in Java is a class that wraps an IOException with an unchecked exception. It is a subclass of RuntimeException and was introduced in Java 8.

So I think we can unwrap UncheckedIOException object using `getCause()`  and it will be a subclass of IOException.

miurahr:
@Kazephil  could you check [new commit](https://github.com/omegat-org/omegat/pull/577/commits/095ca75cd7514c50f480dfcda635c9345b925ffe)?

I believe it solves the issue you suggested.

- Show warning when found link loop and continue loading the project
- Show warning when access permission error found, and continue loading the project.
- Show multi-line message in warning dialog

Kazephil:
@miurahr 

I tried this new commit. It's a big step forward. 

Here are my results after trying this new commit:

> * Show warning when found link loop and continue loading the project

- If the link loop is a file, the project loads normally (without an error message).
- If the link is a directory, the new multiline error message is shown and the project loads ***as an empty project without any source files***.  The error messages says "some source file(s) were skipped", but the actual behaviour is that **all** source files are skipped.

> * Show warning when access permission error found, and continue loading the project.

I'm not sure about this one. I tried creating a project with two source files:
1. The first file only has read/write permissions for root, and no group or other permissions
2. The second file has normal user permissions.

- The project loads as an empty project without showing any error message.

If I remove read permissions on the source folder:
- OmegaT loads an empty project without showing an error message.

If I remove read permissions on a subfolder in the source folder, I get the dialog with the error message, but the link only shows `{0}` instead of the actual path to the file with the access permission error.

> * Show multi-line message in warning dialog

That one shows up properly. It might be nice to add a blank line between the error message and the path to the folder that cause the error, but otherwise it looks good!

miurahr:
> * If the link is a directory, the new multiline error message is shown and the project loads _**as an empty project without any source files**_.  The error messages says "some source file(s) were skipped", but the actual behaviour is that **all** source files are skipped.

It is why Java walk through directory first, so you visit loop link at first, then you see empty.

miurahr:
Here is an idea when walk got error and there is no items, we retry without "follow link" strategy.

miurahr:
> Here is an idea when walk got error and there is no items, we retry without "follow link" strategy.

It is not a solution because it will cause duplication of picked files.

PR #580, opened 2023-05-13
[6.1]chore(gradle): set duplicationStrategy for Zip task, and replace deprecated jar.archiveName property

miurahr:
## Pull request type

## What does this PR change?

- Set INCLUDE strategy for Zip tasks
- Replace deprecated jar.archiveName with jar.archiveFileName.toString()


build.gradle
```diff
@@ -1251,3 +1258,7 @@ tasks.withType(JavaExec).configureEach {
 task printVersion {
     println project.version
 }
+
+tasks.withType(Zip).configureEach {
+    duplicatesStrategy = DuplicatesStrategy.INCLUDE
+}
```
miurahr, line 1264:
```suggestion
}

```

should place last line feed.

PR #581, opened 2023-05-14
Discrepancies between 6.0 UI and manual + review

brandelune:
As I was reviewing the 6.0 FR localization I found a number of discrepancies between the current UI and the manual.
I think they are all fixed here.
I also added a few minor UI review items.


brandelune:
I'm not sure why there are all those merging conflicts...

PR #586, opened 2023-05-14
chore: allow UTF-8 character as source file

miurahr:
- Allow UTF-8 character for comments, especially for author name

## Pull request type

- Build and release changes -> [build/release]
- Other (describe below)

## What does this PR change?

- Add compiler flag to allow UTF-8 in java source files
- Update test case not to check US-ASCII
- A check should be done by spotless and checkstyle, instead of crafted one

### Related topic:  UTF-8 in Bundle.properties

Here is a Dev-ML discussion thread about BUNDLE file.

https://sourceforge.net/p/omegat/mailman/omegat-development/thread/8zfUNKLS058FYMIUPADGzNmrOOu_73IQWvrNRLFGTbfPmZke8l6fR9yAtLT2j1cGOo6ZpeQ4j-hNbT8fNAfgAtaEb2kmScQffrmXhGtZmx4%3D%40northside.tokyo/#msg37844491

PR #587, opened 2023-05-14
[6.1] fix: jackson jaxb parser break SuperTMXMerge library behavior

miurahr:
Some Jackson configuration for JAXB parsing breaks SuperTMXMerge that used for merging translations among team member when concurrently translate in the same project.

When revert and update  some JAXB related part, the bug is gone.

I don't know the fact but a code
```
        XMLInputFactory xmlIn = xmlFactory.getXMLInputFactory();
        xmlIn.setProperty(XMLInputFactory.SUPPORT_DTD, Boolean.TRUE);
```
may modify a system default XMLInputFactory, and might affect to SuperTMXMerge.
It is because `getXMLInputFactory()` returns systems default parser.

## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?

-  Team project failed to merge changes with TMXMerge
  * https://sourceforge.net/p/omegat/bugs/1170/

## What does this PR change?

- revert ProjectFileStorage.java as same as OmegaT 6.0
- remove JaxbXmlMapper.java utility
- Update SRX.java, FileMaster.java and PreferencesXML.java


build.gradle
```diff
@@ -153,6 +153,9 @@ dependencies {
         // macOS integration
         implementation 'org.madlonkay:desktopsupport:0.6.0'
 
+        implementation 'javax.xml.bind:jaxb-api:2.3.1'
+        runtimeOnly 'com.sun.xml.bind:jaxb-impl:2.3.4'
```
miurahr, line 157: Is it really necessary because jackson introduce woodstox for implementation?

miurahr:
A test failed as same reason as master failed.

miurahr:
this wait #586 merge.

build.gradle
miurahr: In my understanding,   JAXB implementation is pluggable and jackson uses woodstox. so we don't need to add 'com.sun.xml:bind:jaxb-impl'.

build.gradle
```diff
@@ -154,6 +154,10 @@ dependencies {
         // macOS integration
         implementation 'org.madlonkay:desktopsupport:0.6.0'
 
+        // jaxb
```
miurahr, line 157:
```suggestion
        // StaX library
```

PR #589, opened 2023-05-15
[6.0]  [BUGS#1168] feat:error message for loop link

miurahr:
- Improve behavior on BUGS#1168 "OmegaT cannot load a project and error dialog with FileSystemLoopException"

## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?

- OmegaT cannot load a project and error dialog with FileSystemLoopException
  * Bugs: https://sourceforge.net/p/omegat/bugs/1168

## What does this PR change?

-
-
-

### DIFF
--- a/src/org/omegat/Bundle.properties
+++ b/src/org/omegat/Bundle.properties
@@ -506,8 +506,9 @@ TF_TM_LOAD_ERROR=Failed to load translation memory {0} for the project.
 TF_LOAD_ERROR=Failed to load the specified project.
 TF_LOAD_ERROR_FILE_ACCESS=Failed to load the project due to a file access issue. Make sure you have \
   the correct permissions or that you can access the files if they are stored in the cloud.
-TF_LOAD_ERROR_SOURCE_LOOP_EXCEPTION=Failed to load the project because the specified source folder contains a looped \
-  symbolic or hard link. Make sure the links to and from the source folder point to the correct folder.
+TF_LOAD_ERROR_SOURCE_LOOP_EXCEPTION=Failed to load the project because the specified source folder contains a \
+  symbolic or hard link that points back to itself or one of its subfolders.\n
+  Make sure all links are outside the source folder and subfolders.
 
 TF_COMPILE_ERROR=Failed to create the translated files.
 
--- a/src/org/omegat/Bundle.properties
+++ b/src/org/omegat/Bundle.properties
@@ -502,8 +502,12 @@ Reload the project now?
 
 TF_SOURCE_LOAD_ERROR=Skipped source file {0} because it could not be loaded.
 
-TF_TM_LOAD_ERROR=Failed to load translation memory {0} for project!
-TF_LOAD_ERROR=Failed to load specified project!
+TF_TM_LOAD_ERROR=Failed to load translation memory {0} for the project.
+TF_LOAD_ERROR=Failed to load the specified project.
+TF_LOAD_ERROR_FILE_ACCESS=Failed to load the project by a file access problem. Please check you have \
+  permissions, or you can access files if there are files in cloud storage.
+TF_LOAD_ERROR_SOURCE_LOOP_EXCEPTION=Failed to load the project because the specified source folder has a looped \
+  symbolic/hard link. Please check your source folder has a link to another folder and vise versa from linked folder.
 
 TF_COMPILE_ERROR=Failed to create the translated files!
 
--- a/src/org/omegat/Bundle.properties
+++ b/src/org/omegat/Bundle.properties
@@ -504,18 +504,18 @@ TF_SOURCE_LOAD_ERROR=Skipped source file {0} because it could not be loaded.
 
 TF_TM_LOAD_ERROR=Failed to load translation memory {0} for the project.
 TF_LOAD_ERROR=Failed to load the specified project.
-TF_LOAD_ERROR_FILE_ACCESS=Failed to load the project by a file access problem. Please check you have \
-  permissions, or you can access files if there are files in cloud storage.
-TF_LOAD_ERROR_SOURCE_LOOP_EXCEPTION=Failed to load the project because the specified source folder has a looped \
-  symbolic/hard link. Please check your source folder has a link to another folder and vise versa from linked folder.
+TF_LOAD_ERROR_FILE_ACCESS=Failed to load the project due to a file access issue. Make sure you have \
+  the correct permissions or that you can access the files if they are stored in the cloud.
+TF_LOAD_ERROR_SOURCE_LOOP_EXCEPTION=Failed to load the project because the specified source folder contains a looped \
+  symbolic or hard link. Make sure the links to and from the source folder point to the correct folder.
 
-TF_COMPILE_ERROR=Failed to create the translated files!
+TF_COMPILE_ERROR=Failed to create the translated files.
 
 TF_MESSAGE_COMPILE=Unable to create the translated files due to missing or damaged tags.\n\
 Use Tools > Check Issues... to correct the tag errors and try again.
 
-TF_COMMIT_ERROR=Failed to commit source files!
-TF_COMMIT_TARGET_ERROR=Failed to commit target files!
+TF_COMMIT_ERROR=Failed to commit source files.
+TF_COMMIT_TARGET_ERROR=Failed to commit target files.
 TF_COMMIT_START=Committing source files
 TF_COMMIT_DONE=Source files committed
 TF_COMMIT_TARGET_START=Committing target files
--- a/src/org/omegat/core/data/RealProject.java
+++ b/src/org/omegat/core/data/RealProject.java
@@ -39,10 +39,12 @@ package org.omegat.core.data;
 import java.io.File;
 import java.io.IOException;
 import java.io.RandomAccessFile;
+import java.io.UncheckedIOException;
 import java.io.Writer;
 import java.nio.channels.FileChannel;
 import java.nio.channels.FileLock;
 import java.nio.charset.StandardCharsets;
+import java.nio.file.FileSystemLoopException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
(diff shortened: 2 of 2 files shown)


Kazephil (commented):
I'll try to find time to test the code on a project with deliberately looped links in a day or two. In the meantime, I rephrased the wording of the error messages a little bit.

Kazephil:
I made a few tests using projects with:
1. looped links inside the `source` folder but not elsewhere, and
2. projects with looped links inside the `target` or `tm` folders (or both), but not in the `source` folder.

## Results

### Current `master`
- In case no. 1,  OmegaT gives a "Failed to load project"
- In case no. 2, if there are files in the folders containing a looped link, OmegaT keeps trying to load until it crashes. It loads the project successfully if the folders are empty.

### Hiroshi's fix (this PR and PR #577)
- In case no. 1, OmegaT shows the new more specific looped link error message (and helpfully shows the path containing that link), and stops loading the project.
- In case no. 1, OmegaT successfully loads the project, even if there are files in the folders containing looped links.

This fix looks good to me (with the caveat that my tests were not exhaustive).

miurahr:
> ### Hiroshi's fix (this PR and PR #577)
> * In case no. 1, OmegaT shows the new more specific looped link error message (and helpfully shows the path containing that link), and stops loading the project.
> * In case no. 1, OmegaT successfully loads the project, even if there are files in the folders containing looped links.

Thank you for the test.  PR #577 can load project but the fix requires Java 11 to build OmegaT.
It is because a fix for 6.0 is just for message.

> projects with looped links inside the target or tm folders (or both), but not in the source folder.

Here is a open space for us to improve. I want to work it on 6.1 development cycle.

Kazephil:
> Thank you for the test. PR #577 can load project but the fix requires Java 11 to build OmegaT. It is because a fix for 6.0 is just for message.

I guess that we don't want to add functionality in 6.0 that we can't also add in 5.8, then?
Shouldn't the new error messages go in 5.8 too, though?

> > projects with looped links inside the target or tm folders (or both), but not in the source folder.
> 
> Here is a open space for us to improve. I want to work it on 6.1 development cycle.

Sounds good to me.

miurahr:
> > Thank you for the test. PR #577 can load project but the fix requires Java 11 to build OmegaT. It is because a fix for 6.0 is just for message.
> 
> I guess that we don't want to add functionality in 6.0 that we can't also add in 5.8, then? Shouldn't the new error messages go in 5.8 too, though?

Exactly yes. @brandelune how is your opinion?

brandelune:
> > > Thank you for the test. PR #577 can load project but the fix requires Java 11 to build OmegaT. It is because a fix for 6.0 is just for message.
> > 
> > 
> > I guess that we don't want to add functionality in 6.0 that we can't also add in 5.8, then? Shouldn't the new error messages go in 5.8 too, though?
> 
> Exactly yes. @brandelune how is your opinion?

My opinion is that it would be nice to have a fix that does not require Java 11.

miurahr:
> the new error messages

The PR here is  proposal of new feature to add a new error message and do not solve the issue that is a loop link cause failure of project open.

> My opinion is that it would be nice to have a fix that does not require Java 11.

Does your opinion menas the PR #589 should not merge into 5.8/6.0?

brandelune:
> > My opinion is that it would be nice to have a fix that does not require Java 11.
> 
> Does your opinion menas the PR #589 should not merge into 5.8/6.0?

I only have a general opinion on the issue:
- bug fixes require people to upgrade their version
- 5.8 is stuck with Java 8
- we can have bug-fix releases for 5.8 if they can be implemented in Java 8
- we can't support 5.8 for ever.

The logical conclusion is that we need to set a date after which 5.8 will not have bug-fixes anymore, and until that we must set ourselves to provide bug fixes that use Java 8.

If a bug cannot be fixed with Java 8 only code, we should find an elegant way to alert the user about that and suggest workarounds or an upgrade.

Since we have decided to keep supporting Java 8 only machines, that's the minimum we can do.

That also means that potentially, some bugs will be fixed in a way in 5.8 and in a different way in 6.n. And the 5.8 fix does not have to be a real fix, but can be just an alert message to propose a workaround.

I think this opinion is reasonable. And if we are to set a date, I'd say 1 year from the 5.8 release date is also reasonable.

Kazephil:
> That also means that potentially, some bugs will be fixed in a way in 5.8 and in a different way in 6.n. And the 5.8 fix does not have to be a real fix, but can be just an alert message to propose a workaround.

I mostly agree, although if the main idea is to provide functional equivalence between 5.8 and 6.0, I'm not entirely certain that fixes for 6.1 and higher need to be ported back to 5.8 unless they are critical security fixes or bugs that cause existing functionality (either available up to 5.7.1, or explicitly implemented for 5.8/6.0) to fail.

Mind you, I also suspect that a large proportion of whatever bugs come up during 6.1 and higher development will be specific to Java 11 or higher and/or affect new funtionality or enhancements that were never planned or intended for 5.8/6.0. I don't think fixes for those need to be considered for 5.8 during the period of continued support for Java 8.

In this specific case, if I understood Hiroshi's explanations correctly, our choices seem to be:
- Only implement the improved error message in 5.8/6.0, and implement the actual fix that enables the project to load in 6.1.
- Implement the fix and improved error messages in 6.0, and expand the error message in 5.8 to add information about a fix being available if the user upgrades to Java 11 and OmegaT 6.0. (My understanding is that Java 8 does not include the functions that make the fix possible in Java 11 and higher.)

I would be fine with either. I don't expect looped symbolic or hard links are a very common problem. (I'm strugging to picture an actual use case for looping a link to a folder within itself...).

brandelune:
> In this specific case, if I understood Hiroshi's explanations correctly, our choices seem to be:
> 
> * Only implement the improved error message in 5.8/6.0, and implement the actual fix that enables the project to load in 6.1.
> * Implement the fix and improved error messages in 6.0, and expand the error message in 5.8 to add information about a fix being available if the user upgrades to Java 11 and OmegaT 6.0. (My understanding is that Java 8 does not include the functions that make the fix possible in Java 11 and higher.)

If there is a way to fix the issue with Java 8 code, we should do that: it does not matter if the fix is different in 5.8 and 6.0.

If there is no way to fix this issue with Java 8 code then we should provide an expanded error message in 5.8.

> I would be fine with either. I don't expect looped symbolic or hard links are a very common problem. (I'm strugging to picture an actual use case for looping a link to a folder within itself...).

Same here 😅

miurahr:
> > * Implement the fix and improved error messages in 6.0, and expand the error message in 5.8 to add information about a fix being available if the user upgrades to Java 11 and OmegaT 6.0. (My understanding is that Java 8 does not include the functions that make the fix possible in Java 11 and higher.)
> 
> If there is a way to fix the issue with Java 8 code, we should do that: it does not matter if the fix is different in 5.8 and 6.0.

The FIX proposed requires **BUILD with Java 11**.
OmegaT 6.0 is built with Java 8 and bundle with Java 11.

brandelune:
> The FIX proposed requires **BUILD with Java 11**. OmegaT 6.0 is built with Java 8 and bundle with Java 11.

That was not my understanding.

My understanding was that 6.0 is built with Java 11 and 5.8 is built with Java 8.

If 6.0 is built with Java 8, I don't see why we need 5.8. We can just ask people to use the generic OmegaT 6.0 version to run on their Java 8 machines.

Releasing 5.8 for Java 8 machines implies that the generic 6.0 does *not* work on Java 8.

And that's the discussion we had on the development list: OmegaT 6.0 is the first to require Java 11 to run.

Kazephil:
> > The FIX proposed requires **BUILD with Java 11**. OmegaT 6.0 is built with Java 8 and bundle with Java 11.
> 
> That was not my understanding.
> 
> My understanding was that 6.0 is built with Java 11 and 5.8 is built with Java 8.

Same here. I thought the whole point of having both 5.8 and 6.0 was that 6.0 marked the full transition to building and running on Java 11, with 5.8 offered as a final Java 8 version for people who need to remain on Java 8 for whatever reason.

I don't understand why you'd build with Java 8 and bundle Java 11.

miurahr:
> > The FIX proposed requires **BUILD with Java 11**. OmegaT 6.0 is built with Java 8 and bundle with Java 11.
> 
> That was not my understanding.
> 
> My understanding was that 6.0 is built with Java 11 and 5.8 is built with Java 8.
> 
> If 6.0 is built with Java 8, I don't see why we need 5.8. We can just ask people to use the generic OmegaT 6.0 version to run on their Java 8 machines.
> 
> Releasing 5.8 for Java 8 machines implies that the generic 6.0 does _not_ work on Java 8.
> 
> And that's the discussion we had on the development list: OmegaT 6.0 is the first to require Java 11 to run.

You define 6.0 is same code with 5.8 with jre 6.0.

brandelune:
> > > The FIX proposed requires **BUILD with Java 11**. OmegaT 6.0 is built with Java 8 and bundle with Java 11.
> > 
> > 
> > That was not my understanding.
> > My understanding was that 6.0 is built with Java 11 and 5.8 is built with Java 8.
> > If 6.0 is built with Java 8, I don't see why we need 5.8. We can just ask people to use the generic OmegaT 6.0 version to run on their Java 8 machines.
> > Releasing 5.8 for Java 8 machines implies that the generic 6.0 does _not_ work on Java 8.
> > And that's the discussion we had on the development list: OmegaT 6.0 is the first to require Java 11 to run.
> 
> You define 6.0 is same code with 5.8 with jre 6.0.

I don't understand what you just wrote.

But I did not define anything.

On the development list, we discussed that 6.0 would be the first version of OmegaT that required Java 11. And we also set that 6.0 and 5.8 would be functionally equivalent.

To me, that means that 6.0 and 5.8 have the same code base but 6.0 is built with Java 11 and 5.8 with Java 8. That was, OmegaT 6.0 cannot be run on Java 8, even its generic version.

What we are discussing here is a bug fix. It is not an added function.

If there is a way to fix this bug with Java 11 on OmegaT 6.0, we should do it. And 5.8 will not have the fix, but will have an expanded error message. That's totally OK.

miurahr:
> On the development list, we discussed that 6.0 would be the first version of OmegaT that required Java 11. And we also set that 6.0 and 5.8 would be functionally equivalent.
> 
> To me, that means that 6.0 and 5.8 have the same code base but 6.0 is built with Java 11 and 5.8 with Java 8. That was, OmegaT 6.0 cannot be run on Java 8, even its generic version.

Thank you for clarifying a release policy.
I've put a pull-request #599 to update target java version to 11 for `releases/6.0` branch.

brandelune:
> > On the development list, we discussed that 6.0 would be the first version of OmegaT that required Java 11. And we also set that 6.0 and 5.8 would be functionally equivalent.
> > To me, that means that 6.0 and 5.8 have the same code base but 6.0 is built with Java 11 and 5.8 with Java 8. That was, OmegaT 6.0 cannot be run on Java 8, even its generic version.
> 
> Thank you for clarifying a release policy. I've put a pull-request #599 to update target java version to 11 for `releases/6.0` branch.

I'm sorry it was not clear before today. Thank you for your understanding.

miurahr:
> If there is no way to fix this issue with Java 8 code then we should provide an expanded error message in 5.8.

I'd like to make a new pull-request against `releaes/5.8` branch that change is as same as here #589, to extend error message.

Kazephil:
> I'd like to make a new pull-request against `releaes/5.8` branch that change is as same as here #589, to extend error message.

Please do. Then I'll work on the actual extended error message.

brandelune:
Let's agree to not wait for the l10n on this message for the release.

miurahr:
I'd like to merge here.

miurahr:
I discovered a way to catch the link loop and continue to load  the project in #577 that can be ported to 5.8/6.0 which is compatible with Java 8.
It is still in review process, so I want to merge the improvement of error message (here) at 6.0.0.
Then when we believe a new way is good enough in 6.1 development, it will be delivered  as 6.0.1.

brandelune:
Regarding the Bundle, only the 2 keys should be modified:

TF_LOAD_ERROR_FILE_ACCESS
TF_LOAD_ERROR_SOURCE_LOOP_EXCEPTION

Modifying the other for style purpose breaks the already produced translations.

This was reported by Kos who is working on the L10N right now.

PR #597, opened 2023-05-17
FR full localization for 6.1

brandelune:


miurahr (approved):
I can not read French but target files changed looks good.

PR #600, opened 2023-05-20
[5.8][BUGS#1168] feat: error message for loop link

miurahr:
Backport error message enhance from 6.0/6.1

## Pull request type

Please mark github LABEL of the type of change your PR introduces:

- Bug fix -> [bug]

## Which ticket is resolved?

- OmegaT cannot load a project and error dialog with FileSystemLoopException
   * https://sourceforge.net/p/omegat/bugs/1168

## What does this PR change?

- Improve behavior on BUGS#1168 by @miurahr "OmegaT cannot load a project and error dialog with FileSystemLoopException"
- style: Reword new error messages by @kazephil: I rephrased the new error messages a little, and replaced a few unnecessary exclamation marks I happened to notice at the same time. (I thought I'd cleaned them all up a little while back, but I guess I missed a few.)
- style: Update file loop error message by @kazephil: I changed the message again to make it easier to understand for people less familiar with computers.

## Other information

The change here is squashed  changes propoesd in #589

PR #601, opened 2023-05-20
feat: display(Error|Warning)RB show multi-line message

miurahr:
Extend MainWindow#displayErrorRB and MainWindow#displayWarningRB to show multiple lines of message according to "\n" in a message.

## Pull request type

- 
## Other information

An enhancement is suggested in #600


src/org/omegat/gui/main/MainWindow.java
```diff
@@ -462,7 +463,12 @@ public void displayWarningRB(final String warningKey, final String supercedesKey
                 }
             }
 
-            lastDialogText = new JLabel(msg);
+            lastDialogText = new JPanel();
```
miurahr, line 466: We can add border config with `BorderFactory.emptyBorder`

PR #602, opened 2023-05-21
[BUGS#1173] fix(team) ask password/passphrase when ssh key

miurahr:
## Pull request type

- bug fix

## Which ticket is resolved?

- Bump JGit @6.6.0

- team: apache HttpClient for git connection 
    * https://sourceforge.net/p/omegat/feature-requests/1706/

-  Infinite authentication when incorrect password entered on git team project that URL scheme is https
    * https://sourceforge.net/p/omegat/bugs/1190/

- Passphrase for SSH key is stored with key= null://null:-1
   * https://sourceforge.net/p/omegat/bugs/1174/
 
- OmegaT ask passphrase to connect git+ssh server but dialog label is user/pass
   * https://sourceforge.net/p/omegat/bugs/1173/

  

## What does this PR change?

- Fix: BUGS#1174 when uri is key path, use path
- Consolidate dialogs to two types: user/pass or pass(word|phrase)
- Catch additional messages from apache sshd and put it on dialog.


miurahr:
Need a regression test for bugs.

Kazephil:
Do you (or anyone else) have an SSH-based Git-based team project I could test?

I tried setting one up, but since my system is already set up with my keys, OmegaT simply connects without asking for authentication...
(Hmm... now that I think about it, I'm pretty sure I set up my keys without a passphrase in the first place. Connecting to GitHub requires my _personal access token_ rather than my password.)

However, I found a bigger problem. Using the build from this branch, I also tried downloading the team project through `https`, and when I did that , OmegaT showed me the dialog asking for my user name and password. 

After I did that:
- A second dialog asking only for the password showed up, two or three times, before I finally got back to the dialog where I can specify the folder for the project and choose the default branch or a specific branch.
- When OmegaT started downloading the project, it immediately asked for only the password again, failed to accept it, and asked again two or three more times before giving up and showing me an authentication error message.

With the latest `master` (commit 82ffdb9984349d3ce1aa2a6b85d5574dfeb3c6ed), the project downloads fine with the same personal access token, but for some reason the `<mapping>` gets deleted from the `omegat.project` file and the project doesn't actually load until they are fixed manually. However, OmegaT deletes the contents of `<mapping>` again when the project is closed.

Earlier commits from master worked as expected.

miurahr:
> With the latest `master` (commit [82ffdb9](https://github.com/omegat-org/omegat/commit/82ffdb9984349d3ce1aa2a6b85d5574dfeb3c6ed)), the project downloads fine with the same personal access token, but for some reason the `<mapping>` gets deleted from the `omegat.project` file and the project doesn't actually load until they are fixed manually. However, OmegaT deletes the contents of `<mapping>` again when the project is closed.

Thank you for feedback.
 The commit may have sub-effect not to output `      <mapping local="" repository=""/>`.

miurahr:
> Do you (or anyone else) have an SSH-based Git-based team project I could test?

You can test with private repository

https://github.com/miurahr/omegat-test

> I tried setting one up, but since my system is already set up with my keys, OmegaT simply connects without asking for authentication... (Hmm... now that I think about it, I'm pretty sure I set up my keys without a passphrase in the first place. Connecting to GitHub requires my _personal access token_ rather than my password.)

Github now has fine graind token 
https://docs.github.com/en/rest/overview/permissions-required-for-fine-grained-personal-access-tokens?apiVersion=2022-11-28

That can limit your token to permit access specified repository.

> However, I found a bigger problem. Using the build from this branch, I also tried downloading the team project through `https`, and when I did that , OmegaT showed me the dialog asking for my user name and password.
> 
> After I did that:
> 
> * A second dialog asking only for the password showed up, two or three times, before I finally got back to the dialog where I can specify the folder for the project and choose the default branch or a specific branch.
> * When OmegaT started downloading the project, it immediately asked for only the password again, failed to accept it, and asked again two or three more times before giving up and showing me an authentication error message.

After removing credentials from Preference > team > credentials I opened test project

1. When starting projec, OmegaT ask master password.
2. When specify  https repository URL onto Download Team dialog , then ask User and Password dialog.
3. After enter user and personal token, go back to download team project dialog.
4. start download
5. Successfully download.

It looks ok with the commit https://github.com/omegat-org/omegat/pull/602/commits/a98b322af11d9acd8a45dfdb9a3c4036e8e35362

Kazephil:
Thank you.

I'll test it tomorrow, or maybe late this evening if I have time.

miurahr:
> Thank you.
> 
> I'll test it tomorrow, or maybe late this evening if I have time.

Could you give me a feedback?

Kazephil:
> Could you give me a feedback?

Sorry about the delay. I had some computer problems and a really busy work week.

I just tried with commit 3eb902527a6f685aa2b0c1bb4af2c75c5d9335b3, and it's looking pretty good. I tried with both the test project you set up and with one of mine, and get the same results with both. I used a fresh config directory with no existing repository credentials.

If I download the project from the `https` link, OmegaT asks me for my user name and password.

For a project on GitHub:
- entering my actual password instead of entering my GitHub token, shows me the same dialog two more times, and then I get the message that my credentials are not valid.
- entering my GitHub token immediately loads the project and everything works normally.

If I download from the `ssh` link,  the project downloads without asking me for any credentials, and everything works normally.

Small changes that might be nice for the user:
- When asking for the user name and password for an `https` connection, it would be nice for the second and third attempts to include an error message informing the user his or her credentials are not valid. Something like: "Your user name or password (token) is incorrect."
- It might be nice to have the password field say "Password (token)" instead, as a reminder that some sites require an access token instead of just a passworld.

Kazephil:
> Thank you for feedback. It is very informative and I'd like to update messages.
> 
> We can update bundles that is
> 
> ```
> TEAM_PASS_WRONG=Wrong password for {0}
> ```
> 
> to be
> 
> ```
> TEAM_PASS_WRONG=Wrong password(token) for {0}
> ```
For the message itself, "incorrect" is better than "wrong": `Incorrect password (token) for {0}` 

-----
However, in my tests using the HTTPS address, that message did not get displayed at all.

I tried again (after deleting the credentials). If I type the wrong password, it shows me this dialog:
image

Even if I enter the correct password in that dialog, it shows it a second time, and then it lets me choose the directory for the project.

When I click OK after choosing the directory, it then shows me the above dialog three more times, and then fails to load the project:
image

The "incorrect password" message never gets shown.

More concerning: if I type the wrong password the first time, even entering the correct token after that doesn't work. It just keeps showing the `Enter the password…` dialog until authentication fails, as explained above.

If delete all my repository credentials and enter the password correctly the first time, then everything works fine.

miurahr:
Thank you for the test. I've reproduced the situation.

miurahr:
A basic logic for http protocol of git connection is not changed, and the http authentication logic in jgit library has  not been changed since 2017.
I think a similar behavior may be observed in previous versions of omegat.

A root cause of multiple asking password is;

http authentication of JGit library has a logic

TransportHttp#sendRequest Line 1758

```java
		CredentialsProvider credentialsProvider = getCredentialsProvider();
		if (credentialsProvider == null) {
			throw new TransportException(uri,	JGitText.get().noCredentialsProvider);
		}
		if (authAttempts > 1) {
			credentialsProvider.reset(currentUri);
		}
		if (3 < authAttempts || !authMethod.authorize(currentUri, credentialsProvider)) {
			throw new TransportException(uri,	JGitText.get().notAuthorized);
		}
		authAttempts++;
```

JGit ask `reset` for `GITCredentialProvider` when user enter incorrect password first and then re-enter password, `authAttempt` is larger than 1.
`authAttempt` is not reset even when authentication is successed.

OmegaT always removes the saved credentials when `reset` is requested, so it shows the dialog again and again.

miurahr:
When watching a behavior in  6.0 branch, it does not act multiple ask of password.
I've enter wrong password first, and then enter correct one.

JGit never call `reset` in the senario. 

```
		if (authAttempts > 1) {
			credentialsProvider.reset(currentUri);
		}
```

It is different than master branch.

When watching `status` in `TransportHttp` in JGit;
```
final int status = HttpSupport.response(conn);
```
It becomes 401(Unauthorized, BASIC) then 200(OK, BASIC) and 401(Unautorized, NONE), 200(OK, BASIC), and 401/200 each come. It seems similar between  6.0 branch and master branch.

miurahr:
The issue is observed in master branch head.
The PR here is not a cause of the problem.

miurahr:
I've create a bug ticket for it.
https://sourceforge.net/p/omegat/bugs/1190/

miurahr:
I believe the problem BUGS#1190 is gone. The bug is in `PassphaseDialog`  I introduced.
I designed `PassphraseDialog` for usage of ssh key passphrase and I don't assume it for usage of http.

I think there are still  a space to improve  there.

miurahr:
> For the message itself, "incorrect" is better than "wrong": `Incorrect password (token) for {0}`

Fix the message.

miurahr:
It is ready to merge now.

miurahr:
I'd like to merge and ask to community for further feedbacks.

PR #604, opened 2023-05-21
feat: make repository connectors as plugin

miurahr:
## Pull request type

- Other (describe below)
refactor

## Which ticket is resolved?

- #1715 Make repository connectors as plugin
  * https://sourceforge.net/p/omegat/feature-requests/1715/

## What does this PR change?

- Add plugin type "repository"
- Introduce RemoteRepositoryFactory.addRepositoryConnector func
- make file, git, svn and http connectors as plugin
- Set description attribute for new plugin type

## Other information

DEV-ML https://sourceforge.net/p/omegat/mailman/message/37803705/

### DIFF
--- a/Plugins.properties
+++ b/Plugins.properties
@@ -7,6 +7,10 @@ plugin=org.omegat.filters3.xml.xliff.XLIFFFilter \
     org.omegat.gui.theme.DefaultFlatTheme \
     org.omegat.gui.theme.DefaultClassicTheme \
     org.omegat.util.gui.HtmlMediaSupport \
+    org.omegat.core.team2.impl.FileRepository \
+    org.omegat.core.team2.impl.HTTPRemoteRepository \
+    org.omegat.core.team2.impl.GITRemoteRepository2 \
+    org.omegat.core.team2.impl.SVNRemoteRepository2 \
     org.omegat.core.dictionaries.LingvoDSL \
     org.omegat.core.dictionaries.StarDict
 #   org.omegat.gui.accesstool.QuickAccess
@@ -101,3 +105,4 @@ plugin.desc.filters3=Bundled filters for simple (linear, unilingual) XML files i
 plugin.desc.filters4=Bundled filters for complex XML files includes MS Office (OpenXML, with complex markups) and XLIFF (bilingual)
 plugin.desc.theme=Bundled default theme plugins
 plugin.desc.externalfinder=Local external search plugin
+plugin.desc.repository=Repository connector plugins
--- a/src/org/omegat/core/team2/RemoteRepositoryFactory.java
+++ b/src/org/omegat/core/team2/RemoteRepositoryFactory.java
@@ -4,6 +4,7 @@
           glossaries, and translation leveraging into updated projects.
 
  Copyright (C) 2014 Alex Buloichik
+               2023 Hiroshi Miura
                Home page: https://www.omegat.org/
                Support center: https://omegat.org/support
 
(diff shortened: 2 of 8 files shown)


t-cordonnier (approved):
Very good idea, did a quick test it seems to work correctly. Approved.

miurahr:
@t-cordonnier could you mind to make git and svn connectors as module based on the changes here?
I think it worth try because you might review it from a view point of plugin/module developer.
There are several point we need to discussed.

1. Is a new plugin category name is ok? 
2. Is it really good about  API/SPI for repository connectors?

If anyone try to add more connector such as mercurial  pluigin,  API design can be evaluated.

t-cordonnier:
> could you mind to make git and svn connectors as module based on the changes here?

Yes, absolutely, these should be modules. I will not have the time right now but I can consider it for the future.

>   Is a new plugin category name is ok?

It is already used from a long time and I have no better idea for the moment.
That said, I would not strictly link to version control because
- strictly speaking OmegaT does only synchronization, no version control: you can select a branch but you are always connected to the head of a branch, without possiblity to switch back to another commit nor create a branch directly from OmegaT
- we already have connectors HTTP and File which do not use version control. I imagine we could extend them for pure synchronization, like rsync does... without branch and always synchronized with last version of a file.
Even in this case no problem for me to use terms like repository or depot. These terms are not strictly linked to VCS: linux distributions also speak about a depot, but it is not versioned.

>  Is it really good about API/SPI for repository connectors?
> If anyone try to add more connector such as mercurial pluigin, API design can be evaluated.

We already have two very different implementations (git and svn) and even some non-vcs implementations, so we can already confirm that it has a good level of factorization.

src/org/omegat/filters2/master/PluginUtils.java
```diff
@@ -96,6 +96,8 @@ public enum PluginType {
         DICTIONARY("dictionary"),
         /** theme, that register Swing Look-and-Feel with OmegaT properties into UIManager. */
         THEME("theme"),
+        /** team repository version control system connector plugins. */
+        TEAMREPOSITORY("teamRepository"),
```
miurahr, line 100: It is better to be  simple "repository", with comment "remote repository connector plugins."

Plugins.properties
```diff
@@ -112,3 +116,4 @@ plugin.desc.filters3=Bundled filters for simple (linear, unilingual) XML files i
 plugin.desc.filters4=Bundled filters for complex XML files includes MS Office (OpenXML, with complex markups) and XLIFF (bilingual)
 plugin.desc.theme=Bundled default theme plugins
 plugin.desc.externalfinder=Local external search plugin
+plugin.desc.teamRepository=Team repository connector plugins
```
miurahr, line 119:
```suggestion
plugin.desc.repository=Repository connector plugins
```

src/org/omegat/core/team2/RemoteRepositoryFactory.java
```diff
@@ -25,10 +26,12 @@
 
 package org.omegat.core.team2;
 
-import org.omegat.core.team2.impl.FileRepository;
+import java.util.HashMap;
+import java.util.Map;
+
 import org.omegat.core.team2.impl.GITRemoteRepository2;
-import org.omegat.core.team2.impl.HTTPRemoteRepository;
 import org.omegat.core.team2.impl.SVNRemoteRepository2;
```
miurahr, line 33: We cannot remove dependency for Git and SVN plugins from OmegaT core because of detection of repository type in Factory class.

```
            if (GITRemoteRepository2.isGitRepository(url)) {
                return "git";
            } else if (SVNRemoteRepository2.isSVNRepository(url)) {
                return "svn";
            }

```

It prevent from making SVN and GIT repository plugins as module.

PR #608, opened 2023-05-22
chore: Realign with l10n and fix typo

Kazephil:
- Fix GLobal --> Global typo spotted during l10n.
- Temporarily revert stylistic improvements to maintain alignment with current l10n translations.

(The fixes will be reapplied in a subsequent minor fix release.)


brandelune (commented):
@kosivantsov will that work for you?

kosivantsov:
> @kosivantsov will that work for you?

If Bundle.properties are the same (or mostly the same, except http/https in URLs) in releases/5.8 and releases/6.0, then yes

miurahr:
I'd like to squash and merge.

kosivantsov:
> I'd like to squash and merge. @Kazephil can I do it?

There are still problems with this PR. It introduces 4 new segments which are not translated if compared to 5.8
5.8/6.0 were supposed to be identical as far a L10n goes, weren't they?
Besides, there's a problem in line 511 in `Bundle.properties` when the file is rendered in OmegaT: the `\n` in the end of the line 510 should be followed by another `\`, otherwise OmegaT sees `Make` as the key, and `sure all links are outside the source folder and subfolders.` as the value.

Kazephil:
> > I'd like to squash and merge. @Kazephil can I do it?

> There are still problems with this PR. It introduces 4 new segments which are not translated if compared to 5.8
5.8/6.0 were supposed to be identical as far a L10n goes, weren't they?

The new segments come from the two new error handling keys added after l10n was mostly fixed, and there's a pending PR (#600) to bring them into 5.8.

The trickier issue is the mismatch for lines 1089 and 1090. That change of key involves a change to fix a bug in the code as well (from PR #526), and it looks like that fix and the corresponding changes in keys never made it to `releases/5.8 `.

I suspect, but can't confirm, that the fix was merged into master after the `releases/5.8` branch was created but before the `releases/6.0` branch was, with the result that the code and new keys only went into 6.0.


> Besides, there's a problem in line 511 in `Bundle.properties` when the file is rendered in OmegaT: the `\n` in the end of the line 510 should be followed by another `\`, otherwise OmegaT sees `Make` as the key, and `sure all links are outside the source folder and subfolders.` as the value. 

I just commited a fix for that to this PR. Thanks for pointing it out.

I realise the problem it causes for l10n, but I think the best course of action is to:
- Merge this PR into `releases/6.0`
- Merge PR #600 and the commits from PR #526 into `releases/5.8`
- Merge the http --> https URL changes into `releases/5.8`

That should bring 5.8 and the current 6.0 in line. We could then freeze the `releases/5.8`and `releases/6.0` branches until the actual release, and apply other minor fixes in `Bundle.properties` and other bug fixes to a subsequent set of 5.8.1/6.0.1 releases.

@brandelune, @miurahr, what do you think?

brandelune:
My opinion is that both #600 and #526 should be in 6.0 and 5.8.

#600 does not need to be localized, there is no rush. It would be better if #526 were but that's up to the translators. I'm totally OK with waiting.

What I especially don't want is translators who have brought 5.8/6.0 to near completeness feel cheated because we put a string before they could react.

Kazephil:
> #600 does not need to be localized, there is no rush. It would be better if #526 were but that's up to the translators. I'm totally OK with waiting.

#600 brings the new error keys and messages into 5.8, so it will have to be localized eventually.
I'm also totally OK with waiting until the translators get a chance to cover the last minute changes in this PR, #600, and #526.
 
> What I especially don't want is translators who have brought 5.8/6.0 to near completeness feel cheated because we put a string before they could react.

That's why I suggest we impose a freeze on changes to 5.8/6.0 once this set of PRs is merged.

Freeze the code, give the translators the time they need to translate/update get the extra/modified `Bundle.properties` segments,  and then release.

Any other fixes that should also go in the 5.8/6.0 series can wait until the 5.8.x/6.0.x (where x >= 1) bug fix releases.

@kosivantsov: Does that sound reasonable?

kosivantsov:
> @kosivantsov: Does that sound reasonable?

I want 5.8 and 6.0 to be frozen, so no more changes are introduced there. If there are new errors discovered after the code is frozen, they need to be addressed in new release (5.8.1)

I have written to our localizers at least 3 times already asking them to finalize their projects for the release, and then a few times committed their reported updates, but the release never happened. I'm OK with writing to them one more time asking for the FINAL update of the l10n data, but I would really appreciate if it is the LAST one before the release.

miurahr:
We freeze the code after merge here.

kosivantsov:
`git diff releases/5.8 releases/6.0 src/org/omegat/Bundle.properties` still produces a number of differences, and the only difference I'm expecting to see between these two versions is `http` vs `https` in the header.

kosivantsov:
Weren't 5.8 and 6.0 supposed to be identical? If it's not the case anymore, I'll make it possible to get 3 versions of the `Bundle.properties` (5.8, 6.0, master), but it sounds like an overkill

miurahr:
The diff come from 
https://github.com/omegat-org/omegat/commit/031aef52a86a0527795d3a9837641855374109b1
@briacp

kosivantsov:
I'm afraid I won't have time or brain power to solve it now, so I'll just wait till `Bundle.properties` in 5.8 and 6.0 is identical

brandelune:
> The diff come from
> [031aef5](https://github.com/omegat-org/omegat/commit/031aef52a86a0527795d3a9837641855374109b1)
> @briacp

Can't we have that in 5.8 ?

briacp:
We could backport it to 5.8, @miurahr do you want me to do it in `releases/5.8`?

miurahr:
I've pick and merge it.

PR #609, opened 2023-05-23
Fire the `COMPILE` project event before committing target files to repository

briacp:
## Which ticket is resolved?

- The COMPILE event is fired after the target commit
  * https://sourceforge.net/p/omegat/bugs/1176/

## What does this PR change?

- The `COMPILE` event is now fired before committing translated files to remote repository. This allows the resulting local files to be modified before sending them to the repository.
- In the unlikely case some scripts were expecting the repo commit to be finished when the event fired, it would make more sense to add a `POST_COMMIT` event.
- The behaviour is the same for non-team projects


src/org/omegat/core/data/RealProject.java
```diff
@@ -706,6 +706,12 @@ public void compileProjectAndCommit(String sourcePattern, boolean doPostProcessi
                 numberOfCompiled++;
             }
         }
+
+        // COMPILE event is fired before committing translated files to remote
+        // repository to be able to modify the resulting files before sending them to
+        // the repository (#1176)
```
miurahr, line 712:
```suggestion
        // the repository (BUGS#1176)
```

miurahr:
Good catch!

PR #618, opened 2023-05-28
chore(gradle): support PKCS#11 signing with HSM

miurahr:
## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?

- #1179 build/release: Windows binary signature with PKCS#11 HSM
  *  https://sourceforge.net/p/omegat/bugs/1179

## What does this PR change?

- Remove a way to sign with Docker because of supporting HSM
- Update local.properties keys for PKCS#11
- Improve unsigned/signed exe generation
   - make unsigned exe generation task not using ant task and use copy and delete task for moving artifact
   - signed exe generation task as Exec type, not sequence of procedure for keeping idempotency.
   - Fix error with GString conversion
 - Tested with my private key stored in HSM with ACR37U-A1 usb dongle
   - I use a proprietary pkcs11 access module

## Other information

When you have private key and certs in HSM, you can test new config wit `winNoJRESigned` task.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -772,6 +772,7 @@ ext.makeWinTask = { args ->
         }
         doFirst {
             delete "$destinationDir/jre"
+            delete installerExe
         }
         doLast {
             project.copy {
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -794,39 +795,49 @@ ext.makeWinTask = { args ->
...
-                        'osslsigncode or Docker not installed'])
+                       [exePresent('osslsigncode'), 'osslsigncode not installed'])
         }
-        def signedExe = "${distsDir}/${installerBasename}_Signed.exe"
+        def pkcs11module = project.hasProperty('pkcs11module')? project.property('pkcs11module') :
+                '/usr/lib/x86_64-linux-gnu/pkcs11/opensc-pkcs11.so'
+        def pkcs11time = project.hasProperty('winCodesignTimestampUrl')? project.property('winCodesignTimestampUrl') :
+                'http://time.certum.pl/'
+        def pkcs11cert = project.hasProperty('winCodesignCert')? project.property('winCodesignCert') : ''
+        def pkcs11pass = project.hasProperty('winCodesignPassword')? project.property('winCodesignPassword') : ''
```
miurahr, line 819: If `winCodesignPassword` not provided, we can use `-askpass` option when it does not run on headless environment, other than CI. we can check project property 'IsCIEnv' .

miurahr: It is better to avoid interaction with developer because of stability of gradle script.

miurahr:
This works for my signing key hardware. I'd like to ask @brandelune  whether it works also for key/cert provided by Comodo.

miurahr:
From Comodo support page;
https://help.comodosslstore.com/support/solutions/articles/22000280787-code-signing-certificate-delivery-methods

It seems we need to know what HSM token hardware is provided from docmodo to know a proper driver file.

> Code Signing Certificate Delivery Methods
> Modified on: Tue, May 16, 2023 at 12:48 PM
> 
> Code Signing certificates are now required to be installed on physical hardware tokens. For standard Organization Validation (OV) or Individual Validation (IV) Code Signing certificates, you can either purchase pre-configured certificate tokens, or you can order a certificate to be installed on an existing hardware device you already own.
> 
> When purchasing your code signing certificate, you can choose the Certificate Delivery Method that works best for you. Please make sure to select the correct certificate delivery method before you complete your purchase, as the method cannot be changed afterward.
> 
> Token + Shipping
> To order a pre-configured token from the Certificate Authority, select the Token + Shipping method. The cost of the hardware and the shipping fees will be included in the purchase price.
> 
> Most users should consider this option.
> 
> Install on Existing HSM (Advanced Users)
> If you already own a compatible Hardware Security Module (HSM) and would prefer to install the code signing certificate on your own device, select the Install on Existing HSM method. You will be required to provide an Attestation bundle for your HSM when you generate the certificate.
> 
> The following HSM brands are supported:
> 
> Yubikey 5 FIPS
> LUNA Network Attached HSM, version 7+

miurahr:
From comodo document, when we got a   certificate file from comodo before 13, May. we can convert it to PFX format.
https://help.comodosslstore.com/support/solutions/articles/22000265839-windows-convert-code-signing-to-pfx-csr-method-

> Windows: Convert Code Signing to PFX (CSR Method)
> Modified on: Tue, May 16, 2023 at 12:49 PM
> 
> This article includes instructions for collecting and converting your code signing certificate into a PFX file for signing. If you purchased your Code Signing certificate after May 14, 2023, you will not need to undergo any of these processes as the certificate will be delivered to you in its final form on physical hardware. This method will be deprecated on June 1, 2023.

Do we need to keep a method to use pfx file?  PR here remove an old method that  use PFX file.

brandelune:
> This works for my signing key hardware. I'd like to ask @brandelune whether it works also for key/cert provided by Comodo.

Sorry, I won't have the time to test anything serious before June 21.

PR #622, opened 2023-05-29
[6.1] refactor(gui/main): make new UI/Toolbar  as an in-source plugin

miurahr:
## Pull request type

- Other (describe below)

refactoring

## What does this PR change?

- Add extension hook in ApperancePreference
    - Define IMenuPreference interface
    - Add QuickAccessUI and MainMenuUI classes
    -  Preferences stores class name of class that implements IMenuPreference
- Add extension hook in MainWindow constructor
    - Allow replacing menu  
    - Allow adding toolbar
- Make newUI components as in-source plugin to allow seperate improvement without affect main code.
    - It cannot be as a module, because OmegaT uses `reflection` in `MenuHandler` that lead `Illegal access` error when accessed from module on Java 9 or later. 
- Add New/Open/Team quick access menu item in project combobox.
    - Project ComboBox has model data as URI.
    - When URI scheme is special "omegat" that is command uri.
 

## Other information

Because OmegaT menu handler use `Reflection` it could not be an external module. It will cause IllegalAccessError.


Plugins.properties
```diff
@@ -3,6 +3,7 @@ plugin=org.omegat.filters3.xml.xliff.XLIFFFilter \
     org.omegat.tokenizer.LuceneEnglishTokenizer \
     org.omegat.tokenizer.HunspellTokenizer \
     org.omegat.core.machinetranslators.BelazarTranslate \
+    org.omegat.gui.main.NewUI \
```
miurahr, line 6: By comment out this, we can hide a plugin component from release in 6.1

src/org/omegat/gui/main/MainWindow.java
```diff
@@ -124,28 +127,31 @@ public class MainWindow extends JFrame implements IMainWindow {
...
 
     /** Creates new form MainWindow */
+    @SuppressWarnings("unchecked")
     public MainWindow() throws IOException {
         MainWindowMenuHandler mainWindowMenuHandler = new MainWindowMenuHandler(this);
-        if (Preferences.isPreference(Preferences.APPLY_BURGER_SELECTOR_UI) && !Platform.isMacOSX()) {
-            menu = new MainWindowBurgerMenu(this, mainWindowMenuHandler);
-            menu.initComponents();
-            mainWindowAccessTools = MainWindowAccessTools.of(menu.mainMenu, mainWindowMenuHandler);
+        // Check are there MainWindowMenu extension is installed.
```
miurahr, line 136:
```suggestion
        // Check if Menu extension installed.
```

miurahr:
There is no change in UI and behavior, but add plugin API for the menu.

miurahr:
There is an expected discussion how extension points should be in MainWindow.

Plugins.properties
```diff
@@ -10,6 +10,7 @@ plugin=org.omegat.filters3.xml.xliff.XLIFFFilter \
     org.omegat.util.gui.HtmlMediaSupport \
     org.omegat.core.dictionaries.LingvoDSL \
     org.omegat.core.dictionaries.StarDict
+#   org.omegat.gui.main.NewUI
```
t-cordonnier, line 13: Like this you have it inactive, coorect?
What is your plan for the future? Do you plan to move it to an external plugin?

src/org/omegat/gui/main/MainWindow.java
```diff
@@ -106,8 +107,10 @@
  */
 @SuppressWarnings("serial")
 public class MainWindow extends JFrame implements IMainWindow {
-    public final BaseMainWindowMenu menu;
+    public static final String menuClassID = "OmegaTMainWindowMenu";
+    public static final String toolbarClassID = "OmegaTMainWindowToolbar";
 
+    public BaseMainWindowMenu menu;
```
t-cordonnier, line 113: Unless it becomes possible to change it on the fly without rebooting the application  I would prefer to keep this variable final (especially if you keep it public!). Yes I know that it requires some changes, see below

src/org/omegat/gui/main/MainWindow.java
```diff
@@ -124,28 +127,32 @@ public class MainWindow extends JFrame implements IMainWindow {
...
+        menuClass = (Class<? extends BaseMainWindowMenu>) UIManager.get(menuClassID);
+        if (menuClass != null) {
+            try {
+                menu = menuClass.getDeclaredConstructor(MainWindow.class, MainWindowMenuHandler.class)
+                                .newInstance(this, mainWindowMenuHandler);
+            } catch (InstantiationException | IllegalAccessException | InvocationTargetException |
+                     NoSuchMethodException ignored) {
+                // fall back to default when loading failed.
+                menu = new MainWindowMenu(this, mainWindowMenuHandler);
+            }
```
t-cordonnier, line 148: To enable the final variable I propose this:
```suggestion
        if (menuClass != null) {
            BaseMainWindowMenu menu1;
            try {
                menu1 = menuClass.getDeclaredConstructor(MainWindow.class, MainWindowMenuHandler.class)
                                .newInstance(this, mainWindowMenuHandler);
            } catch (InstantiationException | IllegalAccessException | InvocationTargetException |
                     NoSuchMethodException ignored) {
                // fall back to default when loading failed.
                menu1 = new MainWindowMenu(this, mainWindowMenuHandler);
            }
            menu = menu1;
        } else {
```

src/org/omegat/gui/main/MainWindowNewMenu.java
```diff
@@ -41,13 +41,14 @@
...
  * @author Hiroshi Miura
  */
-public final class MainWindowBurgerMenu extends BaseMainWindowMenu
+public final class MainWindowNewMenu extends BaseMainWindowMenu
         implements ActionListener, MenuListener, IMainMenu {
 
     JMenu burgerMenu;
 
-    public MainWindowBurgerMenu(MainWindow mainWindow, MainWindowMenuHandler mainWindowMenuHandler) {
+    public MainWindowNewMenu(MainWindow mainWindow, MainWindowMenuHandler mainWindowMenuHandler) {
```
t-cordonnier, line 49: If you add the possibility to add multiple variants like this, then I think the term "new UI" will become obsolete when there will be 3 or 4 implementations.
I would suggest to rename the classical class as MainWindowClassicMenu and find something else for the new UI

t-cordonnier (changes requested):
I did not succeed to switch to the new UI for a test. In an earlier version of 6.1 I remember there was a button somewhere in preferences but it disappeared. Where is it?

miurahr:
> I did not succeed to switch to the new UI for a test. In an earlier version of 6.1 I remember there was a button somewhere in preferences but it disappeared. Where is it?

We decided to be include it in 6.2 release in the Dev ML thread.
https://sourceforge.net/p/omegat/mailman/omegat-development/thread/F3961C2E-EBB0-4BC5-8503-66DAE9B54847%40traduction-libre.org/#msg37844269
 
You can enable it with removing comment-out in `Plugins.properties` and run.

src/org/omegat/gui/main/MainWindow.java
miurahr: Thank you for the suggestion. It is public because `EditorController` class acces menu field to enable/disable menu items.

I think there is a space we can define a better menu API for components/modules, to add/remove menu item, enable/disable specified menu item, etc.

t-cordonnier: Yes but this is not a reason to allow something like ` ((MainWindow) Core.getMainWindow()).menu = new Menu.... `

And yes, you are true, an API would be better to give an abstraction about what is allowed or not to do with the menu.

This conducts me to a point which I did not know where to put, as it is linked to new UI but not to this patch.
Actually both MainWindowMenu and MainWindowNewMenu have a method which quotes the sub-menus one by one. That means that old and new UI _may_ contain a distinct list of menus if the code differs: is it an intentional behaviour? I think you should find a way to put this code in BaseMainWindowMenu, or put the list in BaseMainWindowMenu to ensure it is the same in all sub-implementations. Unless I am wrong and it was an intentional behaviour? Do you have ideas about any menu which should appear in old UI and not in new one or the contrary?

t-cordonnier:
> We decided to be include it in 6.2 release in the Dev ML thread. https://sourceforge.net/p/omegat/mailman/omegat-development/thread/F3961C2E-EBB0-4BC5-8503-66DAE9B54847%40traduction-libre.org/#msg37844269

And how will you proceed? Will you create a new branch for 6.1 where the code is absent or will version 6.1 contain the code but inactivated?

> 
> You can enable it with removing comment-out in `Plugins.properties` and run.

Already tried before sending the previous answer, did not work almost when starting using `gradlew run`. Maybe this command does not check for changes in `Plugin.properties` ?
And as it is not documented, can you remember me in which part of Preferences is this button supposed to be?

src/org/omegat/gui/main/MainWindow.java
miurahr: > I think you should find a way to put this code in BaseMainWindowMenu, or put the list in BaseMainWindowMenu to ensure it is the same in all sub-implementations. Unless I am wrong and it was an intentional behaviour?

I'd like to take a way __puting the list in BaseMainWindowMenu__

src/org/omegat/gui/main/MainWindowNewMenu.java
miurahr: Oh, yes, it is reason I put a name "BurgerMenu" in previous, but the class does not only consist a burger menu also consist with toobar interface. I don't want to change classic and standard menu class name.
The  "new UI" is strongly inspired from "newUI" feature of JetBrains IntelliJ IDEA 2023 version.
https://www.jetbrains.com/help/idea/new-ui.html#new_themes_icons_font

An idea to improve usability by avoiding modal dialog in semi frequent operations;  select target file.

miurahr:
> Already tried before sending the previous answer, did not work almost when starting using `gradlew run`. Maybe this command does not check for changes in `Plugin.properties` ? And as it is not documented, can you remember me in which part of Preferences is this button supposed to be?

You can see a preference > plugins > newUI and checkbox there.

src/org/omegat/gui/main/MainWindowNewMenu.java
t-cordonnier: You could keep the name "burger", and we could refer to it as "burger style" in the documentation: I doubt anybody else will use this name for something else.
And rename the original class becomes necessary to avoid confusion (this class is _a_ menu, not anymore _the_ menu)

t-cordonnier:
> You can see a preference > plugins > newUI and checkbox there.

OK now I see it. Looks strange to see it isolated. Now that you are trying to propose multiple variants, this should not anymore be a checkbox but a combo, probably. Then, even if new UI is a plugin, the basis (BaseMainWindowMenu) is in the core so the choice could be in general preferences or in view menu.

Plugins.properties
miurahr: Now activate for review

src/org/omegat/gui/main/MainWindowNewMenu.java
miurahr: revert name.

miurahr:
Extend `Preferences > Appearance` menu with combobox.
Introduce plugin registration hooks.

miurahr:
> And how will you proceed? Will you create a new branch for 6.1 where the code is absent or will version 6.1 contain the code but inactivated?

Yes, Version 6.1 contain the code but inactivated. Developer still continue to improve it on master branch.

miurahr:
There is no UI effect when mouse is hover on an item in the list. TODO

miurahr:
> There is no UI effect when mouse is hover on an item in the list. TODO

There is no spare time to do it.  It will be implemented in future.

miurahr:
Will be improved another PR.

PR #627, opened 2023-05-29
[BUGS#1180] Support HiDPI on Linux

miurahr:
Here is an explanation how is font scaling works.
https://github.com/JFormDesigner/FlatLaf/issues/590

A mechanism scaling font is different among platforms.
 
Now OmegaT on Java 11 or later support HiDPI. We can observe menu font size, standard widgets are scaled properly. But we also observe editor font does not do scaling on Linux.

It is because mac/windows scales graphics, but Linux scales Font point size.

## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?

- #1180 Font does not scale on HiDPI display on Linux
    * https://sourceforge.net/p/omegat/bugs/1180/

## What does this PR change?

- Introduce UIScale class   derived from FormDev FlatLaf library
  original apache-2
- Scale/Unscale editor pane fonts
- Detect detailed OS version and environment

- Update MainWindow class to use FontUtil class to retrive default/configured font.
- Update MainWindow class to listen Font scale change and fire FontChange event.
- Update preference dialog to use scaled/unscaled font size for display.
- Introduced new three classes;  SystemInfo, UIScale and FontUtil classes.
- Reduce code duplication between SystemInfo and Platform classes


miurahr:
A change here work mostly well also on Windows. There is not scaled for pane titles, preference menu tree fonts and buttons,  attributes table titles.

Patched version
image

miurahr:
Java 11 on Windows seems change a font size when change OS preference to be magnitude other than 100%.

miurahr:
There is no bad effect on Windows, and can support font scaling on Linux.

src/org/omegat/util/Language.java
```diff
@@ -118,7 +118,7 @@ public Language(String str) {
      * user.
      */
     public String getDisplayName() {
-        if (Platform.getJavaVersion() == 8) {
+        if (!SystemInfo.isJava_9_orLater) {
             // work around for java 8 JRE localization bug
             // see https://github.com/OmegaT-L10N/zh_CN/issues/5
             if (Locale.getDefault().equals(Locale.SIMPLIFIED_CHINESE)) {
```
miurahr, line 124: This is workaround for JDK8, so we can remove it.

miurahr:
tested on Ubuntu 22.04.2 it works very well.

Cinnamon desktop on Mint Linux 21 has bug. When increase scaling to 150%, font size of java application become 66%!
Same situation on IntelliJ IDEA on Mint Linux. When restarting omegat, everything is well.

damien-rembert:
Do I understand this correctly: this is supposed to enable the  OmegaT UI to follow the scale in my OS settings? If that's correct then it  doesn't seem to work on my distro, changing the global scale in display configuration doesn't seem to affect OmegaT. I'm on KDE Neon 5.27. I wonder if this might be because KDE Plasma is still using X rather than Wayland? Do you know if the operating systems you tried use X or Wayland?
I will give this a go on a MacOS computer tomorrow.

miurahr:
> Do I understand this correctly: this is supposed to enable the OmegaT UI to follow the scale in my OS settings? If that's correct then it doesn't seem to work on my distro, changing the global scale in display configuration doesn't seem to affect OmegaT. I'm on KDE Neon 5.27. I wonder if this might be because KDE Plasma is still using X rather than Wayland? Do you know if the operating systems you tried use X or Wayland? I'm afraid I don't know much on the subject so I'm trying to guess but I can't really say much more with confidence, sorry. I will give this a go on a MacOS computer tomorrow.

Linux HiDPI support is very complex background.

Here is a best explanation. 
https://wiki.archlinux.org/title/HiDPI

Java Swing GtkLookAndFeel which is a default native LaF on Linux is aware of GTK/GNOME scaling factor that is 2, 3 etc.
The patch try to detect a scaling factor when initializing MainWindow.

When you change magnitude of scaling factor in OS settings, then restart OmegaT, it will follow your setting.
Please try  `Project > restart` .

Why OmegaT does not change font size?  Java runtime 11,17 does not detect scaling factor when gsettings value is `0` - means auto detection.  I put PropertyChangeListener for LaF objects, but Java runtime does not tell change to OmegaT. 

There is a space where we can improve Java Runtime on OpenJDK project!

For KDE/Plasma
https://invent.kde.org/plasma/kde-gtk-config/-/merge_requests/49

- `KScreen/ScaleFactor`
- gtkconfig: `gtk-xft-dpi`
- gtkconfig: `Gdk/WindowScalingFactor`

src/org/omegat/util/SystemInfo.java
damien-rembert, line 1: Just out of curiosity, why not make Platform and SystemInfo one class?

damien-rembert:
It works when using Wayland but I could not get it to work when starting the session with X11. There was no visible problem, everything displayed correctly, OmegaT was just not following the scaling percentage I indicated in the OS settings (even after restarting OmegaT). I couldn't say why but I wouldn't be surprised if the problem came from the distro tbh. 
I tried it on MacOS as well and I did not notice any issue. 
Code looks good to me, approved :)

damien-rembert (approved):

src/org/omegat/util/SystemInfo.java
miurahr: Platform class is originated in OmegaT, but SystemInfo come from Jetbrains copyright and from apache-2 license.
I'd like to unify it into Platfrom class.

damien-rembert: That makes sense, thanks

miurahr:
Now ready for review again. I'd like to squash and merge when approved.

miurahr:
I'd like to squash and merge, then ask community feedback.

PR #629, opened 2023-06-04
[BUGS#1182] Use DumontsHunspellDictionary instead of Hunspell

mohad12211:
Hunspell caches the object, so when we reload the project and free the old object and get a new dictionary object, it will return the same old freed one

## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?

 https://sourceforge.net/p/omegat/bugs/1182/

## What does this PR change?

-  use `org.languagetool.rules.spelling.hunspell.DumontsHunspellDictionary` instead of `org.languagetool.rules.spelling.hunspell.Hunspell` to construct the `HunspellDictionary`


briacp (approved):
Thanks, it's safer to implement this to avoid this blocking issue

PR #630, opened 2023-06-05
FR l10n fix for releases/5.8

brandelune:
Version number is 5.8.1, is that correct @miurahr ?

docs/fr/version_fr.properties


docs/fr/version_fr.properties
```diff
@@ -1,2 +1,2 @@
 # NOI18N
-version=5.8.0
+version=5.8.1
```
miurahr, line 2: it would be 5.8.0

PR #643, opened 2023-06-18
[RFE#1364] feat(preference): Disable restore button when no feature

miurahr:
When there is no field to restore, OmegaT disable restore button.

## Pull request type

## Which ticket is resolved?

- Remove "Restore defaults" button where it has no function
    * https://sourceforge.net/p/omegat/feature-requests/1364/

## What does this PR change?

-  IPreferencesController: add default method canRestoreDefaults
-  Return false when controller do not have restore function.


brandelune:
Please:
- add a line to changes.txt

PR #644, opened 2023-06-18
[RFE#1615]feat(html): preserve html entities

miurahr:
Add option to preserve HTML entities.
When translator want to see HTML entities as-is, this is the option and skip HTML entities conversions.

## Pull request type

## Which ticket is resolved?

- Feature requests: https://sourceforge.net/p/omegat/feature-requests/1615
- show entity unicode character instead delete it

## What does this PR change?

- Add option
- Skip conversion when option enabled

## Other information

I'm not sure the feature is useful.


brandelune:
I read the RFE and I'm not sure I understand why we need an option for that.

The RFE wants spaces expressed as HTML entities to be preserved when the option "preserve white space" is on, correct?

I think we should just do that. No need to add an option here. Did I miss something?

PR #646, opened 2023-06-22
[RFE#1519] Load only the latest plugin versions

briacp:
## Which ticket is resolved?

- If multiple versions of the same plugin are present, only the latest is loaded
  * https://sourceforge.net/p/omegat/feature-requests/1519/

## What does this PR change?

- It adds a check in `org.omegat.filters2.master.PluginUtils.loadPlugins(Map<String, String>)` before adding jars to the application classloader. The check read the Manifest of each jar and determine if it should be loaded based on its name and version.
- The version comparison is done with `org.omegat.util.VersionChecker`, with the qualifier part (after `-`) removed.


miurahr:
I think we should use LFS before adding binary files( *.jar) in the repository.

src/org/omegat/filters2/master/PluginUtils.java
```diff
@@ -167,6 +180,9 @@ public static void loadPlugins(final Map<String, String> params) {
                 Log.log(ex);
             }
         }
+
+        checkForLatestPluginVersion(urlList);
```
miurahr, line 184: method `checkForLatestPluginVersion` manipulate the `urlList` .  I think it is better to have a method `populatePluginUrlList()` that returns `List<URL>`. It may be like

```
protected static List<URL> populatePluginUrlList() {
        // list all jars in /plugins/
        FileFilter jarFilter = pathname -> pathname.getName().endsWith(".jar");
        List<File> fs = pluginsDirs.stream().flatMap(dir -> FileUtil.findFiles(dir, jarFilter).stream())
                .collect(Collectors.toList());
        List<URL> urlList = new ArrayList<>();
        for (File f : fs) {
            try {
                URL url = f.toURI().toURL();
                urlList.add(url);
                Log.logInfoRB("PLUGIN_LOAD_JAR", url.toString());
            } catch (IOException ex) {
                Log.log(ex);
           }
        }

       // body of checkForLatestPluginVersion here..
  
       return urlList;
```

miurahr (changes requested):
It looks good but I think there is a better way for the method.

briacp:
> I think we should use LFS before adding binary files( *.jar) in the repository. could you consider execute before adding *.jar in commits?

No problem I'll do it.

I thought LFS was only required for large files in git, and since the jars are really tiny (they only contain a manifest) I didn't even consider using it.

src/org/omegat/filters2/master/PluginUtils.java
briacp: Good idea, that should be cleaner.

Done in 183ad9eef4ff0a6e576f269d6931f8d3e1fd610a

miurahr:
CI test failed because LFS config may be affected.  A gradle wrapper script do not find gradle-wrapper.jar file.
Could you remove LFS config and push it again?

miurahr:
> I thought LFS was only required for large files in git, and since the jars are really tiny (they only contain a manifest) I didn't even consider using it.

FYI: OmegaT repository consumes Gigabytes because it has many jars in `lib/` without LFS in past days.

miurahr:
I've force pushed to your repository to work around.
I drop lfs commit from your history using `git rebase -i HEAD~4` and put `d` for commit.

test/src/org/omegat/filters2/master/PluginUtilsTest.java
```diff
@@ -0,0 +1,64 @@
...
+        List<File> fileList = pluginsDirs.stream().flatMap(
+                dir -> FileUtil.findFiles(dir, pathname -> pathname.getName().endsWith(".jar")).stream())
+                .collect(Collectors.toList());
+
+        assertEquals(5, fileList.size());
+        List<URL> urlList = PluginUtils.populatePluginUrlList(pluginsDirs);
+        assertEquals(2, urlList.size());
+
+        assertTrue(urlList.get(0).getFile().endsWith("anotherPlugin-0.3.1.jar"));
+        assertTrue(urlList.get(1).getFile().endsWith("testPlugin-1.13-1.45.0.jar"));
```
miurahr, line 61:
```suggestion
        final Set<String> expected = new HashSet<>();
        expected.add("anotherPlugin-0.3.1.jar");
        expected.add("testPlugin-1.13-1.45.0.jar");
        final List<String> result = urlList
                .stream()
                .map(url -> Paths.get(url.getPath()).getFileName().toString())
                .collect(Collectors.toList());
        assertTrue(result.containsAll(expected));
```

miurahr: Please add required imports by yourself.

miurahr: There is a space which can be improved.  It is better to check urlList directly other than producing result.

test/src/org/omegat/filters2/master/PluginUtilsTest.java
```diff
@@ -0,0 +1,70 @@
...
+
+package org.omegat.filters2.master;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+import java.io.File;
+import java.net.URL;
+import java.util.ArrayList;
+import java.util.List;
```
miurahr, line 34:
```suggestion
import java.util.ArrayList;
import java.util.HashSet;
import java.util.List;
import java.util.Set;
```

test/src/org/omegat/filters2/master/PluginUtilsTest.java
```diff
@@ -0,0 +1,70 @@
...
+import org.omegat.util.FileUtil;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+
+public class PluginUtilsTest {
+
+    @Test
+    public void testLoadLatestPluginVersionOnly() throws JsonProcessingException, JAXBException {
+        final List<File> pluginsDirs = new ArrayList<>();
+        pluginsDirs.add(new File("test/data/plugin/jar"));
```
miurahr, line 49:
```suggestion
        File testDataDir = new File("test/data/plugin/jar");
        try {
            URL testDataDirURL = testDataDir.toURL().toURI();
        } catch (MalformedURLException ignored) {
        }
        pluginsDirs.add(testDataDir);
```

test/src/org/omegat/filters2/master/PluginUtilsTest.java
```diff
@@ -0,0 +1,70 @@
...
+                dir -> FileUtil.findFiles(dir, pathname -> pathname.getName().endsWith(".jar")).stream())
+                .collect(Collectors.toList());
+
+        assertEquals(5, fileList.size());
+        List<URL> urlList = PluginUtils.populatePluginUrlList(pluginsDirs);
+        assertEquals(2, urlList.size());
+
+        final Set<String> expected = new HashSet<>();
+        expected.add("anotherPlugin-0.3.1.jar");
+        expected.add("testPlugin-1.13-1.45.0.jar");
```
miurahr, line 62:
```suggestion
        try {
            expected.add(new URL(testDataDirURL, "anotherPlugin-0.3.1.jar"));
            expected.add(new URL(testDataDirURL, "testPlugin-1.13-1.45.0.jar"));
        } catch (MalformedUrlException ignored) {
        }
```

test/src/org/omegat/filters2/master/PluginUtilsTest.java
```diff
@@ -0,0 +1,70 @@
...
+        final List<String> result = urlList
+                .stream()
+                .map(url -> Paths.get(url.getPath()).getFileName().toString())
+                .collect(Collectors.toList());
+        assertTrue(result.containsAll(expected));
```
miurahr, line 67:
```suggestion
        assertTrue(urlList.containsAll(expected));
```

miurahr (commented):
Here is an another approach.

miurahr (approved):

PR #655, opened 2023-06-27
Make the method loadScreenLayout public

briacp:
Making public the following method allows updating the layout with an external file.

I don't think there's a good reason to have it private, since the `resetDesktopLayout` is doing more or less the same thing and is public.

https://github.com/omegat-org/omegat/blob/a8d2077008ab5eb54439346dac75d2e4beaeb29b/src/org/omegat/gui/main/MainWindowUI.java#L324

### DIFF
--- a/src/org/omegat/gui/main/MainWindowUI.java
+++ b/src/org/omegat/gui/main/MainWindowUI.java
@@ -321,7 +321,7 @@ public final class MainWindowUI {
     /**
      * Load the main window layout from the specified file. Will reset to defaults if an error occurs.
      */
-    private static void loadScreenLayout(MainWindow mainWindow, File uiLayoutFile) {
+    public static void loadScreenLayout(MainWindow mainWindow, File uiLayoutFile) {
         try (InputStream in = new FileInputStream(uiLayoutFile)) {
             mainWindow.desktop.readXML(in);
         } catch (Exception ex) {


miurahr:
There is `PerProjectLayoutHandler` which update layout when loading and closing project.
A method `loadScreenLayout` is called from the handler.

Are there any use case other than  current functions?  

1. defult
2. global configuration
3. per project

briacp:
> Are there any use case other than current functions?

I can think of two use cases :

* have different layouts for different tasks. For example, I don't need to have the Machine Translations or Fuzzy Matches panels when I'm reviewing translations, only the Glossary, Notes and Editor
* for a LSP, be able to download and update system-wide config to provide remote translators a consistent config across all the installs

miurahr:
> > Are there any use case other than current functions?
> 
> I can think of two use cases :
> 
> * have different layouts for different tasks. For example, I don't need to have the Machine Translations or Fuzzy Matches panels when I'm reviewing translations, only the Glossary, Notes and Editor
> * for a LSP, be able to download and update system-wide config to provide remote translators a consistent config across all the installs

Both can be implemented without the change proposed.

User script can do with a public API and method;

```
Core.getMainWindow().getDesktop().readXML(in);
```

Are there any use case other than this API?

briacp:
> User script can do with a public API and method;
> 
> ```
> Core.getMainWindow().getDesktop().readXML(in);
> ```

You're right, I don't know why it never occurred to me to do it this way.

PR #659, opened 2023-06-29
[6.0] docs: remove obsolete warning and cleanup image files

brandelune:
See commit list.


miurahr:
There are not enough explaination in commit messages, especially for binary/image changes.
Text change can be shown as diff, but binary/image change can not. So more words are required for the commit.
for example, "fix CA 6.0 layout screenshot \n Use localized UI image\n Remove status message "  etc...

brandelune:
I removed the images or changed their names. There is no diff for that.

miurahr:
> add DE 6.0 layout screenshot

Uhm, if so, a commit message "add DE 6.0 layout screenshot" is incorrect. Is it "Change image file names" correct?

brandelune:
I just rebased a lot of things in one commit so as not to have 20 different commits for trivial things and I grouped them by language.

If you insist that we rebase things only to have a "clean" git history, that's what we get. And the problem with rebasing this way is that we can't change individual commits.

Kazephil (commented):
Looks good to me.

Just one question:

Is using a screenshot with the English interface in all languages deliberate?
I was expecting the localized screenshot.

brandelune:
> Is using a screenshot with the English interface in all languages deliberate? I was expecting the localized screenshot.

It's deliberate inasmuch as we currently don't have a way to automatize screenshot creation, and you'll notice that I either

- removed the other screenshots, or
- replaced them with HTML+CSS mockups

My idea is to recreate this screenshot with HTML+CSS (or SVG, which would be cool, but maybe more complex), so that we can have fully automatized L10N.

Let me go on with the discussion on the dev list though, because it would be interesting to have other people's take.

Kazephil:
> My idea is to recreate this screenshot with HTML+CSS (or SVG, which would be cool, but maybe more complex), so that we can have fully automatized L10N.

I'll add that to my growing list of "things to look into for 6.1 or later" :smiley_cat: 

> Let me go on with the discussion on the dev list though, because it would be interesting to have other people's take.

I hope you get at least a few responses!

Kazephil (approved):

miurahr (approved):

PR #662, opened 2023-07-01
[6.0] test: ProjectFileStorage cases

miurahr:
- Read with entities
- Write team project - with mapping - with excludes attributes

## Pull request type

- Other (describe below)

Add test  to check ProjectFileStorage write cases.

## Other information

maser branch changes dependency for XML handling libraries.
These tests help keeping compatibility between versions.

PR #663, opened 2023-07-01
[6.1][BUGS#1191] fix: jackson custom serializer for ProjectFileStorage

miurahr:
ProjectFileStorage class is use to read and write `omegat.project` file.
This PR change Jackson object mapper configuration using custom serializer for `otherAttributes` attribute.
Current master branch produce `<otherAttributes/>` node in `omegat.project` when there is no attribute, and fix it with custom selializer.

I've added test cases that is as same as `releases/6.0`, so it guarantee a compaitibility between versions.

## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?
 
- ProjectFile has unwanted XML element otherAttributes
  * https://sourceforge.net/p/omegat/bugs/1191/

## What does this PR change?

- Add test as same as  #662
- Add custom serializer for `Map<QName, String> otherAttributes` object.

### DIFF
--- a/test/src/org/omegat/util/ProjectFileStorageTest.java
+++ b/test/src/org/omegat/util/ProjectFileStorageTest.java
@@ -31,25 +31,35 @@ import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 
 import java.io.File;
+import java.net.URL;
 import java.nio.file.Files;
 import java.nio.file.Paths;
+import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
 
 import org.apache.commons.io.FileUtils;
+import org.custommonkey.xmlunit.XMLAssert;
+import org.custommonkey.xmlunit.XMLUnit;
 import org.junit.After;
 import org.junit.Before;
 import org.junit.Test;
+import org.xml.sax.InputSource;
+
 import org.omegat.core.data.ProjectException;
 import org.omegat.core.data.ProjectProperties;
 import org.omegat.tokenizer.LuceneEnglishTokenizer;
 import org.omegat.tokenizer.LuceneFrenchTokenizer;
 
 import gen.core.project.Omegat;
+import gen.core.project.RepositoryDefinition;
+import gen.core.project.RepositoryMapping;
 
 public class ProjectFileStorageTest {
 
     private static final File PROJECT_DIR = new File("test/data/project");
+    private static final File SCHEMA_FILE = new File("src/schemas/project_properties.xsd");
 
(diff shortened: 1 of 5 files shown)


src/org/omegat/util/ProjectOtherAttributesSerializer.java
```diff
@@ -0,0 +1,58 @@
...
+import java.util.Map;
+
+import javax.xml.namespace.QName;
+
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.databind.SerializerProvider;
+import com.fasterxml.jackson.databind.ser.std.StdSerializer;
+import com.fasterxml.jackson.databind.type.MapType;
+
+public class ProjectOtherAttributesSerializer extends StdSerializer<Map<QName, String>> {
```
miurahr, line 38: Can we move the serializer as internal class?
It is only used in ProjectFileStorage class and not considered for extension and used in another class.

test/src/org/omegat/util/ProjectFileStorageTest.java
```diff
@@ -58,11 +68,19 @@ public final void setUp() throws Exception {
         tempDir = Files.createTempDirectory("omegat").toFile().getAbsoluteFile();
         assertTrue(tempDir.isDirectory());
         TestPreferencesInitializer.init(tempDir.getPath());
+        XMLUnit.setIgnoreWhitespace(true);
+        XMLUnit.setIgnoreAttributeOrder(true);
+        XMLUnit.setIgnoreComments(true);
+        // XMLUnit.setControlEntityResolver(null);
+        // XMLUnit.setTestEntityResolver(null);
```
miurahr, line 75:
```suggestion
```

test/src/org/omegat/util/ProjectFileStorageTest.java
```diff
@@ -58,11 +68,17 @@ public final void setUp() throws Exception {
...
+        XMLUnit.setIgnoreAttributeOrder(true);
+        XMLUnit.setIgnoreComments(true);
+
     }
 
     @After
     public final void tearDown() throws Exception {
         FileUtils.deleteDirectory(tempDir);
+        XMLUnit.setControlEntityResolver(null);
+        XMLUnit.setTestEntityResolver(null);
```
miurahr, line 81:
```suggestion
```
No need to reset.

brandelune:
Please:
- add a line to changes.txt

miurahr:
> Please:
> 
> * add a line to changes.txt

It has already been.
https://github.com/omegat-org/omegat/blob/master/release/changes.txt#L61-L62

PR #665, opened 2023-07-03
fix: FiltersCustomizer consistency

miurahr:
- Update Filter customizer change when next edit
- resolve BUGS#1045

## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?

- Changes in "Edit Filter"dialog window not reflected when the window is called up again
  * https://sourceforge.net/p/omegat/bugs/1045/ 

## What does this PR change?

- Update FiltersCustomizer table model class with data POJO class
- Set updated config in FiltersCustomizer table when edit
- Add `setFilter` method for table model


kosivantsov:
I fetched the branch and built it locally. It seems to fix the bug. Thank you so much for looking into it.

PR #519, opened 2023-03-13
[BUGS#558] fix: incorrect display of rtl segments in fuzzy match pane

damien-rembert:
## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?

- Incorrect display of RTL segments in the Fuzzy Matches pane
  * [https://sourceforge.net/p/omegat/bugs/558/](https://sourceforge.net/p/omegat/bugs/558/)

## What does this PR change?

- Consolidate methods used for handling bi-directionality in one file
- Apply an orientation to panels/panes as needed

## Other information

Sorry, I realise it's a lot of moving things around but it was the easiest way to get everything tested.
I have tried to keep the commits fairly small so that might help.

PR #669, opened 2023-07-03
[BUGS#558] fix: incorrect display of rtl segments in fuzzy match pane

miurahr:
Form #519 to be cleaner commits.

## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?

-  Incorrect display of RTL segments in the Fuzzy Matches pane
- https://sourceforge.net/p/omegat/bugs/558/

## What does this PR change?

- Introduce BiDiUtils utility class, gathering all BiDi things
- Add test for MatchesVarExpansion and then extend it to support RtL
- refactor Document3.ORIENTATION enum in BiDiUtils.ORIENTATION
- refactor to remove ORIENTATION from return value of internal API
- Set LtR/RtL properly on editor and another panes
- refactor Langauge#isRTL Language#localeIsRTL to move into BiDiUtils
- All commits author @damien-rembert

### DIFF
--- a/src/org/omegat/filters3/xml/XMLTag.java
+++ b/src/org/omegat/filters3/xml/XMLTag.java
@@ -29,6 +29,7 @@ package org.omegat.filters3.xml;
 
 import org.omegat.filters3.Attribute;
 import org.omegat.filters3.Tag;
+import org.omegat.util.BiDiUtils;
 import org.omegat.util.Language;
 
 /**
@@ -57,7 +58,7 @@ public class XMLTag extends Tag {
     public String toOriginal() {
         StringBuilder buf = new StringBuilder();
 
-        boolean isRtl = Language.isRTL(targetLanguage.getLanguageCode());
+        boolean isRtl = BiDiUtils.isRtl(targetLanguage.getLanguageCode());
         boolean differentDir = isDifferentDirection(isRtl);
         boolean isSpecialDocxTagLTR = isSpecialDocxBidiTag(false);
         boolean isSpecialDocxTagRTL = isSpecialDocxBidiTag(true);
@@ -135,7 +136,7 @@ public class XMLTag extends Tag {
             return false;
         }
 
-        return isRtl != Language.isRTL(sourceLanguage.getLanguageCode());
+        return isRtl != BiDiUtils.isRtl(sourceLanguage.getLanguageCode());
     }
 
     private boolean isSpecialDocxBidiTag(boolean complex) {
--- a/src/org/omegat/gui/editor/AlphabeticalMarkers.java
+++ b/src/org/omegat/gui/editor/AlphabeticalMarkers.java
@@ -47,7 +47,7 @@ import javax.swing.JScrollPane;
 import javax.swing.SwingUtilities;
 
 import org.omegat.core.Core;
-import org.omegat.util.Language;
+import org.omegat.util.BiDiUtils;
 import org.omegat.util.Preferences;
 import org.omegat.util.StaticUtils;
 import org.omegat.util.gui.UIThreadsUtil;
@@ -79,7 +79,7 @@ public abstract class AlphabeticalMarkers extends JPanel {
         this.scrollPane = scrollPane;
         String sourceLang = Core.getProject().getProjectProperties()
                 .getSourceLanguage().getLanguageCode();
-        this.sourceLangIsRTL = Language.isRTL(sourceLang);
+        this.sourceLangIsRTL = BiDiUtils.isRtl(sourceLang);
         this.colorScheme = createColorScheme(scrollPane.getViewport().getView().getBackground());
     }
 
--- a/src/org/omegat/gui/editor/EditorController.java
+++ b/src/org/omegat/gui/editor/EditorController.java
@@ -659,11 +659,11 @@ public class EditorController implements IEditor {
         }
 
         // check if RTL support required for document
-        boolean hasRTL = sourceLangIsRTL || targetLangIsRTL || Language.localeIsRTL()
+        boolean hasRTL = sourceLangIsRTL || targetLangIsRTL || BiDiUtils.isLocaleRtl()
                 || currentOrientation != BiDiUtils.ORIENTATION.ALL_LTR;
         Map<Language, ProjectTMX> otherLanguageTMs = Core.getProject().getOtherTargetLanguageTMs();
         for (Map.Entry<Language, ProjectTMX> entry : otherLanguageTMs.entrySet()) {
-            hasRTL = hasRTL || Language.isRTL(entry.getKey().getLanguageCode().toLowerCase(Locale.ENGLISH));
+            hasRTL = hasRTL || BiDiUtils.isRtl(entry.getKey().getLanguageCode().toLowerCase(Locale.ENGLISH));
         }
 
         Document3 doc = new Document3(this);
@@ -1930,7 +1930,7 @@ public class EditorController implements IEditor {
             String language = detectFirstStepsLanguage();
             introPane = new JTextPane();
             introPane
-                    .setComponentOrientation(Language.isRTL(language) ? ComponentOrientation.RIGHT_TO_LEFT
+                    .setComponentOrientation(BiDiUtils.isRtl(language) ? ComponentOrientation.RIGHT_TO_LEFT
                             : ComponentOrientation.LEFT_TO_RIGHT);
             introPane.setEditable(false);
             DragTargetOverlay.apply(introPane, dropInfo);
--- a/src/org/omegat/gui/editor/EditorController.java
+++ b/src/org/omegat/gui/editor/EditorController.java
@@ -570,45 +570,11 @@ public class EditorController implements IEditor {
      * Decide what document orientation should be default for source/target languages.
      */
     private void setInitialOrientation() {
-        String sourceLang = Core.getProject().getProjectProperties().getSourceLanguage().getLanguageCode();
-        String targetLang = Core.getProject().getProjectProperties().getTargetLanguage().getLanguageCode();
-
-        sourceLangIsRTL = Language.isRTL(sourceLang);
-        targetLangIsRTL = Language.isRTL(targetLang);
-
-        if (sourceLangIsRTL != targetLangIsRTL || sourceLangIsRTL != Language.localeIsRTL()) {
-            currentOrientation = BiDiUtils.ORIENTATION.DIFFER;
-        } else {
-            if (sourceLangIsRTL) {
-                currentOrientation = BiDiUtils.ORIENTATION.ALL_RTL;
-            } else {
-                currentOrientation = BiDiUtils.ORIENTATION.ALL_LTR;
-            }
-        }
-        applyOrientationToEditor();
-    }
-
-    /**
-     * Define editor's orientation by target language orientation.
-     */
-    private void applyOrientationToEditor() {
-        ComponentOrientation targetOrientation = null;
-        switch (currentOrientation) {
-        case ALL_LTR:
-            targetOrientation = ComponentOrientation.LEFT_TO_RIGHT;
-            break;
-        case ALL_RTL:
-            targetOrientation = ComponentOrientation.RIGHT_TO_LEFT;
-            break;
-        case DIFFER:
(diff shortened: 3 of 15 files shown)


brandelune (approved):

src/org/omegat/filters3/xml/XMLTag.java
```diff
@@ -135,7 +136,7 @@ private boolean isDifferentDirection(boolean isRtl) {
             return false;
         }
 
-        return isRtl != Language.isRTL(sourceLanguage.getLanguageCode());
+        return isRtl != BiDiUtils.isRtl(sourceLanguage.getLanguageCode());
```
miurahr, line 139: I think there is no active reason to move the method `isRTL`, because there are other similar methods in `Language` class. It is better to keep there for backward compatibility.

src/org/omegat/filters3/xml/XMLTag.java
```diff
@@ -57,7 +58,7 @@ public XMLTag(String tag, String shortcut, Type type, org.xml.sax.Attributes att
     public String toOriginal() {
         StringBuilder buf = new StringBuilder();
 
-        boolean isRtl = Language.isRTL(targetLanguage.getLanguageCode());
+        boolean isRtl = BiDiUtils.isRtl(targetLanguage.getLanguageCode());
```
miurahr, line 61: Same here.

src/org/omegat/gui/editor/AlphabeticalMarkers.java
```diff
@@ -79,7 +79,7 @@ public abstract class AlphabeticalMarkers extends JPanel {
         this.scrollPane = scrollPane;
         String sourceLang = Core.getProject().getProjectProperties()
                 .getSourceLanguage().getLanguageCode();
-        this.sourceLangIsRTL = Language.isRTL(sourceLang);
+        this.sourceLangIsRTL = BiDiUtils.isRtl(sourceLang);
```
miurahr, line 82: same here.

src/org/omegat/gui/editor/EditorController.java
```diff
@@ -694,11 +670,11 @@ protected void loadDocument() {
...
 
         // check if RTL support required for document
-        boolean hasRTL = sourceLangIsRTL || targetLangIsRTL || Language.localeIsRTL()
-                || currentOrientation != Document3.ORIENTATION.ALL_LTR;
+        boolean hasRTL = sourceLangIsRTL || targetLangIsRTL || BiDiUtils.isLocaleRtl()
+                || currentOrientation != BiDiUtils.ORIENTATION.ALL_LTR;
         Map<Language, ProjectTMX> otherLanguageTMs = Core.getProject().getOtherTargetLanguageTMs();
         for (Map.Entry<Language, ProjectTMX> entry : otherLanguageTMs.entrySet()) {
-            hasRTL = hasRTL || Language.isRTL(entry.getKey().getLanguageCode().toLowerCase(Locale.ENGLISH));
+            hasRTL = hasRTL || BiDiUtils.isRtl(entry.getKey().getLanguageCode().toLowerCase(Locale.ENGLISH));
```
miurahr, line 677: same here.

src/org/omegat/gui/editor/EditorController.java
```diff
@@ -694,11 +670,11 @@ protected void loadDocument() {
         }
 
         // check if RTL support required for document
-        boolean hasRTL = sourceLangIsRTL || targetLangIsRTL || Language.localeIsRTL()
-                || currentOrientation != Document3.ORIENTATION.ALL_LTR;
+        boolean hasRTL = sourceLangIsRTL || targetLangIsRTL || BiDiUtils.isLocaleRtl()
```
miurahr, line 673: similar issue here.

src/org/omegat/gui/editor/SegmentBuilder.java
```diff
@@ -540,7 +541,7 @@ private String addInactiveSegPart(boolean isSource, String text) throws BadLocat
      */
     private void addOtherLanguagePart(String text, Language language) throws BadLocationException {
         int prevOffset = offset;
-        boolean rtl = Language.isRTL(language.getLanguageCode());
+        boolean rtl = BiDiUtils.isRtl(language.getLanguageCode());
```
miurahr, line 544: same here.

src/org/omegat/gui/editor/SegmentBuilder.java
```diff
@@ -588,7 +589,7 @@ private void addModificationInfoPart(TMXEntry trans) throws BadLocationException
         }
 
         int prevOffset = offset;
-        boolean rtl = Language.localeIsRTL();
+        boolean rtl = BiDiUtils.isLocaleRtl();
```
miurahr, line 592: same here.

src/org/omegat/gui/editor/SegmentBuilder.java
```diff
@@ -623,8 +624,8 @@ private String addActiveSegPart(String text) throws BadLocationException {
         // as target text.
         insertDirectionMarker(rtl);
 
-        // the marker itself is in user language
-        insertDirectionEmbedding(Language.localeIsRTL());
+        //the marker itself is in user language
+        insertDirectionEmbedding(BiDiUtils.isLocaleRtl());
```
miurahr, line 628: same here.

src/org/omegat/util/BiDiUtils.java
```diff
@@ -0,0 +1,182 @@
...
+     *            ISO-639-2 language code
+     * @return true if language is RTL
+     */
+    public static boolean isRtl(final String language) {
+        return "ar".equalsIgnoreCase(language) || "iw".equalsIgnoreCase(language)
+                || "he".equalsIgnoreCase(language) || "fa".equalsIgnoreCase(language)
+                || "ur".equalsIgnoreCase(language) || "ug".equalsIgnoreCase(language)
+                || "ji".equalsIgnoreCase(language) || "yi".equalsIgnoreCase(language);
+    }
+
```
miurahr, line 178: Here is one just moving methods from Lanugage class.

src/org/omegat/gui/properties/SegmentPropertiesArea.java
```diff
@@ -341,6 +344,14 @@ private void setProperties(SourceTextEntry ste) {
         viewImpl.update();
     }
 
+    private String addBiDiToTranslation(String translation) {
+        return isTargetRtl ? BiDiUtils.addRtlBidiAround(translation) : translation;
+    }
+
+    private void initialiseIsTargetLangRtl() {
+        isTargetRtl = BiDiUtils.isTargetLangRtl();
```
miurahr, line 352: same here.

miurahr (commented):
Many change does not mandatory. We can shape it minimum.
I think we can consolidate Languages and BiDiUtils classes.

src/org/omegat/filters3/xml/XMLTag.java
miurahr: I checked a reason why we made Language class.
There is javadoc

```
 * This class is here, because the Locale has hard-coded '_' inside, and we must
 * adhere to ISO standard LL-CC.
 * <p>
 * This class tries to follow
 * <a href="https://www.gala-global.org/tmx-14b#xml:lang">TMX Specification on
 * languages</a>, which is based on
 * <a href="http://www.ietf.org/rfc/rfc3066.txt">RFC 3066</a>, i.e.
 * <ul>
 * <li>Language is composed from 1-8 alpha (A-Za-z) chars, then "-", then 1-8
 * alpha/digit chars (A-Za-z0-9).
 * <li>Case insensitive
 * <li>Case is not altered by this class, even though there exist conventions
 * for capitalization (ISO 3166 recommends that country codes are capitalized
 * (MN Mongolia), and ISO 639 recommends that language codes are written in
 * lower case (mn Mongolian)).
 * </ul>
```

It is for ISO standard LL-CC and RtL utility is no reason to place in Language class.

miurahr:
#519 and commits don't have enough explanation of a reason of things.

src/org/omegat/gui/editor/EditorController.java
```diff
@@ -569,54 +570,29 @@ private void setFont(final Font font) {
...
-                currentOrientation = Document3.ORIENTATION.ALL_RTL;
-            } else {
-                currentOrientation = Document3.ORIENTATION.ALL_LTR;
-            }
-        }
-        applyOrientationToEditor();
+    private void getValuesToDetermineOrientation() {
+        sourceLangIsRTL = BiDiUtils.isSourceLangRtl();
+        targetLangIsRTL = BiDiUtils.isTargetLangRtl();
+        currentOrientation = BiDiUtils.getOrientationType();
```
miurahr, line 580: Here is no reason to create private method because user is only single user.

src/org/omegat/gui/editor/mark/RemoveTagMarker.java
```diff
@@ -67,7 +67,7 @@ protected boolean isEnabled() {
 
     @Override
     protected void initDrawers(boolean isSource, boolean isActive) {
-        if (((EditorController) Core.getEditor()).getOrientation() == Document3.ORIENTATION.ALL_LTR) {
+        if (((EditorController) Core.getEditor()).getOrientation() == BiDiUtils.ORIENTATION.ALL_LTR) {
```
miurahr, line 70: This conditional can be a utility method `isOrientationAllLtR` or similar.

src/org/omegat/gui/matches/MatchesVarExpansion.java
```diff
@@ -274,11 +279,29 @@ public String expandVariables(NearString match) {
...
         // Variables
-        result.text = this.expandVariables(match);
+        result.text = expandVariables(match);
+        if (isMixedOrientation && isTargetRtl) {
+            result.text = result.text.replace(VAR_TARGET_TEXT, BiDiUtils.addRtlBidiAround(match.translation));
+        } else if (isMixedOrientation) {
+            result.text = result.text.replace(VAR_TARGET_TEXT, BiDiUtils.addLtrBidiAround(match.translation));
+        } else {
+            result.text = result.text.replace(VAR_TARGET_TEXT, match.translation);
+        }
```
miurahr, line 304: why put expand of VAR_TARGET_TEXT in `expandVariables` method?

miurahr: `expandVariables` is public method and used in 5 places.
It expands `VAR_TARGET_TEXT` in previous code but now It does not expand in the method.
It may lead a bug in another place.

src/org/omegat/gui/editor/EditorController.java
```diff
@@ -569,54 +570,29 @@ private void setFont(final Font font) {
...
-     * returns the orientation of the document
-     * (so we can decide what way of tag colouring we need;
-     * if that has been fixed in an other way, this method can be removed again.).
-     * @return
+     * returns the orientation of the document (so we can decide what way of tag
+     * colouring we need; if that has been fixed in an other way, this method
+     * can be removed again.).
      */
-    public Document3.ORIENTATION getOrientation() {
+    public BiDiUtils.ORIENTATION getOrientation() {
```
miurahr, line 595: Here is a place where public method value type changed. It can be compatibility breaking change.
If there is strong reason to break compatibility, we can keep  it to return `Document3.ORIENTATION`

miurahr: I think we can define

```
    /**
     * The orientation of the document is all RTL.
     * @return true when the orientation is all RtL. otherwise false.
     */
    public boolean isOrientationAllRtl() {
        return currentOrientation.equals(BiDiUtils.ORIENTATION.ALL_RTL);
    }
```

We can avoid return ORIENTATION enum from method.

miurahr:
Could you review it?

1. Is these changes functionally identical with #519  with reviewed adjustments?
2. Does the PR works as intended?

brandelune:
Please:
- add a line to changes.txt

miurahr:
Because there is no response from @damien-rembert  I'd like to merge it and ask community members for feedback.

damien-rembert:
Hi @miurahr , thanks for moving forward with this. I kind of lost momentum on that one and never really managed to get back into it when I tried so I greatly appreciate your carrying it to the end.

PR #670, opened 2023-07-04
fix: display the correct content of the Notes in XLIFF

briacp:
Currently notes are displayed like this :
image

instead of :
image

because we use the default `toString()` of the XMLEvent instead of the actual content of the character event.

+ minor refactor (removing extraneous calls to `.toString()` and replacing `list.size() > 0` by `!list.isEmpty()` to make a clearer intent.

## Which ticket is resolved?

- Content of notes elements are not displayed correctly with Xliff filter
  * https://sourceforge.net/p/omegat/bugs/1192/

### DIFF
--- a/release/changes.txt
+++ b/release/changes.txt
@@ -78,6 +78,9 @@
   - Lucene 5.2 does not work under WebStart
   https://sourceforge.net/p/omegat/bugs/813/
 
+  - Content of notes elements are not displayed correctly with Xliff filter
+  https://sourceforge.net/p/omegat/bugs/1192/
+
 ----------------------------------------------------------------------
  OmegaT 6.0.0 (2023-06-19)
 ----------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


t-cordonnier (approved):
OK for me. Indeed the code worked correctly until we switched to Woodstox. I already corrected some parts where Woodstox has different behaviour than original StaX implementation, but I am not sure there are eventually other ones, we may take the time to check.

t-cordonnier:
I approve the change and I would also recommend to apply it in 5.8 and 6.0 as well, since it is a bug which can happen in these releases as well.

briacp:
* Backported to 5.8.x in 7e8df6b84401107f9030fff9854a44204b6ce2b7
* Backported to 6.0.x in d3034acd9956f70a6f6bc3abb3cf2eacb1178ce5

PR #675, opened 2023-07-04
[RFE#1695] Select output format for statistics output: JSON, XML or Text

briacp:
The preference is set in the "Saving and Output" panel.

image

N.B. This does not include the match stats (`project_stats_match` and `project_stats_match_per_file`), only the general statistics `project_stats`.

## Which ticket is resolved?

- Generate JSON format stastics file when save
  * https://sourceforge.net/p/omegat/feature-requests/1695/


miurahr:
RFE#1965 is raised to provide stats information for project file list when opening project by always save the json format.

src/org/omegat/Main.java
```diff
@@ -417,33 +420,39 @@ private static int runConsoleStats() throws Exception {
...
+            // otherwise XML.
+            if (outputFilename.toLowerCase().endsWith(StatOutputFormat.JSON.getFileExtension())) {
+                statsMode = StatOutputFormat.JSON;
+            } else if (outputFilename.toLowerCase().endsWith(StatOutputFormat.XML.getFileExtension())) {
+                statsMode = StatOutputFormat.XML;
+            } else if (outputFilename.toLowerCase().endsWith(StatOutputFormat.TEXT.getFileExtension())) {
+                statsMode = StatOutputFormat.TEXT;
             } else {
-                statsMode = "XML";
+                statsMode = StatOutputFormat.XML;
```
briacp, line 436:
```suggestion
                statsMode = StatOutputFormat.JSON;
```

If we want to have all JSON stats by default, but it could break scripts/commands that use the CLI to produce stats that are expecting an XML output (current default for CLI).

briacp:
> RFE#1965 is raised to provide stats information for project file list when opening project by always save the json format.

It's mostly the case with this PR (except for CLI mode, see above). By default project stats will be written in JSON if the preference is not set.
I'm not sure it's best to *always* save in JSON as a translator might want to copy/paste the stats text to a LSP or prefer to process it with XML tools.

miurahr:
Is it any way to produce both text (for human) and xml or json (for machine)?

briacp:
> Is it any way to produce both text (for human) and xml or json (for machine)?

Indeed, I made some changes so each format can be selected independently (see new screenshot in the issue description)
The `stat_format` preference is a bit flag (1:text, 2:xml, 4:json)

brandelune (approved):
I don't have much time to check anything now. Sorry.

miurahr:
A  proposed change #676 ,  always commit JSON stats in team project, conflict here.
When no check exist, what happend in team project?

brandelune:
What about this: the "issue checking" checkboxes always have "tags" checked and that option cannot be unchecked.

So:
- if the project is a team project, "json" is checked by default and cannot be unchecked.
- if the project is a normal project, "text" is checked by default and cannot be unchecked.

Or:
- "text" is checked by default and cannot be unchecked.
- if the project is a team project, "json" is also checked by default and cannot be unchecked.

t-cordonnier:
> A proposed change #676 , always commit JSON stats in team project, conflict here. When no check exist, what happend in team project?

See my comment in #676: no conflict here (I did a merge in my computer to check), because #675 acts on SAVE event while #676 acts on COMPILE (commit target files)

miurahr (approved):
All my questions are resolved.

brandelune:
Please:
- ask for UI strings review by Philippe
- add a documentation ticket tagging me or Philippe
- add a line to changes.txt

miurahr:
@brandelune  you can assign reviewer as project manager. 
When you think you need more action before merge, you should not approve it.

brandelune:
> @brandelune you can assign reviewer as project manager.
> When you think you need more action before merge, you should not approve it.

People who merge the code should handle those tasks. Maybe we can have a checklist in the PR template ? Can you remind me where the PR template is ?

It would be better if modifications that impact Bundle.properties could automatically trigger a UI review and a documentation ticket. Also, any modification that refers to an RFE or a bug fix should really trigger a modification of changes.txt.

PR #676, opened 2023-07-07
[RFE#1690] Have "Commit Target Files" also commit statistics file

t-cordonnier:
When we call "commit target files", whenever from console or UI, the TXT and JSON statistics files are re-generated and if you are in a team project, the JSON is also sent to the repository

For explanation about the reason:
https://sourceforge.net/p/omegat/mailman/message/37780726/

## Pull request type

## Which ticket is resolved?

https://sourceforge.net/p/omegat/feature-requests/1690/

### DIFF
--- a/src/org/omegat/core/data/RealProject.java
+++ b/src/org/omegat/core/data/RealProject.java
@@ -716,7 +716,13 @@ public class RealProject implements IProject {
                 && isOnlineMode) {
             tmxPrepared = null;
             glossaryPrepared = null;
-            // commit translations
+            // Ticket 1690 - build project statistics files
+            // so that contents of these files is up to date with target files sent at same moment
+            StatsResult stat = CalcStandardStatistics.buildProjectStats(this);
+            stat.updateStatisticsInfo(hotStat);
+            String fn = config.getProjectInternal() + OConsts.STATS_FILENAME;
+            Statistics.writeStat(fn, stat.getTextData());
+            // commit translations and statistics
             try {
                 Core.getMainWindow().showStatusMessageRB("TF_COMMIT_TARGET_START");
                 remoteRepositoryProvider.switchAllToLatest();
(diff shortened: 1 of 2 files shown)


src/org/omegat/core/data/RealProject.java
```diff
@@ -699,6 +705,13 @@ public void compileProjectAndCommit(String sourcePattern, boolean doPostProcessi
                         null);
                 remoteRepositoryProvider.commitFiles(config.getTargetDir().getUnderRoot(),
                         "Project translation");
+                // Convert stats file name to relative
+                ProjectProperties.ProjectPath path = config.new ProjectPath(true);
+                path.setRelativeOrAbsolute(fn);
+                fn = path.getUnderRoot();
+                // remoteRepositoryProvider.copyFilesFromProjectToRepos(fn,null);
```
miurahr, line 712: To keep comaptibility with previous versions, we should keep commit text stat file.
When user community  recognize json stats file as standard team file, then we remove it from team repository.
 I think we may remove it in OmegaT 6.2.0.

We should notify the planning change in release 6.1.0.

miurahr:
Is this change able to co-exist with #675 ?

src/org/omegat/core/data/RealProject.java
t-cordonnier: No problem for me, I commented out only because after requesting for it Manuel suddenly wanted to "save bandwidth"

t-cordonnier:
> Is this change able to co-exist with #675 ?

Normally it should, since #675 works on SAVE event while my code works on COMPILE event, for that reason they are not in same part of the code.
Just tested a merge between #675 and #676, from a code point of view it works.
I#nitially Manuel wanted me to do both but he also saw #675 and told me that finally part of the work was already done.

That said I see that #675 also introduces configurability, maybe we could think about the possibility to make also the list of files to commit during COMPILE configurable : but then I can do it only after #675 is validated, so that I can re-use the new classes it defines for configuration.

src/org/omegat/core/data/RealProject.java
msoutopico: > No problem for me, I commented out only because after requesting for it Manuel suddenly wanted to "save bandwidth" and only the JSON was useful for him...

It was more a question than a request... In hindsight, since the stats file will always have a very small size, probably there's no need to worry about performance here.

Hence, while JSON is one always needed, I don't think it should be a problem to commit updates to stats files in several formats, if required or convenient.

brandelune:
Please:
- add a documentation ticket tagging me or Philippe
- add a line to changes.txt

src/org/omegat/core/data/RealProject.java
t-cordonnier: OK so I removed last patch then, now let @miurahr decide.

PR #686, opened 2023-07-08
update contributors' list

brandelune:


brandelune:
I don't think this one is necessary.

Issue #694, opened 2023-07-09
UI English to review


brandelune:
https://github.com/omegat-org/omegat/blob/c025843ea2fe3fb286474a9e1fe5977b0838d661/src/org/omegat/Bundle.properties#L505

https://github.com/omegat-org/omegat/blob/c025843ea2fe3fb286474a9e1fe5977b0838d661/src/org/omegat/Bundle.properties#L510

@Kazephil can you check the two blocks above? They look like they need a review. Thank you in advance.

PR #695, opened 2023-07-09
chore: Updated wording for UI blocks

Kazephil:
Here's a proposed updated wording for the blocks from Issue #694.

Let me know if you also want to backport them to 5.8 and 6.0.

### DIFF
--- a/src/org/omegat/Bundle.properties
+++ b/src/org/omegat/Bundle.properties
@@ -509,7 +509,7 @@ TF_LOAD_WARN_SOURCE_LOOP_EXCEPTION=Skipped one or more source files\n\
   Make sure all links are outside the source folder and subfolders.\n\
   Invalid link: {0}
 TF_LOAD_WARN_FILE_ACCESS=Skipped one or more source files due to an access permission \
-  error in the source folder. Skipped file(s): {0}
+  error in the source folder. Check the permissions of {0}
 TF_SOURCE_LOAD_ERROR=Skipped source file {0} because it could not be loaded.
 
 TF_TM_LOAD_ERROR=Failed to load translation memory: {0}\nSome attributes have invalid values. Use a TMX validator to find and fix them.
--- a/src/org/omegat/Bundle.properties
+++ b/src/org/omegat/Bundle.properties
@@ -503,13 +503,13 @@ MW_OPTIONMENU_APPEARANCE_SYNC_WITH_OS_COLOR=Sync with OS colour preference
 MW_REOPEN_QUESTION=OmegaT needs to reload the project to take the modifications into account.\n\
 Reload the project now?
 
-TF_LOAD_WARN_SOURCE_LOOP_EXCEPTION=Skipped some source file(s)\n\
+TF_LOAD_WARN_SOURCE_LOOP_EXCEPTION=Skipped one or more source files\n\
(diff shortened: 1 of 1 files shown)


brandelune:
> Here's a proposed updated wording for the blocks from Issue #694.

Thank you.

> Let me know if you also want to backport them to 5.8 and 6.0.

@miurahr can you confirm that please?

brandelune (approved):

src/org/omegat/Bundle.properties
```diff
@@ -502,13 +502,13 @@ MW_OPTIONMENU_APPEARANCE_THEME_LABEL=Theme:
...
+TF_LOAD_WARN_SOURCE_LOOP_EXCEPTION=Skipped one or more source files\n\
   because the specified source folder contains a symbolic or hard link\n\
   that points back to itself or one of its subfolders.\n\
   Make sure all links are outside the source folder and subfolders.\n\
-  Please check a link: {0}
-TF_LOAD_WARN_FILE_ACCESS=Skipped some source file(s) because of access permission \
-  error in the source folder. Please check a file: {0}
+  Invalid link: {0}
+TF_LOAD_WARN_FILE_ACCESS=Skipped one or more source files due to an access permission \
+  error in the source folder. Skipped file(s): {0}
```
miurahr, line 511: The exception is raised when OmegaT visit problematic file, and report the problematic file.
Please assume there is  file A B C and P, and P is problematic. An order of visit is NOT an order of name, depends on OS.

case 1) 
walk  A ,  B,  P,  C  =>  OmegaT read A and B then raise error with file P.

case 2)
walk A, P, B, C => OmegaT read A then raise error with file P.

case 3)
walk P A B C => OmegaT read nothing and raise error with file P.

when error message said "Skipped file(s): P"   it is incorrect, because OmegaT actually skip file A B C in case 3, and OmegaT has no way  to know which files are skipped. OmegaT only knows files which are already read.

miurahr (changes requested):
The change proposed is incorrect against Java behavior.

miurahr:
No adjustment or update.

miurahr:
@Kazephil  progress?

miurahr:
There is no progress for review comment.

miurahr:
Please feel free to reopen when ready.

brandelune:
@Kazephil let me know if you want me to take over.

src/org/omegat/Bundle.properties
Kazephil: This one had slipped under my radar.

So, in your examples, the `{0}` in the error message would expanded with the name of file P?

In that case:
~~Skipped files(s):~~ --> Unable to access: {0}

If there is there is more than one problematic file (files P1 and P2), will OmegaT pick up both or will it stop at the first problematic file it encounters?

Kazephil:
Sorry, this one flew under the radar and slipped my mind!

See the new reply under the comment in the code.

src/org/omegat/Bundle.properties
Kazephil: Thanks for the explanation.

I had a better idea for the message and just pushed a change. Let me know if that works.

miurahr (approved):
lGTM.

PR #697, opened 2023-07-09
Create CODEOWNERS file

### LINKED ISSUE #696: Automatically add reviewers when certain part of the code changes
> It would be better if modifications that impact Bundle.properties could automatically trigger a UI review and a documentation ticket.

_Originally posted by @brandelune in https://github.com/omegat-org/omegat/issues/675#issuecomment-1627602631_

---

This can be done by adding a [CODEOWNERS](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners) in the `.github` directory.

This way, if a PR modifies `Bundle.properties`, Philippe is automatically asked to review the PR.

briacp:
@Kazephil is responsible for the English UI strings, and @kosivantsov for the translations (but that should be handled in the https://github.com/OmegaT-L10N/ repos)

Maybe we can add @brandelune for the `docs/`?

close #696

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,4 @@
+# Code owners for the OmegaT project
+
+/src/org/omegat/Bundle.properties    @kazephil
+/src/org/omegat/Bundle_*.properties  @kosivantsov
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,4 +1,3 @@
 # Code owners for the OmegaT project
 
-/src/org/omegat/Bundle.properties    @kazephil
-/src/org/omegat/Bundle_*.properties  @kosivantsov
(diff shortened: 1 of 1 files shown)


brandelune:
I don't think it is necessary to assign @kosivantsov to the language versions of the Bundle, since he is basically the one who puts them there and there is little to review once they are committed.

But it would be useful to have him along with @Kazephil (and me) on the Bundle so as to know when new strings have been added.

briacp:
> I don't think it is necessary to assign @kosivantsov to the language versions of the Bundle, since he is basically the one who puts them there and there is little to review once they are committed.

Fair enough, it was more of a fail safe in case someone updated directly translated strings, as they wouldn't be included in the TMX of the l10n project.

> But it would be useful to have him along with @Kazephil (and me) on the Bundle so as to know when new strings have been added.

OK

PR #711, opened 2023-07-25
[BUGS#1206] fix: google2 MT to ignore model property

miurahr:
Google translate v2 api may return response with a model field. OmegaT google2 MT connector has a definition that miss the model field.

## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?

- Google2 Translate MT connector failed to parse response
    - https://sourceforge.net/p/omegat/bugs/1206/

Dev ML post
https://sourceforge.net/p/omegat/mailman/omegat-development/thread/f5c71c67-4cef-0044-735d-f469240aaa8e%40marcprior.de/#msg37874311

## What does this PR change?

- define field "model" in parser

### DIFF
--- a/machinetranslators/google/src/main/java/org/omegat/core/machinetranslators/Google2Translate.java
+++ b/machinetranslators/google/src/main/java/org/omegat/core/machinetranslators/Google2Translate.java
@@ -36,6 +36,7 @@ import java.util.TreeMap;
 
 import javax.swing.JCheckBox;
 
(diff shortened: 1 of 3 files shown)


miurahr:
It should be back-ported to 6.0.

miurahr:
From API document,
https://cloud.google.com/translate/docs/reference/rest/v2/translate?hl=en

there are just three fields;  `detectedLanguage`, `translatedText` and `model`.
It may be better clearly to define  all fields, not just ignore.

PR #715, opened 2023-07-28
[BUGS#1207] fix: catch OOM during auto saving

miurahr:
Update SaveThread class

- Solve BUGS#1207

## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?

- Critical errors are logged, but no longer shown to the user
  * https://sourceforge.net/p/omegat/bugs/1207/

## What does this PR change?

- Catch OOM during auto saving, and exit for emergencies
- Catch network error and leave a message-

### DIFF
--- a/src/org/omegat/Bundle.properties
+++ b/src/org/omegat/Bundle.properties
@@ -2332,7 +2332,12 @@ PROJECT_W_GLOSSARY=No writable glossary file in the glossary folder.
 PROJECT_LOCKED=Project locked. Perhaps it is already opened in another OmegaT instance
 PROJECT_INVALID=The specified folder is not an OmegaT project, or the project file is broken, or there is no valid backup.
 
-#Team work
+# Auto Saving exceptions
+AUTOSAVE_TIMEOUT_TAKING_LOCK=Lock trying timeout during autosave
+AUTOSAVE_GENERIC_ERROR=Got error during autosaving: {0}
+AUTOSAVE_INTERRUPTED=Save thread interrupted: {0}
+
+# Team work
 SVN_START=SVN '{0}' execution start
 SVN_FINISH=SVN '{0}' execution finished successfully
 SVN_ERROR=SVN '{0}' error: {1}
--- a/src/org/omegat/Bundle.properties
+++ b/src/org/omegat/Bundle.properties
@@ -2334,8 +2334,7 @@ PROJECT_INVALID=The specified folder is not an OmegaT project, or the project fi
 
 # Auto Saving exceptions
 AUTOSAVE_TIMEOUT_TAKING_LOCK=Lock trying timeout during autosave
-AUTOSAVE_GENERIC_ERROR=Got error during autosaving: {0}
-AUTOSAVE_INTERRUPTED=Save thread interrupted: {0}
+AUTOSAVE_GENERIC_ERROR=Error while autosaving: {0}
 
 # Team work
 SVN_START=SVN '{0}' execution start
--- a/src/org/omegat/Bundle.properties
+++ b/src/org/omegat/Bundle.properties
@@ -2333,7 +2333,7 @@ PROJECT_LOCKED=Project locked. Perhaps it is already opened in another OmegaT in
 PROJECT_INVALID=The specified folder is not an OmegaT project, or the project file is broken, or there is no valid backup.
 
 # Auto Saving exceptions
-AUTOSAVE_TIMEOUT_TAKING_LOCK=Lock trying timeout during autosave
+AUTOSAVE_LOCK_ACQUISITION_TIMEOUT=Unable to get exclusive lock for autosave.
 AUTOSAVE_GENERIC_ERROR=Error while autosaving: {0}
 
 # Team work
--- a/src/org/omegat/core/threads/SaveThread.java
+++ b/src/org/omegat/core/threads/SaveThread.java
@@ -74,12 +74,12 @@ public class SaveThread extends Thread implements IAutoSave {
     }
 
     public synchronized void disable() {
-        Log.logDebug(LOGGER, "Disable autosave"); // NOI18N
+        LOGGER.fine("Disable autosave");
         enabled = false;
     }
 
     public synchronized void enable() {
-        Log.logDebug(LOGGER, "Enable autosave"); // NOI18N
+        LOGGER.fine("Enable autosave");
         enabled = true;
         needToSaveNow = false;
         notify();
@@ -99,7 +99,7 @@ public class SaveThread extends Thread implements IAutoSave {
                 if (needToSaveNow && enabled) {
                     // Wait finished by time and autosaving enabled.
                     IProject dataEngine = Core.getProject();
-                    Log.logDebug(LOGGER, "Start project save from SaveThread"); // NOI18N
+                    LOGGER.fine("Start project save from SaveThread");
                     try {
                         Core.executeExclusively(false, () -> {
                             dataEngine.saveProject(false);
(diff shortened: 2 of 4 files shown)


miurahr:
In core team discussion in 2014, we leave FINE and FINER message in English, non-localized, but all others should be localized.
https://sourceforge.net/p/omegat/mailman/omegat-dev-tech/thread/CAHQuqxF5o3_wmjf781m3Ss4bDJRva965%2BvM-Veh7sLHq--1GoQ%40mail.gmail.com/#msg32928899

src/org/omegat/Bundle.properties
```diff
@@ -2258,7 +2258,7 @@ SCW_SCRIPT_DONE=done ({0} ms elapsed)
 SCW_SCRIPT_CANCELED=canceled ({0} ms elapsed)
 SCW_SCRIPT_ERROR=An error occurred
 SCW_SCRIPT_LOAD_ERROR=An error occurred when loading the script: {0}
-SCW_LIST_ENGINES=Available script engines:
+SCW_LIST_ENGINES=Available script enginesLock trying timeout during autosave:
```
miurahr, line 2261:
```suggestion
SCW_LIST_ENGINES=Available script engines:
```

miss operation

miurahr (commented):
Here is typo.

src/org/omegat/util/Log.java
```diff
@@ -324,4 +323,65 @@ public static void logDebug(Logger logger, String message, Object... parameters)
...
+     * Writes a warning message through a specified logger.
+     * 
+     * @param logger
+     *            logger object
+     * @param key
+     *            The key of the error message in the resource bundle
+     * @param parameters
+     *            Parameters for the error message.
+     */
+    public static void logWarningRB(Logger logger, String key, Object... parameters) {
```
miurahr, line 337: Are there any good ways to  abstract a logger without referencing `java.util.logging.Logger` ?

src/org/omegat/Bundle.properties
```diff
@@ -2332,7 +2332,12 @@ PROJECT_W_GLOSSARY=No writable glossary file in the glossary folder.
 PROJECT_LOCKED=Project locked. Perhaps it is already opened in another OmegaT instance
 PROJECT_INVALID=The specified folder is not an OmegaT project, or the project file is broken, or there is no valid backup.
 
-#Team work
+# Auto Saving exceptions
+AUTOSAVE_TIMEOUT_TAKING_LOCK=Lock trying timeout during autosave
+AUTOSAVE_GENERIC_ERROR=Got error during autosaving: {0}
+AUTOSAVE_INTERRUPTED=Save thread interrupted: {0}
+
```
Kazephil, line 2339: Questions and suggestions:

- `AUTOSAVE_TIMEOUT_TAKING_LOCK`
   What does "lock trying timeout" mean? What situation is it meant to describe?

- `AUTOSAVE_GENERIC_ERROR`
   --> "Error while autosaving: {0}" is a better phrasing, I think.

- `AUTOSAVE_INTERRUPTED`
   If this is a user facing message, it should just be "Autosaving interrupted: {0}"
   If it's only a log message for developers, it can remain as is.

src/org/omegat/core/threads/SaveThread.java
```diff
@@ -89,16 +90,16 @@ public void run() {
         try {
             while (true) {
                 synchronized (this) {
-                    // Set flag for saving. If somebody will reset time, he will
-                    // clear this flag also.
+                    // Set the flag for saving. If somebody resets the timer,
+                    // one should clear the flag.
                     needToSaveNow = true;
```
Kazephil, line 95: I don't want to get overly picky about the wording of comments, but:

`Set the flag for saving. Clear the flag if the timer is reset.`

would probably be clearer the next time a developer needs to look at this part of the code.

src/org/omegat/Bundle.properties
miurahr: > AUTOSAVE_TIMEOUT_TAKING_LOCK
> What does "lock trying timeout" mean? What situation is it meant to describe?

I don't know. It is as same as text in source code.
All other messages are also as same as original.

miurahr: Original author is @alex73  
https://github.com/omegat-org/omegat/commit/dfa8097ac66df2e9b12d8acfca2aeced35511aa5

What's mean, @alex73  ?

miurahr: These message is raised on a code chunk here.

```
                   try {
                        Core.executeExclusively(false, () -> {
                            dataEngine.saveProject(false);
                            dataEngine.teamSyncPrepare();
                        });
                        Core.getMainWindow().showStatusMessageRB("ST_PROJECT_AUTOSAVED",
                                DateFormat.getTimeInstance(DateFormat.SHORT).format(new Date()));
                    } catch (TimeoutException ex) {
                        Log.logWarningRB("AUTOSAVE_TIMEOUT_TAKING_LOCK");
                    } catch (KnownException ex) {
                        Core.getMainWindow().showStatusMessageRB(ex.getMessage(), ex.getParams());
                    } catch (IRemoteRepository2.NetworkException ex) {
                        Log.logWarningRB("TEAM_NETWORK_ERROR", ex.getMessage());
                    } catch (OutOfMemoryError oome) {

                    } catch (Exception ex) {
                        Log.logWarningRB("AUTOSAVE_GENERIC_ERROR", ex.getMessage());
                    }
```

- `AUTOSAVE_TIMEOUT_TAKING_LOCK` is shown when `saveProject`  throws `TimeoutException` that may be rarely happened when user start saving manually and it had not finished before timeout duration.

miurahr: I've checked `Core.executeExclusively`  function  that is called  in the code chunk;

```
public static void executeExclusively(boolean waitForUnlock, RunnableWithException run)
            throws Exception {
        if (!EXCLUSIVE_RUN_LOCK.tryLock(waitForUnlock ? 180000 : 1, TimeUnit.MILLISECONDS)) {
            Exception ex = new TimeoutException("Timeout waiting for previous exclusive execution");
            Exception cause = new Exception("Previous exclusive execution");
            if (runningStackTrace != null) {
                cause.setStackTrace(runningStackTrace);
                ex.initCause(cause);
            }
            throw ex;
        }
```

The timeout exception will be thrown when multiple critical code are running and can not get `EXCLUSIVE_RUN_LOCK`.

Kazephil: Yes, I noticed you simply moved the original hardcoded messages to `Bundle.properties`, which means they can be localized. It's probably still a good idea to improve on the phrasing when we can, especially when the original message was written by a non-native speaker doing their best.

(No slight on their efforts, which are highly appreciated. It's just very difficult to write in a non-native language.)

And speaking of localization, is keeping the FINE and FINER messages non-localized still relevant and appropriate? 

I don't know if @alex73 is still around. Does looking at the code give you an idea of what it accomplishes?

Some situations I can imagine (wild guesses):

If the autosave function locks the project temporarily while saving:
- Situation: The autosave function failed because its attempt lock the project timed out.

If it's the timeout function that gets locked:
- Situation: The attempt to lock the timeout function failed before the autosave completed.

Something else?

Kazephil: > The timeout exception will be thrown when multiple critical code are running and can not get `EXCLUSIVE_RUN_LOCK`.

Looks like we ended up typing at the same time.

Thank you for investigating. I'll take a closer look later this evening.

Kazephil:
I've looked at your explanation of your code investigation more closely. Based on that, if I understand everything correctly, I would propose the following:

- Key name: AUTOSAVE_LOCK_ACQUISITION_TIMEOUT
- Message: Unable to get exclusive lock for autosave.

What do you think?

Kazephil (approved):
As far as the changes to `Bundle.properties` go, this looks good to me.

I can't really comment on the changes to the code itself, though, so you might want to ask @t-cordonnier or @briacp to take a look at those parts.

miurahr:
Change against logger part will be conflict with #717  so I'd like revert these.

PR #716, opened 2023-07-28
refactor: logging API

miurahr:
Related #715 

## Pull request type

- Other (describe below)

Refactor logging API and prepare for improvement of logging backend

## Which ticket is resolved?

- Title ex. fix...
  * URL ex. https://...
 
- Title
  * URL

Dev ML
https://sourceforge.net/p/omegat/mailman/message/37876949/

## What does this PR change?

- Introduce `o.o.u.logging.Logger`
- fix out of index error when custom logging configuration given
- make `o.o.u.Log` as a thin proxy of `o.o.u.l.Logger` instantiated with name "global"
- unit test
- fix usage of `Log.getLogFileLocation`  that result is not used
- rewrite `Log.logDebug` with new `Logger#logDebug` that remove dependency for JDK1.4 logging

### DIFF
--- a/src/org/omegat/util/logging/Logger.java
+++ b/src/org/omegat/util/logging/Logger.java
@@ -234,23 +234,6 @@ public final class Logger {
         }
     }
 
-    /**
-     * Writes an informational message.
-     *
-     * @param key
-     *            The key of the error message in the resource bundle
-     * @param parameters
-     *            Parameters for the error message.
-     */
-    public void log(String key, Object... parameters) {
-        if (logger.isLoggable(Level.INFO)) {
-            LogRecord rec = new LogRecord(Level.INFO, key);
-            rec.setResourceBundle(OStrings.getResourceBundle());
-            rec.setParameters(parameters);
-            rec.setLoggerName(logger.getName());
-            logger.log(rec);
-        }
-    }
 
     /**
      * Writes a throwable message.
--- a/src/org/omegat/util/logging/Logger.java
+++ b/src/org/omegat/util/logging/Logger.java
@@ -1,0 +1,333 @@
+/*
+ *  OmegaT - Computer Assisted Translation (CAT) tool
+ *           with fuzzy matching, translation memory, keyword search,
+ *           glossaries, and translation leveraging into updated projects.
+ *
+ *  Copyright (C) 2023 Hiroshi Miura
+ *                Home page: https://www.omegat.org/
+ *                Support center: https://omegat.org/support
+ *
+ *  This file is part of OmegaT.
+ *
+ *  OmegaT is free software: you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation, either version 3 of the License, or
+ *  (at your option) any later version.
+ *
+ *  OmegaT is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU General Public License for more details.
+ *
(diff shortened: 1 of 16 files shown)


miurahr:
1. commit 67c1693d3a45b890bec7b852f7708b8173c805b2 introduce `o.o.u.logging.Logger` class that abstract JUL(java.util.Logger). 
2. commit d708dba update `o.o.u.Log` class to be a "global" logger using `o.o.u.l.Logger` and avoid code duplication.
3.  commit 8da3123  update all `Log.logDebug` to rewrite with `o.o.u.l.Logger` to avoid direct dependency for JUL.

miurahr:
I've tried  to port `o.o.u.l.Logger` to use SLF4J instead of JUL.

https://github.com/omegat-org/omegat/pull/717/commits/10fd77e2072a36627a82e096ef75ec5764597156

src/org/omegat/util/logging/Logger.java
```diff
@@ -0,0 +1,333 @@
...
+        if (!loaded) {
+            // load built-in logger settings
+            try (InputStream in = Log.class.getResourceAsStream("/org/omegat/logger.properties")) {
+                init(in);
+            } catch (IOException ex) {
+                logger.log(Level.SEVERE, "Can't open file for logging", ex);
+            }
+        }
+    }
+
```
miurahr, line 72: We should load "logger.properties" only once.

src/org/omegat/util/logging/Logger.java
```diff
@@ -0,0 +1,336 @@
...
+     */
+    public void log(String key, Object... parameters) {
+        if (logger.isLoggable(Level.INFO)) {
+            LogRecord rec = new LogRecord(Level.INFO, key);
+            rec.setResourceBundle(OStrings.getResourceBundle());
+            rec.setParameters(parameters);
+            rec.setLoggerName(logger.getName());
+            logger.log(rec);
+        }
+    }
```
miurahr, line 253:
```suggestion
```

Unnecessary method.

t-cordonnier (commented):
For the moment, I don't see what the class org.omegat.util.logging.Logger gives better than org.omegat.util.Log
You did a try to port it to SLF4J but finally it is not in the last version of your branch
Was it impossible to port org.omegat.util.Log to SLF4J?
(eventually forgetting the logDebug method as a first step, and considering it only after)

miurahr:
> (that said, not sure it will require less changes to migrate from o.o.u.Log to o.o.u.l.Logger than directly to slf4j)

There are two custom requirement and one compatibility issue to port logging abstraction layer.

1. an API to get a log file path to provide  Help>Log... dialog
2. provide logDebug that give class/thread name in log file.
3. JUL template use "{0}" form, but SLF4J is "{ }". there need to template function.
 
> Now what happened with commit [10fd77e](https://github.com/omegat-org/omegat/commit/10fd77e2072a36627a82e096ef75ec5764597156)?

o.o.u.l.Logger is a thin wrapper of org.slf4j.Logger,  and give convenient methods
 such as `logErrorRB` and `getLogFileName`
It also has a private utility method to support "{0}" style from Bundle `getMessage`

miurahr:
> (that said, not sure it will require less changes to migrate from o.o.u.Log to o.o.u.l.Logger than directly to slf4j)

I've tried at first to migrate directly to slf4j. It requires overhaul of all `Log.*` calls everywhere.
So I take a strategy to use a proxy design pattern to hide under logging layer from code space and replace it with slf4j.
If we can make  `o.o.u.Log` with slf4j directly, it will be better.

miurahr:
I've tried to migrate directly SLF4J without making intermediate o.o.u.l.Logger class.
You can see.
https://github.com/omegat-org/omegat/pull/717

PR #719, opened 2023-08-01
fix(resx): avoid NPE and add test cases

miurahr:
Improve ResX unit test with "translate" test case.

## Pull request type

- Other (describe below)

QA

## Which ticket is resolved?

- Title
  * URL

## What does this PR change?

- There is no feature change.
- Compare whether output is identical with original as in XML.

### DIFF
--- a/src/org/omegat/filters3/xml/resx/ResXFilter.java
+++ b/src/org/omegat/filters3/xml/resx/ResXFilter.java
@@ -67,25 +67,26 @@ public class ResXFilter extends XMLFilter {
     }
 
     /**
-     * The default list of filter instances that this filter class has. One filter class may have different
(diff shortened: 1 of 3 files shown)


src/org/omegat/filters3/xml/resx/ResXFilter.java
```diff
@@ -120,7 +121,7 @@ public void tagEnd(String path) {
         if ("/root/data/comment".equals(path)) {
             comment = text;
         } else if ("/root/data".equals(path)) {
-            if (entryParseCallback != null) {
+            if (entryParseCallback != null && entryText !=null) {
```
miurahr, line 124:
```suggestion
            if (entryParseCallback != null && entryText != null) {
```

to be better code style.

PR #726, opened 2023-08-22
BUG#1213 - Ensure file order consistency between platforms

briacp:
## Which ticket is resolved?

- Selection of match for pre-translation varies through different OS's
  * https://sourceforge.net/p/omegat/bugs/1213/

## What does this PR change?

- When TMX are loaded, the order is determined by `java.io.File.listFiles()` which can return files in a directory in a different order depending on the underlying OS. This fix sort the resulting files by name, case-insensitive.

### DIFF
--- a/src/org/omegat/util/DirectoryMonitor.java
+++ b/src/org/omegat/util/DirectoryMonitor.java
@@ -143,10 +143,8 @@ public class DirectoryMonitor extends Thread {
         }
 
         // find new files
-        List<File> foundFiles = FileUtil.findFiles(dir, pathname -> true);
-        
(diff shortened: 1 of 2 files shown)


miurahr:
@briacp is it feasible to extend  `FileUtil.findFile` method to take a compareter?

briacp:
> @briacp is it feasible to extend `FileUtil.findFile` method to take a compareter?

Great idea, thanks. It's done in 9f4a06adf1feff85353ed7c4c7254c2ce8a3dc21

miurahr (approved):

PR #727, opened 2023-08-27
[BUGS#1194] fix: allow empty custom tag

miurahr:
Fix the bug OmegaT does not allow disable tag regex "\\d+"

## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?

Bugs: https://sourceforge.net/p/omegat/bugs/1194

## What does this PR change?

- Add special configuration value `"__EMPTY__"` to express ""
- When remove tag value in preference, OmegaT store it as `"__EMPTY__"`


miurahr:
Any comments?
@t-cordonnier

miurahr:
It may be better to use a regex that should not match anything.

`"empty^"` is one not match.

miurahr:
It is better to have a preference option to disable custom tag.

PR #728, opened 2023-08-27
[BUGS#1090] fix: improve error message.

miurahr:
Update error message to suggest users to fix their TMX using validation tool.

## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?

- Fails to load TM using <ph> tags
- https://sourceforge.net/p/omegat/bugs/1090

## What does this PR change?

- Update an error message in Bundle


Kazephil (commented):
I've proposed a reworded error message.

I removed the "before trying again" part because I think it's obvious that the memory won't load until the errors are fixed.

PR #645, opened 2023-06-20
[BUGS#1164] Fix reloading TMX after team synchronization

briacp:
## Which ticket is resolved?

-  Translation memories are not correctly reloaded after team synchronization
  * https://sourceforge.net/p/omegat/bugs/1164

## What does this PR change?

- The TMX content refresh is placed inside the `saveAndRebase`callback to make sure all the TMX data is present.

PR #730, opened 2023-08-29
[BUGS#1164] Fix properly reloading TMX after team synchronization

miurahr:
PR #645 , that try to resolve BUGS#1164,  breaks team feature so integration test become failure.
So it is revertd in PR #643 .

## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?

- Translation memories are not correctly reloaded after team synchronization
- https://sourceforge.net/p/omegat/bugs/1164

## What does this PR change?

- Extend `RebaseAndCommit.IRebase` interface to have `reload` method
- `RebaseAndCommit#rebaseAndCommit` properly call `reload` with target file when it is ready to load 
- Improve integration-test container configuration

### DIFF
--- a/src/org/omegat/core/data/RealProject.java
+++ b/src/org/omegat/core/data/RealProject.java
@@ -1039,16 +1039,18 @@ public class RealProject implements IProject {
                         }
 
                         @Override
-                        public void rebaseAndSave(File out) throws Exception {
+                        public void rebaseAndSave(File tempOut) throws Exception {
+                            // rebase-merge and immediately save to
+                            // the temporary TMX
                             mergeTMX(baseTMX, headTMX, commitDetails);
+                            projectTMX.exportTMX(config, tempOut, false, false, true);
+                        }
 
+                        @Override
+                        public void reload(File file) throws Exception {
                             ProjectTMX newTMX = new ProjectTMX(config.getSourceLanguage(), config.getTargetLanguage(),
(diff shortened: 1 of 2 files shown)


miurahr:
10 minutes integration test passed.

test-integration/src/org/omegat/core/data/TestTeamIntegrationChild.java
```diff
@@ -699,51 +704,6 @@ boolean checkMergeInput(ProjectTMX base, ProjectTMX other) {
                     && base.alternatives.keySet().stream().allMatch(other.alternatives::containsKey);
         }
 
-        protected void mergeTMXOld(ProjectTMX baseTMX, ProjectTMX headTMX) {
```
miurahr, line 702: We should not change not  relevant thing.

briacp:
Thanks a lot for your work Hiroshi, I was getting a bit desperate on this bug.

miurahr:
@briacp I am wondering whether the change is really resolve the bug BUGS#1164 reported.
RealProject.java line 1158
`        projectTMX.replaceContent(mergedTMX);
`
 can be a root cause of the bug.` mergedTMX `may not have auto/enforced entries.

miurahr:
The change here may lead a new bug and will be  fixed in https://github.com/omegat-org/omegat/pull/733

PR #736, opened 2023-09-02
[RFE#1713] feat(filter/bundle): make "don't translate NOI18N comment" feature optional

miurahr:
Implement "TODO" commented feature of resource bundle filter.

## Pull request type

## Which ticket is resolved?

- #1713 ResourceBundle filter: "Don't translate NOI18N comment" feature optional
- https://sourceforge.net/p/omegat/feature-requests/1713/

## What does this PR change?

- make "don't translate NOI18N comment" feature optional
- improve unicode escape
    - output standard Capital hex for escape
    - improve string manipulation performance


miurahr:
@Kazephil  could you leave a review progress note or approvement?

miurahr:
There is no objection or comment other than a message change. I would like to merge it now.

PR #737, opened 2023-09-03
feat(filter/bundle): change default not to do a Unicode escape (java 9-21 standard)

miurahr:
- not to force escape Unicode
- Java 9 and later change the default of resource bundle charset to UTF-8, so there is no need to escape it now. 

## Pull request type

## Which ticket is resolved?

- #1716 ResourceBundle Filter: default not to escape Unicode character
- https://sourceforge.net/p/omegat/feature-requests/1716/

## What does this PR change?

- Change option default


miurahr:
@kosivantsov  @Kazephil  @brandelune 
Could you tell me a review progress or change an approve status?

brandelune (approved):

Kazephil (approved):

miurahr:
Thank you for approvals.

PR #740, opened 2023-09-06
[BUGS#1185] Saving does not register pre-translation in new projects

t-cordonnier:
Actually if you create a project containing something in tm/auto, the segments are auto-populated but when you click on SAVE OmegaT does consider the project as non-modified and saves nothing, despite the fact that importing segments against empty file is a modification.

## Pull request type

Bug fix -> [bug]

## Which ticket is resolved?

https://sourceforge.net/p/omegat/bugs/1185/
https://sourceforge.net/p/omegat/bugs/1200/

## What does this PR change?

Beginning of tm/auto import a new boolean didAnyChange is false. If any segmennt has been imported it is set to true
At the end of the process, if didAnyChange = true then project.isModified = true

### DIFF
--- a/release/changes.txt
+++ b/release/changes.txt
@@ -2,7 +2,7 @@
  OmegaT 6.1.0
 ----------------------------------------------------------------------
   22 Enhancement
-  30 Bug fixes
+  31 Bug fixes
 ----------------------------------------------------------------------
 6.1.0 vs 6.0.0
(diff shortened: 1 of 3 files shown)


t-cordonnier:
> @t-cordonnier Please check SpotBugs error and fix the problem.

Done

PR #743, opened 2023-09-08
refactor: avoid script engine initialization in static context

miurahr:
## Pull request type

refactor

## What does this PR change?

- refactor Scripting code to avoid initialization of scripting engine instance in static context
-

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -265,7 +265,7 @@ dependencies {
         runtimeOnly 'org.apache.ivy:ivy:2.5.1'
 
         // Javascript used for scripts
-        implementation 'org.openjdk.nashorn:nashorn-core:15.4'
+        runtimeOnly 'org.openjdk.nashorn:nashorn-core:15.4'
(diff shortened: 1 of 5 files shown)


build.gradle
```diff
@@ -265,7 +265,7 @@ dependencies {
         runtimeOnly 'org.apache.ivy:ivy:2.5.1'
 
         // Javascript used for scripts
-        implementation 'org.openjdk.nashorn:nashorn-core:15.4'
+        runtimeOnly 'org.openjdk.nashorn:nashorn-core:15.4'
```
miurahr, line 268: Required for compilation of deprecated `JsonParser` class on Java 17 and later.

miurahr: Because OmegaT 6.1 is complied with Java 11, so it can be `runtimeOnly`.
There is a pending discussion for future version when we remove deprecated `o.o.u.JsonParser` class.

PR #763, opened 2023-10-09
docs(devel): add guideline for welcoming beginners

miurahr:
There are many issue tickets and bug reports in source forge issue manager and development mailing list from beginners or not well-experienced users.

Many of them does not have enough information to reproduce the issue and identify a target source code. To reply and support these post consumes limited times and motivations of contributors.
To help these members, I'd like to add a guidelines and text snippet to reply easily.

## Pull request type


brandelune (approved):
Excellent idea. Thank you.

I'm also thinking of adding a section in the user manual with precise steps to produce the log file and the version information.

PR #764, opened 2023-10-09
[BUGS#1146] ui: TagProcessingOptionsPanel to TextArea

miurahr:
There are two text field in TagProcessingOptionsPanel. This change these to textarea.

## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?

 
- #1146 Custom tag field does not resize based on contents
- Bugs: https://sourceforge.net/p/omegat/bugs/1146/

## What does this PR change?

-
-
-

### DIFF
--- a/src/org/omegat/gui/preferences/view/TagProcessingOptionsPanel.form
+++ b/src/org/omegat/gui/preferences/view/TagProcessingOptionsPanel.form
@@ -21,10 +21,11 @@
     <AuxValue name="FormSettings_generateFQN" type="java.lang.Boolean" value="true"/>
     <AuxValue name="FormSettings_generateMnemonicsCode" type="java.lang.Boolean" value="true"/>
     <AuxValue name="FormSettings_i18nAutoMode" type="java.lang.Boolean" value="false"/>
+    <AuxValue name="FormSettings_layoutCodeTarget" type="java.lang.Integer" value="1"/>
     <AuxValue name="FormSettings_listenerGenerationStyle" type="java.lang.Integer" value="0"/>
     <AuxValue name="FormSettings_variablesLocal" type="java.lang.Boolean" value="false"/>
     <AuxValue name="FormSettings_variablesModifier" type="java.lang.Integer" value="2"/>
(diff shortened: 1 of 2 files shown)


miurahr:
There is a different request to solve similar issue that regex is longer than the field.

The request "BUGS#1146 Custom tag field does not resize based on contents" asks size of text field.

[RFE#1445 Multiple-value input/editor for custom tags](https://sourceforge.net/p/omegat/feature-requests/1445/) propose  changing custom tag fields to be a table.

brandelune (changes requested):
The text area needs to wrap otherwise the result is similar to what we had before: the need to scroll to the right to see the end of the expression.

miurahr:
> The text area needs to wrap otherwise the result is similar to what we had before: the need to scroll to the right to see the end of the expression.

Any change suggestions?

brandelune:
> > The text area needs to wrap otherwise the result is similar to what we had before: the need to scroll to the right to see the end of the expression.
> 
> Any change suggestions?

Yes:

> > The text area needs to wrap

miurahr:
> > > The text area needs to wrap otherwise the result is similar to what we had before: the need to scroll to the right to see the end of the expression.
> > 
> > 
> > Any change suggestions?
> 
> Yes:
> 
> > > The text area needs to wrap

I do not know how to do it.

miurahr:
There is similar ticket that request for custom tag edit.
https://sourceforge.net/p/omegat/feature-requests/1445/

brandelune:
> There is similar ticket that request for custom tag edit. https://sourceforge.net/p/omegat/feature-requests/1445/

I saw it. It's interesting.

Right now the function works as expected. I think that the priority should go to have "project specific" custom tags, before having all the nice refinements for editing.

miurahr:
> I think that the priority should go to have "project specific" custom tags, before having all the nice refinements for editing.

There is a big gap that OmegaT have never support project specific `preferences` yet,  because segmentation(SRX) and filter(filters.xml) has its specific config files. 

To implement the feature "project specific" custom tags, it is necessary to make big changes in OmegaT core of preferences management subsystem.

brandelune:
> To implement the feature "project specific" custom tags, it is necessary to make big changes in OmegaT core of preferences management subsystem.

I think we can do it more smartly, by using what OmegaT already provides.

OmegaT provides user with the ability to set a configuration folder on the command line. But using the command line is not trivial for users.

So, if we had a dialog in the project properties that has "use project specific settings", OmegaT would simply use a project specific configuration folder.

We can discuss that on dev because it is slightly off topic here.

miurahr:
@brandelune when you realize what you want, please raise PR.

PR #769, opened 2023-10-14
feat: selector credentials per host/repo

miurahr:
- Add checkbox in user/pass dialog to ask credentials for per-host or per-repository

## Pull request type

## Which ticket is resolved?

- RFE#1719  Give a selection of perHost/PerRespo credentials treatment
- https://sourceforge.net/p/omegat/feature-requests/1719/

## What does this PR change?

- Refactor save/loadCredentials methods
- When per-repo selected
   - Store "scheme://host(:port)/path" as a key

### DIFF
--- a/src/org/omegat/core/team2/impl/GITCredentialsProvider.java
+++ b/src/org/omegat/core/team2/impl/GITCredentialsProvider.java
@@ -389,6 +389,7 @@ public class GITCredentialsProvider extends CredentialsProvider {
             if (userPassDialog.getReturnStatus() == UserPassDialog.RET_OK) {
                 credentials.username = userPassDialog.getUsername();
                 credentials.password = userPassDialog.getPassword();
+                credentials.perHost = userPassDialog.isPerHost();
                 return credentials;
             }
         }
@@ -412,6 +413,8 @@ public class GITCredentialsProvider extends CredentialsProvider {
             }
             char[] pass = console.readPassword(OStrings.getString("TEAM_PASS_FIRST", uri.getHumanishName()));
             credentials.password = Arrays.toString(pass);
+            String yesNo = console.readLine(OStrings.getString("TEAM_CREDENTIALS_PER_HOST_CUI"));
+            credentials.perHost = "yes".equalsIgnoreCase(yesNo) || "y".equalsIgnoreCase(yesNo);
             return credentials;
         }
         Log.log("No console found.");
--- a/src/org/omegat/core/team2/impl/GITCredentialsProvider.java
+++ b/src/org/omegat/core/team2/impl/GITCredentialsProvider.java
@@ -47,13 +47,12 @@ import org.eclipse.jgit.transport.URIish;
 
 import org.omegat.core.Core;
 import org.omegat.core.KnownException;
-import org.omegat.core.team2.TeamSettings;
 import org.omegat.core.team2.gui.PassphraseDialog;
 import org.omegat.core.team2.gui.UserPassDialog;
+import org.omegat.core.team2.impl.TeamUtils.Credentials;
 import org.omegat.gui.main.IMainWindow;
 import org.omegat.util.Log;
 import org.omegat.util.OStrings;
-import org.omegat.util.StringUtil;
 
 /**
  * Git repository credentials provider. One credential provider created for all
@@ -83,9 +82,6 @@ import org.omegat.util.StringUtil;
 
 public class GITCredentialsProvider extends CredentialsProvider {
 
-    private static final String KEY_USERNAME_SUFFIX = "username";
-    private static final String KEY_PASSWORD_SUFFIX = "password";
-    private static final String KEY_FINGERPRINT_SUFFIX = "fingerprint";
     private static final Pattern[] fingerPrintRegex = new Pattern[] {
             Pattern.compile("The authenticity of host '" + /* host */ ".*" + "' can't be established\\.\\n" +
             /* key_type */ "(RSA|DSA|ECDSA|EDDSA)" + " key fingerprint is " +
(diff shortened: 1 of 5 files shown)


src/org/omegat/core/team2/impl/GITCredentialsProvider.java
```diff
@@ -144,57 +140,11 @@ public void setPredefinedCredentials(String url, String predefinedUser, String p
...
-    }
-
-    private void saveFingerprint(URIish uri, String fingerprint) {
-        String url = uri.toString();
-        try {
-            TeamSettings.set(url + "!" + KEY_FINGERPRINT_SUFFIX, fingerprint);
-        } catch (Exception e) {
-            Core.getMainWindow().displayErrorRB(e, "TEAM_ERROR_SAVE_CREDENTIALS", null, "TF_ERROR");
-        }
+        TeamUtils.saveCredentials(uri.toString(), uri.getScheme(), uri.getHost(), uri.getPath(), uri.getPort(), credentials);
```
miurahr, line 147:
```suggestion
        TeamUtils.saveCredentials(uri.getRawPath(), uri.getScheme(), uri.getHost(), uri.getPath(), uri.getPort(), credentials);
```

src/org/omegat/core/team2/impl/TeamUtils.java
```diff
@@ -28,16 +28,97 @@
...
+
+    public static void saveCredentials(String url, String scheme, String host, String path, int port,
+                                       Credentials credentials) {
+        String key;
+        if (scheme == null) {
+            key = path;
+        } else if (credentials.perHost && host != null) {
+            key = scheme + "://" + host + (port != -1 ? ":" + port : "");
+        } else {
+            key = url; // uri.getRawPath();
```
miurahr, line 98:
```suggestion
            key = url;
```

t-cordonnier (approved):
I agree. When Manuel asked me to implement sharing I thought he would mostly use GitHub which shares the password. But later he came with samples where the password differs from one repo to another even in same host so I had the same idea but failed to make him understand. Good that you finally did it.

miurahr:
Waiting review

- [ ]  message
- [ ]  UI

brandelune:
Could you add a screenshot?

miurahr:
> Could you add a screenshot?

Do not you have a capablity to run it? Here is a a way to produce the feature dialog
1. create new test repository on github
2. create access token that is limited to the repository
3. create team project with omegat team init command
4. start omegat using  ./gradlew run command
5. launch File > download team project > enter https protocol type of the repository url 
6. show the dialog and enter access token
7. put check mark on per-repository checkbox

miurahr:
image

miurahr:
I'd like to make more changes for UI and logics.

miurahr:
image

brandelune:
> > Could you add a screenshot?
> 
> Do not you have a capablity to run it? Here is a a way to produce the feature dialog
> 
> 1. create new test repository on github
> 2. create access token that is limited to the repository
> 3. create team project with omegat team init command
> 4. start omegat using  ./gradlew run command
> 5. launch File > download team project > enter https protocol type of the repository url
> 6. should show the dialog, and enter access token
> 7. put check mark on per-host/per-repository checkbox
> 
> When you already have per-host access token, step 6 may not be shown

I have the capability to run it, but very limited time.

If you add a screenshot from when you build/run the thing, it makes it easier for me to have an opinion on what I can understand (UI/UX) and I don't have to take time to deal with things that I don't understand (code).

brandelune:
> image

@Kazephil: "Apply to all repositories on the host" ?

t-cordonnier:
Maybe useful to display the host after the message?
"Applies to repositories on the host https://github.com/"
"Applies to repositories on https://github.com/"

Kazephil:
Does this
- only apply to https access,
- only apply to git repositories,
- apply to any host (not just GitHub)?

I'll take a closer look tomorrow. I'm typing this on mobile from the train back from a long work trip.

t-cordonnier:
It also applies to SVN and not only to GitHub. "the host" means the one of the current repository, the part before the 3rd /

miurahr:
image

brandelune:
> image

@Kazephil : "Enter credentials for ..." ?

Kazephil:
Two more general comments based on this discussion in general:

1. For GitHub, the "password" to enter in the dialog is actually the **_access token_**.
   I wonder if we should modify the label to something like _Password (access token):_, of if a note in the manual is enough.
   Is the access token approach common in services other than GitHub?

2. Should we create a set "test team projects" repositories under the omegat.org GitHub organization so reviewers can build branches like these and test them more easily?
   Like Jean-Christophe, I have limited time, and having to set up a dummy test project (with specific features) to test a team project-related branch can make the difference between taking a look quickly and holding up my review for a couple of days before I can make time for it.

miurahr:
> Two more general comments based on this discussion in general:
> 
> 1. For GitHub, the "password" to enter in the dialog is actually the **_access token_**.
>    I wonder if we should modify the label to something like _Password (access token):_, of if a note in the manual is enough.
>    Is the access token approach common in services other than GitHub?

Services which 2FA/MFA enabled requires provide personal access token because service should not permit only with password in 2FA/MFA rule. Because Git https protocol does not support 2FA/MFA, it requires access token.

Github, Gitlab, SourceHut, Codeberg etc use access token.
 
In other word, services which is in-secure, and no support for 2FA, may keep using password.  
SourceForge use password.

brandelune:
Github (I don't know for others) is clear that we need to use the token instead of the password. So let's keep password and let's have a general word on the issue in the manual.

t-cordonnier:
When as a developer I use the git command line and push something in GitHub repositories, the UI requires my "ssh pass phrase".  I did not check whenever it uses the same terminology for repositories which are not in Github
In any case they don't seem to use the term "access token". I admit it was a source of confusion for me when they migrated to this solution but not sure I would like to change the term, especially if there are sill other providers which use passwords.

Kazephil:
> When as a developer I use the git command line and push something in GitHub repositories, the UI requires my "ssh pass phrase". I did not check whenever it uses the same terminology for repositories which are not in Github In any case they don't seem to use the term "access token". I admit it was a source of confusion for me when they migrated to this solution but not sure I would like to change the term, especially if there are sill other providers which use passwords.

Access token is for **_https_** access, passphrase is for **_ssh_** access.

Hiroshi mentioned this was specifically for **_https_** access. I'm not sure how—or even if—individual repository access works with **_ssh_**.

miurahr:
You can find  GUI parts `PassphraseDialog` to ask  passphrase 
https://github.com/omegat-org/omegat/blob/master/src/org/omegat/core/team2/gui/PassphraseDialog.java

t-cordonnier:
> Access token is for **_https_** access, passphrase is for **_ssh_** access.
> 
> Hiroshi mentioned this was specifically for **_https_** access. I'm not sure how—or even if—individual repository access works with **_ssh_**.

Good remark but I use **_https_** to access Github, if I want to use ssh I need a pair of keys.

t-cordonnier:
> In other word, services which is in-secure, and no support for 2FA, may keep using password. SourceForge use password.

Do you know whenver there is a way to detect which kind of authentication protocol is used? I doubt we will be able to use a specific terminology for each provider if it is not the case.
GitHub, if I correctly remember, accepted both passwords and access tokens until 2021. If we create a provider-specific message depending on our knowledge rather than on a detection algorithm, the risk is that it becomes obsolete before we think about changing it.

miurahr:
When user enter a ssh host string such as `git@github.com:/foo/boo` OmegaT detect it and use `passphrase` dialog.
https://github.com/omegat-org/omegat/blob/master/src/org/omegat/core/team2/impl/GITCredentialsProvider.java#L224

Kazephil (approved):
I've now had the chance to test this branch. Looks good!

PR #770, opened 2023-10-14
CI: enforce checkstyle as default

miurahr:
Enforce checkstyle as default and mark several checks as warning.

To enable it, also doing refactoring to solve checktyle errors.

Once we  enfoce the check, all Pull-Requests to be checked before merge.

## Pull request type

- Other (describe below)
    - refactoring

## What does this PR change?

- Even default to be error for checkstyle, several checks are to be warning
    - Line length check
    - Constant name rule  like  `CONSTANT_NAME`
    - Member name rule like `memberName`
        - there are several old fashion like `m_variable` suppressed
    -  TypeName
    - Operator wrap;  wrap line before operator such as `+`
    - Avoid nested blocks
    - hidden fields;  don't name method parameter as same as fields
    - Todo Comment:  PR with TODO comment generally considered as not finished
  
- Many classes are targeted to suppress error with checkstyle
    - tests
    - generated classes which are generated by NetBeans GUI editor
    - script features; there are so many violations
    - Searcher.java; so many violations
    - machine translations connectors share same violation type
    - Stax filters; so many violations


src/org/omegat/core/segmentation/datamodels/MappingRulesModel.java
```diff
@@ -105,6 +106,8 @@ public void setValueAt(Object aValue, int rowIndex, int columnIndex) {
                 fireException(pse);
             }
             break;
+        default:
+            // we have only two columns
```
miurahr, line 110:
```suggestion
            throw new IllegalArgumentException();
```

src/org/omegat/core/segmentation/datamodels/SegmentationRulesModel.java
```diff
@@ -60,8 +60,9 @@ public Object getValueAt(int rowIndex, int columnIndex) {
             return rule.getBeforebreak();
         case 2:
             return rule.getAfterbreak();
+        default:
+            return null;
```
miurahr, line 64:
```suggestion
            // return null if called before ready
            return null;
```

src/org/omegat/core/tagvalidation/ErrorReport.java
```diff
@@ -176,7 +176,8 @@ public static String localizedTagError(TagError error) {
             return OStrings.getString("TAG_ERROR_ORPHANED");
         case UNSPECIFIED:
             return OStrings.getString("TAG_ERROR_UNSPECIFIED");
+        default:
+            return null;
```
miurahr, line 180:
```suggestion
            return "Unknown tag error is occurred!";
```
we can `return null` as same as before, but it may be better to show message that is invalid and continue, rather than NPE.

src/org/omegat/core/tagvalidation/TagValidation.java
```diff
@@ -257,7 +257,9 @@ protected static void inspectOrderedTags(List<Tag> srcTags, List<Tag> locTags,
                 }
                 break;
             case SINGLE:
-                // Ignore
+                break;
```
miurahr, line 260: It is better to leave an original comment
```suggestion
                // Ignore
                break;
```

src/org/omegat/core/tagvalidation/ErrorReport.java
miurahr: Double check with `TagError` enum class. Here is never come condition, so we should put throwing exception.

PR #771, opened 2023-10-14
chore: generate mac distribution without JRE

miurahr:
launcher for mac `appbundler` seems support an environment without bundled JRE.
This PR add a new task to generate distribution zip for mac without JRE.

## Pull request type

- Bug fix
- Refactor

## What does this PR change?

- Introduce `macDistZip`, `macSignedDistZip` and `macInstallDist` tasks
- refactor `build.gradle` script
- Update `info.plist`
    -  set `JVMVersion` to '11+' which is '11' in previous
        - it may break distribution with JRE 17 failed to launch
    - set `JVMRuntime`
        -  `jre.bundle` when   bundles JRE
        - `MinimumJava11` when without JRE


miurahr:
@brandelune  could you try to run `macInstallDist` task and check `build/install/OmegaT-Mac_without_JRE` folder and click `build/install/OmegaT-Mac_without_JRE/MacOS/OmegaT` to launch?

You can check when you installed JRE 11 and 17 and you can run on both JRE on mac.
I hope it also works on JRE 21.

brandelune:
I will.

brandelune:
Screen Shot 2023-10-17 at 23 24 29

Not working.

brandelune:
Screen Shot 2023-10-17 at 23 26 46

When I double click on OmegaT.jar I get this error, but OmegaT launches.

miurahr:
@brandelune 
Could you suggest which parameter in `Info.plist`  is wrong for launch?

JRE bunlded version of OmegaT's Info.plist has

- [ ] JVMRequired  "11+"
- [ ] JVMRuntime "jre.bundle"

I modified following  parameter for disto w/o JRE

- [ ] JVMRuntime      set to "MinimumJava11"

Could you tell me whether  macOS require any other changes, or  a better value?

miurahr:
further information from reviewer?

brandelune:
> @brandelune Could you suggest which parameter in `Info.plist` is wrong for launch?
> 
> JRE bunlded version of OmegaT's Info.plist has
> 
> * [ ]  JVMRequired  "11+"
> * [ ]  JVMRuntime "jre.bundle"
> 
> I modified following parameter for disto w/o JRE
> 
> * [ ]  JVMRuntime      set to "MinimumJava11"
> 
> Could you tell me whether macOS require any other changes, or a better value?

Where did you get the information on the possible values?

miurahr:
> Where did you get the information on the possible values?

ChatGPT tells

> JVMRuntime:
> 
> jdkhome: Specifies the location of the JDK runtime. You would set this if you have a specific JDK bundled with your application.
> default: Indicates that the default system JDK should be used.
>

miurahr:
Another suggestion
> 
> To specify a specific Java runtime version in the `Info.plist` file, you'll need to use the `JVMVersion` key. This key is used to indicate the minimum required version of the Java Runtime Environment (JRE) that your application needs.
> 
> Here's an example of how you would specify a Java runtime version in an `Info.plist` file:
> 
> ```xml
> <key>JVMVersion</key>
> <string>1.8+</string>
> ```
> 
> In this example, the value `1.8+` indicates that the application requires at least Java 8 or a higher version.
> 
> Keep in mind that this is an example and you should replace `1.8+` with the specific version range that your application requires.

brandelune:
I'm looking for official Apple reference regarding this item.

miurahr:
> I'm looking for official Apple reference regarding this item.

good practice.

miurahr:
Here is a official document which had published in (closed) java.net.
https://cs.longwood.edu/java/docs/technotes/guides/jweb/packagingAppsForMac.html

miurahr:
@brandelune do you have any insight in 2 weeks?

brandelune:
> @brandelune do you have any insight in 2 weeks?

No.

miurahr:
I'd like to close here because of no success.

brandelune:
I'm sorry we could not make it work.

PR #775, opened 2023-10-17
[BUGS#1193] fix: allow custom tag empty

miurahr:
## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?

- Regular expression for custom tags
- https://sourceforge.net/p/omegat/bugs/1193/

## What does this PR change?

- Allow empty custom tag pattern
- When not exists a configuration key of custom tag pattern, assume default from PatternsConsts.DEFAULT_CHECK_CUSTOM_PATERN


brandelune (approved):
It looks like it works.

PR #705, opened 2023-07-14
chore(gradle): build manual zip packs

miurahr:
- Add new task `manualPacks` to produce `build/manuals/manual<Lang>.zip`
- Pack manual zips in OmegaT distribution/installation instead of plain files
- Tweak task dependencies to avoid unwanted duplicated run

## Pull request type

- Build and release changes -> [build/release]

## What does this PR change?

- Clean docs/${lang}/*.html, docs/${lang}/images and docs/${lang}/_wh other than instant-start and OmegaT.css
- Update doc_src/build.xml to accept target folder
- Update doc_src/build.xml to add instant-start-pictures to copy images for instant-start

## Other information

Discussed in dev-list
https://sourceforge.net/p/omegat/mailman/omegat-development/thread/xe9mv6M8stttjRgudIj5zEUiGO4UNyIknfxEEJeJIa6eqvTsuXIqt0fge3G5S0avzGovvJcyAKZ7zVyV5s-FxTEG1pWj-qsfFasIeekM3wk%3D%40northside.tokyo/#msg37870550

PR #778, opened 2023-10-20
chore(ci): publish webManual in snapshot folder

miurahr:
Publish snapshot version of manual  when writers update files under `doc_src`.
the manual will be appeared in https://omegat.sourceforge.io/manual-snapshot/

## Pull request type

## What does this PR change?

- Define CI task to build manual HTMLs and publish at web site.
- Spin out PR from #705 that have a same change
- When #705 merged, document writer have a difficulty to check resulted manuals. The DocCI will help writers to check resulted HTML manuals on web browser without running build job.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -39,6 +39,30 @@ variables:
   GRADLE_USER_HOME: '$(Pipeline.Workspace)/.gradle'
 
 stages:
+- stage: DocCI
+  pool:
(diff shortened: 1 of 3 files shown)


brandelune:
Does the task automatically publish to the web ?
It would be nice to have the task create a local version to check things and then a separate task to publish.

miurahr:
You can create local manual  by `./gradlew updateManuals`

PR #779, opened 2023-10-20
[bug #1219] OmegaT does not handle line breaks correctly on Windows

WynnChen:
## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?

- #1219 OmegaT does not handle line breaks correctly on Windows
  * [https://sourceforge.net/p/omegat/bugs/1219/](https://sourceforge.net/p/omegat/bugs/1219/)
 

## What does this PR change?

- Changes line separator from operating system dependent `System.lineSeparator()` to `"\n"`, since TMXWriter2 and TMXReader2 can not handle `"\r"` within Windows-style line break (`"\r\n"`) correctly.
- 
-
-

### DIFF
--- a/src/org/omegat/util/TMXWriter2.java
+++ b/src/org/omegat/util/TMXWriter2.java
@@ -62,7 +62,7 @@ import org.omegat.core.data.ITMXEntry;
  * @author Aaron Madlon-Kay
  */
 public class TMXWriter2 implements AutoCloseable {
-    static String lineSeparator = System.lineSeparator();
+    static String lineSeparator = "\n";
 
     public static final String PROP_ID = "id";
 


miurahr:
See #780

miurahr:
- We need to add CI test on Windows. It is not valid without testing on Windows.
- TMXWriter2/TMXReader2 is not used only for project_save.tmx There should be sub-effect for export feature.
- Root cause of the problem on library woodstox behavior. #780 changes the behavior, but #780 is work in progress to add a provitioning CI test on windows

src/org/omegat/util/TMXWriter2.java
```diff
@@ -62,7 +62,7 @@
  * @author Aaron Madlon-Kay
  */
 public class TMXWriter2 implements AutoCloseable {
-    static String lineSeparator = System.lineSeparator();
```
miurahr, line 65: Could you extend `TMXWtier2` class to configure to use `system` line separator or `unix style` line separator with argument in constructor?  Please keep backward compatibility because plugins may use it.

miurahr (changes requested):

WynnChen:
Since @miurahr has been working on another PR (#780)  for the same bug, and obviously his solution is much better than mine, this PR is closed.

PR #784, opened 2023-10-22
feat: StaX filter produce self-contained empty element

miurahr:
Set Woodstox property to allow self-contained empty element in StaX filters

## Pull request type

### DIFF
--- a/src/org/omegat/filters4/xml/AbstractXmlFilter.java
+++ b/src/org/omegat/filters4/xml/AbstractXmlFilter.java
@@ -24,38 +24,40 @@
  **************************************************************************/
 package org.omegat.filters4.xml;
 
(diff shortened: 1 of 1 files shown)


t-cordonnier:
The idea is good but we must check what DOCX format requires. Even if in theory <x></x> and <x/> are strictly equivalent it sometimes happens that some tools (like Word, or Trados!) do not consider as such. I must do some tests before to validate it (even if strictly speaking I approve)

miurahr:
@t-cordonnier  yes or no?

miurahr:
progress?

t-cordonnier:
I found an example where a document containing something like `<x></x>` was rejected by Trados Studio while accepted by Word.
So in principle I would say yes.
Now, before validating, are you capable of giving me an example where OmegaT produces a different document after your patch than before? If yes, I will test the document against various tools and tell you wherever your change improves compatibility or can make a regression.

PR #791, opened 2023-10-26
fix: add test case of BUGS#1221

miurahr:
## Pull request type

- Bug fix -> [bug]

## What does this PR change?

- Add test case reported in bug ticket 1221
- A problem reported is not reproduced from reported data.
-


miurahr:
No further information given from reporter.

PR #793, opened 2023-10-26
[BUGS#459] fix: check URL before doWikiGet

miurahr:
- Resolved BUGS#459 Print Exception with Invalid URL in "Import From MediaWiki..

## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?

 
- Print Exception with Invalid URL in "Import From MediaWiki..
- https://sourceforge.net/p/omegat/bugs/459/

## What does this PR change?

- Extend HttpConnectionUtils to have checkUrl method
- Refactor JTextPaneLinkifier to move URL_PATTERN to HttpConnectionUtils class

### DIFF
--- a/src/org/omegat/Bundle.properties
+++ b/src/org/omegat/Bundle.properties
@@ -287,6 +287,7 @@ TF_MENU_WIKI_IMPORT=Add MediaWiki Pa&ge...
 TF_WIKI_IMPORT_TITLE=Add MediaWiki Page
 TF_WIKI_IMPORT_PROMPT=URL
 TF_WIKI_IMPORT_URL_ERROR=Provided URL is invalid.
+TF_WIKI_IMPORT_URL_ERROR_TITLE=URL Error
 TF_WIKI_IMPORT_FAILED=Failed to download page
 MAIN_ERROR_File_Import_Failed=Failed to copy files
 
--- a/src/org/omegat/Bundle.properties
+++ b/src/org/omegat/Bundle.properties
@@ -286,6 +286,7 @@ TF_FILE_IMPORT_TITLE=Add Files to Source Folder...
 TF_MENU_WIKI_IMPORT=Add MediaWiki Pa&ge...
 TF_WIKI_IMPORT_TITLE=Add MediaWiki Page
 TF_WIKI_IMPORT_PROMPT=URL
+TF_WIKI_IMPORT_URL_ERROR=Provided URL is invalid.
 TF_WIKI_IMPORT_FAILED=Failed to download page
 MAIN_ERROR_File_Import_Failed=Failed to copy files
 
--- a/src/org/omegat/gui/main/ProjectUICommands.java
+++ b/src/org/omegat/gui/main/ProjectUICommands.java
@@ -1307,6 +1307,11 @@ public final class ProjectUICommands {
             return;
         }
         try {
+            if (!HttpConnectionUtils.checkUrl(remoteUrl)) {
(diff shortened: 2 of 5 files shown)


brandelune:
I don't understand the dialog when the URL is supposed to be invalid
Screen Shot 2023-10-27 at 8 52 23

If that means the user can still use the URL, we should have:

```
The provided URI looks invalid. Do you still want to use it?
[Yes] [No]
```

There is no need for [Cancel].

If the user hits [Yes] and the URL was really invalid, then there should be a dialog like this:

```
The provided URI was invalid.
[OK]
```

src/org/omegat/gui/main/ProjectUICommands.java
```diff
@@ -1307,6 +1307,11 @@ public static void doWikiImport() {
             return;
         }
         try {
+            if (!HttpConnectionUtils.checkUrl(remoteUrl)) {
+                JOptionPane.showConfirmDialog(Core.getMainWindow().getApplicationFrame(),
+                        OStrings.getString("TF_WIKI_IMPORT_URL_ERROR"));
```
miurahr, line 1312:
```suggestion
                JOptionPane.showMessageDialog(Core.getMainWindow().getApplicationFrame(),
                        OStrings.getString("TF_WIKI_IMPORT_URL_ERROR"));
```

miurahr: @brandelune  Is it what you want to say?
https://docs.oracle.com/javase/tutorial/uiswing/components/dialog.html

brandelune: The question is: do we allow users to try an invalid URI or not?

If yes, the current dialog must be adapted and there should be a second dialog that says whether the resource existed or not.

If no, then the dialog you now suggest is what we need.

miurahr: When allowd invalid URL, you can enjoy exception stack trace.

src/org/omegat/Bundle.properties
```diff
@@ -286,6 +286,7 @@ TF_FILE_IMPORT_TITLE=Add Files to Source Folder...
 TF_MENU_WIKI_IMPORT=Add MediaWiki Pa&ge...
 TF_WIKI_IMPORT_TITLE=Add MediaWiki Page
 TF_WIKI_IMPORT_PROMPT=URL
+TF_WIKI_IMPORT_URL_ERROR=Provided URL is invalid.
```
miurahr, line 289:
```suggestion
TF_WIKI_IMPORT_URL_ERROR=Provided URL is invalid.
TF_WIKI_IMPORT_URL_ERROR_TITLE=Url Error
```

src/org/omegat/gui/main/ProjectUICommands.java
brandelune: Ok.

src/org/omegat/Bundle.properties
brandelune: Do we use URL or URI ?

miurahr: Input dialog use "URL" for prompt.

```
TF_MENU_WIKI_IMPORT=Add MediaWiki Pa&ge...
TF_WIKI_IMPORT_TITLE=Add MediaWiki Page
TF_WIKI_IMPORT_PROMPT=URL
```

brandelune: Yes. I thought URI was a better choice for our UI in general but that does not seem to be the case. So let's leave URL.

PR #798, opened 2023-11-08
filter/latex: add reproducible of bugs#743

miurahr:
Refactor filter/latex based on proposal of BUGS#743 and its case reporoducibles.

## Pull request type

## Which ticket is resolved?

- The LaTeX filter should use \item as a delimiter
- https://sourceforge.net/p/omegat/bugs/743/

## What does this PR change?

- refactoring filter/latex code
- Add reproducible as unit test
-


miurahr:
Branch name is wrong.

PR #803, opened 2023-11-16
[RFE#1725] feat: UserPassDialog extra space char

miurahr:
Many translators may paste password with training space character that lead authentication error.
The enhance here to check whether password text has extra space, show warning and disable OK button.

## Pull request type

## Which ticket is resolved?

- #1725 Trim authentication details in team projects
https://sourceforge.net/p/omegat/feature-requests/1725/

## What does this PR change?

- Check validity of username and password
- When there is extra space character, warn it
- Don't manipulate username or password text in application to keep privacy habit


brandelune:
Although an automatic check is nice, what would be better is that the password is displayed in the field so that users can check what they pasted right away.

Using black dots for the password field makes no security sense at all and keeps users from checking what they type/paste.

miurahr:
> Although an automatic check is nice, what would be better is that the password is displayed in the field so that users can check what they pasted right away.
> 
> Using black dots for the password field makes no security sense at all and keeps users from checking what they type/paste.

see Oracle genuine Java manual https://docs.oracle.com/javase/tutorial/uiswing/components/passwordfield.html

brandelune:
> > Although an automatic check is nice, what would be better is that the password is displayed in the field so that users can check what they pasted right away.
> > Using black dots for the password field makes no security sense at all and keeps users from checking what they type/paste.
> 
> see Oracle genuine Java manual https://docs.oracle.com/javase/tutorial/uiswing/components/passwordfield.html

"For security reasons, a password field does not show the characters that the user types. Instead, the field displays a character different from the one typed, such as an asterisk '*'."

-> That makes no security sense. Especially in a copy/paste context on a personal computer.

"As another security precaution, a password field stores its value as an array of characters, rather than as a string."

I understand that you might not want to reimplement that.

miurahr:
Implemented show/hide button

miurahr:
Now all the feedback from reviewer are adjusted.

PR #804, opened 2023-11-16
[BUGS#1220] Support XLIFF as external TM

miurahr:
There is a bug that filters4/Xliff1Filter does not claim as bilingual filter.
 
## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?

- https://sourceforge.net/p/omegat/bugs/1220/
- #1220 Unable to use XLIFF file as TM

## What does this PR change?

- Add test case that failed to load bilingual XLIFF1 file as external TM
- add  `filters4/Xliff1Filter#isBilingual` method.


src/org/omegat/filters4/xml/xliff/Xliff1Filter.java
```diff
@@ -60,6 +60,11 @@ protected final String versionPrefix() {
         return "1."; // can be 1.0, 1.1 or 1.2
     }
 
+    @Override
+    public boolean isBilingual() {
```
t-cordonnier, line 64: You can put it in AbstractXliffFilter instead, I imagine that if later there is an XLIFF 3 it will be bilingual as well

t-cordonnier (approved):
Sorry I thought I had already done it. Seems ok for me

PR #806, opened 2023-11-18
feat: glossary: encode URL in glossary comment and linkify percent escaped URLs with human readable link text

miurahr:
When showing glossary which contains URL in comment, OmegaT decode percent-encoded URL into human readable URL. 

When OmegaT accept URL which contains non-ascii characters, OmegaT encode URL into percent-encoded one, and write to writable glossary TSV file.

## Pull request type

## Which ticket is resolved?

- https://sourceforge.net/p/omegat/bugs/1229

## What does this PR change?

- Show percent escaped URL as a decoded path with Unicode characters
- Introduce HttpsConnectionUtils#decodeHttpsURLs utility method
- Add commons-validator dependency to validate URLs
- Enhance HttpConnectionUtils#checkUrl utility to use commons-validator to validate URL.
- Enhance IGlossaryRenderer.HtmlTarget#append to decode percent-encoded URLs
- Update test cases
- Document writer should suggest users to put URL with encoded form in glossary file.
- Encode URL when adding glossary term with comment that contains URL

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -219,6 +219,7 @@ dependencies {
     } else {
         implementation(libs.commons.io)
         implementation(libs.commons.lang3)
+        implementation(libs.commons.validator)
         api(libs.slf4j.api)
         implementation(libs.slf4j.format.jdk14)
         runtimeOnly(libs.slf4j.jdk14)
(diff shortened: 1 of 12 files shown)


brandelune:
Thank you very much. It seems to work with any type of character.
I tried

https://ja.wikipedia.org/wiki/ドミノ・ピザ
https://fr.wikipedia.org/wiki/Système_solaire

both worked fine.

PR #815, opened 2023-11-30
chore: use hash of commit when git export

miurahr:
Use .gitattribute substitution for git-archival.properties file to determine version of git export archive.

## Pull request type

## What does this PR change?

-
-
-

### DIFF
--- a/.git-archival.properties
+++ b/.git-archival.properties
@@ -1,0 +1,3 @@
+node=$Format:%H$
+date=$Format:%cI$
+describe=$Format:%(describe:tags=true,match=v[0-9]*)$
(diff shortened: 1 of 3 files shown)


miurahr:
When running ` ./gradlew sourceDistZip` we can get a source distribution zip file `build/distributions/OmegaT_6.1.0_Beta_Source.zip`  which has a `Version.properteis`
```
# Version must always be 3 parts; do not abbreviate trailing zeros
version=6.1.0
# Update is deprecated; do not use any value other than zero.
update=0
# Revision is managed by tooling; do not adjust manually
revision=e0c25a71f208d5fe7f8ea454c8c8b8a065464ca2
# Use an empty string for non-beta versions (Standard flavor) and "_Beta" for
# beta versions (Latest flavor)
beta=_Beta
```

It is intended results.

PR #832, opened 2023-12-07
chore: fix jpackage dependency duplication

miurahr:
## Pull request type

## What does this PR change?

- Fix an issue that deb package has unwanted dependency library


miurahr:
There are still multiple versions  of trie4j and dictzip

PR #841, opened 2023-12-13
[BUGS#1235] fix: hardening plugin installer utility class

miurahr:
Hardening plugin installer class

## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?

-  PluginInstaller throws NPE in certain condition
- https://sourceforge.net/p/omegat/bugs/1235/

dev-ml
https://sourceforge.net/p/omegat/mailman/omegat-development/thread/78b810a2-34f6-4416-b28e-6e1dda659f4b%40northside.tokyo/#msg58712660

## What does this PR change?
- Refactor PluginUtils#loadPlugins to register "jarFile!className" URL even when it is main jar file.
- Check nullity of `currentInfo.getUrl()` and skip retrieving of jarFile object when null.
- Check jar file path that is not under installDir system directory, when try to remove it.

### DIFF
--- a/src/org/omegat/Bundle.properties
+++ b/src/org/omegat/Bundle.properties
@@ -2906,12 +2906,18 @@ PREFS_PLUGINS_COL_VERSION=Version
 PREFS_PLUGINS_COL_AUTHOR=Author
 PREFS_PLUGINS_COL_DESCRIPTION=Description
 PREFS_PLUGINS_INSTALL_FROM_DISK=&Install plugin from disk
-PREFS_PLUGINS_TITLE_CONFIRM_INSTALLATION=Install plugin file
-PREFS_PLUGINS_CONFIRM_UPGRADE=Upgrade/override plugin {0}\n\
-   current: {1}\n\
-   new version: {2}
-PREFS_PLUGINS_CONFIRM_INSTALL=Install plugin {0} {1}
-PREFS_PLUGINS_INSTALLATION_FAILED=Plugin installation failed.
+PREFS_PLUGINS_TITLE_CONFIRM_INSTALLATION=Confirmation of a plugin installation
+PREFS_PLUGINS_CONFIRM_UPGRADE=Do you want to upgrade a plugin "{0}"?\n\
+   installed version: {1}\n\
+   specified plugin version: {2}\n\
+   if select CANCEL, the process will be aborted.
+PREFS_PLUGINS_CONFIRM_OVERRIDE=Do you want to override a plugin "{0}"?\n\
+   installed version: {1}\n\
+   specified plugin version: {2}\n\
+   if select CANCEL, the process will be aborted.
+PREFS_PLUGINS_CONFIRM_INSTALL=Do you want to install plugin {0} {1} ?
+PREFS_PLUGINS_INSTALLATION_FAILED=Plugin installation is failed.
+PREFS_PLUGINS_INSTALLATION_SUCCEED=Plugin installation is succeeded. Please restart OmegaT.
 PREFS_PLUGINS_UNKNOWN_ARCHIVE=This file is not a plugin archive.
 
 # PREFERENCES - UPDATES
--- a/src/org/omegat/Bundle.properties
+++ b/src/org/omegat/Bundle.properties
@@ -2136,7 +2136,7 @@ GUI_DICTIONARY_INSTALLER_AVAILABLE=Available languages:
 
 # org.omegat.gui.dialogs.PluginInstallerDialog
 GUI_PLUGIN_INSTALLER_TITLE=Plugin Installer
-GUI_PLUGIN_OPEN=Select a plugin file (*.jar, *.zip)
+GUI_PLUGIN_OPEN=Select a plugin file (*.jar), or a zip archive include a plugin jar (*.zip)
 GUI_PLUGIN_INSTALLER_INSTALL=Install
 
 # matches text area popup menu
--- a/src/org/omegat/Bundle.properties
+++ b/src/org/omegat/Bundle.properties
@@ -2136,7 +2136,7 @@ GUI_DICTIONARY_INSTALLER_AVAILABLE=Available languages:
 
 # org.omegat.gui.dialogs.PluginInstallerDialog
 GUI_PLUGIN_INSTALLER_TITLE=Plugin Installer
-GUI_PLUGIN_OPEN=Select a plugin file (*.jar), or a zip archive include a plugin jar (*.zip)
+GUI_PLUGIN_OPEN=Select a plugin file (*.jar) or a zip archive (*.zip) that contains one or more plugin jar files.
 GUI_PLUGIN_INSTALLER_INSTALL=Install
 
 # matches text area popup menu
--- a/src/org/omegat/Bundle.properties
+++ b/src/org/omegat/Bundle.properties
@@ -2906,18 +2906,15 @@ PREFS_PLUGINS_COL_VERSION=Version
 PREFS_PLUGINS_COL_AUTHOR=Author
 PREFS_PLUGINS_COL_DESCRIPTION=Description
 PREFS_PLUGINS_INSTALL_FROM_DISK=&Install plugin from disk
-PREFS_PLUGINS_TITLE_CONFIRM_INSTALLATION=Confirmation of a plugin installation
-PREFS_PLUGINS_CONFIRM_UPGRADE=Do you want to upgrade a plugin "{0}"?\n\
-   installed version: {1}\n\
-   specified plugin version: {2}\n\
-   if select CANCEL, the process will be aborted.
-PREFS_PLUGINS_CONFIRM_OVERRIDE=Do you want to override a plugin "{0}"?\n\
-   installed version: {1}\n\
-   specified plugin version: {2}\n\
-   if select CANCEL, the process will be aborted.
-PREFS_PLUGINS_CONFIRM_INSTALL=Do you want to install plugin {0} {1} ?
-PREFS_PLUGINS_INSTALLATION_FAILED=Plugin installation is failed.
-PREFS_PLUGINS_INSTALLATION_SUCCEED=Plugin installation is succeeded. Please restart OmegaT.
+PREFS_PLUGINS_TITLE_CONFIRM_INSTALLATION=Install Plugin
+PREFS_PLUGINS_CONFIRM_UPGRADE=Upgrade the {0} plugin?\n\
+   Current version: {1}\n\
+   New version: {2}
+PREFS_PLUGINS_CONFIRM_OVERWRITE=Version {1} is allready installed.\n\
+   Reinstall the {0} plugin?\n\   
+PREFS_PLUGINS_CONFIRM_INSTALL=Install the {0} {1} plugin?
+PREFS_PLUGINS_INSTALLATION_FAILED=Plugin installation failed.
+PREFS_PLUGINS_INSTALLATION_SUCCEED=Plugin installation successful. Rstart OmegaT to activate it.
 PREFS_PLUGINS_UNKNOWN_ARCHIVE=This file is not a plugin archive.
 
 # PREFERENCES - UPDATES
--- a/src/org/omegat/Bundle.properties
+++ b/src/org/omegat/Bundle.properties
@@ -2136,7 +2136,7 @@ GUI_DICTIONARY_INSTALLER_AVAILABLE=Available languages:
 
 # org.omegat.gui.dialogs.PluginInstallerDialog
 GUI_PLUGIN_INSTALLER_TITLE=Plugin Installer
-GUI_PLUGIN_OPEN=Select a plugin file (*.jar) or a zip archive (*.zip) that contains one or more plugin jar files.
+GUI_PLUGIN_OPEN=Select a plugin file (*.jar) or a zip archive (*.zip) that contains a plugin jar file.
 GUI_PLUGIN_INSTALLER_INSTALL=Install
 
 # matches text area popup menu
--- a/src/org/omegat/Bundle.properties
+++ b/src/org/omegat/Bundle.properties
@@ -2917,6 +2917,9 @@ PREFS_PLUGINS_CONFIRM_INSTALL=Install the {0} {1} plugin?
 PREFS_PLUGINS_INSTALLATION_FAILED=Plugin installation failed.
 PREFS_PLUGINS_INSTALLATION_SUCCEED=Plugin installation successful. Restart OmegaT to activate it.
 PREFS_PLUGINS_UNKNOWN_ARCHIVE=This file is not a plugin archive.
+PREFS_PLUGINS_TITLE_NAME=Plugin Name
+PREFS_PLUGINS_TITLE_CURRENT_VERSION=Installed
+PREFS_PLUGINS_TITLE_TARGET_VERSION=Selected
 
 # PREFERENCES - UPDATES
 VERSION_CHECK_UP_TO_DATE=You already have the latest version.
--- a/src/org/omegat/Bundle.properties
+++ b/src/org/omegat/Bundle.properties
@@ -2907,14 +2907,15 @@ PREFS_PLUGINS_COL_AUTHOR=Author
 PREFS_PLUGINS_COL_DESCRIPTION=Description
 PREFS_PLUGINS_INSTALL_FROM_DISK=&Install plugin from disk
 PREFS_PLUGINS_TITLE_CONFIRM_INSTALLATION=Install Plugin
+PREFS_PLUGINS_TITLE_CONFIRM_UPGRADE=Upgrade Plugin
 PREFS_PLUGINS_CONFIRM_UPGRADE=Upgrade the {0} plugin?\n\
    Current version: {1}\n\
    New version: {2}
-PREFS_PLUGINS_CONFIRM_OVERWRITE=Version {1} is allready installed.\n\
-   Reinstall the {0} plugin?\n\   
+PREFS_PLUGINS_CONFIRM_OVERWRITE=Version {1} is already installed.\n\
+   Reinstall the {0} plugin?\n
 PREFS_PLUGINS_CONFIRM_INSTALL=Install the {0} {1} plugin?
 PREFS_PLUGINS_INSTALLATION_FAILED=Plugin installation failed.
-PREFS_PLUGINS_INSTALLATION_SUCCEED=Plugin installation successful. Rstart OmegaT to activate it.
+PREFS_PLUGINS_INSTALLATION_SUCCEED=Plugin installation successful. Restart OmegaT to activate it.
 PREFS_PLUGINS_UNKNOWN_ARCHIVE=This file is not a plugin archive.
 
 # PREFERENCES - UPDATES
--- a/src/org/omegat/MainClassLoader.java
+++ b/src/org/omegat/MainClassLoader.java
@@ -42,20 +42,9 @@ public final class MainClassLoader extends URLClassLoader {
         registerAsParallelCapable();
     }
 
-    private String name;
-
-    /**
-     * Java9 compatible class loader constructor.
-     * @param name name of class loader
-     * @param parent
-     */
-    public MainClassLoader(String name, ClassLoader parent) {
-        this(parent);
-        this.name = name;
-    }
-
     /*
-     * Required when this classloader is used as the system classloader
+     * Java9 compatible class loader constructor.
+     * @param parent system class loader.
      */
     public MainClassLoader(ClassLoader parent) {
         this(new URL[0], parent);
@@ -73,13 +62,12 @@ public final class MainClassLoader extends URLClassLoader {
      * Main class can add jar classpath for plugins in dynamic manner.
      * @param url
      */
-    synchronized void add(URL url) {
+    public synchronized void add(URL url) {
         addURL(url);
     }
 
-    synchronized void addJarToClasspath(String jarName)
-            throws MalformedURLException {
-        URL url = new File(jarName).toURI().toURL();
+    public synchronized void addJarToClasspath(File jarFile) throws MalformedURLException {
+        URL url = jarFile.toURI().toURL();
         addURL(url);
     }
 
--- a/src/org/omegat/MainClassLoader.java
+++ b/src/org/omegat/MainClassLoader.java
@@ -59,14 +59,19 @@ public final class MainClassLoader extends URLClassLoader {
     }
 
     /**
-     * Main class can add jar classpath for plugins in dynamic manner.
-     * @param url
+     * Add URL to classpath.
+     * @param url to added.
      */
-    public synchronized void add(URL url) {
+    public void addJarToClasspath(URL url) {
         addURL(url);
     }
 
-    public synchronized void addJarToClasspath(File jarFile) throws MalformedURLException {
+    /**
+     * Add Jar file into classpath.
+     * @param jarFile JAR file to add to classpath.
+     * @throws MalformedURLException when a malformed File object is passed.
+     */
+    public void addJarToClasspath(File jarFile) throws MalformedURLException {
         URL url = jarFile.toURI().toURL();
         addURL(url);
     }
@@ -86,6 +91,6 @@ public final class MainClassLoader extends URLClassLoader {
      */
     @SuppressWarnings("unused")
     private void appendToClassPathForInstrumentation(String jarfile) throws IOException {
-        add(Paths.get(jarfile).toRealPath().toUri().toURL());
+        addJarToClasspath(Paths.get(jarfile).toRealPath().toUri().toURL());
     }
 }
--- a/src/org/omegat/MainClassLoader.java
+++ b/src/org/omegat/MainClassLoader.java
@@ -91,6 +91,6 @@ public final class MainClassLoader extends URLClassLoader {
      */
     @SuppressWarnings("unused")
     private void appendToClassPathForInstrumentation(String jarfile) throws IOException {
-        addJarToClasspath(Paths.get(jarfile).toRealPath().toUri().toURL());
+        addJarToClasspath(Paths.get(jarfile).toRealPath().toFile());
     }
 }
--- a/src/org/omegat/filters2/master/PluginUtils.java
+++ b/src/org/omegat/filters2/master/PluginUtils.java
@@ -194,12 +194,10 @@ public final class PluginUtils {
                 try (InputStream in = mu.openStream()) {
                     Manifest m = new Manifest(in);
                     if ("org.omegat.Main".equals(m.getMainAttributes().getValue("Main-Class"))) {
-                        // found main manifest - not in development mode
+                        // found a main manifest - not in development mode
                         foundMain = true;
-                        loadFromManifest(m, pluginsClassLoader, null);
-                    } else {
-                        loadFromManifest(m, pluginsClassLoader, mu);
                     }
+                    loadFromManifest(m, pluginsClassLoader, mu);
                     if ("theme".equals(m.getMainAttributes().getValue("Plugin-Category"))) {
                         String target = mu.toString();
                         for (URL url : urlList) {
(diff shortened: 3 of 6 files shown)


src/org/omegat/filters2/master/PluginUtils.java
```diff
@@ -194,12 +194,10 @@ public static void loadPlugins(final Map<String, String> params) {
                 try (InputStream in = mu.openStream()) {
                     Manifest m = new Manifest(in);
                     if ("org.omegat.Main".equals(m.getMainAttributes().getValue("Main-Class"))) {
-                        // found main manifest - not in development mode
+                        // found a main manifest - not in development mode
                         foundMain = true;
-                        loadFromManifest(m, pluginsClassLoader, null);
```
miurahr, line 199: Here is a place we can understand `getUrl()` in `PluginInstaller` can be NULL because we pass `null` here.

miurahr:
I hope this change solve the issue  @brandelune experienced

brandelune:
The install process works (the plugin is copied to the configuration folder / plugins subfolder) but the weird error dialog is still displayed.
Screen Shot 2023-12-12 at 9 28 25

miurahr:
> The install process works (the plugin is copied to the configuration folder / plugins subfolder) but the weird error dialog is still displayed. Screen Shot 2023-12-12 at 9 28 25

This is not error dialog, confirmation dialog. no problem

brandelune:
But the meaning of this dialog is not clear. What is its purpose? What is its meaning?

miurahr:
You override existing plugin. It confirm a replacement, means remove old one and install new one.

miurahr:
FYI @brandelune you examine the dialog when the feature was proposed and  merged in 5.8 development cycle.
https://github.com/omegat-org/omegat/pull/109

brandelune:
> FYI @brandelune you examine the dialog when the feature was proposed and merged in 5.8 development cycle. #109

I don't remember seeing this dialogue. If I had seen it in the context of a plugin install, I would have equally complained about its lack of clarity.

here is what I saw then:

https://github.com/omegat-org/omegat/pull/109#issuecomment-1280002278

It's very different from the new dialog.

In any case, if the plugin version is the same, the message should not be "upgraded".

miurahr:
> In any case, if the plugin version is the same, the message should not be "upgraded".

So it said "upgrade/override"

brandelune:
> > In any case, if the plugin version is the same, the message should not be "upgraded".
> 
> So it said "upgrade/override"

Which is confusing.

There are 3 cases:
- no previous version of a plugin is installed
  the message should be 
  "Version ... of the ... plugin was properly copied to the configuration folder."
- an older version of the plugin is installed, there should be a validation first:
  "OmegaT has detected version ... of the ... plugin. Do you want to replace it with version ...?"
  and when the new version is installed we should have the same message as above.
If OmegaT detected that there was an *older* version, there should be a different message first:

miurahr:
> > > In any case, if the plugin version is the same, the message should not be "upgraded".
> > 
> > 
> > So it said "upgrade/override"
> 
> Which is confusing.
> 
> There are 3 cases:
> 
> * no previous version of a plugin is installed
>   the message should be
>   "Version ... of the ... plugin was properly copied to the configuration folder."

There is no dialog shown.

> * an older version of the plugin is installed, there should be a validation first:
>   "OmegaT has detected version ... of the ... plugin. Do you want to replace it with version ...?"
>   and when the new version is installed we should have the same message as above.
> * the same version of the plugin is installed :
>   "The plugin is already installed. Do you want to reinstall it?"
>   and when the plugin is reinstalled we should have the same bessage as above.

There is no logic to compare versions, because there is no rule in plugin versions. Only user can decide it.

miurahr:
https://github.com/omegat-org/omegat/blob/release/5.7.1/docs_devel/HowToCreateFilterPlugin.md#manifest

You can find the fact in OmegaT V5.7.1 developer document;

> There must be a manifest file that indicates that it is an OmegaT plugin. 
>
> There are two flavors, see below. Omegat 5.3.0 also supports to provide additional information (valid for both flavors) that can be displayed in the UI. You can optionally provide name, **version**, author and description.

**There can be no version string in plugin.**

brandelune:
> > > > In any case, if the plugin version is the same, the message should not be "upgraded".
> > > 
> > > 
> > > So it said "upgrade/override"
> > 
> > 
> > Which is confusing.
> > There are 3 cases:
> > 
> > * no previous version of a plugin is installed
> >   the message should be
> >   "Version ... of the ... plugin was properly copied to the configuration folder."
> 
> There is no dialog shown.

There should be one to indicate that the action ended with success.

> > * an older version of the plugin is installed, there should be a validation first:
> >   "OmegaT has detected version ... of the ... plugin. Do you want to replace it with version ...?"
> >   and when the new version is installed we should have the same message as above.
> > * the same version of the plugin is installed :
> >   "The plugin is already installed. Do you want to reinstall it?"
> >   and when the plugin is reinstalled we should have the same bessage as above.
> 
> There is no logic to compare versions, because there is no rule in plugin versions, that means plugin author can give arbitrary version string. Only user can decide it.

If the version is optional, you can have a different case for plugins that don't have a version.

The point of this discussion is that the current dialogue does not mean anything.

If you don't want to deal with cases, just have one dialogue that covers everything :  "The ... plugin was successfully copied to the configuration folder", and we're done.

miurahr:
> > > > > In any case, if the plugin version is the same, the message should not be "upgraded".
> > > > 
> > > > 
> > > > So it said "upgrade/override"
> > > 
> > > 
> > > Which is confusing.
> > > There are 3 cases:
> > > 
> > > * no previous version of a plugin is installed
> > >   the message should be
> > >   "Version ... of the ... plugin was properly copied to the configuration folder."
> > 
> > 
> > There is no dialog shown.
> 
> There should be one to indicate that the action ended with success.
> 
> > > * an older version of the plugin is installed, there should be a validation first:
> > >   "OmegaT has detected version ... of the ... plugin. Do you want to replace it with version ...?"
> > >   and when the new version is installed we should have the same message as above.
> > > * the same version of the plugin is installed :
> > >   "The plugin is already installed. Do you want to reinstall it?"
> > >   and when the plugin is reinstalled we should have the same bessage as above.
> > 
> > 
> > There is no logic to compare versions, because there is no rule in plugin versions, that means plugin author can give arbitrary version string. Only user can decide it.
> 
> If the version is optional, you can have a different case for plugins that don't have a version.

What do you mean? I don't understand what is "case". I have already show you the fact, OmegT team have instructed developers it is optional. Whether  you want to change a specifications or not, you should keep compatibility.

> The point of this discussion is that the current dialogue does not mean anything.

The dialog is informative purpose. If message is not good, you can improve it.

> If you don't want to deal with cases, just have one dialogue that covers everything : "The ... plugin was successfully copied to the configuration folder", and we're done.

This dialog is shown before copying plugin file in the folder. You request to show some dialog after copying jar file and removing old one.  This is last chance to stop the process before losing old plugin file from users disk.

brandelune:
> This dialog is shown before copying plugin file in the folder.

The user has no idea where she is in the process when that dialog shows up.
This dialogue reads like an error message after the process, not like an alert before the process.

It should be written in plain English with a confirmation question.

src/org/omegat/util/PluginInstaller.java
```diff
@@ -109,17 +107,17 @@ public static boolean install(final File pluginFile) {
...
 
         // confirm installation
-        if (JOptionPane.YES_OPTION == JOptionPane.showConfirmDialog(Core.getMainWindow().getApplicationFrame(),
-                    message,
-                    OStrings.getString("PREFS_PLUGINS_TITLE_CONFIRM_INSTALLATION"),
-                    JOptionPane.OK_CANCEL_OPTION, JOptionPane.ERROR_MESSAGE)) {
+        if (JOptionPane.YES_OPTION == JOptionPane.showConfirmDialog(
+                Core.getMainWindow().getApplicationFrame(), message,
+                OStrings.getString("PREFS_PLUGINS_TITLE_CONFIRM_INSTALLATION"), JOptionPane.OK_CANCEL_OPTION,
+                JOptionPane.ERROR_MESSAGE)) {
```
miurahr, line 113:
```suggestion
                JOptionPane.WARNING_MESSAGE)) {
```

The message is warning not error. Also we should update message

"PREFS_PLUGINS_CONFIRM_UPGRADE"

and 

"PREFS_PLUGINS_CONFIRM_INSTALL"

brandelune: Thank you. @Kazephil could you take a look at that?

Kazephil: Sure, but it might take a few days. (It looks like I won't get a whole lot of time at home this weekend.)

miurahr: also "PREFS_PLUGINS_TITLE_CONFIRM_INSTALLATION"

miurahr:
The message is not a scope of the proposal here.
Are any issues in the fix?
If not, I'd like to merge a logic hardening and wait @Kazephil to send PR for message improvement.

src/org/omegat/util/PluginInstaller.java
miurahr: also "PREFS_PLUGINS_INSTALLATION_FAILED" that dialog will be shown when installation error occurred.

Kazephil:
I've found time to test the plugin installation from disk in both the current `master` branch and this one, using the same ODT Review plugin as Jean-Christophe .

For reference, this is on Arch Linux, where the standard OmegaT installation is in `/usr/share/java/omegat`. I've set up `~/.omegat-test` and `~/.omegat-master` configuration directories, and I start the application with the `--config-dir` option.

With the latest master (using the `~/.omegat-master` configuration directory), the installation fails silently. (The plugin installer seems to be trying to install the plugin to the `plugins` directory in the `/usr/share/java/omegat/plugins` OmegaT installation directory, but fails because it doesn't have write permission for that directory.) The installation does not create a `plugins` directory under `~/.omegat-master`.

With the branch for this PR (using the `~/.omegat-test` configuration directory), a `plugins` directory ***is*** created under `~/.omegat-test` and the plugin seems to be installed correctly. In that respect, the fix seems to be correct in terms of functionality.

One thing I haven't been able to test:
- What happens if a user inadvertently installs different versions of the plugin, one in the application directory (`/usr/share/java/omegat/plugins` in my example) manually, and one under the configuration directory (`~/.omegat-test/plugins` in my example) using the `Install from file` option?

The dialog will need more than just changing the messages. I see several issues that need to be solved:
- The icon definitely makes it look like an _error_ or (strong) _warning_ message, not a _confirmation_ one.
- I don't know if it's a problem in the plugin itself or in the installer logic, but the dialog should show the actual plugin name ("ODT Review") rather than the category (or whatever "GUI extensions" represents).
- Having the same version shown (for plugins that have one) for the current and new versions is very confusing (especially under a phrase that starts with "Upgrade". Wording concerns aside, if these lines are shown, it should only be when the currently installed plugin and the plugin the user is trying to install are different. If the user is installing a plugin for the first time, the dialog should either show no version, or only the version being installed.
- If (and only if) the plugin is already installed, the dialog should add additional information about versions and make it clear that installing the chosen plugin file will overwrite the existing one.
- The dialog should show the user where the plugin will be installed, and probably give the option to change the location.
  (On an unrelated note, the plugins section of the Preferences should probably let the user set the location of the plugin directory (defaulting to the configuration directory) in a way similar to how users can set the location of the scripts directory. The plugin installer could then default to the directory set by the user.

I think some of these issues probably affect the code logic (although perhaps logic for a separate PR) as well, at least to some extent.

I'll handle the message keys in a separate PR, but I think we need to agree on the above first.
My own rough proposal:

- Title bar: "Install plugin" (or better, if possible, "Install _`<name>`_ plugin")
- Icon: Either an information type icon or none.
- Main message: "The _`<name>`_ plugin will be installed in _`<directory name>`_".
  If possible, there should be a `Browse` button to let the user choose a different directory, or a drop down list of available plugins directories recognized by OmegaT. If a system directory is chosen, the user should ideally be asked to enter the administrator password, but that might be handled at the system level rather than by OmegaT itself.
- (Secondary message, if the same plugin already exists in the plugins directory: 
   1. If the versions are different: "This will install version _`<version string for file to install>`_ of _plugin name_ over _`<version string of already installed plugin>`_")
   2. If the version strings are the identical: "This plugin is already installed."
- Buttons: `Install` and `Cancel`

It think it's worth taking the time to make sure the hardening logic and the dialog are in sync before proceeding with merging, even if several small PRs are the best way to implement everything.

@brandelune:
Any more feedback on the overall workflow?

@miurahr:
Which part of the differences in dialog logic fit with the logic for this fix? Which ones can (or should) be a separate PR?

brandelune:
> @brandelune: Any more feedback on the overall workflow?

I think you got everything right.

miurahr:
Here use [JOptionPane](https://docs.oracle.com/javase/8/docs/api/javax/swing/JOptionPane.html) for message dialog, then we have several choices.


Method Name |	Description
----------------------  | -----------------------
showConfirmDialog	| Asks a confirming question, like yes/no/cancel.
showInputDialog	| Prompt for some input.
showMessageDialog	| Tell the user about something that has happened.
showOptionDialog	| The Grand Unification of the above three.

Here we use a `showConfirmDialog` 


## messageType

Defines the style of the message. The Look and Feel manager may lay out the dialog differently depending on this value, and will often provide a default icon. The possible values are:
- ERROR_MESSAGE
- INFORMATION_MESSAGE
- WARNING_MESSAGE
- QUESTION_MESSAGE
- PLAIN_MESSAGE

Here we use ERROR_MESSAGE when installation failed, and  WARNING_MESSAGE before actual copy.

## optionType
Defines the set of option buttons that appear at the bottom of the dialog box:

- DEFAULT_OPTION
- YES_NO_OPTION
- YES_NO_CANCEL_OPTION
- OK_CANCEL_OPTION

When we want to show "install" or something other than "yes/no", we cannot use the method `showConfirmDialog`.

miurahr:
> * I don't know if it's a problem in the plugin itself or in the installer logic, but the dialog should show the actual plugin name ("ODT Review") rather than the category (or whatever "GUI extensions" represents).

We use pluginName for message.  I don't know why you see category.
```
        String pluginName = info.getName();
       String version = info.getVersion();
        // detect current installation
        PluginInformation currentInfo = getInstalledPlugins().getOrDefault(info.getClassName(), null);
        String message;
        if (currentInfo != null) {
            message = StringUtil.format(OStrings.getString("PREFS_PLUGINS_CONFIRM_UPGRADE"), pluginName,
                    currentInfo.getVersion(), version);
        } else {
            message = StringUtil.format(OStrings.getString("PREFS_PLUGINS_CONFIRM_INSTALL"), pluginName,
                    version);
        }
```

miurahr:
I found that jar file loader,  URLClassLoader, does not work on Java 11 (9) and later. `PluginInstaller#parsePluginJarFileManifest` returns wrong information for the jar file specified.

brandelune:
> I found that jar file loader, URLClassLoader, does not work on Java 11 (9) and later. `PluginInstaller#parsePluginJarFileManifest` returns wrong information for the jar file specified.

Thank you for your finding. Let's not forget that 6.0 targets Java 11.

miurahr:
PluginInstaller#parsePluginJarFileManifest method depends on classloader instrument external jar file manifest entries. It is impossible with standard URLClassLoader in Java9 and later, we have already had a custom class loader MainClassLoader class but PluginInstaller don't use it.
This fix is to use MainClassLoader instead of URLClassLoader. We also need to change MainClassLoader#addJarToClasspath method to be public.

miurahr:
All the review feedback are adjusted. I'd like to wait for comments for dialog messages.

Kazephil:
I'll try to test this latest version and propose updated messages in the next day or two.

Kazephil:
I had a little time to test the latest version of this branch.

This is the dialog I see now:

Plugin_installation_confirmation

It no longer looks like an error message, but I don't think everything works as intended yet.

I get the exact same dialog when I try to install a new plugin and when I choose a plugin that is already installed.

If I understand the code in lines 100 to 111 of `PluginInstaller.java` correctly, I should see a different dialog depending on whether or a version of the plugin is already installed.

However, even if I delete the contents of my test configuration folder and try to install the plugin, I still get the upgrade dialog.
I also tried with a plugin I had never installed in my test configuration, just in case information about the plugin having been installed in the past came from elsewhere, but I got the same results.

It looks like the plugin installer does not recognize that the plugin is not installed (at least when installing from a file), and assumes it is an update every time.

Also, it looks like the plugin information is not taken from the plugin's manifest.
As you can see in the screenshot, the dialog shows "GUI extensions" for the name, and "6.1.0" for the version, but neither matches the information in the manifest:

```
Manifest-Version: 1.0
Plugin-Description: Export/Import translations for external reviews. S
 ponsored by California State University, Long Beach.
OmegaT-Plugins: net.briac.omegat.plugin.odtreview.ODTReviewPlugin
Plugin-Category: plugin
Plugin-Author: Briac Pilpré
Plugin-Version: 0.4.0
Plugin-Link: https://github.com/briacp/plugin-odt-review
Created-By: 1.8.0_332 (Temurin)
Plugin-Date: 2023-06-26T10:25:35+0200
Plugin-Name: ODT Review
```

Getting the correct information from the plugin appears to be tricky.

src/org/omegat/util/PluginInstaller.java
```diff
@@ -180,34 +208,39 @@ static Path unpackPlugin(File sourceFile, Path targetPath) throws IOException {
...
+        ClassLoader cl = ClassLoader.getSystemClassLoader();
+        try (URLClassLoader pluginsClassLoader = new URLClassLoader(urls, cl)) {
             for (Enumeration<URL> mlist = pluginsClassLoader.getResources("META-INF/MANIFEST.MF"); mlist
                     .hasMoreElements();) {
                 URL mu = mlist.nextElement();
-                pluginInfo.addAll(parsePluginJarFileManifest(mu));
+                if (Files.isSameFile(pluginJarFile.toPath(),
+                        Paths.get(PluginUtils.getJarFileUrlFromResourceUrl(mu).getFile()))) {
+                    pluginInfo.addAll(parsePluginJarFileManifest(mu));
+                }
```
miurahr, line 228: I've added a check whether parsed `META-INF/MANIFEST.MF` is exactly from the file given.
I don't know why `URLClassLoader` returns resource other than given URL, but it will avoid a problem reported.

src/org/omegat/MainClassLoader.java
```diff
@@ -73,13 +62,12 @@ public MainClassLoader() {
      * Main class can add jar classpath for plugins in dynamic manner.
      * @param url
      */
-    synchronized void add(URL url) {
+    public synchronized void add(URL url) {
```
miurahr, line 65: Should not be `synchronized` to avoid dead-lock. 
```suggestion
    public  void add(URL url) {
```

src/org/omegat/MainClassLoader.java
```diff
@@ -73,13 +62,12 @@ public MainClassLoader() {
...
         addURL(url);
     }
 
-    synchronized void addJarToClasspath(String jarName)
-            throws MalformedURLException {
-        URL url = new File(jarName).toURI().toURL();
+    public synchronized void addJarToClasspath(File jarFile) throws MalformedURLException {
```
miurahr, line 69:
```suggestion
    public void addJarToClasspath(File jarFile) throws MalformedURLException {
```

src/org/omegat/MainClassLoader.java
miurahr: It would be better to change method name to `addJarToClasspath(URL)` as same as  a `addJartoClasspath(String)` and `addJarToClasspath(File)`

Kazephil:
@miurahr 

Thanks for working on this some more.

I'll try to test it sometime this week. (Vacation is over, and I expect to be busy in the first few days back to work.)

miurahr:
I would like to port some improvement to raise robustness of plugin management from a proposed update #906 to here.

Kazephil:
I have finally been able to test the latest version of this branch. The code seems to work as intended now.

If I install a new plugin, I get the following dialog:
Fresh_install

The dialog now shows the correct plugin name.
Excellent! :tada:

I didn't bother to take a screenshot, but I got the dialog with the "Installation successful. Restart OmegaT" message after installing the plugin.

If I try to install the same plugin again, I get a different dialog:
Update_install

Question:
Would it be possible to show different messages when the already installed plugin and the plugin to install:
- have ***different*** version numbers in the `META-INF/MANIFEST.MF` file, and
- have ***identical*** version numbers in the `META-INF/MANIFEST.MF` file.

Ideally, in the example above where the version of the plugin is the same, the dialog should show something like:
```
This version of the plugin is already installed. Reinstall?"
```
When the versions are different, the message would show both versions:
```
Update the ABC plugin?
Current version: 0.x.0
Updated version: 0.x.1
```
(I think the `Cancel` button is sufficiently self-explanatory and that there is no need for the last line about clicking it to abort.)

Also, for good measure, I also created a fake `.jar` (I changed the extension of a shell script to `.jar`) to test for failure, and got this message:
Failed_install

In terms of functionality, it looks pretty good now. I'll work on the messages and dialog titles later this evening.

brandelune:
Thank you @Kazephil for checking. The behavior seems much improved. Thank you @miurahr. I'll let the two of you discuss further details.

Kazephil:
Replying to myself as I start to work on the messages themselves...

> Question: Would it be possible to show different messages when the already installed plugin and the plugin to install:
> 
> * have _**different**_ version numbers in the `META-INF/MANIFEST.MF` file, and
> * have _**identical**_ version numbers in the `META-INF/MANIFEST.MF` file.
> 
> Ideally, in the example above where the version of the plugin is the same, the dialog should show something like:
> 
> ```
> This version of the plugin is already installed. Reinstall?"
> ```

I just realized that maybe the `PREFS_PLUGINS_CONFIRM_OVERRIDE` key may be intended to cover cases where the existing plugin and the plugin to install have the same version. If that's the case, the question is moot, and it's only a matter of changing that message.

(Minor note: it should be "overwrite", not "override", so I'll be changing the key name as well.)

src/org/omegat/Bundle.properties
```diff
@@ -2906,12 +2906,15 @@ PREFS_PLUGINS_COL_VERSION=Version
 PREFS_PLUGINS_COL_AUTHOR=Author
 PREFS_PLUGINS_COL_DESCRIPTION=Description
 PREFS_PLUGINS_INSTALL_FROM_DISK=&Install plugin from disk
-PREFS_PLUGINS_TITLE_CONFIRM_INSTALLATION=Install plugin file
-PREFS_PLUGINS_CONFIRM_UPGRADE=Upgrade/override plugin {0}\n\
-   current: {1}\n\
-   new version: {2}
-PREFS_PLUGINS_CONFIRM_INSTALL=Install plugin {0} {1}
+PREFS_PLUGINS_TITLE_CONFIRM_INSTALLATION=Install Plugin
```
Kazephil, line 2909: I would also like to add a key for an upgrade dialog message:
`PREFS_PLUGINS_TITLE_CONFIRM_UPGRADE=Upgrade Plugin`

I not confident about how to modify `PluginInstaller.java` to implement that change.

I think I would have to add an extra line of code to assign the title above the `message=...` statement in line 103 and line 106 in the current file?

At the current line 103:

```java
title = PREFS_PLUGINS_TITLE_CONFIRM_INSTALLATION
message = StringUtil.format(OStrings.getString("PREFS_PLUGINS_CONFIRM_OVERWRITE"), pluginName,
                        currentInfo.getVersion(), version);
```

At the current line 106:

```java
title = PREFS_PLUGINS_TITLE_CONFIRM_UPGRADE
message = StringUtil.format(OStrings.getString("PREFS_PLUGINS_CONFIRM_UPGRADE"), pluginName,
                        currentInfo.getVersion(), version);
```

And then change line 117 to:
```java
OStrings.getString(title), JOptionPane.OK_CANCEL_OPTION,
```

Am I on the right track?

miurahr: It is better that make line 106  `title = OStrings.getString("PREFS_PLUGINS_TITLE_CONFIRM_UPGRADE")` then line 117 to `title, JOptionPane.OK_CANCEL_OPTION,` because OmegaT project test set checks `OStrings.getString("PREFS_PLUGINS_TITLE_CONFIRM_UPGRADE")` syntax and distill key and check Bundles for missing keys.

Kazephil: OK, thanks!

By the same logic, should line 103 then be `title = OStrings.getString("PREFS_PLUGINS_TITLE_CONFIRM_INSTALLATION")?`

Kazephil:
Thanks to Hiroshi's help, I was able to update the properties and code to get slightly easier to understand dialogs based on context.

I created a fake version 0.4.1 of the plugin and tested again with version 0.4.0 and version 0.4.1.

On a brand new install of 0.4.0:
pluginFreshInstall

And then this confirmation after clicking `OK`:
pluginInstallSuccessful

If I try to upgrade to 0.4.1:
pluginUpgrade

Clicking OK shows the same installation successful message as above.

If I "forget" I've already installed 0.4.1 and try to install it again, I get:
pluginSameVersion

If I try to install a `.jar` file that isn't a plugin, I get:
pluginInstallFailed

All of that looks pretty good to me!

I have two questions left:
1. Line 2919 of `Bundle.properties` has a `PREFS_PLUGINS_UNKNOWN_ARCHIVE=This file is not a plugin archive.` key, but I never see it come up if I try to install a file that isn't a plugin file. It's not shown in the status bar or the log either. When (and where) is it used?
2. The file dialog says you can install a plugin from a `.jar` file or a `.zip` file, but if I try to install a plugin from a `.zip` file, the file dialog just closes abruptly and nothing happens. This behaviour is the same regardless of whether the file is a valid plugin (I tried renaming the ODT plugin to `.zip`) or a file that isn't a valid plugin.

miurahr:
> 1. Line 2919 of `Bundle.properties` has a `PREFS_PLUGINS_UNKNOWN_ARCHIVE=This file is not a plugin archive.` key, but I never see it come up if I try to install a file that isn't a plugin file. It's not shown in the status bar or the log either. When (and where) is it used?

You can test with  a zip file that does not have any jar.

miurahr:
> The file dialog says you can install a plugin from a .jar file or a .zip file, but if I try to install a plugin from a .zip file, the file dialog just closes abruptly and nothing happens. This behaviour is the same regardless of whether the file is a valid plugin (I tried renaming the ODT plugin to .zip) or a file that isn't a valid plugin.

OmegaT can install a plugin from zip file that contains jar file. 
OmegaT cannot handle a zip file renaming a jar file.

miurahr:
@Kazephil I have updated a manual page with warning for a common misunderstanding.

miurahr:
There can be multiple filters in a single plugin jar file. A test case `PluginInstallerTest#testParseLegacyManifest` illustrate how it is realized.  A legacy stax filter which is distributed as 3rd party plugin, contains 6 plugin manifest entries such as `{"Xliff1Filter", "Xliff2Filter", "SdlXliff", "SdlProject", "MqXliff", "MsOfficeFileFilter"}`

Current implementation only  show first plugin name and version in the dialog.
It is better to show all the entries names and versions against current installed version.

When current plugin contains 5 entries and new plugin jar contains 6 entries, it is better to show a difference.

Current dialog just show a comparison of a name and a version, but it can be a table form.

miurahr:
image

miurahr:
image

Kazephil:
> @Kazephil I have updated a manual page with warning for a common misunderstanding.

@miurahr 
Thank you.

I've just pushed some commits to update the wording for the file dialog and the documentation.

I like the table. This is what I get now for a new install:
New_install_(table)

And when I upgrade the plugin (using a `.jar` file inside a `.zip` file):
Upgrade_install_(table)

Looks good to me.
I also did get the "This file is not a plugin" error dialog when I tried to install a fake `.jar` file that wasn't a plugin, so I'm not sure what I did wrong the first time I tried and didn't see that message.

PR #842, opened 2023-12-16
fix: MT execution condition

miurahr:
OmegaT execute MT connector to translate even when auto-fetch is disabled, if some connectors are enabled. This changes logic to be better to understand a condition and respect the configuration.

## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?

Report on azure translator plugin issue
https://github.com/omegat-org/azure-translate-plugin/issues/7

## What does this PR change?

- Refactor MachineTranslateTextArea#getTranaslation
- Don't go to MT#getTranslate in default action.


miurahr:
I recognize that I misunderstand the behavior of the MT logic. I want to close here without merge.

PR #844, opened 2023-12-18
chore: update linux-deb/rpm build

miurahr:
A jpackage plugin author declined to fix the bug on gradle cache feature with my proposal to change, and express the project won't support cache feature.
The plugin has no more benefit other than a syntax sugar, so I decided to remove the plugin and migrate it as Exec task type.

Now we can enable a configuration cache and speed-up our build.

## Pull request type

## What does this PR change?

-
-
-

### DIFF
--- a/gradle/utils.gradle
+++ b/gradle/utils.gradle
@@ -74,14 +74,17 @@ ext {
     }
 
     exePresent = { exe ->
-        ["where $exe", "which $exe"].any {
-            try {
-                def findExe = it.execute()
-                findExe.waitForProcessOutput()
-                return findExe.exitValue() == 0
-            } catch (any) {
-                return false
-            }
+        def osName = System.getProperty('os.name').toLowerCase()
+        if (osName.contains('windows')) {
+            providers.exec {
+                ignoreExitValue = true
+                commandLine('where', exe)
+            }.result.get() == 0
+        } else {
+            providers.exec {
+                ignoreExitValue = true
+                commandLine('which', exe)
+            }.result.get() == 0
         }
     }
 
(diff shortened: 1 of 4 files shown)


Kazephil:
@miurahr 

This commit seems to break the build!

I can't build OmegaT at revision `b28bcb71c` at all.
I tried both the `installDist` and `distZip`, and building with the docs and without using the `-PforceSkipDocumentBuild` option introduced in an earlier commit.

I also tried running the tasks after running the `clean` task, but it still fails.

I get a lot of errors along the lines of:
```
(path)/build.gradle' line: 509

* What went wrong:
Could not evaluate onlyIf predicate for task ':manualHtmlEn'.
> Could not evaluate spec for 'Task satisfies onlyIf closure'.
```
I get a similar message for just about every manual and first step task, with a different line number and a variation on the task name.

Incidentally, everything builds fine at revision `ee2c65a45`.

miurahr:
It is very strange. CI build is fine and successfully done.

miurahr:
@Kazephil  what version of JDK do you use  for running Gradle?

gradle/utils.gradle
```diff
@@ -74,14 +74,17 @@ ext {
...
+            providers.exec {
+                ignoreExitValue = true
+                commandLine('where', exe)
+            }.result.get() == 0
+        } else {
+            providers.exec {
+                ignoreExitValue = true
+                commandLine('which', exe)
+            }.result.get() == 0
         }
```
miurahr, line 88: A change here may change a behavior of  `onlyIf` block that test existence of executables.
It was not recommended for configuration-cache enabled build, so I changed to use recommended way that use `providers.exec` gradle command.

Kazephil:
> @Kazephil what version of JDK do you use for running Gradle?

@miurahr 
I use OpenJDK 11 (set as the JDK for the IDEA project). I just realized that at the command line, my system was set to OpenJDK 21. I set it back to OpenJDK 11, but got the same eror.

I also don't have any installation of Gradle other than the one bundled with OmegaT.

This is on Arch Linux.

miurahr:
I've updated (revert) changes in utils.gradle and turn off the configuration cache  #844 
I hope it fix the issue.

Kazephil:
It did. Everything builds fine now.

Thanks!

PR #845, opened 2023-12-20
chore: Update gradle plugin ben-manes.version@0.50.0

miurahr:
## Pull request type

## What does this PR change?

- Bump gradle plugin com.github.ban-manes.version@0.50.0 
- Support for Gradle 8.x


miurahr:
It seems better to migrate Dependabot out of gradle script, so we can simplify it.

PR #883, opened 2023-12-24
[BUGS#1233] fix: Give a proper feedback when filter got character encoding error

miurahr:
When character encoding error happened,  it is just logging but  no notification.
It is a fix to notify the problem for translators, because resulted translation files can not be completed.

## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?

- java.nio.charset.UnmappableCharacterException: Input length = 1
- Bugs: https://sourceforge.net/p/omegat/bugs/1233/

## What does this PR change?

- Catch UnsupportedEncodingException, and UnmappableCharacterException and MalformedInputException in `FilterMaster#translateFile`
- The latter 2 exceptions are inherited from  CharacterCodingException.
- Raise a dialog to notify translator the problems.

### DIFF
--- a/src/org/omegat/Bundle.properties
+++ b/src/org/omegat/Bundle.properties
@@ -1252,6 +1252,9 @@ FILTERMASTER_ERROR_LOADING_FILTER=Filter '{0}' could not be loaded from '{1}'
 
 FILTERMASTER_ERROR_SRC_TRG_SAME_FILE=Translated file path cannot be identical to the source file path:\n{0}
 
+FILTERMASTER_ERROR_UNKNOWN_ENCODING=Detect a wrong character encoding or wrong character in specified encoding when \
+  output to translation file. Please check your project configuration and translations.
+
 # Filters.java
 
 FILTERS_FILE_FORMAT=File format
--- a/src/org/omegat/filters2/master/FilterMaster.java
+++ b/src/org/omegat/filters2/master/FilterMaster.java
@@ -37,6 +37,8 @@ package org.omegat.filters2.master;
 
 import java.io.File;
 import java.io.IOException;
+import java.io.UnsupportedEncodingException;
+import java.nio.charset.CharacterCodingException;
 import java.nio.charset.Charset;
 import java.text.SimpleDateFormat;
 import java.util.ArrayList;
@@ -55,6 +57,7 @@ import com.fasterxml.jackson.dataformat.xml.ser.ToXmlGenerator;
 import com.fasterxml.jackson.module.jaxb.JaxbAnnotationModule;
 import org.apache.commons.io.FileUtils;
 
+import org.omegat.core.Core;
 import org.omegat.filters2.AbstractFilter;
 import org.omegat.filters2.FilterContext;
 import org.omegat.filters2.IAlignCallback;
@@ -269,6 +272,9 @@ public class FilterMaster {
         IFilter filterObject = lookup.filterObject;
         try {
             filterObject.translateFile(inFile, outFile, lookup.config, fc, translateCallback);
+        } catch (UnsupportedEncodingException | CharacterCodingException ex) {
+            Log.logErrorRB(ex, "FILTERMASTER_ERROR_UNKNOWN_ENCODING");
+            Core.getMainWindow().displayErrorRB(ex, "FILTERMASTER_ERROR_UNKNOWN_ENCODING");
         } catch (Exception ex) {
             Log.log(ex);
         }


miurahr:
Please check a log and warning dialog message.

miurahr:
Any comment?

src/org/omegat/Bundle.properties
brandelune, line 1: @Kazephil 

"OmegaT detected an incorrect encoding when creating the translated files. Please check your project configuration."

-> "wrong character in specified encoding" is equivalent to wrong encoding, since the selected encoding should include the character.

-> "check your translations" in which case should checking the translation  be a requirement?

Kazephil: > -> "wrong character in specified encoding" is equivalent to wrong encoding, since the selected encoding should include the character.

It seems unlikely, but…

…could "wrong character in specified encoding" refer to using a character supported in the font used inside OmegaT, but not in the font and encoding of the output translated file?

@miurahr: Is that what you meant? And if so, is that why you thought you needed the "check your translations" comment?

brandelune: > …could "wrong character in specified encoding" refer to using a character supported in the font used inside OmegaT, but not in the font and encoding of the output translated file?

That would be "wrong character set", wouldn't it? Or glyph missing, etc.

miurahr: The message will be shown when `UnsupportedEncodingException` or `CharacterCodingException` Exception occurred in `processFile` in filter modules, that process both reading source files and writing target files.
 
When a source file is bundle properties file,  and configured as ASCII in bundle-properties filter configuration, and it has characters out of ascii, it easily cause the error.

brandelune: Thank you for the specifics. I think that we can safely summarize that as an encoding problem. Either the encoding covers the characters used, or it does not (regardless of how the user has translated the document).

What do you think?

miurahr: > Thank you for the specifics. I think that we can safely summarize that as an encoding problem. Either the encoding covers the characters used, or it does not (regardless of how the user has translated the document).
> 
> What do you think?

I do not have an answer but some file format that filter supports is standardized in its specification as ASCII, ISO-8859-1 or others, and user can put character out of a range on text editor.
 This is input file problem.

If translator use "emoji" for translation of "text" file which  is made on old Windows and encoding is Windows-1251, then file write will be failed.
 This is output file problem.

Kazephil: @miurahr 
Thanks for the additional details and explanations.

So we need a message that tells the user the problem could be in:
1. the source file,
2. the filter configuration, or
3. the target file

depending on the encoding of the original file, the encoding configured in the filter, or the use of unsupported characters by the translator?

One more question: 

Let's say a file filter supports UTF-8 (for example) in addition to ASCII, ISO-8859-1, Windows-1251, or other encodings with a more limited set of characters.

If the source file is encoded in ASCII, and the translator configures the filter to output in UTF-8, and then uses emoji or other characters outside the ASCII range, will the filter convert the encoding and create the output file correctly?

Or are filters limited to creating output files in the same encoding as the source file?

brandelune: That depends on the filter.

If OmegaT can know which is the issue (the encoding setting was wrong *or* the input characters were not appropriate) then we should separate the messages.

Regarding the error message displayed in the dialogue, we want to keep it simple, while giving explanations in the manual.

Kazephil: Indeed!

If there are a lot of "it depends" and multiple possibilities, I'll come up with a fairly generic "encoding problem found" error message with a note to check the manual for details.

If I can provide a little more information in the message to point the user in the right direction without complicating things, I'd like to do that.

miurahr: The exception is caught at `FilterMaster` where details are not known. Filter should handle the exception cases,  if the filter want to tell user something based on conditions.

src/org/omegat/Bundle.properties
```diff
@@ -1252,6 +1252,9 @@ FILTERMASTER_ERROR_LOADING_FILTER=Filter '{0}' could not be loaded from '{1}'
 
 FILTERMASTER_ERROR_SRC_TRG_SAME_FILE=Translated file path cannot be identical to the source file path:\n{0}
 
+FILTERMASTER_ERROR_UNKNOWN_ENCODING=Detect a wrong character encoding or wrong character in specified encoding when \
+  output to translation file. Please check your project configuration and translations.
+
```
Kazephil, line 1257: How about:
`FILTERMASTER_ERROR_UNKNOWN_ENCODING=Incorrect encoding or unsupported character detected when generating the translated file. Check your source files or filter configuration.`?

brandelune: @miurahr When OmegaT displays the message, does it know the encoding or the character? Maybe displaying the information would be helpful?

miurahr: > The exception is caught at `FilterMaster` where details are not known.

Kazephil: OK. Thanks.

In that case, my suggestion remains unchanged.

Kazephil (approved):

PR #895, opened 2023-12-29
refactor: MainClassLoader as ContextClassLoader

miurahr:
It make `Core.initializeGUI` not to take class loader. It is as same signature as OmegaT 5.7.1, and keep support Java9+.

- Main class - set MainClassLoader in main threads context class loader - always instantiate MainClassLoader as application class loader
- MainClassLoader class - add `appendToClassPath` method - make synchronized `add` method - `addJarToClasspath` use `add` method
- Core class - `initializeGUI` get class loader from thread context
- UIDesignManager - `initialize` get clas loader from thread context
- PluginUtils - fix search order 1. user plugin 2 system plugin 3 module - use class loader given from Main class - remove `getThemePluginJars` method
- TeamTool to call `PluginUtils.loadPlugins` w/o args
- AlignFilePickerContoller.main to call `PluginUtils.loadPlugins` w/o args

## Pull request type

- Other (describe below)
refactor

## What does this PR change?

-
-
-

### DIFF
--- a/aligner/src/main/java/org/omegat/gui/align/AlignFilePickerController.java
+++ b/aligner/src/main/java/org/omegat/gui/align/AlignFilePickerController.java
@@ -462,7 +462,7 @@ public class AlignFilePickerController {
         System.setProperty("apple.laf.useScreenMenuBar", "true");
 
         Preferences.init();
-        PluginUtils.loadPlugins(Collections.emptyMap());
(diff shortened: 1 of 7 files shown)


miurahr:
After leaning things of class loader and java internals in #899, I decided here is not good approach for solution. I'd like to close.

PR #904, opened 2024-01-12
[BUGS#1239] feat: glossary support optional sort by src length

miurahr:
## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?

- Glossary popups behavior changed unintended way
- https://sourceforge.net/p/omegat/bugs/1239/

## What does this PR change?

- Support sort by length of a source text when one contains another.
- It prioritizes a longer match to be shown in tooltips
- Introduce a new preference GLOSSARY_SORT_BY_SRC_LENGTH
- Configurable on glossary pane configuration button
- Configurable on preference dialog
- Update manual to add description of preference -> glossary

## Other information

glossary_sort_order_config

### DIFF
--- a/doc_src/en/OmegaT_Preferences.xml
+++ b/doc_src/en/OmegaT_Preferences.xml
@@ -286,7 +286,24 @@
           displayed on the same line.</para>
         </listitem>
       </varlistentry>
-    </variablelist>
+
+	  <varlistentry id="dialogs.preferences.glossary.sort.by.source.term.length">
+		  <term id="dialogs.preference.glossary.sort.by.source.term.length.title"><option>Sort by source
+			  text term length</option></term>
+		  <listitem>
+			  <para>If some glossary items are similar in a source term text,
+				  and one contains another term text, sort by a length of source text.</para>
+		  </listitem>
+	  </varlistentry>
+
+		<varlistentry id="dialogs.preferences.glossary.sort.by.target.term.length">
+			<term id="dialogs.preference.glossary.sort.by.target.term.length.title"><option>Sort by source
+				text term length</option></term>
+			<listitem>
+				<para>If a glossary item has multiple definitions, sort by a length of definition text.</para>
+			</listitem>
+		</varlistentry>
+	</variablelist>
   </section>
 
   <section id="dialogs.preferences.dictionary">
--- a/src/org/omegat/Bundle.properties
+++ b/src/org/omegat/Bundle.properties
@@ -1886,6 +1886,7 @@ GUI_GLOSSARYWINDOW_insertselection=Insert Selection
 GUI_GLOSSARYWINDOW_SETTINGS_NOTIFICATIONS=Notify Me When a Segment Has Glossary Hits
 
 GUI_GLOSSARYWINDOW_SETTINGS_OPEN_FILE=Open Writable Glossary File
+GUI_GLOSSARYWINDOW_SETTINGS_SORT_BY_SRC_LENGTH=Sort by Source Term Length
 GUI_GLOSSARYWINDOW_SETTINGS_SORT_BY_LENGTH=Sort by Target Term Length
 
 GUI_MATCHWINDOW_SUBWINDOWTITLE_Dictionary=Dictionaries
@@ -2764,6 +2765,8 @@ PREFS_GLOSSARY_STEMMING=Use &stemming
 PREFS_GLOSSARY_REPLACE_ON_INSERT=Replace matches when inserting source text
 PREFS_GLOSSARY_REQUIRE_SIMILAR_CASE=&Ignore matches with very different case (e.g. FOO vs foo)
 PREFS_GLOSSARY_MERGE_ALTERNATE_DEFINITIONS=Merge alternate definitions of the same term
+PREFS_GLOSSARY_SORT_BY_SRC_LENGTH=Sort by Source Term Length
+PREFS_GLOSSARY_SORT_BY_LENGTH=Sort by Target Term Length
 
 PREFS_GLOSSARY_LAYOUT=&Layout:
 GLOSSARY_RENDERER_DEFAULT=Default
--- a/src/org/omegat/gui/glossary/GlossarySearcher.java
+++ b/src/org/omegat/gui/glossary/GlossarySearcher.java
@@ -181,8 +181,7 @@ public class GlossarySearcher {
 
     protected static boolean isCjkMatch(String fullText, String term) {
         // This is a CJK word and our source language is not space-delimited, so
-        // include if
-        // word appears anywhere in source string.
+        // include if word appears anywhere in source string.
         IProject project = Core.getProject();
         return project.isProjectLoaded()
                 && !project.getProjectProperties().getSourceLanguage().isSpaceDelimited()
@@ -191,8 +190,7 @@ public class GlossarySearcher {
 
     private static List<Token[]> getCjkMatchingTokens(String fullText, String term) {
         // This is a CJK word and our source language is not space-delimited, so
-        // include if
-        // word appears anywhere in source string.
+        // include if word appears anywhere in source string.
         IProject project = Core.getProject();
         if (!project.isProjectLoaded()
                 || project.getProjectProperties().getSourceLanguage().isSpaceDelimited()) {
--- a/src/org/omegat/gui/glossary/GlossarySearcher.java
+++ b/src/org/omegat/gui/glossary/GlossarySearcher.java
@@ -4,6 +4,7 @@
           glossaries, and translation leveraging into updated projects.
 
  Copyright (C) 2017 Aaron Madlon-Kay
+               2024 Hiroshi Miura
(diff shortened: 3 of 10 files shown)


src/org/omegat/gui/glossary/GlossarySearcher.java
```diff
@@ -173,18 +180,22 @@ private static boolean keepMatch(Token[] tokens, String srcTxt, String locTxt) {
     }
 
     protected static boolean isCjkMatch(String fullText, String term) {
-        // This is a CJK word and our source language is not space-delimited, so include if
+        // This is a CJK word and our source language is not space-delimited, so
+        // include if
         // word appears anywhere in source string.
```
miurahr, line 185:
```suggestion
        // include if word appears anywhere in source string.
```

src/org/omegat/gui/glossary/GlossarySearcher.java
```diff
@@ -173,18 +180,22 @@ private static boolean keepMatch(Token[] tokens, String srcTxt, String locTxt) {
...
+        return project.isProjectLoaded()
+                && !project.getProjectProperties().getSourceLanguage().isSpaceDelimited()
                 && StringUtil.isCJK(term) && fullText.contains(term);
     }
 
     private static List<Token[]> getCjkMatchingTokens(String fullText, String term) {
-        // This is a CJK word and our source language is not space-delimited, so include if
+        // This is a CJK word and our source language is not space-delimited, so
+        // include if
         // word appears anywhere in source string.
```
miurahr, line 195:
```suggestion
        // include if word appears anywhere in source string.
```

miurahr (commented):
comment style can be improved

miurahr:
This adds a glossary configuration options. There is also a place to improve in preferences dialog and glossary pane configurations on user manuals.

miurahr:
I can not find a place to write text about a glossary pane configuration which I paste a capture to "other information" in manual.

miurahr:
Dear reviewer, please check and give me a suggestion for `doc_src/en/OmegaT_Preferences.xml` and new messages in `src/org/omegat/Bundle.properties`

miurahr:
Dear reviewer, you will understand a behavior change,  a sort order change, by looking unit test expectations.
`test/src/org/omegat/gui/glossary/FindGlossaryThreadTest.java`

A detailed investigation and expectations are  in bug ticket.

src/org/omegat/gui/glossary/GlossarySearcher.java
```diff
@@ -240,14 +251,21 @@ static void sortGlossaryEntries(List<GlossaryEntry> entries) {
             int p1 = o1.getPriority() ? 1 : 2;
             int p2 = o2.getPriority() ? 1 : 2;
             int c = p1 - p2;
+            if (c == 0 && Preferences.isPreferenceDefault(Preferences.GLOSSARY_SORT_BY_SRC_LENGTH, false)
```
miurahr, line 254: From discussion on the issue ticket, we can enable it in default.
```suggestion
            if (c == 0 && Preferences.isPreferenceDefault(Preferences.GLOSSARY_SORT_BY_SRC_LENGTH, true)
```

miurahr:
source_tem-in-tooltip

does this interface acceptable?

source term in tooltip of context menu.

doc_src/en/OmegaT_Preferences.xml
brandelune, line 1: @Kazephil can you propose corrections please?

src/org/omegat/Bundle.properties
brandelune, line 1: If the items are in Preferences > Glossaries, they should be "sentence case", like the other items surrounding them.

doc_src/en/OmegaT_Preferences.xml
Kazephil: Give me a few days to read through the details in the SourceForge ticket and actually try the branch so I can make informed suggestions.

Kazephil:
I have now had time to try this branch with the test project provided by Julian in the SourceForge ticket.

If I choose `Sort by Source Term Length`, I get the same results as those posted by @miurahr above, but based on the name, of the option, I would expect the following order:

- 向上 = enhance
- 結果 = result
- 確実 = reliable
- 上 = up
- 向 = direct
- 果 = achieve
- 実 = fruit
- 確 = surely

Then I thought that maybe it's actually `Sort Related Source Terms by Length` because the individual _kanji_ for each term are neatly grouped under the compound word using those _kanji_ in Hiroshi's screenshot. Then I realized that there was no entry for the "結" in "結果", so I added one to the glossary:
- 結 = tie

And I got the following result:
sortBySourceTermLength

I was expecting `結 = tie` to come either just before or just after `果 = achieve`, but instead it ended up at the end of the list. Then I thought maybe the order in the glossary file has something to do with it, so I edited the glossary file and moved the new entry just below the "achieve" entry, but I got the same result.

Similarly, if I sort by target term, I get this:
sortByTargetTermLength

This is clearly not sorted by target term length in either ascending or descending order, and it also does not group target terms related to the same source term.

I don't know if this is related to the fact that the source terms are in Japanese. I'll try to find time to make a test using French and English later, and see what the results for those are.

One more thing that caught my attention:
bothSortOptionsOn

The sort by term length options are checkboxes, and it's possible to check both at the same time. Is that intentional?
When both options are checked, the target length option seems to simply be ignored (at least, I don't see any difference in the sort order with only sort by source checked and with both checked, but I'll have to try it on a project with a more extensive glossary).

I think these options are mutually exclusive and should be radio buttons rather than checkboxes.

I'll try to do some more trials with different projects and languages over the next few days.

miurahr:
> The sort by term length options are checkboxes, and it's possible to check both at the same time. Is that intentional?

This is intentional.

Please assume there are three glossary definition for single source term "terminal" and definitions "command window" , "railway station",  and "end of line". There are three length is 14, 15 and 11.

When sort by term option is on, we will find "railway station", "command window" and "end of line".

This will happened when there are multiple-definition on one source term.

   please read a comment and source 
```
        // After the matched entries have been tokenized and listed,
        // we reorder entries as
        // 1) by priority
        // 2) by length of source text if one contains another (optional)
        // 3) by alphabet of source term
        // 4) by length of localized term (optional)
        // 5) by alphabet of localized term
        // Then remove the duplicates and combine t
```

When we can sort by alphabet of source term, we don't need to more comparison.

miurahr:
Please read a code. 
```
   static void sortGlossaryEntries(List<GlossaryEntry> entries) {
        entries.sort((o1, o2) -> {
            int p1 = o1.getPriority() ? 1 : 2;
            int p2 = o2.getPriority() ? 1 : 2;
            int c = p1 - p2;
            if (c == 0 && Preferences.isPreferenceDefault(Preferences.GLOSSARY_SORT_BY_SRC_LENGTH, true)
                    && (o2.getSrcText().contains(o1.getSrcText())
                            || o1.getSrcText().contains(o2.getSrcText()))) {
                // longer is better if one contains another
                c = o2.getSrcText().length() - o1.getSrcText().length();
            }
            // sort source text alphabetically, first ignore a case, then
            // consider a case
            if (c == 0) {
                c = o1.getSrcText().compareToIgnoreCase(o2.getSrcText());
            }
            if (c == 0) {
                c = o1.getSrcText().compareTo(o2.getSrcText());
            }
            if (c == 0 && Preferences.isPreferenceDefault(Preferences.GLOSSARY_SORT_BY_LENGTH, false)) {
                c = o2.getLocText().length() - o1.getLocText().length();
            }
            if (c == 0) {
                c = o1.getLocText().compareToIgnoreCase(o2.getLocText());
            }
            return c;
        });
    }
```

Kazephil:
> > The sort by term length options are checkboxes, and it's possible to check both at the same time. Is that intentional?
> 
> This is intentional.
> 
> Please assume there are three glossary definition for single source term "terminal" and definitions "command window" , "railway station", and "end of line". There are three length is 14, 15 and 11.
> 
> When sort by term option is on, we will find "railway station", "command window" and "end of line".

OK. I'll try to test this with a more comprehensive glossary that includes terms with multiple definitions over the next few days.
I'll also take a closer look at the code and its comments at the same time.

If both options can be on at the same time, I think we'll need to at least add a sentence in the Preferences pane to explain that sorting by source term length takes priority over sorting by target term length if both are selected.

The detailed description will go in the manual, of course, but the interface should at least give the user a hint about the priority.
I'll try to come up with concise wording that provides that hint.

miurahr:
Any further review comment or feedback? If no, I would like to merge.

brandelune:
I guess we need to produce documentation for this item. Would you mind creating a tracker item on SF with this PR in reference  and me as owner?

miurahr:
https://sourceforge.net/p/omegat/documentation/420/

miurahr:
Are there any further concern that @brandelune cannot approve the PR? Please take an action to give a feedback or approve.

brandelune:
Please go ahead.

miurahr:
image

Approval is here.

PR #908, opened 2024-01-14
docs_devel: update  contribution manual about  document build  and others

miurahr:
## Pull request type

## What does this PR change?

- Fix URL typo
- Add section about manual build tasks and explanation of container
- Update about documentation project rules and procedures


miurahr:
There are almost a technology side of document build, not about contribution rules and procedures.

A missing part is a build property option to skip html manual build when building and testing features.

Kazephil:
@miurahr
 like the idea of splitting the developer documentation into a contribution guide and a build manual.

There's a lot of material in the developer documentation, so I'm going to wait until I can set aside a good chunk of time to do a proper review, which probably means sometime early in April, when work gets quieter.

Until then, I think we can still push the updates as is (maybe with a DRAFT label) to make information available. (It'll also be easier to review if it's up on the `readthedoc.io` site you set up.)

@brandelune What do you think?

miurahr:
There is very small section about a document contribution rule and procedure 

https://github.com/omegat-org/omegat/blob/d875fb7795f7c1233d63d4afd3f18a3a55912718/docs_devel/docs/40.ContributingDocument.md

docs_devel/docs/40.ContributingDocument.md
```diff
@@ -2,9 +2,26 @@
...
 
-You can see built html pages on `docs/` folder and its source is located at `doc_src/` folder in OmegaT project
-source code repository.
+You can see a document source in`doc_src/` folder in OmegaT project source code repository.
 
 ## Documentation rules and procedures
 
-(TBD)
+Please contact to [Documentation manager](https://omegat.org/about) at [omegat-development mailing list]
+(https://sourceforge.net/projects/omegat/lists/omegat-development) for details.
```
miurahr, line 10:
```suggestion
Please contact to [Documentation manager](https://omegat.org/about) at [omegat-development mailing list](https://sourceforge.net/projects/omegat/lists/omegat-development) for details.
```

brandelune: Why would people need to contact the documentation manager ? People can do as they do for the code. Send PRs after creating issues on SF, on the documentation tracker.

miurahr: Because I cannot find any document to describe rules and procedure in omegat.org website, and also in source tree. Please give a description. Otherwise, I can revert it to previous sentence; "[TBD]".

docs_devel/docs/45.LocalizeApplicationAndManuals.md
```diff
@@ -1,12 +1,13 @@
...
+# Localize applications and manuals
 
 All the localization of OmegaT application, manuals and website are handed on L10N project
 located at `https://github.com/omegat-L10N/<lang>`
 
 OmegaT is a tool to assist translation, L10N project uses OmegaT for the localization.
 
-You need a permission to write the project repository, please ask l10n manager at dev mailing list.
+If You need permission to write the project repository, please ask l10n manager at [omegat-development mailing list]
+(https://sourceforge.net/projects/omegat/lists/omegat-development).
```
miurahr, line 9: There is unwanted extra space character.
```suggestion
If You need permission to write the project repository, please ask localization manager at [omegat-development mailing list](https://sourceforge.net/projects/omegat/lists/omegat-development).
```

docs_devel/docs/45.LocalizeApplicationAndManuals.md
```diff
@@ -1,12 +1,13 @@
...
 
 ## L10N Project rules and Procedures
 
-(TBD)
+Please contact to [Localization manager](https://omegat.org/about).
```
miurahr, line 13:
```suggestion
Please contact to [localization manager](https://omegat.org/about).
```

docs_devel/docs/40.ContributingDocument.md
miurahr: This may be valuable to add from https://omegat.org/support

> documentation issues are handled on SourceForge by the [Documentation tracking system](https://sourceforge.net/p/omegat/documentation/). Please do not use these tracking systems before first raising the issue on the user group and/or mailing lists.

brandelune:
@miurahr @Kazephil I have amended all the documents directly on the branch. There are a few comments (check lines that start with "->").

Thank you very much Hiroshi for this tremendously important work.

docs_devel/docs/07.ManualContainerTaskBuild.md
```diff
@@ -304,7 +248,11 @@ same location as `docbook-xml-4.5`.
        └────/doc_src_paths.xml
 	   └────/docbook-utf8.xsl
 
+
 # Fonts
+
+-> Aren't the fonts installed in the container?
```
Kazephil, line 254: They should be. This would apply to building the documentation without using a container, which requires the user to manually install all font and software dependencies.

docs_devel/docs/07.ManualContainerTaskBuild.md
```diff
@@ -315,6 +263,8 @@ The following fonts must be installed:
 
 # Usage
 
+-> Is this part still relevant?
+
 All the scripts require the language folder as argument and must be run from the
 `doc_src` directory.
```
Kazephil, line 270: I think this one also applies to building manually without a container.
@miurahr Can you confirm this?

docs_devel/docs/40.ContributingDocument.md
```diff
@@ -1,27 +1,24 @@
...
-on web server. You can see snapshot version of HTML manual at https://omegat.sourceforge.io/manual-snapshot/
+## Manual snapshot release
+
+When the document source is modified, the CI/CD processes automatically build the HTML manuals and publish that snapshot
+on the web server at [https://omegat.sourceforge.io/manual-snapshot/](https://omegat.sourceforge.io/manual-snapshot/).
 
 ### License
 
-All the files in `doc_src` folder and below are under the terms of the GNU General
+All the files in `doc_src` folder and below are published under the terms of the GNU General
```
Kazephil, line 20: @brandelune 
I'm not a licence expert, but do software licenses apply to documentation?
Should this be one of the Creative Commons licenses?

docs_devel/docs/93.BuildingInstallerPackage.md
```diff
@@ -132,75 +120,77 @@ OmegaT project prepares tasks to sign to packages for macOS.
...
-with a familiar way to install and launch your applications.
+You can provide platform-specific packages for Linux, macOS and Windows to users that come
+with familiar ways to install and launch the packaged applications.
 
-`jpackage` bundles java runtime which is used to run Gradle or is used to compile a project
-that has `jpackage` utility. Because OmegaT project uses Java 11 to compile, you need to
+`jpackage` bundles the java runtime which is used to run Gradle or the one which is used to compile the project
+that has the `jpackage` utility. Because OmegaT project uses Java 11 to compile, you need to
 run Gradle with JDK17.
+-> Why ???
```
Kazephil, line 171: @miurahr 
Is JDK 17 required by the version of Gradle we use?

docs_devel/docs/40.ContributingDocument.md
brandelune: The GPL is not specifically a software licence. It is a source code distribution licence. Our docbook sources are not supposed to have issues that can only be solved by using documentation specific licences. But if you find such issues, we need to investigate how to fix them, because changing the licence would be a major pain.

docs_devel/docs/07.ManualContainerTaskBuild.md
brandelune: Sure, and if we still have the possibility to make manual build, I don't understand why we have a container requirement.

Kazephil: It should probably be called a container _option_.

In theory, a container makes building the documentation easier than the manual process because it… well… ***contains*** everything!

Something strange seems to be going on with the manual build process, though. I'm getting everything inside a literal `${target}` directory in `doc_src\en\` instead of in the correct `docs\en\` folder if I try to build with Ant. More investigation required.

docs_devel/docs/40.ContributingDocument.md
Kazephil: If XML is considered code, then there's probably no problem because the docbook sources can be defined as "source code".

I'll trust that whoever first chose the licence checked this at the time! :)

brandelune: Originally, the manual was in HTML. There was no build process involved. The whole set was distributed under the GPL, and it just stayed like that. The Emacs manuals use the GFDL 1.3+ which makes sense, if I remember well, because they can also be distributed as books (cf. invariant clauses, etc.)

Kazephil: Interesting.

I found [this interesting thread](https://opensource.stackexchange.com/questions/741/with-gpl-what-does-source-code-mean-for-non-software) that's relevant.

If I understand it correctly, GPL for the docbook sources is fine. The actual manual(s) produced from those sources should probably be under the GFDL or one of the Creative Commons licences.

brandelune: What practical difference would that make? Or rather, what problem is that a solution for?

Kazephil: If I understood the thread (and GNU GPL FAQ) correctly, licensing the actual manual (output HTML, PDF, etc.) under the GPL would require including the contents of `doc_src` with the manual itself, which I think would be (a) cumbersome for us, and (b) tedious for users.

Of course, I might simply be misunderstanding the answers in the thread and FAQ.

brandelune:
> licensing the actual manual (output HTML, PDF, etc.) under the GPL would require including the contents of doc_src with the manual itself, which I think would be (a) cumbersome for us, and (b) tedious for users.

The requirement is only and minimally access to the code upon request. Which we do by default since the code is always available. Just like we do not distribute the OmegaT code with the binaries but keep it available online.

docs_devel/docs/93.BuildingInstallerPackage.md
miurahr: `jpackage` is a JDK feature that is general available in Java 16 and later.
See https://openjdk.org/jeps/392

JEP is [Java Enhancement Proposal](https://en.wikipedia.org/wiki/JDK_Enhancement_Proposal), that is java committer's specification note.

It is why deb/rpm package build requires Java 17 (LTS) or later.

miurahr: I think contribution manual already explain why.
> OmegaT has tasks to generate Linux native packages in "DEB" and "RPM" archive formats.
> We use the `jpackage` packaging tool which is provided by Java Development Kits.
> The tool has been introduced in Java 14. OmegaT project requires JDK 17.

docs_devel/docs/07.ManualContainerTaskBuild.md
miurahr: I moved these clauses from `doc_src/Readme.md`  I do not check these descriptions, it can be obsoleted.

docs_devel/docs/40.ContributingDocument.md
miurahr: FSF recommend GFDL for document rather than GPL in several reason.
https://www.gnu.org/licenses/license-recommendations.html

It is something extra-ordinal to apply GPL to a large manual document like OmegaT manual, so I try to add explanation here.

brandelune: I understand that our choice of licence may not be the best, but we currently do not have issues that need to be solved by a change of licence.

It could be a fun exercise to track down the various authors that share the copyright on the current document though.

Kazephil: Well, the statement in the developer documentation here only applies to the docbook sources, so as far as those go, I think the developer documentation is fine as is in terms of contents.

The broader licensing issue may (or may not) be something to look at further down the road in the larger context of the entire project.
Have we explicitly applied the GPL (or any other licence) to the products output from those sources?

brandelune: The GPL applies to the deliverables (the "program") and forces the developers to provide sources on demand so that users can modify the deliverable. The GPL is not a source file distribution licence.

docs_devel/docs/93.BuildingInstallerPackage.md
Kazephil: OK. So packaging an OmegaT release requires JDK 17, but compiling the application itself only requires JDK 11.

brandelune: That was not clear when I read the document. Would you mind rewording it so that it is clearer?

Kazephil: I just committed a proposed rewording.

docs_devel/docs/93.BuildingInstallerPackage.md
```diff
@@ -129,75 +119,78 @@ OmegaT project prepares tasks to sign to packages for macOS.
...
+`jpackage` bundles the java runtime used to run Gradle, but can also be run with the runtime
+used to compile the project that relies on the `jpackage` utility. Fpr OmegaT, you need to
+run Gradle with JDK 17 to create packages, because the application itself is compiled with
+Java 11, which is not compatible with `jpackage`.
 
-OmegaT use `gradle-jpackage-plugin` that integrate `jpackage` command line utility
-which is bundled in JDK with Gradle build utility.
+OmegaT uses the `gradle-jpackage-plugin`, which integrates the `jpackage` command line utility
+bundled in the JDK into the Gradle build tool.
+-> I do not understand this sentence
```
Kazephil, line 175: > -> I do not understand this sentence
I just took a stab at rewriting it.
@miurahr: Is this what you meant?

docs_devel/docs/93.BuildingInstallerPackage.md
```diff
@@ -158,27 +158,28 @@ The following packages are recognized as Linux JREs:
...
-with familiar ways to install and launch the packaged applications.
+You can provide platform-specific packages for Linux, macOS or Windows that can be
+installed and launched in ways familiar to user.
 
-`jpackage` bundles the java runtime which is used to run Gradle or the one which is used to compile the project
-that has the `jpackage` utility. Because OmegaT project uses Java 11 to compile, you need to
-run Gradle with JDK17.
--> Why ???
+`jpackage` bundles the java runtime used to run Gradle, but can also be run with the runtime
+used to compile the project that relies on the `jpackage` utility. Fpr OmegaT, you need to
```
miurahr, line 169:
```suggestion
used to compile the project that relies on the `jpackage` utility. For OmegaT, you need to
```

docs_devel/docs/93.BuildingInstallerPackage.md
miurahr: Yes

docs_devel/docs/93.BuildingInstallerPackage.md
```diff
@@ -158,27 +158,28 @@ The following packages are recognized as Linux JREs:
...
+bundled in the JDK into the Gradle build tool.
 -> I do not understand this sentence
 
 You can use the following tasks to produce the packages.
 
 - `linuxDebDist`
 - `linuxRpmDist`
 
-You can debug the contents of the package by running the `jpackage` taks to produce the
+You can debug the contents of the package by running the `jpackage` tasks to produce the
```
miurahr, line 182:
```suggestion
You can debug the contents of the package by running the `jpackage` task to produce the
```

The `jpackage` task, `$./gradlew jpackage`, generate directory structure in `build/jpacakge/app-image/OmegaT` like `app-image/OmegaT/bin/OmegaT`, `app-image/OmegaT/lib/app/OmegaT.jar` , `app-image/OmegaT/lib/app/languagetool-core-6.1.jar`, and so on. 

There are another `JPackage` type tasks, `distDeb` and `distRpm`,  generate final `deb` and `rpm` files.

Kazephil: Got it. Thanks for the extra explanation.

docs_devel/docs/93.BuildingInstallerPackage.md
Kazephil: Thank you, @miurahr.

@brandelune: If this change works for you, I think we're ready to delete your "->" comment line and merge this set of changes.

docs_devel/docs/07.ManualContainerTaskBuild.md
Kazephil: > Something strange seems to be going on with the manual build process, though. I'm getting everything inside a literal `${target}` directory in `doc_src\en\` instead of in the correct `docs\en\` folder if I try to build with Ant. More investigation required.

OK. It looks like building the documentation manually now requires passing both the `-Dtarget` and the `-Dlanguage` arguments to Ant.

I'm not sure why it didn't need to be passed explicitly before but is necessary now.

I think it might be a good idea to split this file into separate "using a container to build the documentation" and "building the documentation manually" files.

brandelune: > I think it might be a good idea to split this file into separate "using a container to build the documentation" and "building the documentation manually" files.

I agree. Thank you.

docs_devel/docs/02.HowToBuild.md
```diff
@@ -97,3 +91,281 @@ general use on Maven Central:
...
+that are handled in a parallel way. Gradle will use as many cores as the computer has.
+
+### Complex features in dependencies
+
+-> This paragraph is not very clear
+OmegaT has many filters, machine translation connectors, dictionary drivers and so on. A core part of these features
+is considered as useful for other Free Software projects. It is the reason why core developers create library projects
+which support file formats, Web services, and GUI parts.
+These have already built in another place, and been downloaded from maven central repository.
+
```
miurahr, line 371:
```suggestion
```

I'd like to remove it.

docs_devel/docs/07.ManualContainerTaskBuild.md
miurahr: I will work it.

miurahr:
Now all the reviewer feedback to be fixed are resolved.

PR #913, opened 2024-01-15
[6.0] docs/l10n: update it and de from L10N project,  generate html,  and fix version number of ca

miurahr:
## Pull request type

## What does this PR change?

- Update doc_src for it and de
- Fix version number in ca
- Remove unnecessary files from docs
- Generate HTML manuals and update

## Other information

I have built HTML manuals by using a build configuration with #914 change


kosivantsov:
It looks good to me, but my only objections to changing l10n data directly is that next time the localizers open their respective projects, they won't have these changes because they are not in their `project_save.tmx`. They may or may not reflect these changes in their translations. If there is a change in localized strings in the source code tree, the same change should be there in the OmegaT-L10n project for that locale.

miurahr:
I see. I would revert de and it imports, and releave the fix of problems only in ca about version number, and en/ja to  update html manuals from current source of document.

kosivantsov:
The only reason for my above comment was that if the change is introduced in localized string in the source code directly, we may lose that change at some point when it is updated from the respective l10n project.

miurahr:
> The only reason for my above comment was that if the change is introduced in localized string in the source code directly, we may lose that change at some point when it is updated from the respective l10n project.

I think the tweak is necessary to build html properly. I hope L10N-de and L10N-it team fix  the translations for the 6.0 manuals.

PR #925, opened 2024-01-23
docs: fix pdf generation and restructure manual generations

miurahr:
Fix issue in manual PDF generation

## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?

-  `OmegaT_EditorPanes.xml` has duplicated "id" that cause XML error on PDF generation
- https://sourceforge.net/p/omegat/documentation/419/

## What does this PR change?

- document: Manual source XML has duplicated "id" value which should be unique.
- chore: Update `doc_src/build.xml` not to generate files in source folder but write to under `build/`.  The task accept "target" argument and use it for the output folder.
- chore: Update `docgen` shell script to use `docge:apline` and `docgen:pdfgen` container image
- chore: Change font to "Noto" for PDF generation.
- chore: Fix `manualsPdfs` task properly detect output file.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -490,19 +490,16 @@ manualIndexXmls.each { xml ->
     def lang = xml.parentFile.name
     def pdfTaskName = "manualPdf${lang.capitalize()}"
     tasks.register(pdfTaskName, Exec) {
-        inputs.files fileTree(dir: "doc_src/${lang}", include: '**/*.xml')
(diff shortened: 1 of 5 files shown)


miurahr:
Please review about 
> Manual source XML has duplicated "id" value which should be unique.

Kazephil (approved):
Looks good to me!

PR #933, opened 2024-01-28
Correct SDLXLIFF comments

t-cordonnier:
## Pull request type

Bug

## Which ticket is resolved?

- sdlxliff filter does not display (some) comments
- https://sourceforge.net/p/omegat/bugs/1244/

Problem mentioned by JC, but he sent me sample in private - see with him if the sample is confidential or not, in which case we may add a unit test.

## What does this PR change?

StaX filter for SDLXLIFF improperly manages comments when used with WoodstoX parser. We already corrected similar problems with other StaX filters but we forgot this one.

To be corrected in all branches (5.8, 6.0 and 6.1)

### DIFF
--- a/src/org/omegat/filters4/xml/xliff/SdlXliff.java
+++ b/src/org/omegat/filters4/xml/xliff/SdlXliff.java
@@ -351,7 +351,7 @@ public class SdlXliff extends Xliff1Filter {
     @Override
     protected boolean processCharacters(Characters event, XMLStreamWriter writer) throws XMLStreamException {
         if (commentBuf != null) {
-            commentBuf.append(event.toString());
+            commentBuf.append(event.getData());
             if ((writer != null) && isCurrentSegmentTranslated(currentMid)) {
                 if ("last_modified_by".equals(currentProp)) {
                     writer.writeCharacters(Preferences.getPreferenceDefault(Preferences.TEAM_AUTHOR,


brandelune (approved):

miurahr:
There are several another `toString()` to event. Is it ok?

```
SdlXliff.java:432:                                writer.write(ev.toString());
```

It can be like
```
if (ev.getEventType() == XMLStreamConstants.CHARACTERS) {
     writer.write(((Characters) ev).getData());
} else {
     weriter.write(ev.toString());
}
```

miurahr:
@brandelune  Could you open a bug ticket?

>not opened, as far as I remember
>Problem mentioned by JC, but he sent me sample in private - see with him if the sample is confidential or not, in which >case we may add a unit test.

brandelune:
Isn't it related to this one ?

https://sourceforge.net/p/omegat/bugs/1192/

t-cordonnier:
> Isn't it related to this one ?
> 
> https://sourceforge.net/p/omegat/bugs/1192/

This one was about filter for standard xliff, but yes it is similar topic. However it is already closed, so I let you decide if you want to open a new one for sdlxliff or not.

t-cordonnier:
> There is another `toString()` to event. Is it ok?
> 
> ```
> SdlXliff.java:432:                                writer.write(ev.toString());
> ```
> 
> It can be like
> 
> ```
> if (ev.isCharacters()) {
>      writer.write(ev.asCharacters().getData());
> } else {
>      weriter.write(ev.toString());
> }
> ```

Yes we can do like that. This is a case for which I would have to find a better solution later.

brandelune:
> @brandelune Could you open a bug ticket?

I did and updated @t-cordonnier comment.

miurahr (approved):

t-cordonnier:
@miurahr As it is a bug correction, which @brandelune explicitly mentioned for 6.0.1, I think it should be pushed also in 5.8 and 6.0

miurahr:
I will cherry-pick this to releases/6.0 branch.

PR #937, opened 2024-02-01
[BUGS#1247]  xliff1: multiple original file source produce broken output

miurahr:
xliff1 filter produce a broken output file when XLIFF1 file has multiple `<file>` element.

## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?

-  XLIFF1 filter produce broken trasnalated file
- https://sourceforge.net/p/omegat/bugs/1247/

## What does this PR change?

- `AbstractXmlFilter#processFile`: when it calls `checkCurrentPosition`, it should be checks `isEventMode` immediately.
- Several style improvements
- Add regression tests

### DIFF
--- a/src/org/omegat/filters4/xml/AbstractXmlFilter.java
+++ b/src/org/omegat/filters4/xml/AbstractXmlFilter.java
@@ -4,6 +4,7 @@
           glossaries, and translation leveraging into updated projects.
 
  Copyright (C) 2017 Thomas Cordonnier
+               2024 Hiroshi Miura
                Home page: https://www.omegat.org/
                Support center: http://groups.yahoo.com/group/OmegaT/
 
@@ -189,16 +190,17 @@ public abstract class AbstractXmlFilter extends AbstractFilter {
             try {
                 strReader = iFactory.createXMLStreamReader(inReader);
                 eventReader = iFactory.createXMLEventReader(strReader);
-                isEventMode = false; // always start like this, even with new
-                                     // file
+                // always start like this, even with new file
+                isEventMode = false;
                 if (writer == null) {
                     while (strReader.hasNext()) {
+                        // First check if not event mode.
                         if (!isEventMode) {
                             checkCurrentCursorPosition(strReader, false);
                         }
-                        if (isEventMode) { // calculated after
-                                           // checkCurrentCursorPosition, may
-                                           // have changed!
+                        // calculated after checkCurrentCursorPosition may
+                        // have changed!
+                        if (isEventMode) {
                             XMLEvent event = eventReader.nextEvent();
                             if (event.isStartElement()) {
                                 processStartElement(event.asStartElement(), null);
@@ -237,17 +239,17 @@ public abstract class AbstractXmlFilter extends AbstractFilter {
                             strReader.next();
                             continue;
                         }
+
+                        // First check if not event mode.
                         if (!isEventMode) {
                             checkCurrentCursorPosition(strReader, true);
-                            // in non-event mode, we always write exactly what
-                            // was in the source
-                            fromReaderToWriter(strReader, strWriter);
                         }
-                        if (isEventMode) { // calculated after
-                                           // checkCurrentCursorPosition, may
-                                           // have changed!
-                            XMLEvent event = eventReader.nextEvent();
+
+                        // calculated after checkCurrentCursorPosition may
+                        // have changed!
+                        if (isEventMode) {
                             boolean keep;
+                            XMLEvent event = eventReader.nextEvent();
                             if (event.isStartElement()) {
                                 keep = processStartElement(event.asStartElement(), strWriter);
                             } else if (event.isEndElement()) {
@@ -262,6 +264,9 @@ public abstract class AbstractXmlFilter extends AbstractFilter {
                                 fromEventToWriter(event, strWriter);
                             }
                         } else {
+                            // still in non-event mode, we always write
+                            // exactly what was in the source
+                            fromReaderToWriter(strReader, strWriter);
                             strReader.next();
                         }
                     }
@@ -331,9 +336,9 @@ public abstract class AbstractXmlFilter extends AbstractFilter {
             break;
         case XMLEvent.START_ELEMENT:
             final String localName = xmlr.getLocalName(), namespaceURI = xmlr.getNamespaceURI();
-            if (namespaceURI != null && namespaceURI.length() > 0) {
+            if (namespaceURI != null && !namespaceURI.isEmpty()) {
                 final String prefix = xmlr.getPrefix();
-                if (prefix != null && prefix.length() > 0) {
+                if (prefix != null && !prefix.isEmpty()) {
                     writer.writeStartElement(prefix, localName, namespaceURI);
                 } else {
                     writer.writeStartElement(namespaceURI, localName);
@@ -346,7 +351,7 @@ public abstract class AbstractXmlFilter extends AbstractFilter {
             }
             for (int i = 0, len = xmlr.getAttributeCount(); i < len; i++) {
                 String attUri = xmlr.getAttributeNamespace(i);
-                if (attUri != null && attUri.length() > 0) {
+                if (attUri != null && !attUri.isEmpty()) {
                     writer.writeAttribute(attUri, xmlr.getAttributeLocalName(i), xmlr.getAttributeValue(i));
                 } else {
                     writer.writeAttribute(xmlr.getAttributeLocalName(i), xmlr.getAttributeValue(i));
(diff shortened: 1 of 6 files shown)


miurahr:
I would like to rebased on #938 is merged into master.

miurahr:
I've raised a bug ticket for the issue.

miurahr:
The test data file is not compliant with XLIFF 1.2 specification. So I would like to update it but the issue is not solved when file is updated.

miurahr:
The output is strange.

1. missing `<file>` element and there are double `<body>`

```xml
<file target-language="be" source-language="en" original="text3.txt" datatype="plaintext">
    <body><body>
```

2. double end element of `<body>`

```xml
    </body>
</body>
```

3. double end of  `</file>`

```xml
</file>
</file></xliff>
``` 

When observing with debugger,  I see StaX start-element events;  "file", "body", "trans-unit", "file", "body", "trans-unit" accordingly.

miurahr:
I found when argument `writer` is not null,  `AbstractXmlFilter#processFile` does 

```java
                        if (!isEventMode) {
                            checkCurrentCursorPosition(strReader, true);
                            // in non-event mode, we always write exactly what
                            // was in the source
                            fromReaderToWriter(strReader, strWriter);
                        }
                        if (isEventMode) { 

```
always call `fromReaderToWriter` method. As a comment tells
```
  // calculated after checkCurrentCursorPosition may
  // have changed!
```
We should check `isEventMode` immediately after called `checkCurrentCursorPosition`. So it should  be like
```java
                        if (!isEventMode) {
                            checkCurrentCursorPosition(strReader, true);
                       }
                       if (!isEventMode) {
                            // in non-event mode, we always write exactly what
                            // was in the source
                            fromReaderToWriter(strReader, strWriter);
                        }
                        if (isEventMode) { 

```

test/data/filters/xliff/filters3/file-XLIFFFilter-cdata-bugs418.xlf
```diff
@@ -1,25 +1,25 @@
...
+            <trans-unit id="id1">
                 <source>This is test2</source>
             </trans-unit>
-            <trans-unit id="2">
+            <trans-unit id="id2">
                 <source>test21</source>
             </trans-unit>
         </body>
     </file>
     <file target-language="be" source-language="en" original="text.txt">
```
miurahr, line 13:
```suggestion
    <file target-language="be" source-language="en" original="text.txt" datatype='plaintext'>
```
datatype is mandatory.

test/src/org/omegat/filters4/Xliff1FilterTest.java
```diff
@@ -119,4 +120,24 @@ protected void translate(AbstractFilter filter, String filename) throws Exceptio
         translate(filter, filename, Collections.emptyMap(),
                 Collections.singletonMap("Should translate in result.", "Devrait traduire dans le r\u00E9sultat."));
     }
+
+    @Test
+    public void testBugs1247() throws Exception {
```
miurahr, line 125: We can add a regression test for bugs#418

miurahr:
This should backport to 6.0.1.

t-cordonnier:
@miurahr  sorry that I could not answer before. We always must be very careful when we modify AbstractXmlFilter, as the code is potentially used by multiple subclasses. Let me do some tests and I will come back to this soon. In the meantime better don't push anything to master.

t-cordonnier:
OK, so my first hypothesis is a problem due to migration from Oracle's StaX to Jackson. To validate this, I tried to open this file with OmegaT 5.7 + plugin (this one does not use Jackson) and create target file, the result is correct. If I do the same with OmegaT 6.1, the bug appears.

Now I will detail more in answer to your change.

t-cordonnier:
> always call fromReaderToWriter method. 

Yes. The goal of the non-event mode, or stream mode, is to save memory when the document contains lot of non-translatable content. This is the case in particular of SDLXLIFF files, which contain a big Base64 string to be kept as is in the result: so I prefer not to create events and let the stream reader and writer speak each other.

Now the problem we have is in the transition between the stream and event mode, this transition seems managed differently from Oracle to Jackson. I will have a look.

PS. Sorry for the modification of your comment, it was not intentional I thought I was doing a quote reply.

t-cordonnier:
> This should backport to 6.0.1.

Does OmegaT 6.0 use Jackson?

miurahr:
> > This should backport to 6.0.1.
> 
> Does OmegaT 6.0 use Jackson?

No, OmegaT 6.0 use Sun's stack and JAXB for XML processing. So the issue is specific to 6.1.0.

miurahr:
> The error seems linked to ` eventReader = iFactory.createXMLEventReader(strReader);`
> The StaX specification only advice not to use the stream reader after that, but it does not describe its behaviour. And indeed, it seems that Oracle's implementation introduces a shift between the stream reader and the event reader, which Jackson does not: when stream reader is at the `<body>`, event reader's nextEvent() will return the following ` <trans-unit>` in Oracle's StaX - and my implementation was based on this shift - but it seems that it returns the `<body>` in Jackson. As a consequence `<body>` is read twice - once by the stream reader as isEventMode is false, and second time by the event reader when isEventMode suddenly becomes true.

Thank you for the explanation.  Especially,

> The StaX specification only advice not to use the stream reader after that,

It means that a behavior can depend on implementation, so we encounter a regression on it. 
OmegaT 6.1.0 configured to use  `woodstox` for StaX framework, because we enabled jackson's xml-bind.

src/org/omegat/filters4/xml/AbstractXmlFilter.java
```diff
@@ -262,6 +264,9 @@ public void processFile(BufferedReader inReader, BufferedWriter writer, FilterCo
                                 fromEventToWriter(event, strWriter);
                             }
                         } else {
+                            // still in non-event mode, we always write
+                            // exactly what was in the source
+                            fromReaderToWriter(strReader, strWriter);
```
t-cordonnier, line 269: Yes I wanted to suggest to move fromReaderToWriter in the else part (rather than having a specific if as in the comment) but apparently you saw this possiblity.

OK for the moment. I would prefer a solution which works for all StaX implementations (in case you decide later for a new change...) but I will think about this later.  You can merge like this now.

t-cordonnier (approved):
I will search later whenever I can find a solution which works for all StaX implementations.
For the moment you can merge like that, I would only suggest that we find a location where to document this, in case a new change introduces a regression later.

src/org/omegat/filters4/xml/AbstractXmlFilter.java
miurahr: > OK for the moment. I would prefer a solution which works for all StaX implementations (in case you decide later for a new change...) but I will think about this later. You can merge like this now.

Good news!   >> **works for all StaX implementations**

t-cordonnier: No, I tested with OmegaT 5.7 + plugin, the <body> completely disappears in this case

miurahr:
I found both Oracle and Woodstox StaX implementation respect a steam file pointer when we make an event reader from stream reader.

I also found Woodstox StaX implementation sync an element pointer between stream reader and event reader.
Oracle implementation independently has an element pointer  in stream reader and event reader.

A perfect solution is that we can create an event reader when we decide to move to event mode.
We can make it like

```
                        if (!isEventMode) {
                            if (checkCurrentCursorPosition(strReader, false)) {
                                eventReader = iFactory.createXMLEventReader(strReader);
                            }
                        }
```

miurahr:
I've proposed the above idea in #945

t-cordonnier:
> I found both Oracle and Woodstox StaX implementation respect a steam file pointer when we make an event reader from stream reader.

Yes but the problem is that eventReader.nextEvent() returns ` ++pointer`  for Oracle and `pointer++` for Woodstox.

> A perfect solution is that we can create an event reader when we decide to move to event mode.

The idea is good I will do tests. For the moment I still see some strange behaviour with the document you used for tests, but it may be due to something else.

t-cordonnier:
OK, after test,

```
                        if (!isEventMode) {
                            checkCurrentCursorPosition(strReader, true);
                            if (isEventMode) { // changed during checkCurrentCursorPosition
                                eventReader = iFactory.createXMLEventReader(strReader);
                            }
                        }
```

seems to work (but with `fromReaderToWriter` in the second if, as in your last implementation)
We can do like this.

miurahr:
> OK, after test,
> 
> ```
>                         if (!isEventMode) {
>                             checkCurrentCursorPosition(strReader, true);
>                             if (isEventMode) { // changed during checkCurrentCursorPosition
>                                 eventReader = iFactory.createXMLEventReader(strReader);
>                             }
>                         }
> ```
> 
> seems to work (but with `fromReaderToWriter` in the second if, as in your last implementation) We can do like this.
> 
> PS. Don't forget to do the change in both loops - there is one when streamWriter is null (when importing a file to OmegaT) and one when streamWriter is not null (when producing target file)

#945 implement it w/ small improvement. @t-cordonnier

PR #938, opened 2024-02-01
fix: properly handle bilingual in filter tests

miurahr:
## Pull request type

- Other (describe below)
refactor

## Which ticket is resolved?

Dev ML discussion

## What does this PR change?

- Switch behavior in a mock translate callback
- PoFilterTest: properly set options

### DIFF
--- a/test/src/org/omegat/filters/TestFilterBase.java
+++ b/test/src/org/omegat/filters/TestFilterBase.java
@@ -299,7 +299,8 @@ public abstract class TestFilterBase extends TestCore {
     }
 
     /**
-     * Helper function for testing the translateFile method of a filter. Translation equals the source.
+     * Helper function for testing the translateFile method of a filter.
+     * Translation equals the source when monolingual filter or monolingual option.
      * Translation is written to {@link #outFile}.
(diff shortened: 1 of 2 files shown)


test/src/org/omegat/filters/TestFilterBase.java
```diff
@@ -299,22 +299,29 @@ protected void translate(AbstractFilter filter, String filename) throws Exceptio
...
      * Translation is written to {@link #outFile}.
      * @param filter the filter to test
      * @param filename the file to use as input for the filter
      * @param config the filter options/config to use
      * @throws Exception when the filter throws an exception
      */
     protected void translate(AbstractFilter filter, String filename, Map<String, String> config) throws Exception {
+        boolean allowBlank;
+        if (filter.isBilingual()) {
+            allowBlank = !"true".equalsIgnoreCase(config.get("monolingualFormat"));
```
miurahr, line 313: The POFilter specific logic should be in POFilterTest class as overriding base class method.

t-cordonnier (approved):
lgtm

PR #939, opened 2024-02-03
[BUGS#903] fix: not-undo a pre-existed note

miurahr:
- Fix BUGS#903

## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?

- Ctrl+Z incorrectly erases contents of segment Note
- https://sourceforge.net/p/omegat/bugs/903/

## What does this PR change?

- When visiting segment and update note pane when segment has note, we reset a status of undoManager after put note contents.


miurahr:
Currently, only a `EditorController#activateEntry` calls `SetNoteText` method but it is defined in `INote` interface and published to plugins, scripts. It may not be good idea to reset history here but we can prepare  internal method for `EditorContoller` to reset history.

PR #940, opened 2024-02-04
[BUGS#858] change alternate modifier to alt

miurahr:
Java Swing menu key accelerator is set with `Ctrl-<Key>` or `Ctrl-Alt-<Key>`(PC) or `Meta-<Key>` or `Meta-Alt-<Key>` (Mac), so accessing menu item with shortcut key always with a `Ctrl`/`Meta` modifier which is defined as `MenuShortcutKeyMask`.

OmegaT menu handler checks  whether a modifier is  `MenuShortcutKeyMask`. It is always "true" ,  when accessed from a "shortcut key"

A change is intended to behave

1. Use the menu item normally to open the file
2. Hold "Alt" while using the menu item to reveal the file
3. Shortcut key "Ctrl+<key>", "Ctrl+<mod>+<key>" open the file

## Pull request type
- Bug fix -> [bug]

## Which ticket is resolved?

- Manually assigned shortcuts open parent folders instead of files
- https://sourceforge.net/p/omegat/bugs/858/

## What does this PR change?

- Skip alternation of feature when shortcut key modifier is pressed
- alternate open behavior with Alt key

## Other information

- May need to update manual


brandelune:
It looks like 6.1 behaves properly already. There must have been changes between 4.1 and 6.1, or maybe it is moving from Java 8 to Java 11. In any case, I can't reproduce this bug anymore.

Have you confirmed the faulty behavior on your side?

miurahr:
When select  `File > Access Project Contents > Current Source File` on the project that sources are HTML, I see a web browser opens and show a source HTML file contents.
When select it by mouse with "Ctrl" key pushed, it is expected that file explore, Nemo on mint linux, opens and how the subfolder which hold HTML file, but it opens web browser on linux.   

I think the feature is not working at least on linux.

src/org/omegat/gui/main/MainWindowMenuHandler.java
```diff
@@ -324,7 +324,7 @@ public void projectAccessCurrentSourceDocumentMenuItemActionPerformed(int modifi
             return;
         }
         File toOpen = new File(root, path);
-        if ((modifier & Java8Compat.getMenuShortcutKeyMaskEx()) != 0) {
```
miurahr, line 327: Here is a bug that we compare with `KeyEvent.META_DOWN_KEY` or `KeyEvent.CTRL_DOWN_KEY` which is returned by `Java8Compat.getMenuShortcutKeyMaskEx()` but caller call here with `ActionEvent.CTRL_KEY` or `ActionEvent.META_KEY`. It is a reason that a modifier has never been worked.

miurahr: The bug is introduced in v3.5.1 at [a first commit](https://github.com/omegat-org/omegat/commit/fab16c8b1d11ce49d872aab86c7b52ad47267e4b) when adding the feature. At that time, `KeyEvent.META_KEY` and `ActionEvent.META_KEY`  has same value, then it works like as expected.

in v4.2.0, when supporting Java 11 runtime [a change](https://github.com/omegat-org/omegat/commit/98127c352a2b1bb63f1a831389aaa7fc79a83966) breaks the feature. Then compiling with Java11 it stop working.

brandelune: Interesting that nobody noticed that, even though the feature was discussed in 4.1. Thank you for checking the history of the bug.

miurahr:
There is no explanation the feature of clicking menu item with modifier key.

https://omegat.sourceforge.io/manual-standard/en/chapter.menus.html#menus.project.access.project.contents

>Current Source File
> 
>     The current source file is located in the [source](https://omegat.sourceforge.io/manual-standard/en/chapter.project.folder.html#project.folder.source) folder.
> 
>     Opens the file in the associated application.
> Current Target File
> 
>     The current target file is located in the [target](https://omegat.sourceforge.io/manual-standard/en/chapter.project.folder.html#project.folder.target) folder.
> 
>     Opens the file in the associated application if the file exists.
> Writable Glossary
> 
>     The writable glossary is located in the [glossary](https://omegat.sourceforge.io/manual-standard/en/chapter.project.folder.html#project.folder.glossary) folder.
> 
>     Opens the file in the associated application if the file exists.

It is necessary to add an explanation such as `If you press down ALT key during selecting menu, it opens a folder enclosing the file, instead of opening the file using finder / file explore`

brandelune:
> It is necessary to add an explanation such as `If you press down ALT key during selecting menu, it opens a folder enclosing the file, instead of opening the file using finder / file explore`

Yes, the feature needs to be documented otherwise it will be forgotten. I'm putting that on my TODO stack.

doc_src/en/Menus_Project.xml
```diff
@@ -324,7 +324,10 @@
 			  linkend="project.folder.source"
 			  endterm="project.folder.source.title"/> folder.</para>
 
-			  <para>Opens the file in the associated application.</para>
+              <para>Opens the file in the associated application if the file exists.
+               If you press down ALT key during selecting the menu item, opens a folder enclosing the file, by
+               OS standard file manager such as finder or file explorer, instead of opening the file.
+              </para>
```
miurahr, line 330: Manual is updated. 
@brandelune  please check and rewording when necessary.

miurahr:
Are there any concern that @brandelune cannot approve the PR? Please take an action to give a feedback or approve.

brandelune:
Please go ahead.

PR #941, opened 2024-02-04
[BUGS#1131] fix: Redraw duplicate segments only when changed

chelobaka:
## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?

- Very slow segment switching with a lot of duplicates
  * https://sourceforge.net/p/omegat/bugs/1131/
 

## What does this PR change?

- A check is added if a segment or its note is changed before redrawing identical segments 

## Other information

This bug seems to be specific to some environments. OmegaT on Debian testing x64 with OpenJDK from Debian repository is affected for sure. I also have a report that Arm Mac build suffers as well.

I've been using this fix for a year and never found any regressions.


miurahr:
LGTM.

miurahr (approved):

PR #942, opened 2024-02-05
refactor: test for XLIFF2 glossary module

miurahr:
## Pull request type

- Other (describe below)
Improve unit tests

## What does this PR change?

- Improve test case with extra-ordinary cases.
- Add test case when source XLIFF file has contents with an unsupported glossary extension module.


miurahr:
Failed with BUGS#1247  
#937 will solve the issue.

t-cordonnier:
I see no problem but I don't understand what you want to test here. The filter does not use the internal glossary for the moment.

miurahr:
> I see no problem but I don't understand what you want to test here. The filter does not use the internal glossary for the moment.

There are not enough variation of test cases.
The test is to check the case when xliff extension module exists, and it does not break a normal feature.

t-cordonnier (approved):

PR #945, opened 2024-02-05
[BUGS#1247] fix: filters4: create event reader when entering event mode

miurahr:
- Complete solution to avoid StaX implementation difference.

## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?

- BUGS#1247

## What does this PR change?

- Create event reader when entering event mode

## Other information

#937


miurahr:
#947 back-port the improvement here into 6.0 branch.

PR #951, opened 2024-02-07
refactor: migrate markers as first class plugins

miurahr:
- Deprecate Core.registerMarker and Core.getMarkers
- Remove MarkerController.init and WhitespaceMarkerFactory.init

## Pull request type

- Other (describe below)
refactor

## What does this PR change?

- Make `NBSPMarker` as true plugin
- NBSP marker menu are created in the plugin
- Removal of hard coded menu creation
- When registering maker plugins with its class, and loading the plugin, static fields are initialized when `loadPlugins` static function is called. it breaks colors of markers.

### DIFF
--- a/src/org/omegat/gui/editor/MarkerController.java
+++ b/src/org/omegat/gui/editor/MarkerController.java
@@ -32,28 +32,26 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.Queue;
 
+import javax.swing.JCheckBoxMenuItem;
 import javax.swing.SwingUtilities;
 import javax.swing.text.BadLocationException;
 import javax.swing.text.Highlighter;
 import javax.swing.text.Position;
 
-import org.omegat.core.Core;
-import org.omegat.core.spellchecker.SpellCheckerMarker;
+import org.openide.awt.Mnemonics;
+
 import org.omegat.filters2.master.PluginUtils;
-import org.omegat.gui.editor.mark.BidiMarkers;
 import org.omegat.gui.editor.mark.CalcMarkersThread;
-import org.omegat.gui.editor.mark.ComesFromAutoTMMarker;
 import org.omegat.gui.editor.mark.ComesFromMTMarker;
 import org.omegat.gui.editor.mark.EntryMarks;
-import org.omegat.gui.editor.mark.FontFallbackMarker;
 import org.omegat.gui.editor.mark.IMarker;
 import org.omegat.gui.editor.mark.Mark;
 import org.omegat.gui.editor.mark.NBSPMarker;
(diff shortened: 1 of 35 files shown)


src/org/omegat/gui/editor/MarkerController.java
```diff
@@ -116,6 +135,26 @@ public static void init() throws Exception {
...
         }
+
+        // default markers
+        JCheckBoxMenuItem menuItem = new JCheckBoxMenuItem();
+        nbspMarker.setEnabled(Preferences.isPreference(Preferences.MARK_NBSP));
+        menuItem.setSelected(nbspMarker.isEnabled());
+        menuItem.setIcon(MainMenuIcons.newColorIcon(Styles.EditorColor.COLOR_NBSP.getColor()));
+        Mnemonics.setLocalizedText(menuItem, OStrings.getString("MW_VIEW_MENU_MARK_NBSP"));
+        menuItem.addActionListener(e -> {
+            isNBSP = menuItem.isSelected();
```
miurahr, line 146: It is necessary to improve more.

miurahr:
Now these 6 maker become true plugins.

```
    org.omegat.core.spellchecker.SpellCheckerMarker \
    org.omegat.gui.glossary.TransTipsMarker \
    org.omegat.gui.editor.mark.BidiMarkers \
    org.omegat.gui.editor.mark.ComesFromAutoTMMarker \
    org.omegat.gui.editor.mark.FontFallbackMarker \
    org.omegat.gui.editor.mark.ReplaceMarker \
```

Menu Items are generated dynamically using  marker plugins IMarker interface.

miurahr:
I propose addition of method `    void reloadDocument();` in `IEditor` core interface, that reload project contents and redraw markers when project is loaded.

brandelune:
@t-cordonnier review appreciated ! :-)

t-cordonnier:
> @t-cordonnier review appreciated ! :-)

@brandelune  I already sent some comments in this ticket. 

I notice that now the menus are more dynamic as I suggested, good point. So, there was an evolution since my last comments, which I was not alerted about, so for me it was still pending.

The rest of the code seems unchanged so I maintain what I said: I would prefer one or more sub-interfaces to distinguish between non-skippable markers (the ones which are always active), menu markers (the ones which have a menu with an icon to activate/inactivate) and eventually, those which are skippable but whose configuration is done in another location than in a menu.
I remember in the past Aaron wanted to avoid casts, but maybe @miurahr can have a different position. I let to him the final decision but for the moment I saw not even a reaction to my comments.

miurahr:
The change does not provide a benefit over the task.

PR #952, opened 2024-02-08
refactor: markers: avoid unnecessary static

miurahr:
There are unnecessary static variable initialization and static functions. These break registering Markers as plugins.

## Pull request type

- Other (describe below)
refactor

## What does this PR change?

- avoid all `Styles` static reference in static initialization, because the reference should be happened after theme modules initialization.  
-
-

## Other information

Bring improvements from #951


miurahr:
Need unit tests for these markers

PR #953, opened 2024-02-08
refactor: whitespace marker to be with single class

miurahr:
This reduces a number of threads and may improve performance. It also help make WhitespaceMarkers to be implemented as plugins.

## Pull request type

- Other (describe below)
refactor

## What does this PR change?

- Avoid registering three (LF, SP, and TAB) markers but single white space marker class.

## Other information

Brings from #951


src/org/omegat/gui/editor/MarkerController.java
```diff
@@ -77,7 +77,7 @@ public class MarkerController {
     private final Highlighter highlighter;
 
     public static void init() throws Exception {
-        WhitespaceMarkerFactory.init();
+        Core.registerMarker(new WhitespaceMarkers());
```
miurahr, line 80: The class name should be `WhitespaceMarker` not plural.

miurahr: No need to register it here.

PR #959, opened 2024-02-13
[BUGS#1250] fix: korean terms are not found in glossary

miurahr:
## Pull request type

- Bug fix

## Which ticket is resolved?

- [Short glossary items are not recognized with Korean source](https://sourceforge.net/p/omegat/bugs/1250/)
- https://sourceforge.net/p/omegat/bugs/1250/)

## What does this PR change?

- Add reproducible of the bug
- Fix the collision detection logic
-

## Other information

glossary token collisions in Japanese
 https://sourceforge.net/p/omegat/bugs/1034/

### DIFF
--- a/src/org/omegat/gui/glossary/GlossarySearcher.java
+++ b/src/org/omegat/gui/glossary/GlossarySearcher.java
@@ -150,10 +150,8 @@ public class GlossarySearcher {
 
     private static boolean rawMatch(Token[] tokens, String srcTxt, String term) {
         for (Token token : tokens) {
-            if (term.length() > token.getLength()) {
-                if (term.contains(token.getTextFromString(srcTxt))) {
-                    return true;
-                }
+            if (term.length() > token.getLength() && term.contains(token.getTextFromString(srcTxt))) {
+                return true;
             }
         }
         return false;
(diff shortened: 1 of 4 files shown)


miurahr:
The bug is reproduced by a  test case `GlossarySearcherTest#testGlossarySearcherKorean`.

miurahr:
Now we can find a test case failed and the patch fixes the issue and pass the tests.

miurahr:
The fix here should be backported to OmegaT 6.0.1

src/org/omegat/gui/glossary/GlossarySearcher.java
```diff
@@ -150,8 +150,7 @@ private List<Token[]> getMatchingTokens(Token[] fullTextTokens, String fullText,
 
     private static boolean rawMatch(Token[] tokens, String srcTxt, String term) {
         for (Token token : tokens) {
-            if (term.length() > token.getLength()
-                    && token.getTextFromString(srcTxt).equals(term.substring(token.getLength()))) {
+            if (term.length() > token.getLength() && term.contains(token.getTextFromString(srcTxt))) {
```
miurahr, line 153: We don't need to check length

```suggestion
            if (term.contains(token.getTextFromString(srcTxt))) {
```

PR #961, opened 2024-02-16
feat: separate segment matcher configurable

miurahr:
## Pull request type

- Bugs

## Which ticket is resolved?

- FindMatcher always search separate segment match without penalty 
- https://sourceforge.net/p/omegat/bugs/1251/

## What does this PR change?

- Add preference to enable separate segment matcher
- Add panalty value for separate segment matcher
- Make penalty configurable for separate segment matcher
- UI: add checkbox and spinner on TMMatcher preference dialog
- Add TMX path as "from_separate_segment_matcher" when match it.

## Other information

- tm/penalty-nnn value is not respected
- https://sourceforge.net/p/omegat/bugs/1248/


brandelune:
We don't need to rush a modification. Now that we understand where the issue comes from we need to discuss it and see if it is worth modifying. We can consider that as a undocumented feature and not as a bug since that's the way Alex intended to implement it.

Also, it seems to appear only in very specific cases. Here, the separate string was `(1)` and the issue did not appear with longer strings.

Maybe the feature implementation is not correct (obviously I should not have a 100% match here, or it should be displayed differently). But I don't think adding another preference is the solution.

miurahr:
There is successor #963

PR #964, opened 2024-02-18
feat: support GUI test

miurahr:
## Pull request type

## Which ticket is resolved?

dev-ML
https://sourceforge.net/p/omegat/mailman/omegat-development/thread/47347769-ccd4-4272-9155-9eaf96c0296c%40northside.tokyo/#msg58743384

## What does this PR change?

- refacotr main gui classes to allow GUI test - MainWindowMenuHandler class inherited from BaseMenuHanlder abstract class - Tester should implement MenuHandler class for test inherited from BaseMenuHandler class
- Add testGui configuration on gradle
- Set gui test source under test/gui
- Introduce TestCoreGUI class
- Add BasicMenuItemTest

-
-
-

### DIFF
--- a/src/org/omegat/gui/main/BaseMainWindowMenu.java
+++ b/src/org/omegat/gui/main/BaseMainWindowMenu.java
@@ -37,6 +37,7 @@
 
 package org.omegat.gui.main;
 
+import java.awt.Component;
 import java.awt.Font;
 import java.awt.event.ActionEvent;
 import java.awt.event.ActionListener;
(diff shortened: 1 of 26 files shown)


src/org/omegat/gui/main/BaseMainWindowMenu.java
```diff
@@ -738,6 +738,7 @@ protected JMenu createMenu(String titleKey) {
 
     protected JMenu createMenu(String titleKey, MenuExtender.MenuKey menuKey) {
         JMenu result = new JMenu();
+        result.setName(titleKey);
```
miurahr, line 741: It is not good idea to give menu name from localization key.

miurahr:
- It seems better to add definitions of UI names at end of class definition because of test purpose.

PR #969, opened 2024-02-28
fix: always use MainWindow#getApplicationFrame

miurahr:
fix `MainWindowMenuHandler`  to use `MainWindow#getApplicationFrame`  to get Swing JFrame component from `MainWindow` object.

## Pull request type

- Other (describe below)
refactor

## Which ticket is resolved?

Dev-ML
https://sourceforge.net/p/omegat/mailman/message/58742917/

## What does this PR change?
- do not assume MainWindow class as JFrame object
- Use MainWindow#getApplicationFrame when expect parent frame
- The enforcement helps an internal change of MainWindow class

## Other information

Spin out from #964 topic branch


brandelune:
Is there anything specific that you want us to check, besides for knowing whether the code builds/runs?

damien-rembert (approved):
Looks good to me :)

PR #974, opened 2024-03-09
refactor: Introduce IProjectFilesList interface

miurahr:
Technical debt repayment.

## Pull request type

- Other (describe below)
refactor

## What does this PR change?

- add IProjectFilesList interface
- add Core.getProjectFilesList core function
- create ProjectFilesList object in Core.initilizeGUI function
- reduce a direct MainWindow object access w/o using IMainWindow API from MenuHandler

## Other information

Spin off PR from #964


miurahr:
There is no object from community members in 2 weeks.

PR #976, opened 2024-03-11
feat: IMainMenu#getMenuItemForNames API

miurahr:
Extend `IMainMenu` API and use the API in `MainWindowMenuHandler` instead of accessing `MainWindowMenu` object methods.

## Pull request type

- Other (describe below)
refactor

## Which ticket is resolved?

dev-ML
https://sourceforge.net/p/omegat/mailman/message/58747200/

## What does this PR change?

- Define menu and menu items "name"
- Add new API getMenuItemForNames to help manipulating or retrieving status of menu item.

## Other information

TODO:

- [ ] unit tests
- [ ] define all component names

### DIFF
--- a/src/org/omegat/gui/main/MainWindowMenuHandler.java
+++ b/src/org/omegat/gui/main/MainWindowMenuHandler.java
@@ -43,6 +43,8 @@ import java.io.File;
 import java.util.List;
 import java.util.regex.Pattern;
 
+import javax.swing.JDialog;
+import javax.swing.JMenuItem;
 import javax.swing.JOptionPane;
 import javax.swing.SwingWorker;
 import javax.swing.text.JTextComponent;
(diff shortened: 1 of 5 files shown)


src/org/omegat/gui/main/MainWindowMenuHandler.java
```diff
@@ -1023,7 +1039,8 @@ public void helpContentsMenuItemActionPerformed() {
      * Shows About dialog
      */
     public void helpAboutMenuItemActionPerformed() {
-        new AboutDialog(mainWindow).setVisible(true);
+        JDialog aboutDialog = new AboutDialog(mainWindow.getApplicationFrame());
```
miurahr, line 1042: It should not create new object every calls, we can recycle the dialog.

miurahr:
A feature handles a GUI parts. We need a robust GUI test framework to implement unit test for the feature.
It is why the proposal is keeping in draft.

I've working on Assertj-Swing testing framework to help the proposal tested with it.
https://codeberg.org/miurahr/assertj-swing

Now "tokyo.northside:assertj-swing:4.0.0-beta-1" published, so I would like to resume the activity.

PR #983, opened 2024-03-16
refactor: call MenuHandlers without reflection

miurahr:
refactor MainWindowMenuHandler and MainWindowMenu not to use reflection to invoke action. Otherwise, we use `Action` interface appeared in Java 1.6. 
OmegaT project have not used it,  because lib-mnemonics does not support `Action` interface in previous versions.
Now lib-mnemonics library version 1.2 adds `AbstractmnemonicsAction` class that help application define `Acton` with mnemonics.

## Pull request type

- Other (describe below)
refactor

## Which ticket is resolved?
dev-ML https://sourceforge.net/p/omegat/mailman/message/58749145/

## What does this PR change?

- Replace `MainWindowMenuHandler#braBrabrahActionperformed` method with `Action` classes that extends with `AbstractMnemonicAction` class
- Change `BaseMainWindowMenu` to use `Action` class to construct UI components.
- Deprecated IMainMenu#invokeAction but calls Action in action class.

## Other information
Implementation depends on lib-mnemonics enhancement

https://github.com/omegat-org/lib-mnemonics/pull/8


miurahr:
rebased on current master.

miurahr:
Now ready for merge.

miurahr:
`MainWindowMenuHanlder.getActions` is used only in `BaseMainWindowMenu` class. We can instantiate Action classes directly in `BaseMainWindowMenu` class.

miurahr:
It looks not good to use Map for instantiate Action objects and get it from MainWindowMenu  class.
It would be better to call constructor in MainWindowMenu class directly.

PR #984, opened 2024-03-16
feat: allow inject to JRE translations

miurahr:
We can inject RU translations through UIManager.
This proposal introduces `org.omegat.gui.basic` resource that is loaded when initialize GUI.

## Pull request type

## What does this PR change?

-
-
-


miurahr:
There is a suggestion the feature should be external library and OmegaT uses it.

PR #985, opened 2024-03-16
feat: send whole locale to Apertium MT

MarcRiera:
Originally, the Apertium API was strict about pair language codes. For example, if `en-US` to `es-ES` translation was requested but only `en` to `es` translation was available, it would not fall back and instead report "The requested pair is not installed". For this reason, OmegaT stripped the country/variant part of the locale when calling the Apertium API.

From the current most recent version of the API (for reference: https://github.com/apertium/apertium-apy/releases/tag/v0.12.1), released a couple of weeks ago and already deployed to https://apertium.org, the API has a fallback mode. Now it tries to find an appropriate pair if the requested locale combination is not found directly.

For this reason, OmegaT should now send the whole locale and let the Apertium server choose a suitable pair.

## Pull request type

## What does this PR change?

- Apertium MT now sends whole locale code to server instead of only the language code


miurahr:
LGTM.

miurahr (approved):

PR #992, opened 2024-03-20
[bugs:#965] Delay autocompleter popup until 100ms after last input

miurahr:
@aaron proposed the fix for OmegaT 5.3 development window,  but it was not merged.

> I've already cut the release, so it won't be included. The next release it could be included in would be 5.3.0, which won't be released probably for a while.

## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?

-  Autosuggestion and IME malfunctioning
- https://sourceforge.net/p/omegat/bugs/965/

## What does this PR change?

- autocomplete pop-ups raised 100ms after input.
-
-

### DIFF
--- a/src/org/omegat/gui/editor/autocompleter/AutoCompleter.java
+++ b/src/org/omegat/gui/editor/autocompleter/AutoCompleter.java
@@ -37,7 +37,7 @@ import javax.swing.JLabel;
 import javax.swing.JPopupMenu;
 import javax.swing.JScrollPane;
 import javax.swing.KeyStroke;
-import javax.swing.SwingUtilities;
+import javax.swing.Timer;
 import javax.swing.UIManager;
 import javax.swing.border.CompoundBorder;
 import javax.swing.border.EmptyBorder;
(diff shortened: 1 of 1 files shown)


miurahr:
This would be worth merging back to 6.0.1

miurahr:
A topic branch was abandoned because original reporter does not give a feedback for the proposed fix.

miurahr:
I got a same situation and I think it is valuable for Japanese translations.

brandelune (approved):

src/org/omegat/gui/editor/autocompleter/AutoCompleter.java
```diff
@@ -93,6 +94,13 @@ public class AutoCompleter implements IAutoCompleter {
     public AutoCompleter(EditorTextArea3 editor) {
         this.editor = editor;
 
+        popupTimer = new Timer(100, e -> {
```
miurahr, line 97: It is better to define constant with a value "100" and replace it with constant.

PR #998, opened 2024-03-27
chore: gradle: replace deprecated convention

miurahr:
## Pull request type

## What does this PR change?

- update build.gradle
- Replace deprecated distsDirName with distsDirectory base plugin convension
- refactor win dists signing tasks to use lazy evaluation with function
- rename mac dist build task names
- add checks for up-to-date status for custom tasks
- release with checksums
- 
## Other information

- [x] check signing task

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1389,13 +1389,13 @@ processResources {
 tasks.register('checksums') {
     def algos = ['SHA-512']
     description = "Generate ${algos.join(', ')} checksums for distribution files"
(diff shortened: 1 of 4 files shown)


miurahr:
This changes replace "doFirst" and "doLast" scripting in gradle as individual tasks. The change speed-ups build process. Gradle detect "up-to-date" status, if so, gradle skip execution. It is why the change here is proposed.

PR #1014, opened 2024-04-27
Improve a few UI items and their corresponding documentation

brandelune:
\<!-- Note: The OmegaT project uses GitHub pull requests for code review only.
The "real" code base is hosted on SourceForge; the GitHub project is a mirror.

If your PR is accepted it will be applied to the SourceForge repository by a
core contributor and the PR ticket will be closed. It will appear to have been
"closed without merging" but that is normal. ---\>

↑ maybe we can remove this part from the template now ?

## What does this PR change?

- The PR improves some UI items that were not finalized for the 6.0 release
- It rewords a number of new items

## Other information

We need to have a style for error messages, with indication of a subject, a verb, etc. so that have a better understanding of what is going on.


brandelune:
Please review the commits separately and I'll squash all when I commit.
Also, most of the items are portable to 6.0.1. So we'll need to handle that too.

Kazephil:
> Please review the commits separately and I'll squash all when I commit.

I'll try to get started on this today, but my wife has plans to steal most of my free time while she's off today and tomorrow, so the bulk of the commits might not get done before Wednesday, when she's at work :sweat_smile:

doc_src/en/OmegaT_Preferences.xml
kosivantsov, line 1: The whole shebang about switching from the clear but long "project-specific" to a much more confusing but shorter "local" was the length of the string, IIRC. Now, "locally-defined" is only one character shorter than "project-specific", but still not nearly as clear.

brandelune: No, it was not the length of the string that brought the modification but the fact that the UI did not have a word for the "default" which is what we call now "global". And *that* was extremely confusing.

Now we have a clear opposition between "global" and "local" all over the UI.

This modification is not about relabeling all the "local" properties as "locally-defined" but just to remove the confusion that comes from having "local" next to "external" in one item. Maybe the second one is not necessary. And maybe there is a better way to fix what I think is confusing about having "local external things".

I'm fine if you oppose the modification. That's the reason why there is a review. But it was not a "shebang" that we had but quite a few thoughtful exchanges over the course of a few months and I don’t have the feeling that OmegaT is worse off with the modifications that we made to the UI and the extensive work that we did on the manual to clarify everything.

brandelune: Don’t worry. I’m fine :-) If you think the changes are not necessary, please say so. I don’t mind removing that part of the proposal and I’m OK with proposals that help "unweirden" that label.

src/org/omegat/Bundle.properties
```diff
@@ -2369,14 +2368,13 @@ TEAM_UPDATE_REPO_ERROR=Error updating repository. Skipping. Error: {0}
 # Save options
 SAVE_DIALOG_TITLE=Saving and Output Options
 PREFS_TITLE_SAVING_AND_OUTPUT=Saving and Output
-SAVE_DIALOG_INTERVAL_LABEL=Project data saving interval
-SAVE_DIALOG_DESCRIPTION=This option sets the project data saving interval, which also applies to team data synchronization.
+SAVE_DIALOG_INTERVAL_LABEL=Project save and team synchronization interval
```
Kazephil, line 2371: "save interval" is better than "saving interval", and I think data is redundant.

brandelune: Ok, but it’s not the whole project that’s saved. And with teams, it’s not the whole project and it’s more than the project memory that’s sync'ed...

Kazephil: Does that matter from the user's point of view?

The end result is that everything that matters is saved/synched, and from an end user point of view, that essentially means the project is saved even if you're not technically saving every single file that's part of the project.

(If this was in the developer documentation, we'd have to be more specific, but either way "project data" is vague, and keeping it it would probably warrant restoring and editing the additional explanation you explicitly took out!)

doc_src/en/OmegaT_EditorsPanes.xml
```diff
@@ -729,14 +729,17 @@ When enabled, all formatting in the source segments are hidden.</programlisting>
...
     <para>When you enter a segment, the first fuzzy match (the one with the
     highest match percentage) is automatically selected. You can press
-    <keycombo><keycap>C</keycap><keycap>2, 3, 4, or 5</keycap>
-    </keycombo> to select a specific match, or cycle through the available matches with
-    <keycombo><keycap>C</keycap><keycap>↑ or ↓</keycap></keycombo>
-    (<keycombo><keycap>A</keycap><keycap>↑ or ↓</keycap></keycombo> on macOS).
-    Alternatively, you can use <link linkend="menus.edit" endterm="menus.edit.title"/>
+    <keycombo><keycap>C</keycap><keycap>2, 3, 4, or 5</keycap> </keycombo> to
+    select a different match. Alternatively, you can use <link
+    linkend="menus.edit" endterm="menus.edit.title"/>
```
Kazephil, line 734: I think I prefer the original version's approach of keeping the keyboard related information together. It could be split into two sentences, with a change of paragraph for the menu alternative, if length was the main concern for the change:

```xml
<para>When you enter a segment, the first fuzzy match (the one with the
 highest match percentage) is automatically selected. You can press
<keycombo><keycap>C</keycap><keycap>2, 3, 4, or 5</keycap>
</keycombo> to select a specific match. You can also cycle through the available matches with
<keycombo><keycap>C</keycap><keycap>↑ or ↓</keycap></keycombo>
(<keycombo><keycap>A</keycap><keycap>↑ or ↓</keycap></keycombo> on macOS).</para>

<para>Alternatively, you can use <link linkend="menus.edit" endterm="menus.edit.title"/>
<link linkend="menus.edit.select.match" endterm="menus.edit.select.match.title"/>
to select a different match.</para>
```

src/org/omegat/Bundle.properties
```diff
@@ -1769,9 +1769,9 @@ SEG_NEW_LN_CO=New language and country
 # TeamOptions
 PREFS_TITLE_TEAM=Team
 
-GUI_TEAM_DESCRIPTION=Your name/ID is used to identify your translations.
+GUI_TEAM_DESCRIPTION=This nickname will be used to identify your translations in the project TMX.
 
-WF_OPTION_INSERT_AUTHOR_PREFIX=&Name/ID:
+WF_OPTION_INSERT_AUTHOR_PREFIX=Nickname:
```
Kazephil, line 1774: I'm not entirely sure about "nickname". It's not wrong, but something about it bugs me a little, though I can't quite put my finger on why. Maybe "alias" would be better?

Given the context, I think line 1772 should probably be:
`Enter the alias you want to use to identify your translations in the memory.`

("Enter the nickname…" if we decide to keep that)

Line 1774 then either becomes `Alias:` or stays as is.

Interestingly, this alias/nickname applies even in non team projects, so this entire section should perhaps be under the `General` heading, but that's beyond the scope of this specific PR.

Kazephil:
OK, I've gone through each individual commit and…
1. added extra commits to propose additional tweaks,
2. added a comment to discuss whether further tweaks are necessary, or
3. did nothing.

The last one simply means that particular commit is fine as is.

On a slightly different note, I just noticed in looking at some of these tweaks that we have a mix of "directory" and "folder". We should probably pick just one of those. If memory serves, "directory" is favoured in Linux, and "folder" is preferred in Windows. Not sure which one macOS tends to use more.

I don't have a strong preference either way, though I suppose that "folder" has the benefit of being a few characters shorter, if that's a consideration. (I haven't checked, but I have a hunch that we have more uses of "folder" than "directory", so keeping the former may be the path of least resistance.)

src/org/omegat/Bundle.properties
brandelune: Now that I think about it, in terms of UI, what the user sees is that the translation is saved (and that's what they want to have saved), so what about "Translation save and synchronization interval"?

doc_src/en/OmegaT_EditorsPanes.xml
brandelune: Ok.

Kazephil:
> ↑ maybe we can remove this part from the template now ?

Just noticed this comment. 

Indeed. We should remove that!

src/org/omegat/gui/preferences/view/SaveOptionsPanel.java
```diff
@@ -56,7 +56,6 @@ private void initComponents() {
         java.awt.GridBagConstraints gridBagConstraints;
 
         intervalDescriptionPanel = new javax.swing.JPanel();
-        intervalDescriptionTextArea = new javax.swing.JTextArea();
```
miurahr, line 59: You should not directly edit `SaveOptionsPanel.java` but you edit it through `NetBeans` GUI editor.
The change break `SaveOptionsPanel.form` divert with generated Java code.

miurahr (changes requested):

src/org/omegat/gui/preferences/view/SaveOptionsPanel.java
brandelune: It builds properly here. I'm OK with editing things in Netbeans but I asked how to use the GUI Editor in the past and nobody answered.

miurahr: I remember a question was too general, and I could not say [RTBM](https://netbeans.apache.org/tutorial/main/kb/docs/java/quickstart-gui/) in public ML for core team.

But I need to say readthedocs of [Coding Styles](https://omegat.readthedocs.io/en/latest/30.CodingStyles/) which can be improved to explain clearer.

brandelune: I don’t know if that's the information I need, but that's the kind of link you could have posted when I asked for information.

Regarding the coding style file, we may need to improve it a bit but I need first to identify the reason why I was stuck.

Thank you for the reference.

brandelune:
There are items here that need to go to releases/6.0 and others that must stay here. Some are not relevant anymore.

I am closing the PR and will rework it.

PR #1027, opened 2024-05-09
[6.0]Temporarily inactivate XLSX for StaX filter (known bug ticket #1260)

t-cordonnier:
Temporarily inactivate Stax filter for Excel until I can debug it
(it is still here but user would have to manually add it in the filters list to use it)

### DIFF
--- a/src/org/omegat/filters4/xml/openxml/MsOfficeFileFilter.java
+++ b/src/org/omegat/filters4/xml/openxml/MsOfficeFileFilter.java
@@ -239,7 +239,7 @@ public class MsOfficeFileFilter extends AbstractZipFilter {
 
     @Override
     public Instance[] getDefaultInstances() {
-        return new Instance[] { new Instance("*.doc?"), new Instance("*.dotx"), new Instance("*.xls?"),
+        return new Instance[] { new Instance("*.doc?"), new Instance("*.dotx"), //new Instance("*.xls?"),
                 new Instance("*.ppt?"), new Instance("*.vsdx") };
     }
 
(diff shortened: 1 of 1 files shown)


src/org/omegat/filters4/xml/openxml/MsOfficeFileFilter.java
```diff
@@ -239,7 +239,7 @@ protected java.util.Comparator<ZipEntry> getEntryComparator() {
 
     @Override
     public Instance[] getDefaultInstances() {
-        return new Instance[] { new Instance("*.doc?"), new Instance("*.dotx"), new Instance("*.xls?"),
+        return new Instance[] { new Instance("*.doc?"), new Instance("*.dotx"), //new Instance("*.xls?"),
```
miurahr, line 242: It  is commented out because of a known bug, I think it is better to add a reference of the bug ticket.
```suggestion
        // disable a link for MS Excel files: https://sourceforge.net/p/omegat/bugs/1260
        return new Instance[] { new Instance("*.doc?"), new Instance("*.dotx"), //new Instance("*.xls?"),
```

miurahr (changes requested):

src/org/omegat/filters4/xml/openxml/MsOfficeFileFilter.java
t-cordonnier: There was a reference to the ticket in the commit message, but ok, no problem for me

miurahr (approved):

PR #1035, opened 2024-05-22
feat: support Unix-like OSes

trombik:
As discussed in [1746](https://sourceforge.net/p/omegat/feature-requests/1746/), here is the patch.

In general, `Platform.isUnix()` is preferred, which will not break Linux support.

`isX11()` was suggested. I did search in the code and the web. `(Unix|UNIX)` is used in several places in the code and the documentation. It looks to me that `UNIX(R)` is used when referring to the registered trademark, and `Unix` when referring to Unix-like and Unix variants. Thus, `isUnix()` is used in the patch.

## Pull request type

## Which ticket is resolved?

- Title #1746 Support U*ix variants
  * URL https://sourceforge.net/p/omegat/feature-requests/1746

## What does this PR change?

- Support *BSD variants.

## Other information

The build was successful, and I am using it. `xdg-open` is working.

### DIFF
--- a/src/org/omegat/util/Platform.java
+++ b/src/org/omegat/util/Platform.java
@@ -53,6 +53,9 @@ public final class Platform {
         WIN64,
         // os.arch=x86, os.name=Windows 7, os.version=6.1
         WIN32,
+        // BSD variants
+        BSD64,
+        BSD32,
         // unknown system
         OTHER
     }
@@ -61,6 +64,8 @@ public final class Platform {
     public static final boolean isWindows;
     public static final boolean isMacOS;
     public static final boolean isLinux;
+    public static final boolean isBSD;
+    public static final boolean isUnix;
     // OS versions
     public static final long osVersion;
     public static final boolean isWindows_10_orLater;
@@ -87,6 +92,8 @@ public final class Platform {
         isWindows = osName.startsWith("windows");
         isMacOS = osName.startsWith("mac");
         isLinux = osName.startsWith("linux");
+        isBSD = osName.startsWith("bsd");
+        isUnix = isLinux || isBSD;
 
         // OS versions
         osVersion = scanVersion(System.getProperty("os.version"));
@@ -107,7 +114,7 @@ public final class Platform {
         isJava_17_orLater = (javaVersion >= toVersion(17, 0, 0, 0));
 
         // UI toolkits
-        isKDE = (isLinux && System.getenv("KDE_FULL_SESSION") != null);
+        isKDE = (isUnix && System.getenv("KDE_FULL_SESSION") != null);
 
         // Windows 11 detection is implemented in Java 8u321, 11.0.14, 17.0.2
         // and 18 (or later).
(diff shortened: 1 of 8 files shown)


src/org/omegat/util/Platform.java
```diff
@@ -87,6 +92,8 @@ public enum OsType {
         isWindows = osName.startsWith("windows");
         isMacOS = osName.startsWith("mac");
         isLinux = osName.startsWith("linux");
+        isBSD = osName.startsWith("bsd");
+        isUnix = isLinux || isBSD;
```
miurahr, line 96: `isUnixLike` is better abstraction. see https://github.com/search?q=isUnixLike&type=code

```suggestion
        isUnixLike = isLinux || isBSD;
```

src/org/omegat/util/Platform.java
```diff
@@ -144,6 +153,20 @@ public static boolean isLinux() {
...
+     */
+    public static boolean isBSD() {
+        return isBSD;
+    }
+
+    /**
+     * Returns true if running on Unix variants
+     */
+    public static boolean isUnix() {
+        return isUnix;
```
miurahr, line 167:
```suggestion
    public static boolean isUnixLike() {
        return isUnixLike;
```

src/org/omegat/util/Platform.java
```diff
@@ -107,7 +114,7 @@ public enum OsType {
         isJava_17_orLater = (javaVersion >= toVersion(17, 0, 0, 0));
 
         // UI toolkits
-        isKDE = (isLinux && System.getenv("KDE_FULL_SESSION") != null);
+        isKDE = (isUnix && System.getenv("KDE_FULL_SESSION") != null);
```
miurahr, line 117:
```suggestion
        isKDE = (isUnixLike && System.getenv("KDE_FULL_SESSION") != null);
```

src/org/omegat/util/Platform.java
```diff
@@ -61,6 +64,8 @@ public enum OsType {
     public static final boolean isWindows;
     public static final boolean isMacOS;
     public static final boolean isLinux;
+    public static final boolean isBSD;
+    public static final boolean isUnix;
```
miurahr, line 68:
```suggestion
    public static final boolean isUnixLike;
```

src/org/omegat/util/StaticUtils.java
```diff
@@ -312,7 +312,7 @@ public static String getConfigDir() {
             }
             // Check for UNIX varieties
             // Solaris is generally detected as SunOS
-        } else if (Platform.isLinux()) {
+        } else if (Platform.isUnix()) {
```
miurahr, line 315:
```suggestion
        } else if (Platform.isUnixLike()) {
```

miurahr (changes requested):

src/org/omegat/util/Platform.java
trombik: should I `git push --force` the fix, or simply commit another one with the fix?

miurahr: It is ok to simply commit another one. I will squash and merge when all is ok.

trombik:
built and tested on my machine.

miurahr (approved):

PR #1039, opened 2024-05-26
refactor: EditorController to access menu through IMainMenu API

miurahr:
In order to implement acceptance GUI test, it is important to decouple UI components from real main window object.
EditorController enable/disable some menu items through `mainwindow.menu` access.
This change decouple it.

To make things working, I add `IMainMenu#enableMenuItem` API. 
This allow other components to enable/disable menu item like as `Core.getMainWindow().getMainMenu().enableMenuItem(name, true)` 

## Pull request type

- Other (describe below)
refactor

## What does this PR change?

- Add IMainMenu#enableMenuItem API
- Implement enableMenuItem methods
- EditorContoller enable/disable menu item through the API
-
-
-

### DIFF
--- a/src/org/omegat/gui/main/IMainMenu.java
+++ b/src/org/omegat/gui/main/IMainMenu.java
@@ -6,6 +6,7 @@
  Copyright (C) 2008 Alex Buloichik
                2011 Didier Briel
                2016 Aaron Madlon-Kay
+ 2024 Hiroshi Miura
                Home page: https://www.omegat.org/
                Support center: https://omegat.org/support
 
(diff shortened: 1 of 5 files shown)


src/org/omegat/gui/main/IMainMenu.java
```diff
@@ -37,23 +38,70 @@
...
 
+    /**
+     * Enable/Disable menu item.
+     * When it is JCheckBoxMenuItem and JRadioButtonMenuItem, a specified
+     * item will be selected.
+     *
+     * @param name    a component name of menu item.
+     * @param enabled true when make it enabled, otherwise make it disabled.
+     */
+    void enableMenuItem(String name, boolean enabled);
```
miurahr, line 104: Now we have two type of menu manipulation interface.
1. `getMenu(MenuKey)` 
2. `enableMenuItem(String name, boolean)`

When we define `name` for all JMenu components, we can replace No.1 with `getMenu(String name)`

miurahr: We can implement menu updates in Action class for these menu items.
The object can detect changes through ProperyChangeListener interface.

PR #1046, opened 2024-05-31
refactor: GlossaryTextArea to use Actions classes

miurahr:
## Pull request type

- Bug fix -> [bug]
- Other (describe below)

## Which ticket is resolved?

- Title ex. fix...
  * URL ex. https://...
 
- Title
  * URL

## What does this PR change?

-
-
-

### DIFF
--- a/src/org/omegat/gui/glossary/GlossaryTextArea.java
+++ b/src/org/omegat/gui/glossary/GlossaryTextArea.java
@@ -37,8 +37,6 @@ import java.awt.Frame;
 import java.awt.GraphicsEnvironment;
 import java.awt.Point;
 import java.awt.Window;
-import java.awt.event.ActionEvent;
-import java.awt.event.ActionListener;
 import java.awt.event.KeyEvent;
 import java.awt.event.MouseAdapter;
 import java.awt.event.MouseEvent;
@@ -70,6 +68,10 @@ import org.omegat.core.data.StringEntry;
 import org.omegat.gui.common.EntryInfoThreadPane;
 import org.omegat.gui.dialogs.CreateGlossaryEntry;
 import org.omegat.gui.editor.EditorUtils;
+import org.omegat.gui.glossary.actions.AddEntryAction;
+import org.omegat.gui.glossary.actions.InsertSectionAction;
+import org.omegat.gui.glossary.actions.SettingsNotifications;
+import org.omegat.gui.glossary.actions.SettingsOpenFileAction;
 import org.omegat.gui.main.DockableScrollPane;
 import org.omegat.gui.main.IMainWindow;
 import org.omegat.gui.shortcuts.PropertiesShortcuts;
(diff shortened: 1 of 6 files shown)


src/org/omegat/gui/main/ProjectUICommands.java
```diff
@@ -1315,6 +1316,40 @@ public static void doWikiImport() {
         }
     }
 
+    public static void openWritableGlossaryFile(boolean parent) {
```
miurahr, line 1319:
```suggestion
    public static void openWritableGlossaryFile(boolean parentDirectory) {
```

It is better argument name to avoid ambiguous against parent window.

src/org/omegat/gui/glossary/GlossaryTextArea.java
```diff
@@ -451,4 +432,73 @@ public void actionPerformed(ActionEvent e) {
...
+
+        private SettingsOpenFileAction() {
+            super(OStrings.getString("GUI_GLOSSARYWINDOW_SETTINGS_OPEN_FILE"), OStrings.getLocale());
+            final String key = "projectAccessWriteableGlossaryMenuItem";
+            putValue(Action.ACTION_COMMAND_KEY, key);
+            try {
+                KeyStroke keyStoroke = PropertiesShortcuts.getMainMenuShortcuts().getKeyStroke(key);
+                putValue(Action.ACCELERATOR_KEY, keyStoroke);
+            } catch (Exception ignored) {
+            }
```
miurahr, line 494:
```suggestion
```
No need to set shortcut key for pop-ups

miurahr:
Now ready for review and merge.

miurahr:
Updated with a feedback in dev-ML discussion that @t-cordonnier suggests lambda instead of old fashioned Action class

src/org/omegat/gui/glossary/GlossaryTextArea.java
```diff
@@ -451,4 +446,29 @@ public void actionPerformed(ActionEvent e) {
                 .setPreference(Preferences.GLOSSARY_SORT_BY_LENGTH, sortOrderLocLength.isSelected()));
         menu.add(sortOrderLocLength);
     }
+
+    private static JMenuItem createMenuItem(String titleKey, String command, Runnable action) {
```
t-cordonnier, line 450: If you plan to use the same methodology for main menu, you should move these methods to a common class, for example org.omegat.gui.common.MenuFactory

src/org/omegat/gui/glossary/GlossaryTextArea.java
```diff
@@ -451,4 +446,29 @@ public void actionPerformed(ActionEvent e) {
...
+        return createMenuItem(titleKey, command, ev -> {
+            Log.logInfoRB("LOG_MENU_CLICK", ev.getActionCommand());
+            action.run();
+        });
+    }
+
+    private static JMenuItem createMenuItem(String titleKey, String command, ActionListener actionListener) {
+        JMenuItem result = new JMenuItem();
+        setLocalizedText(result, OStrings.getString(titleKey));
+        result.setActionCommand(command);
```
t-cordonnier, line 460: I think also action command is not necessary. In the past it was used when same action listener was shared by multiple menus, which is the case in MainWindowMenu actually.
But with a distinct lambda for each menu, does not seem necessary

t-cordonnier (changes requested):
For action command I let you decide, but if you plan to re-use the methods in the future you should move them in a dedicated class right now.

miurahr:
I want to give up the change here. It is just a refactoring.

PR #1049, opened 2024-06-02
feat: update tmx writer, escaper, and unit test

miurahr:
## Pull request type

## Which ticket is resolved?

- TMX writer output in 6.1 should be the same as in 6.0
  * https://sourceforge.net/p/omegat/feature-requests/1749/

## What does this PR change?

- update TMX test expectation for level2 write
    - use double quote for attribute
    - format indentation and empty line
- Add TmxEscapingWriterFactory class and unit test with ordinary string, "< & >", Surrogate Pair and invalid one.
- Use TmxEscapingWriterFactory in TMXWriter2 for TEXT writer

## Other information

`com.sun.xml.bind.marshaller#escape` which is ultimately OmegaT 6.0 and before use is as follows;
```java
public void escape(char[] ch, int start, int length, boolean isAttVal, Writer out) throws IOException {
        int limit = start + length;

        for(int i = start; i < limit; ++i) {
            char c = ch[i];
            if (c == '&' || c == '<' || c == '>' || c == '\r' || c == '\n' && isAttVal || c == '"' && isAttVal) {
                if (i != start) {
                    out.write(ch, start, i - start);
                }

                start = i + 1;
                switch (ch[i]) {
                    case '\n':
                    case '\r':
                        out.write("&#");
                        out.write(Integer.toString(c));
                        out.write(59);
                        break;
                    case '"':
                        out.write("&quot;");
                        break;
                    case '&':
                        out.write("&amp;");
                        break;
                    case '<':
                        out.write("&lt;");
                        break;
                    case '>':
                        out.write("&gt;");
                        break;
                    default:
                        throw new IllegalArgumentException("Cannot escape: '" + c + "'");
                }
            }
        }

        if (start != limit) {
            out.write(ch, start, limit - start);
        }

    }

```

### DIFF
--- a/src/org/omegat/util/TMXWriter2.java
+++ b/src/org/omegat/util/TMXWriter2.java
@@ -30,6 +30,7 @@ package org.omegat.util;
 
 import static com.ctc.wstx.api.WstxOutputProperties.P_OUTPUT_ESCAPE_CR;
 import static org.codehaus.stax2.XMLOutputFactory2.P_AUTOMATIC_EMPTY_ELEMENTS;
+import static org.codehaus.stax2.XMLOutputFactory2.P_TEXT_ESCAPER;
 
 import java.io.BufferedOutputStream;
 import java.io.File;
@@ -109,6 +110,7 @@ public class TMXWriter2 implements AutoCloseable {
         factory = XMLOutputFactory.newInstance();
         factory.setProperty(P_OUTPUT_ESCAPE_CR, false);
         factory.setProperty(P_AUTOMATIC_EMPTY_ELEMENTS, true);
+        factory.setProperty(P_TEXT_ESCAPER, new TmxEscapingWriterFactory());
 
         out = new BufferedOutputStream(Files.newOutputStream(file.toPath()));
         xml = factory.createXMLStreamWriter(out, StandardCharsets.UTF_8.name());
--- a/src/org/omegat/util/TmxEscapingWriterFactory.java
+++ b/src/org/omegat/util/TmxEscapingWriterFactory.java
@@ -28,7 +28,6 @@ package org.omegat.util;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.io.OutputStreamWriter;
-import java.io.UnsupportedEncodingException;
 import java.io.Writer;
 import java.nio.charset.StandardCharsets;
 import java.util.Arrays;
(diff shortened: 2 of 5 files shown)


miurahr:
@brandelune  this PR is created by your request. Have you already lost the interest?

miurahr:
The change here make a modification and an override of a deep and core part of XML writing, marshal. When there is a potential bug, it can break things easily.  To reduce sub effect in another place, it only apply a TMXWriter2, TMX writing class. So there is no change in preferences, and other XML related features.

brandelune:
> @brandelune this PR is created by your request. Have you already lost the interest?

Thank you for the nice reminder :-)

brandelune:
> The change here make a modification and an override of a deep and core part of XML writing, marshal. When there is a potential bug, it can break things easily. To reduce sub effect in another place, it only apply a TMXWriter2, TMX writing class. So there is no change in preferences, and other XML related features.

@t-cordonnier do you have an opinion on this ?

miurahr:
> @t-cordonnier do you have an opinion on this ?

Maybe no opinion.

brandelune:
> The change here make a modification and an override of a deep and core part of XML writing, marshal. When there is a potential bug, it can break things easily. To reduce sub effect in another place, it only apply a TMXWriter2, TMX writing class. So there is no change in preferences, and other XML related features.

Do you mean that there is no option in the new library to make the behavior the same as in 6.0?

I don’t think we should take risks with such overrides. So I’m still eager to have @t-cordonnier’s opinion.

miurahr:
@t-cordonnier what your opinion?

t-cordonnier:
Sorry for late reply but I had not seen this and I was buzy.
Globally I can agree with this suggestion, I have only added some margin comments.

miurahr:
@brandelune what is your conclusion?

brandelune (approved):

miurahr:
What do you mean? @brandelune do you want to take a risk?

I don't want to merge it if members don't have substantial demand and a commitment to responding to the issues the change brings.

 I have a position that we don't need to make byte-level compatibility with the old, buggy behavior of the XML output.
The project manager should use the XMLDiff tool to monitor the changes instead of the standard diff tool.

brandelune:
> What do you mean? @brandelune do you want to take a risk?
> 
> I don't want to merge it if members don't have substantial demand and a commitment to responding to the issues the change brings.
> 
> I have a position that we don't need to make byte-level compatibility with the old, buggy behavior of the XML output. The project manager should use the XMLDiff tool to monitor the changes instead of the standard diff tool.

My position is :

if Thomas agrees with this suggestion of yours, then I agree with it too.

That is the reason why I approved.

Since you are in charge of making sure that we have robust code, my approval does not mean that I am forcing you to make choices that you don’t like, but that I am not able to make a technical judgement on this issue, hence I consider that the most qualified persons here (you and Thomas), should have the last word.

Here is what I wrote on the list, in june 2024:

_If the library has standard options regarding XML output, it would be a 
nice thing to have backward compatibility, but as you say, as long as 
the output is XML compliant, it should not be a requirement. For ex. if 
we find an extremely efficient XML library that does not offer backward 
compatibility, I think that we should prioritise efficiency over 
backward compatibility._

miurahr:
Ok, I will merge it. Please take risks with you.

PR #1052, opened 2024-06-07
feat: introduce Lucene-Hunspell spellchecker module

miurahr:
## Pull request type

## What does this PR change?

-
-
-


miurahr:
There is a problem for dependency version conflict. OmegaT 6.1 have a mechanism to encapsulate modules/plugins with individual class loader and avoid  referencing from other module/core to look unwanted version of dependency.
It is mandatory to merge here lucene-hunspell module to OmegaT, OmegaT core should separate LT and Lucene to individual modules.

miurahr:
When this PR is merged into LT project, we can use it.
https://github.com/languagetool-org/languagetool/pull/10810

miurahr:
Now it should be able to replace the current Hunspell spellchecker. To do so, it should support the dictionary that the current Hunspell spellchecker supports.

PR #1055, opened 2024-06-12
Refector: MainMenuHandler: action handlers takes ActionEvent object

miurahr:
## Pull request type

- Other (describe below)
refactor

## What does this PR change?

- `MainMenuHandler`: avoid direct reference to real MainWindow object when getting a menu item status of action event occurred, by passing ActionEvent object to handler method, by using `ActionEvent#getSource`. 
- Make static fields of `Core`  as a  private to allow further refactoring.
- Instantiate `MainWindowMenu` in `Core` and put it in private field
- Add `Core#getMainMenu` and change `MainWindow#getMainMenu` to proxy it


src/org/omegat/core/Core.java
```diff
@@ -299,10 +313,58 @@ public static void initializeConsole(final Map<String, String> params) throws Ex
         mainWindow = new ConsoleWindow();
     }
 
+    @SuppressWarnings("unchecked")
+    private static void initMainMenu(MainWindow mw) {
```
miurahr, line 317: It can be placed in MainWindowUI utility class.

src/org/omegat/gui/main/MainWindowMenuHandler.java
```diff
@@ -760,7 +764,7 @@ public void gotoEditorPanelMenuItemActionPerformed() {
      */
     public void gotoSegmentMenuItemActionPerformed() {
```
miurahr, line 765: We can unify all the action handler to take ActionEvent object.
It makes things simpler, even a patch become large.

miurahr:
There are too many refactoring in single PR. It is required to split into  series of Pull-Requests.

miurahr:
this is over-engineered, so I think I can close here and raise another PR in proper size.

PR #1059, opened 2024-06-13
refactor: add IMainWindow#resetDesktopLayout

miurahr:
extend IMainWindow API to add `resetDesktopLayout` to help menu handler to avoid direct dependency for raw MainWindow object to help test code mocking it.
 
## Pull request type

- Other (describe below)
refactor

## What does this PR change?

- Extend IMainWindow API
- Remove dependency to "raw" mainWindow from MainWindowMenuHandler
- Update test codes according to API

### DIFF
--- a/src/org/omegat/gui/main/IMainWindow.java
+++ b/src/org/omegat/gui/main/IMainWindow.java
@@ -204,4 +204,9 @@ public interface IMainWindow {
      * Retrieve main docking desktop.
      */
     DockingDesktop getDesktop();
+
+    /**
+     * Restores the main window layout to the default values.
+     */
+    default void resetDesktopLayout() {}
 }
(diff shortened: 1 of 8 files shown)


src/org/omegat/gui/main/IMainWindow.java
```diff
@@ -204,4 +204,9 @@ public interface IMainWindow {
      * Retrieve main docking desktop.
      */
     DockingDesktop getDesktop();
+
+    /**
+     * Restores the main window layout to the default values.
+     */
+    default void resetDesktopLayout() {}
```
miurahr, line 211: We add it as `default`  which keeps backward compatibility for plugins compiled with previous versions of OmegaT.

PR #1063, opened 2024-06-15
refactor: extend MainWindow API, and move MenuHandler functions in ProjectUICommands, then introduce SearchWindowManager

miurahr:
## Pull request type

- Other (describe below)
refactor 

## What does this PR change?

- add IMainWindow#saveDesktopLayout and IMainWindow@resetDesktopLayout
- define MenuHandler functions in ProjectUICommands  
- introduce SearchWindowManager and move related functions to it

### DIFF
--- a/src/org/omegat/gui/main/MainWindow.java
+++ b/src/org/omegat/gui/main/MainWindow.java
@@ -75,6 +75,7 @@ import org.omegat.core.events.IApplicationEventListener;
 import org.omegat.core.events.IProjectEventListener;
 import org.omegat.core.matching.NearString;
 import org.omegat.gui.matches.IMatcher;
+import org.omegat.gui.search.SearchWindowManager;
 import org.omegat.util.OStrings;
 import org.omegat.util.Preferences;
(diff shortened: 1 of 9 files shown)


src/org/omegat/gui/main/MainWindow.java
```diff
@@ -231,6 +233,14 @@ public void windowDeactivated(WindowEvent we) {
         }
     }
 
+    /**
+     * {@inheritDoc}
+     */
+    @Override
```
miurahr, line 239:
```suggestion
    @Override
```

src/org/omegat/gui/main/MainWindow.java
```diff
@@ -179,6 +180,7 @@ public void onApplicationShutdown() {
     /**
      * {@inheritDoc}
      */
+    @Override
```
miurahr, line 183:
```suggestion
    @Override
```

miurahr (commented):
IhhreitDoc automatically done, so we do not need it.

PR #1077, opened 2024-07-06
feat: show only a current session log when Help>Log

miurahr:
"Help >  Log..."  dialog shows current session log.

## Pull request type

## Which ticket is resolved?

- Add timestamp to log
- https://sourceforge.net/p/omegat/feature-requests/1734/

- Log filename to be date-time based
- https://sourceforge.net/p/omegat/feature-requests/1753/

- Dev-ML
- https://sourceforge.net/p/omegat/mailman/omegat-development/thread/6631DFBB-8304-4C28-B08D-5730371F39F1%40traductaire-libre.org/#msg58792827

## What does this PR change?

- Change log file name to "OmegaT_(session id)_yyyyMMddHHmmss.log"
- Log Retention configuration: default 10 days
- Refactoring LogDialog class, introduce LogDialogController class


src/org/omegat/gui/dialogs/LogDialog.java
```diff
@@ -104,6 +102,33 @@ protected void done() {
         }.execute();
     }
 
+    private String loadCurrentSessionLogFromFile(File logLocation) {
```
miurahr, line 105: I should add a unit test for the function.

src/org/omegat/gui/dialogs/LogDialog.java
```diff
@@ -87,11 +89,7 @@ public LogDialog(java.awt.Frame parent) {
         new SwingWorker<String, Void>() {
             @Override
             protected String doInBackground() throws Exception {
-                try (FileInputStream fis = new FileInputStream(logLocation)) {
-                    return IOUtils.toString(fis, StandardCharsets.UTF_8);
-                } catch (Exception e) {
-                    return "";
-                }
+                return loadCurrentSessionLogFromFile(logLocation);
```
miurahr, line 92: Is this feature consist with SaveAs feature? the latter saves all the current file, but here show current session.

src/org/omegat/gui/dialogs/LogDialog.java
```diff
@@ -29,15 +29,17 @@
 
 import java.awt.Font;
 import java.io.File;
-import java.io.FileInputStream;
 import java.io.IOException;
-import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.stream.Stream;
```
miurahr, line 35: This file should be generated by NetBeans GUI editor

miurahr:
image
This is an example of updated dialog.

src/org/omegat/logger.properties
```diff
@@ -8,16 +8,19 @@ org.omegat.level = ALL
...
 #org.omegat.util.logging.OmegaTLogFormatter.mask=$time: $threadName [$level] $key $text
+#org.omegat.util.logging.OmegaTLogFormatter.mask=$mark: $level: $text $key
 
-org.omegat.util.logging.OmegaTLogFormatter.timeFormat=HH:mm:ss,SSSS
+org.omegat.util.logging.OmegaTLogFormatter.timeFormat=HH:mm:ss.SSS
 
 java.util.logging.ConsoleHandler.formatter = org.omegat.util.logging.OmegaTLogFormatter
 
 org.omegat.util.logging.OmegaTFileHandler.formatter = org.omegat.util.logging.OmegaTLogFormatter
 org.omegat.util.logging.OmegaTFileHandler.size=1048576
```
t-cordonnier, line 20: Are these lines still useful if we change the retention policy? What happens if the file os over 1 MB?

t-cordonnier (approved):
Except the small comment I added (to be tested) I approve this change

brandelune (commented):
We’ll need additions to the relevant places in the manual:

https://omegat.sourceforge.io/manual-standard/en/chapter.appendices.html#configuration.folder

src/org/omegat/logger.properties
miurahr: These two property values are used in removed `rotate` method in the handler. Now it is not used.

miurahr: adjust it by removing two property in the file.

miurahr:
> We’ll need additions to the relevant places in the manual:
> 
> https://omegat.sourceforge.io/manual-standard/en/chapter.appendices.html#configuration.folder

Add sentence

src/org/omegat/logger.properties
miurahr: We can respect `count` property when removing old files, we always keep `count` number of recent files, even if retention date passed.

miurahr: When respecting `count`, it is not good for file name convention because it is not able to sort of creation date by file name, which consist of session id first.

Now I would like to propose a change of file name format to be `OmegaT_yyyyMMdd-hhmmss_sessionId.log`

src/org/omegat/util/logging/OmegaTFileHandler.java
```diff
@@ -131,48 +141,44 @@ public synchronized void close() throws SecurityException {
...
+            return;
+        }
+
+        long cutoffTime = System.currentTimeMillis() - retention;
+
+        for (File file : files) {
+            try {
+                Path filePath = Paths.get(file.getAbsolutePath());
+                BasicFileAttributes attrs = Files.readAttributes(filePath, BasicFileAttributes.class);
+                if (attrs.creationTime().toMillis() < cutoffTime) {
```
miurahr, line 176: Here I use `creationTime`, Is it better to use date time from file name?

brandelune:
> > We’ll need additions to the relevant places in the manual:
> > https://omegat.sourceforge.io/manual-standard/en/chapter.appendices.html#configuration.folder
> 
> Add sentence

Thank you.

I fear that the time stamp format is quite difficult to read. If we can add separation marks it would be better:

OmegaT_12943_20240709-074234.log could be something like:

OmegaT_12943_2024.07.09-07.42.34.log

@t-cordonnier what do you think ?

src/org/omegat/logger.properties
t-cordonnier: Good idea, I agree

t-cordonnier: > Now I would like to propose a change of file name format to be OmegaT_yyyyMMdd-hhmmss_sessionId.log

For me the best retention policy would be based on the size of the directory, but I suppose it is harder to implement?

t-cordonnier:
> @t-cordonnier what do you think ?

Usually I avoid to have dots in file names, as some tools may interpret them as beginning of extension. You can play with alternance between - and _ instead
Also in this timestamp we can remove seconds: I don't think lot of people remember so precisely when they start the application, nor that they start lot of instances in the same second. That is different in file contents, where milliseconds are useful to check whenever some processes are longer than expected.

brandelune:
> > @t-cordonnier what do you think ?
> 
> Usually I avoid to have dots in file names, as some tools may interpret them as beginning of extension. You can play with alternance between - and _ instead Also in this timestamp we can remove seconds: I don't think lot of people remember so precisely when they start the application, nor that they start lot of instances in the same second. That is different in file contents, where milliseconds are useful to check whenever some processes are longer than expected.

Ok, so would that work better for you ?

`OmegaT_12943_2024-07-09_07-42.log`

I think readability is still good enough for the purpose.

@miurahr what do you think ?

t-cordonnier:
> Ok, so would that work better for you ?
> 
> `OmegaT_12943_2024-07-09_07-42.log`
> 

OK for me

src/org/omegat/logger.properties
miurahr: > For me the best retention policy would be based on the size of the directory, but I suppose it is harder to implement?

I don't have an idea how to implement it. You are welcome to propose new  method to do it by a new property.

miurahr:
https://github.com/omegat-org/omegat/pull/1077#discussion_r1669759507 and https://github.com/omegat-org/omegat/pull/1077#issuecomment-2217118829

`OmegaT_2024-07-09_07-42_12943.log`

src/org/omegat/logger.properties
miurahr: Which behavior is more natural?

1. always keep minimum 10 log files and then retention based on time
2. retention based on file date but maximum files to be kept is 10

PR #950, opened 2024-02-06
[RFE#1736] feat: Alternative translation highlight

chelobaka:
## Pull request type

Please mark github LABEL of the type of change your PR introduces:

- Feature enhancement -> [enhancement]

## Which ticket is resolved?

- Alternative translation highlight
  * https://sourceforge.net/p/omegat/feature-requests/1736/
 

## What does this PR change?

- This PR implements new highlighter for marking segments with an alternative translation.
- GUI bits one might expect for a highlighter like changing color in settings and default color values for light/dark themes.

PR #1082, opened 2024-07-11
[RFE#1736] feat: Alternative translation highlight

miurahr:
Re-working on the proposal from @chelobaka 

I have reverted some changes from my review comment, because we now understand mark is not perfect to be a plugin.

I tweak several places to fix test errors which is caused by the proposed changes.

## Pull request type

## Which ticket is resolved?

- Alternative translation highlight
  * https://sourceforge.net/p/omegat/feature-requests/1736/
 

## What does this PR change?

- This PR implements new highlighter for marking segments with an alternative translation.
- GUI bits one might expect for a highlighter like changing color in settings and default color values for light/dark themes.

## Other information

Re-submit PR to modify #950

### DIFF
--- a/Plugins.properties
+++ b/Plugins.properties
@@ -95,6 +95,7 @@ tokenizer=org.omegat.tokenizer.DefaultTokenizer \
     org.omegat.tokenizer.LuceneSwedishTokenizer \
     org.omegat.tokenizer.LuceneThaiTokenizer \
     org.omegat.tokenizer.LuceneTurkishTokenizer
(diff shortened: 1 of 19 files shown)


miurahr:
got menu handling api error that return wrong menu item position.
```
expected:<[Display Modification Info]> but was:<[Use Aggressive Font Fallback]>	
```

It is caused by inserting menu item in the menu.

miurahr:
Bundle have already been reviewed in #950 so I think it is ok.

brandelune (approved):

PR #1106, opened 2024-08-18
chore: add unescape diff tool

miurahr:
Developers can use the unescaped diff tool by configuring git like

```bash
git config difftool.udiff.cmd 'config/git/udiff.sh $LOCAL $REMOTE'
git config alias.udiff 'difftool --tool=udiff'
```

then developers can run like

```bash
git udiff src/org/omegat/Bundle_(lang).properties
```

Developers and localizers happily see a diff with native languages.

## Type

Others

## Other information

Developers must install `perl`, `diff`, and `bash` on the local environment.

### DIFF
--- a/config/git/udiff.sh
+++ b/config/git/udiff.sh
@@ -1,0 +1,21 @@
+#!/bin/bash
+
+# Unescape Unicode sequences in a properties file
+unescape() {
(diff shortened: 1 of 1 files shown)


miurahr:
It is better to update developer manual.

PR #1117, opened 2024-08-24
refactor: simplify IParseCallback#addEntry

miurahr:
Use default for old addEntry method signature to reduce code duplications.

## Pull request type

- Other (describe below)
refactor

## What does this PR change?

- Update IParseCallback#addEntry to be default
- Remove overrides of it
- Update ILIASFilter and MagentoFilter to use new one

### DIFF
--- a/src/org/omegat/filters2/IParseCallback.java
+++ b/src/org/omegat/filters2/IParseCallback.java
@@ -38,10 +38,10 @@ import org.omegat.core.data.ProtectedPart;
  */
 public interface IParseCallback {
     /**
-     * Read entry from source file, with arbitrary (optional) properties
+     * Read entry from a source file, with arbitrary (optional) properties
      *
      * @param id
-     *            ID in source file, or null if ID not supported by format
+     *            ID in a source file, or null if ID not supported by format
      * @param source
      *            source entry text
      * @param translation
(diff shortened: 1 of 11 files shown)


src/org/omegat/filters2/IParseCallback.java
```diff
@@ -55,24 +55,46 @@ public interface IParseCallback {
...
+     *            generated 'reference' TMX, a special TMX that is used as extra
+     *            reference during translation.
+     * @param comment
+     *            entry's comment, if format supports it
+     * @param filter
+     *            filter which produces entry
      */
-    void addEntry(String id, String source, String translation, boolean isFuzzy, String comment,
-            IFilter filter);
+    default void addEntry(String id, String source, String translation, boolean isFuzzy, String comment,
```
miurahr, line 94: Compatibility is important.  We need to check whether a change from standard method to `default` keeps its backward compatibility.

src/org/omegat/filters2/IParseCallback.java
```diff
@@ -55,24 +55,46 @@ public interface IParseCallback {
      * @param filter
      *            filter which produces entry
      * @param protectedParts
-     *            (since 3.0.6) protected parts
+     *            protected parts
      */
     void addEntryWithProperties(String id, String source, String translation, boolean isFuzzy, String[] props,
```
miurahr, line 60: A method `addEntryWIthProperties` has 8 parameters, that breaks a checkstyle rule.  Exceeding a number of argument is a signature of design error.

src/org/omegat/filters2/IParseCallback.java
miurahr: Binary compatibility guide said
https://docs.oracle.com/javase/specs/jls/se17/html/jls-13.html

> 13.5.7. Interface Method Declarations
> (snip)
> Adding a default method, **or changing a method from abstract to default, does not break compatibility with pre-existing binaries**, but may cause an IncompatibleClassChangeError if a pre-existing binary attempts to invoke the method. This error occurs if the qualifying interface of the method invocation, K, is a subinterface of two interfaces, I and J, where both I and J declare a default method with the same signature and result, and neither I nor J is a subinterface of the other.

miurahr: It considered as safe, because the interface `IParseCallback` is for OmegaT core internal, and rarely, or nowhere, used outside the core.

miurahr: Question is resolved.

PR #1125, opened 2024-08-26
fix: LTmod's IllegalArgument on en-AU

miurahr:
- When loading "en-AU" LT language, it failed with IllegalArgumentException because of missing "en-US".

## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?

- Failed to load "en-AU" LT language by IllegalArgumentException
- https://sourceforge.net/p/omegat/bugs/1265/

## What does this PR change?

- LanguageManager#getLTLanguage loads not only specified language, such as "en-AU" also loads variants such as "en-US", "en-GB" and "en"

## Other information

- [x] write test case with "en-AU"

### DIFF
--- a/src/org/omegat/languagetools/LanguageManager.java
+++ b/src/org/omegat/languagetools/LanguageManager.java
@@ -44,18 +44,39 @@ public final class LanguageManager {
         LT_LANGUAGE_CLASSES.put(lang, fqcn);
     }
 
+    /**
+     * Get LanguageTool language.
+     * <p>
+     *     This code made side effect to load LT modules eventually.
+     *     It loads a specified language-country.
+     *     It also loads a variant of countries, such as "en-US" for "en-AU".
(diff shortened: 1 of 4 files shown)


src/org/omegat/languagetools/LanguageManager.java
```diff
@@ -44,18 +44,39 @@ public static void registerLTLanguage(String lang, String fqcn) {
...
     }
 
+    /**
+     * Get LanguageTool language.
+     * <p>
+     *     This code made side effect to load LT modules eventually.
+     *     It loads a specified language-country.
+     *     It also loads a variant of countries, such as "en-US" for "en-AU".
+     *     And also it loads language code module such as "en".
+     *     It is because sone language definition depends on another language,
```
miurahr, line 54:
```suggestion
     *     It is because some language definition depends on another language,
```
typo

PR #1127, opened 2024-08-26
Locked segments

t-cordonnier:
This PR implements the notion of locked segment: when you are in such a segment inside the editor, any attempt to add or delete characters will be rejected - only the navigation keys work, you can escape the segment of course!

## Pull request type

## Which ticket is resolved?

https://sourceforge.net/p/omegat/feature-requests/1745/ (for tm/enforce)

I am not sure there is a specific ticket for xliff / final but we already discussed it in the mailing lists

About https://sourceforge.net/p/omegat/bugs/1215/ : it does not directly solve the bug, which is in Okapi filter, but once the patch is applied we may have some info we can give to them so that they can use the new feature to solve the bug.


src/org/omegat/Bundle.properties
```diff
@@ -1803,6 +1805,7 @@ WG_SAVE_ORIGIN=Save origin of &translation
 WG_INITIAL_SEGMENT_LOAD_COUNT=Initial number of loaded segments:
 WG_INITIAL_SEGMENT_LOAD_COUNT_TOOLTIP=When a single file has more segments than this, additional segments will be loaded on-demand as you scroll.
 WG_PARA_MARK=Paragraph delimiter:
+WG_SUPPORT_LOCKED_SEGMENTS=Support for locked segments
```
Kazephil, line 1808: I think "Allow segment locking" would be better for this one, if I understand the intent correctly.

miurahr:
I am waiting @t-cordonnier  to make the PR to pass the CI tests. any progress?
And also the proposal conflict with master branch. Please merge and resolve.

t-cordonnier:
Sorry but as I said, I am in holidays now so not before next week

Hiroshi, In the meantime you can have a look to the comments I had left in answer to your change requests: I remember that for some of them your intentions we're not clear. Or if you only want me to   work on what github says, let me know

miurahr:
@t-cordonnier you can click GitHub interface of the test list a place "GitHub CI"
image

Then you are going to https://github.com/omegat-org/omegat/pull/1127/checks?check_run_id=36114418664
And you will observe the error detail as like
image

The test case `testKeepProperties`, that you added,   failed because the expectation is different with actual on Windows environment.

I am waiting you to adjust your code to pass it.

src/org/omegat/Bundle.properties
brandelune: What about "lock segments translated from enforced TM" ?

t-cordonnier: This pull request includes 2 situations where a segment can be locked:
- when it comes from tm/enforce
- when it comes from xliff with status = final
The second one is open: it is easy inside a filter to implement locking for another format if you can define any rule for segments which should not be modified (but preferably only for bilingual formats)
When you enter in a locked segment, any attempt to modify it will be rejected but the status bar will tell you the reason (xliff/final or tm/enforce)

For that reason your proposal is not correct; else I would have considered it as perfectly valid.
Still open to discussion, I did not reject Kazephil's proposal, only sent a comment. If it is validated by others we can use it.

Kazephil (commented):
> " if I understand the intent correctly."
> 
> To be sure of that I would prefer that people test the feature, not only comment based on the words. 

Ah, the things I could do if only I had unlimited time or resources! :grinning:

I would prefer that people provide a small sample project to test the feature. Unfortunately, they clearly also don't have unlimited time or resources and are not able to do so. :grin: 

Anyhow, my initial suggestion does not necessarily mean individual segments, but it is ambiguous and I can definitely see how you can arrive at that interpretation, so some tweaking will be required.

Give me a little time to think of something that:
1. is unambiguous about locking all segments meeting a given criterion (or set of criteria), and
2. is general enough to convey the idea for any and all current (and potential future) conditions leading to segment lock.

Kazephil:
Some ideas:

- "Lock preconfirmed segments."
- "Lock pre-accepted segments."
- "Lock pre-validated segments."
- "Lock already confirmed segments."
- "Lock already accepted segments."
- "Lock already validated segments."
- "Lock externally confirmed segments."
- "Lock externally accepted segments."
- "Lock externally validated segments."

I think all of those do a decent job of suggesting that we're talking all segments that have already been validated and confirmed as correct in some way, without being too specific about the underlying details of that confirmation.

src/org/omegat/Bundle.properties
t-cordonnier: > "lock translations from enforced TM and from XLIFF where status is "final"" 

As I said, the feature is more open, the case of xliff is only an example: once it is applied, any writer for a bilingual filter has a way to define a rule for locked segments. So this is actually enforce + xliff.final but in the future it may be any format which has a rule for locked segments

brandelune:
> > The three variations using that use "externally" would, of course, not work if there's a possibility of introducing a locking mechanism within OmegaT at some point in the future.
> 
> Externally seems good to me as it emphases the fact that the validation has been done in another tool prior to using OmegaT. For the other parts of your proposals, all seems correct to me. I prefer "already" than prefix "pre-" but I let other answer. Oh and maybe we could speak about "definitively" accepted/confirmed segments, to emphase the fact that we explicitly want the user not to touch them.

We don’t need to create a new concept for such matches. In fact, the "native" concept is already here. It is called "enforced translations". That’s obvious to the user and they know how to create/use them. So the obvious solution is "Lock enforced translations". That will considerably reduce the need to access the manual of reach out for user support.

Now, we can define the "enforced translations" as a wider concept in the manual, that is based on what we currently have, describe how to create them natively (our TM folder structure), and extend it by indicating that "XLIFF where status is "final"" behave the same (or not), indicate that there is the possibility to lock them, and link to the relevant option description in the manual.

Also, if we use the Editor category for that option, it should be under "Insert the best fuzzy match", and we should have "Save autopopulated status" and "Save origin of translation" below it. Not at the bottom of the list.

Kazephil:
> We don’t need to create a new concept for such matches. In fact, the "native" concept is already here. It is called "enforced translations". That’s obvious to the user and they know how to create/use them. So the obvious solution is "Lock enforced translations". That will considerably reduce the need to access the manual of reach out for user support.

Or maybe "Lock all segments containing an enforced translation." to make it a little more explicit that this applies in bulk and not to individual segments.

> Now, we can define the "enforced translations" as a wider concept in the manual, that is based on what we currently have, describe how to create them natively (our TM folder structure), and extend it by indicating that "XLIFF where status is "final"" behave the same (or not), indicate that there is the possibility to lock them, and link to the relevant option description in the manual.

I'm not entirely convinced that this will make a huge difference in terms of the need (or lack thereof) for user support. However, the term "enforced" *does* have the advantage of conveying the concept more strongly and clearly than any of the more general suggestions I made earlier, so broadening the definition of that concept is a better starting point.

brandelune:
> "Lock enforced translations" seems a good label to me.
> 
> Labels should be short and clear, and consistent with other labels. It seems reasonable to consider that target's "final" status is the XLIFF way to _enforce_ translations. My two cents.

@Kazephil I’m all for short labels too 😊.

brandelune:
> > Now, we can define the "enforced translations" as a wider concept in the manual, that is based on what we currently have, describe how to create them natively (our TM folder structure), and extend it by indicating that "XLIFF where status is "final"" behave the same (or not), indicate that there is the possibility to lock them, and link to the relevant option description in the manual.
> 
> I'm not entirely convinced that this will make a huge difference in terms of the need (or lack thereof) for user support. However, the term "enforced" _does_ have the advantage of conveying the concept more strongly and clearly than any of the more general suggestions I made earlier, so broadening the definition of that concept is a better starting point.

By defining the term as a "wider concept", I mean indicating the fact that it is not all about "tm/enforce" but goes into XLIFF territory and eventually beyond.

(On a side note, I’m also for including in the manual links to the developer documentation that Hiroshi is writing separately)

Kazephil (commented):
> @Kazephil I’m all for short labels too 😊. 

@brandelune 
:smile_cat: I have nothing against short labels, with the caveat that sometimes "short" can also be "ambiguous".

Having had a little bit more time to let my backbrain mull this over, I think the two best options are:

1. Lock enforced translations.
2. Lock all enforced translations.

The first is fine if you feel it's unambiguous enough.
The second is still short but makes it a little more explicit that it's an all-or-nothing option.

I'm happy with either version.

Kazephil (commented):
> By defining the term as a "wider concept", I mean indicating the fact that it is not all about "tm/enforce" but goes into XLIFF territory and eventually beyond.

Yes, that's how I understood it. Even just that would be enough to confuse some of the less tech-savvy people I know. :grin:
 
> (On a side note, I’m also for including in the manual links to the developer documentation that Hiroshi is writing separately) 

I think I would prefer one link in an appropriate section inviting users interested in the technical side of things to read that documentation and join the development team. I think multiple links could potentially be distracting, maybe even annoying, from the point of view of non-technically minded users. (Although, if we could make collapsible "Technical details" notes that those interested can open and read, that might be a good compromise.)

brandelune:
> > @Kazephil I’m all for short labels too 😊.
> 
> @brandelune 😸 I have nothing against short labels, with the caveat that sometimes "short" can also be "ambiguous".
> 
> Having had a little bit more time to let my backbrain mull this over, I think the two best options are:
> 
> 1. Lock enforced translations.
> 2. Lock all enforced translations.
> 
> The first is fine if you feel it's unambiguous enough. The second is still short but makes it a little more explicit that it's an all-or-nothing option.
> 
> I'm happy with either version.

@t-cordonnier would you mind choosing the version you prefer?

brandelune (approved):

miurahr:
This PR is closed because it seems abandoned. If you are alive, please reopen and fix the issues.

PR #1148, opened 2024-09-26
bump vldocking@3.1.1

miurahr:
- update VLDocking customization with new UIDefaults API which is LaF friendly.

## Pull request type

- Other (describe below)
refactoring

## What does this PR change?

- There is a long standing TO-DO in the theme subsystem that `vldocking` library does not respect LaF defaults so  `UIDesignManager` need to override `vldocking` configuration with `UIManager.put` instead of `defaults.put`
- This bump `vldocking` library to 3.1.1 and improve the customization of `vldocking` settings


miurahr:
The change does not affect the current theme plugins. When plugins configure vldocking customization with UIDefaults, the plugin can only support OmegaT 6.1 and later versions.

PR #1165, opened 2024-10-23
feat: filter(po): plural as comment in japanese

miurahr:
Japanese is nplural=1 language.
When there is no marker in singular msgid, translator can not see plural form of msgid when translate in OmegaT panes. This add plural form of msgid as a comment of a segment which is a singular form.

## Pull request type

## Which ticket is resolved?

- N.A.
- https://sourceforge.net/p/omegat/feature-requests/

## What does this PR change?

- Add plural form of msgid string as comment when singular msgid is a source

## Other information

ref: discussion on https://sourceforge.net/p/omegat/bugs/1269/

### DIFF
--- a/src/org/omegat/Bundle.properties
+++ b/src/org/omegat/Bundle.properties
@@ -2257,6 +2257,7 @@ POFILTER_TRANSLATOR_COMMENTS=Translator comments:
 POFILTER_EXTRACTED_COMMENTS=Extracted comments:
 POFILTER_REFERENCES=References:
 POFILTER_PLURAL_FORM_COMMENT=Translate this plural as a plural form #{0}.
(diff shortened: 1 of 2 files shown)


Kazephil (approved):

brandelune (commented):
I am not sure this modification will help the translator in the case at hand.
The resulting file will still trigger errors.

Also, in the case of plurals, their display location could be under the source, at the location of the alternative source.

brandelune:
> Please provide a reproducible of the error. I observe the checker pass well.

I can’t do much at the moment. Give me some time.

brandelune:
Documentation ticket:
https://sourceforge.net/p/omegat/documentation/436/

miurahr:
I think it is worth merging because it does not break anything except for notifying user the item is plural setup.

PR #963, opened 2024-02-17
[BUGS#1248] feat: add regression test

miurahr:
## Pull request type

Please mark github LABEL of the type of change your PR introduces:

- Bug fix -> [bug]
- Feature enhancement -> [enhancement]
- Documentation -> [documentation]
- Build and release changes -> [build/release]
- Other (describe below)

## Which ticket is resolved?

- Title ex. fix...
  * URL ex. https://...
 
- Title
  * URL

## What does this PR change?

-
-
-

PR #1174, opened 2024-11-04
refactor: FindMatches to split looong method into multiple private method, use lambda, and relax suppression of checkstyle

miurahr:
FindMatches class has been introduced in 2013 and there are several place to improve coding styles.

## Pull request type

- refactoring

## What does this PR change?

- Update javadoc contents
- style: remove final from arguments
- revert haveChanceToAdd logic to noChanceToAdd because all conditions are negated
- Replace inline ` new DefaultTranslationsIterator()` and `new MultiTranslationsIterator()` to lambda expression
- Add  NearString and FindMatches ctors with small number of arguments, with PrepareTMXEntries and NearString.Scores data classes.


miurahr:
#963 is also modifying the `FindMatches` class. We should make it coherent.

PR #1175, opened 2024-11-06
feat: EditorUtils use document's locale to splitting words

miurahr:
There are many TODO comments in the source.
This try to implement one of TODO in EditorUtils class.

## Pull request type

## Which ticket is resolved?

- Word selection according to the language of Source/Target by clicking in the editor pane
- https://sourceforge.net/p/omegat/feature-requests/1770/

## What does this PR change?
- Update javadoc for getWordEnd and getWordStart utility functions
- Implement "TODO: change to use document's locale"
- grab word with source and target locale of the project
-

### TODO
- [x] add UI test


src/org/omegat/gui/editor/EditorUtils.java
```diff
@@ -99,6 +100,41 @@ public static int getWordEnd(JTextComponent c, int offs) throws BadLocationExcep
...
+        Element line = Utilities.getParagraphElement(c, offs);
+        if (line == null) {
+            throw new BadLocationException("No word at " + offs, offs);
+        }
+        int lineStart = line.getStartOffset();
+        Document doc = c.getDocument();
+        int lineEnd = Math.min(line.getEndOffset(), doc.getLength());
+        if  (lineEnd - lineStart > 0) {
+            String lineString = doc.getText(lineStart, lineEnd - lineStart);
+            Locale locale = c.getLocale();
```
miurahr, line 114: It is better to leave a comment that explain an intention. 

```suggestion
            // Detection of target string locale.
            // A default is UI component locale, and when OmegaT project
            // is loaded, it uses a source or a target language as a
            // processing locale.
            Locale locale = c.getLocale();
```

test-acceptance/src/org/omegat/gui/editor/EditorUtilsTest.java
```diff
@@ -0,0 +1,123 @@
...
+            } catch (InterruptedException ignored) {
+            }
+            //
+            final JTextComponent editPane = window.panel("Editor - source.txt").textBox().target();
+            // select word from a source text
+            int offs = 102;
+            int posStart = EditorUtils.getWordStart(editPane, offs);
+            int posEnd = EditorUtils.getWordEnd(editPane, offs);
+            String word = editPane.getText(posStart, posEnd - posStart);
+            assertEquals("太平寺中的文笔塔", word);
```
miurahr, line 113: Chinese does not work correctly. It is because java has a bug which marked as "Won't-Fix".
https://bugs.openjdk.org/browse/JDK-4095325

There is a comment in Stack-overflow about it, and suggest to use ICU4j for that.
https://stackoverflow.com/a/44508826

src/org/omegat/gui/editor/EditorUtils.java
```diff
@@ -99,6 +101,47 @@ public static int getWordEnd(JTextComponent c, int offs) throws BadLocationExcep
...
+        Document doc = c.getDocument();
+        int lineEnd = Math.min(line.getEndOffset(), doc.getLength());
+        if  (lineEnd - lineStart > 0) {
+            String lineString = doc.getText(lineStart, lineEnd - lineStart);
+            // Detection of target string locale.
+            // A default is UI component locale, and when OmegaT project
+            // is loaded, it uses a source or a target language as a
+            // processing locale.
+            Locale locale = c.getLocale();
+            if (c instanceof EditorTextArea3 && Core.getProject().isProjectLoaded()) {
```
miurahr, line 120: It is better to pass boolean `isTranslation` by caller because we need to detect caller component here, and use `(EditorTextArea3) c.isActivaTranslation(offs)` and the `offs` is passed by caller.

brandelune:
This commit is breaking OmegaT.

brandelune:
We need to revert to the previous working state and rework this feature.

PR #1184, opened 2024-11-10
refactor: guard future LT method deprecation

miurahr:
LT 6.5 and later deprecate JLanguageTool.VERSION
This guards it from future deprecation.

## Pull request type

- Other (describe below)
refactor

## What does this PR change?

- Add @suppresswarnings("deprecation") for JLanguageTool.VERSION access
-
-


miurahr:
This can be done when requirement raised.

PR #1186, opened 2024-11-10
chore: Some tokenizer as language module

miurahr:
Tokenizer is used for the project in which the source or target language is the same as the provided one.
It is feasible to move tokenizers to the corresponding language module.

## Pull request type

- Build/release
- refactoring

## What does this PR change?

- Move Japanese, Chinese and Polish tokenizers to the corresponding language module.
- There are derived from 3rd party analyzer library, that are not mandatory for other languages.
-

## Other information

Depends on #1183


miurahr:
There is no obvious benefit to move  tokenizer classes to language-module and change package paths. Most tokenizers  come from Lucene-analyzer-common library that is a mandatory library of OmegaT core.  There are three languages that can be justificated for the change; Japanese, Chinese and Portugues that are provided their special analyzer.

miurahr:
This is not feasible to split tokenizer to module. I want to give up here.

PR #1190, opened 2024-11-17
refactor: reduce Core dependency for aligner

miurahr:
The Aligner depends on the Core configuration of segmentation and the filters.
It is not good to access Core API throughout the code, but it can be taken in ctor once.

## Pull request type

- Other (describe below)
refactor

## What does this PR change?

- Add a constructor to take the segmenter as an argument for testing.
- Logger is instantiated with Bundle.
- Rewrite action listeners with lambdas.
- Aligner class holds its own the segmenter and the filter master objects.

## Other information
- [x] Acceptance test

### DIFF
--- a/aligner/src/main/java/org/omegat/gui/align/AlignFilePickerController.java
+++ b/aligner/src/main/java/org/omegat/gui/align/AlignFilePickerController.java
@@ -51,6 +51,8 @@ import javax.swing.filechooser.FileFilter;
 import javax.swing.text.JTextComponent;
 
 import org.omegat.core.Core;
+import org.omegat.core.segmentation.SRX;
+import org.omegat.core.segmentation.Segmenter;
 import org.omegat.filters2.master.FilterMaster;
 import org.omegat.filters2.master.PluginUtils;
 import org.omegat.util.Language;
--- a/aligner/src/main/java/org/omegat/gui/align/AlignFilePickerController.java
+++ b/aligner/src/main/java/org/omegat/gui/align/AlignFilePickerController.java
@@ -50,13 +50,11 @@ import javax.swing.event.DocumentListener;
 import javax.swing.filechooser.FileFilter;
 import javax.swing.text.JTextComponent;
 
-import tokyo.northside.logging.ILogger;
-import tokyo.northside.logging.LoggerFactory;
-
 import org.omegat.core.Core;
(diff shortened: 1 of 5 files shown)


miurahr:
#1191 implements acceptance test case.

aligner/src/main/java/org/omegat/gui/align/AlignFilePickerController.java
```diff
@@ -499,16 +494,13 @@ public boolean accept(File f) {
...
+     * @param args command arguments.
      */
-    public static void main(String[] args) throws Exception {
+    public static void main(String[] args) {
         System.setProperty("apple.laf.useScreenMenuBar", "true");
 
         Preferences.init();
         PluginUtils.loadPlugins(Collections.emptyMap());
-        Core.setFilterMaster(new FilterMaster(FilterMaster.createDefaultFiltersConfig()));
-        Core.setSegmenter(new Segmenter(SRX.getDefault()));
```
miurahr, line 504: This part should be reverted.

```suggestion
        Core.setFilterMaster(new FilterMaster(FilterMaster.createDefaultFiltersConfig()));
        Core.setSegmenter(new Segmenter(SRX.getDefault()));
```

aligner/src/main/java/org/omegat/gui/align/AlignFilePickerController.java
```diff
@@ -499,16 +494,13 @@ public boolean accept(File f) {
      * <li>Target file path
      * </ol>
      *
-     * @param args
-     * @throws Exception
+     * @param args command arguments.
      */
-    public static void main(String[] args) throws Exception {
+    public static void main(String[] args) {
```
miurahr, line 499:
```suggestion
    public static void main(String[] args) throws Exception {
```

aligner/src/main/java/org/omegat/gui/align/AlignFilePickerController.java
```diff
@@ -499,16 +494,13 @@ public boolean accept(File f) {
      * <li>Target file path
      * </ol>
      *
-     * @param args
-     * @throws Exception
+     * @param args command arguments.
```
miurahr, line 497:
```suggestion
     * @param args command arguments.
     * @throws Exception when failed to ininitalize OmegaT core. 
```

PR #1194, opened 2024-11-20
Refactor: IProject: add getter and setter of Segmenter and removal of static fileds of the objects in Core class

miurahr:
##  Move segmenter object to be hold by RealProject object.

There are Preferences and configurations of SRX in User preferences and project-specific custom segmentation rules.
There is a segmenter object which is for the current live project, which is runtime property of the live project.

Current implementation hold the segmenter in the Core class as static field.
Here propose it to move into RealProject class, and extend IProject interface to have getter and setter of the segmenter.

## Pull request type

- Other (describe below)
refactoring

## Which ticket is resolved?

Discussed on DEV-ML thread
- https://sourceforge.net/p/omegat/mailman/omegat-development/thread/95b050f8-a375-4ed9-b0fe-e4deccca045b%40northside.tokyo/#msg58842059

## What does this PR change?

- Extend IProject to have getter and setter for Segmenter and FilterMaster
- add segmenter and filterMaster fields in RealProject class, and put setter and getter
- Update Core class: 
       * removal of static fileds of segmenter
       * removal of initialization of these fields in initilizeGuiImpl method
       * Update getters to return value from `currentProject#getSegmenter` 
       * Update setters as well
       * We don't need to reset Core static fields value when closing project.
- Update test cases
       * Removal of `Core.setSegmenter`. We define private filed of the test class for these objects.
       * We extend mock project class with these private fields
       * These changes make the test case does not depend on the static context, so there will be no interference among test cases.
- Extent  `AlignFilesCallback` ctor to take segmenter object.
       * the callback object does not depend on static context of the OmegaT project.
       * We can test `ExternalTMFactory` without Core static context.


src/org/omegat/gui/main/ProjectUICommands.java
```diff
@@ -796,6 +796,7 @@ public static void projectClose() {
         // commit the current entry first
         Core.getEditor().commitAndLeave();
 
+        final IProject project = Core.getProject();
```
miurahr, line 799:
```suggestion
```

This is orphan auto variable. We can remove it.

src/org/omegat/core/data/RealProject.java
```diff
@@ -1951,21 +1944,26 @@ protected String getSegmentTranslation(String id, int segmentIndex, String segme
     }
 
     static class AlignFilesCallback implements IAlignCallback {
-        AlignFilesCallback(ProjectProperties props) {
+        AlignFilesCallback(ProjectProperties props, FilterMaster filterMaster, Segmenter segmenter) {
```
miurahr, line 1947: The order of argument.
`ExternalTMFactory.load` takes `segmenter` then `filterMaster` but `AlignFilesCallback` ctor takes `filtermaster` then `segmenter`.

test/src/org/omegat/core/data/ExternalTMFactoryTest.java
```diff
@@ -87,7 +85,7 @@ public void testLoadTMX() throws Exception {
 
         assertTrue(ExternalTMFactory.isSupported(tmxFile));
 
-        ExternalTMX tmx = ExternalTMFactory.load(tmxFile);
+        ExternalTMX tmx = ExternalTMFactory.load(tmxFile, props, segmenter, filterMaster);
```
miurahr, line 88: The order of argument.
`ExternalTMFactory.load` takes `segmenter` then `filterMaster` but `AlignFilesCallback` ctor takes `filtermaster` then `segmenter`.

miurahr (commented):
There are space to improve an order of arguments.

src/org/omegat/core/data/RealProject.java
t-cordonnier: As an alternative, is it really relevant to keep this class static?

miurahr:
refactoring FilterMaster does not work well. I want to revert it.

miurahr:
There is unrelated refactorings. I want to skip it.

miurahr:
The test failure is happened because an aligner test uses `Core.setSegmenter()` without initializing `Core.setProject` with `NotLoadedProject` or a real project.

miurahr:
The issue in Aligner is dealt with #1190. Now merge it.

src/org/omegat/core/Core.java
```diff
@@ -215,11 +214,11 @@ public static IDictionaries getDictionaries() {
     }
 
     public static Segmenter getSegmenter() {
-        return segmenter;
+        return currentProject.getSegmenter();
```
miurahr, line 217: `currentProject`  can be null

miurahr: I push the commit which initialize current project in static fields initializer.

miurahr:
It should be better to integrate into PR #1463 
#1463 is a topic to manage Core and Project state.

PR #1220, opened 2024-12-17
feat: introduce matchSource match template variable

miurahr:
split of #963 

## Pull request type

## Which ticket is resolved?

RFE#1251

## What does this PR change?

- Change default fuzzy match pane template with matchSource
- Add method MatchesVarExpansion#expandMatchSource
- Extend NearString.MATCH_SOURCE to have TM_SUBSEG
- Update test expectations of MatchesTextAreaTest, and FindMatchesTest
- Add human-readable names of MATCH_SOURCE in Bundle.properties


src/org/omegat/core/statistics/FindMatches.java
```diff
@@ -385,7 +416,7 @@ public void processEntry(EntryKey key, ITMXEntry entry, String tmxName,
         }
 
         // BUGS#1236 - stat display does not use threshold config check
-        if (applyThreshold && similarityStem < fuzzyMatchThreshold
+        if (fuzzyMatchThreshold > 0 && similarityStem < fuzzyMatchThreshold
```
miurahr, line 419:
```suggestion
        if (applyThreashold && similarityStem < fuzzyMatchThreshold
```

src/org/omegat/Bundle.properties
brandelune, line 1: MATCHES → MATCH

If we use "From" in 2 items, why not use it in the 2 others?

brandelune (changes requested):

src/org/omegat/Bundle.properties
miurahr: Please propose alternative

miurahr: > MATCHES → MATCH
> 

The items are used in `MatchesVarExpansion` class as you can look in the PR.
Why not use "MATCHES_VAR_EXPANSION_SOMETHING"?

brandelune: > > MATCHES → MATCH
> 
> The items are used in `MatchesVarExpansion` class as you can look in the PR. Why not use "MATCHES_VAR_EXPANSION_SOMETHING"?

No intent to sound pompous, but anything that is grammatically correct is fine :-)

matches (plural) comes (singular) X
match (singular) comes (singular) O

Since we’re talking about the origin of singular matches "MATCH_COMES_FROM..." should work.

brandelune: > Please propose alternative

I’m not sure what "from files" means in the context of a match. Also, maybe we don’t need "From" since it is clear that we talk about the origin of the match.

Where would the strings be displayed ? Could you attach a screenshot?

miurahr: > > Please propose alternative
> 
> I’m not sure what "from files" means in the context of a match. Also, maybe we don’t need "From" since it is clear that we talk about the origin of the match.
> 
> Where would the strings be displayed ? Could you attach a screenshot?

Because this PR is a split of mother PR, there is no exact screenshot.
 
Please check an acceptance test expectation for the screen contents.

https://github.com/omegat-org/omegat/pull/963/files#diff-7372ede97de3491b9f3d0402b5e3d800b9d9afb7b0fc7fe621dde3b1ebda6a0cR54-R56

miurahr:
```java
        Pattern pattern = Pattern.compile("1. Error while reading MT results\\n"
                + "Erreur lors de la lecture des résultats de TA\\n"
                + "<\\d+/\\d+/\\d+%\\s*" + OStrings.getString("MATCHES_COMES_FROM_MEMORY") + "\\s*>");
        window.textBox("matches_pane").requireText(pattern);
```

looks like 
```
1. Error while reading MT results
    Erreur lors de la lecture des résultats de TA
  <90/90/90% From Project>  File_path_here
```

brandelune: Ok. It makes sense.

So what does "Files" mean in that context ?

```
<90/90/90% Files>  File_path_here

Comment (miurahr) [reply src/org/omegat/Bundle.properties:1]: > Ok. It makes sense.
> 
> So what does "Files" mean in that context ?
> 
> ```
> <90/90/90% Files>  File_path_here
> ```

I do not know. There are no any place to use it in OmegaT. For API compatibility reason, I don't remove it.

Comment (t-cordonnier) [reply src/org/omegat/Bundle.properties:1]: If I am not wrong, FILES it used when the match comes from bilingual file in the source, but I must check. I will confirm it to you later

Comment (brandelune) [reply src/org/omegat/Bundle.properties:1]: That would make sense. For ex PO/XLIFF files.
In fact, I don’t remember seeing anything indicating that such matches come from the file and not from the project memory (and I guess that’s also correct since they are loaded in the project memory). But it would be nice to systematically have them be specified as such.

Comment (miurahr) [reply src/org/omegat/Bundle.properties:1]: Translations from PO/XLIFF files has been treated as `MATCH_SOURCE.MEMORY` from the time the feature implemented, at least 11 years ago the file `FindMatches.java` introduced.

**Comment by miurahr** [comment]
I have updated bundle keys and test expectation.
During process I merged the master branch.

File: src/org/omegat/Bundle.properties
Comment (miurahr) [reply src/org/omegat/Bundle.properties:1]: > ```
> 1. Error while reading MT results
>     Erreur lors de la lecture des résultats de TA
>   <90/90/90% From Project>  File_path_here
> ```

It is my mistake. The field is defined as
```java
public static final String DEFAULT_TEMPLATE = VAR_ID + ". " + VAR_FUZZY_FLAG + VAR_SOURCE_TEXT + "\n"
            + VAR_TARGET_TEXT + "\n" + "<" + VAR_SCORE_BASE + "/" + VAR_SCORE_NOSTEM + "/"
            + VAR_SCORE_ADJUSTED + "%" + VAR_MATCH_SOURCE + VAR_FILE_PATH + ">";
```
so It will become look like
```
1. Error while reading MT results
    Erreur lors de la lecture des résultats de TA
  <90/90/90%Project C:\data\project\tm\path>
```

miurahr: > If I am not wrong, FILES it used when the match comes from bilingual file in the source, but I must check. I will confirm it to you later

I have just pushed new PR to realize your idea as PR #1227 
Please file an issue your idea as RFE on source forge ticket system.

miurahr: > Now, as the value exists I suggest that we can use it that way, to make distinction between segments coming from tmx and from source files. But maybe we should rename it to SOURCE_FILES to avoid any ambiguity?

Changing enum item name will break the JVM binary compatibility against 3rd party plugins which use `NearString` class.
When breaking change is acceptable, we can do it.

miurahr: > So, we cannot be sure what was Alex's intention when he added this line, as he documented nowhere. Either it was a totally different idea, or it is a bug (not an RFE)

It has never been implemented to indicate source of matches on the fuzzy match pane, even Alex had had an idea, it has not been any feature for users.
Please put your idea to indicate the matches from source files as that on the fuzzy match pane.  I want to propose use the unused value FILES for it.

brandelune: > > So, we cannot be sure what was Alex's intention when he added this line, as he documented nowhere. Either it was a totally different idea, or it is a bug (not an RFE)
> 
> It has never been implemented to indicate source of matches on the fuzzy match pane, even Alex had had an idea, it has not been any feature for users. Please put your idea to indicate the matches from source files as that on the fuzzy match pane. I want to propose use the unused value FILES for it.

I would suggest that FILES be used for matches coming from bilingual files. The fact that such matches are currently labeled as coming from the project is confusing.

miurahr:
> I have updated bundle keys and test expectation. During process I merged the master branch.

With new changes, the part of "<50/50/60% >" will become like 

1. "<50/50/60% Project>"
2. "<50/50/60% External TMX tm/penalty-010/segment.tmx>"
3. "<50/50/60% Source files source/foo.po>"
4. "<50/50/60% Sub-segmented match tm/file1.tmx,tm/file2.tmx>"

@t-cordonnier  @brandelune  is the resulted change suitable for your assumption?

brandelune:
> > I have updated bundle keys and test expectation. During process I merged the master branch.
> 
> With new changes, the part of "<50/50/60% >" will become like
> 
> 1. "<50/50/60% Project>"
> 2. "<50/50/60% External TMX tm/penalty-010/segment.tmx>"
> 3. "<50/50/60% Source files source/foo.po>"
> 4. "<50/50/60% Sub-segmented match>"
> 
> @t-cordonnier @brandelune is the resulted change suitable for your assumption?
> 
> When #1226 is merged, No4 will become like "<50/50/60% Sub-segmented match tm/file1.tmx,tm/file2.tmx>"

Ok. So we should first modify the second item in that representation of the match pane this way: 

```
<50/50/60% >
```
→ 
```
<50/50/60% Project>
```
Is that correct ?

In the case of orphan segments (we should not has a plural here, right @Kazephil), what will be the display ? Since orphan matches all come from the project file, there will be a double indication.

Regarding the various labels:

> 1. "<50/50/60% Project>"

Ok for now.

> 2. "<50/50/60% External TMX tm/penalty-010/segment.tmx>"

should be "External TM: <path>" since all bilingual files are accepted in the /tm folder, not exclusively TMX files.

> 3. "<50/50/60% Source files source/foo.po>"

should be "Source file: <path>"

> 4. "<50/50/60% Sub-segmented match>"

not easy to understand. What about

"Reconstructed match", "Merged match", etc.

@t-cordonnier, what do you think about all that?

miurahr:
Answer the question.

> In the case of orphan segments (we should not has a plural here, right @Kazephil), what will be the display ? Since orphan matches all come from the project file, there will be a double indication.

Current master branch has been programmed to store localized of "Orphan segments" as a TMX **name** field.
The message is defined in the Bundle with the key "CT_ORPHAN_STRINGS". 

see. https://github.com/omegat-org/omegat/blob/3ea0c12f1b6f9cf92799f3cc9dde23eeaaed2491/src/org/omegat/core/statistics/FindMatches.java#L247

It may become fuzzy match pane to show

```
<50/50/60% Project Orphan segments>
```

src/org/omegat/core/matching/NearString.java
```diff
@@ -49,37 +49,81 @@
...
-        MEMORY, TM, FILES
-    };
+        /** From project memory */
+        MEMORY,
+        /** From external TM in project tm/ folder */
+        TM,
+        /** From files */
+        FILES,
+        /** From sub-segmented match */
+        TM_SUBSEG
```
t-cordonnier, line 59: Why this prefix TM_ ?
Subsegments can come from any other source, including project memory, tmx files or external files.  They can even be combination of segments which do not come from same source.

miurahr: > Why this prefix TM_ ? Subsegments can come from any other source, including project memory, tmx files or external files. They can even be combination of segments which do not come from same source.

What is your proposal? `SUBSEG` only?

t-cordonnier: Yes, or SUBSEGMENTS as I did in PR #1224 ; in any case no reason why to keep this prefix which gives a wrong info about where data come from.

src/org/omegat/core/matching/NearString.java
```diff
@@ -49,37 +49,81 @@
  */
 public class NearString {
     public enum MATCH_SOURCE {
-        MEMORY, TM, FILES
-    };
+        /** From project memory */
+        MEMORY,
+        /** From external TM in project tm/ folder */
+        TM,
+        /** From files */
```
t-cordonnier, line 56: /* From source files */
(let this line give a useful information, remember that nobody remembered its meaning...)

src/org/omegat/core/matching/NearString.java
miurahr, line 59:
```suggestion
        SUBSEGMENTS
```

src/org/omegat/core/matching/NearString.java
miurahr, line 56:
```suggestion
        /** From source files */
```

src/org/omegat/gui/matches/MatchesVarExpansion.java
```diff
@@ -222,6 +224,22 @@ private String getPropValue(List<TMXProp> props, String type) {
...
 
+    private String expandMatchSource(String localTemplate, NearString.MATCH_SOURCE comesFrom) {
+        switch (comesFrom) {
+        case TM:
+            return localTemplate.replace(VAR_MATCH_SOURCE, OStrings.getString("MATCHES_VAR_EXPANSION_MATCH_COMES_FROM_TM") + " ");
+        case FILES:
+            return localTemplate.replace(VAR_MATCH_SOURCE, OStrings.getString("MATCHES_VAR_EXPANSION_MATCH_COMES_FROM_FILES"));
+        case MEMORY:
+            return localTemplate.replace(VAR_MATCH_SOURCE, OStrings.getString("MATCHES_VAR_EXPANSION_MATCH_COMES_FROM_MEMORY"));
+        case TM_SUBSEG:
```
miurahr, line 235:
```suggestion
        case SUBSEGMENTS:
```

src/org/omegat/gui/matches/MatchesVarExpansion.java
```diff
@@ -222,6 +224,22 @@ private String getPropValue(List<TMXProp> props, String type) {
...
+        switch (comesFrom) {
+        case TM:
+            return localTemplate.replace(VAR_MATCH_SOURCE, OStrings.getString("MATCHES_VAR_EXPANSION_MATCH_COMES_FROM_TM") + " ");
+        case FILES:
+            return localTemplate.replace(VAR_MATCH_SOURCE, OStrings.getString("MATCHES_VAR_EXPANSION_MATCH_COMES_FROM_FILES"));
+        case MEMORY:
+            return localTemplate.replace(VAR_MATCH_SOURCE, OStrings.getString("MATCHES_VAR_EXPANSION_MATCH_COMES_FROM_MEMORY"));
+        case SUBSEGMENTS:
+            return localTemplate.replace(VAR_MATCH_SOURCE, OStrings.getString(
+                    "MATCHES_VAR_EXPANSION_MATCH_COMES_FROM_TM_SUBSEG") + " ");
```
miurahr, line 237:
```suggestion
                    "MATCHES_VAR_EXPANSION_MATCH_COMES_FROM_SUBSEGMENTS") + " ");
```

src/org/omegat/Bundle.properties
```diff
@@ -2947,3 +2947,8 @@ DICTIONARY_LOAD_FILE=Loaded dictionary from '{0}': {1} ms
 DICTIONARY_LOAD_ERROR=Error load dictionary from '{0}': {1} 
 DICTIONARY_MANAGER_ERROR_SAVE_IGNORE=Error saving ignore words"
 EDITOR_CONTROLLER_EXCEPTION=bad location exception when changing case
+
+MATCHES_VAR_EXPANSION_MATCH_COMES_FROM_TM=External TM
+MATCHES_VAR_EXPANSION_MATCH_COMES_FROM_FILES=Source Files
+MATCHES_VAR_EXPANSION_MATCH_COMES_FROM_MEMORY=Project
+MATCHES_VAR_EXPANSION_MATCH_COMES_FROM_TM_SUBSEG=Sub-segmented match
```
miurahr, line 2954:
```suggestion
MATCHES_VAR_EXPANSION_MATCH_COMES_FROM_SUBSEGMENTS=Sub-segmented match
```

brandelune:
Documentation ticket: https://sourceforge.net/p/omegat/documentation/432/

brandelune (approved):

PR #1223, opened 2024-12-18
Apply max penalty for sub-segment match

t-cordonnier:
During a subsegment search, if some of the subsegments have a penalty, the global segment receives as penalty the highest value

## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?

https://sourceforge.net/p/omegat/bugs/1251/

### DIFF
--- a/src/org/omegat/core/statistics/FindMatches.java
+++ b/src/org/omegat/core/statistics/FindMatches.java
@@ -7,7 +7,7 @@
                2008 Alex Buloichik
                2012 Thomas Cordonnier, Martin Fleurke
                2013 Aaron Madlon-Kay, Alex Buloichik
-               2024 Hiroshi Miura
+               2024 Hiroshi Miura, Thomas Cordonnier
                Home page: https://www.omegat.org/
                Support center: https://omegat.org/support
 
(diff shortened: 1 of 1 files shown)


miurahr:
Why don't you use the `maxPanelty` variable to process the entry, but use `0`?
```
processEntry(null, entry, "", NearString.MATCH_SOURCE.TM, false, 0);
```

t-cordonnier:
> Why don't you use the `maxPanelty` variable to process the entry, but use `0`?
> 
> ```
> processEntry(null, entry, "", NearString.MATCH_SOURCE.TM, false, 0);
> ```

You are true, I had modified this line in the other PR only, I forgot to do this change when I splitted in two pull requests. I will correct it

t-cordonnier:
> > Why don't you use the `maxPanelty` variable to process the entry, but use `0`?
> > ```
> > processEntry(null, entry, "", NearString.MATCH_SOURCE.TM, false, 0);
> > ```
> 
You are true, I had modified this line in the other PR only, I forgot to do this change when I splitted in two pull requests. 

Corrected now

PR #1224, opened 2024-12-18
Change label for subsegments

t-cordonnier:
Actually a subsegment is identified as an empty string, which can make confusion with segments coming from project

## Pull request type

    Feature enhancement -> [enhancement]

## What does the PR change

Display first entry as file name instead of empty string

Adds new variables for "comes from":
 
- ${matchSource} contains a string like 'From TM'
- ${matchSource-dingbat} contains a symbol, better than ${matchSource} when added in a very long line
- ${matchSource-emoji} contains an emoji, better than ${matchSource} but may be unavailable in some fonts

## Which ticket is resolved?

https://sourceforge.net/p/omegat/bugs/1251/


miurahr:
Please adjust for master and update with feedbacks from #1220

t-cordonnier:
> Please adjust for master and update with feedbacks from #1220

Just done but each time I do a commit I receive alerts from checkstyle. However my impression is that it is not linked to the part I am working on. Any idea?

miurahr:
> > Please adjust for master and update with feedbacks from #1220
> 
> Just done but each time I do a commit I receive alerts from checkstyle. Seems linked to the size of the method, but it was long even before I modified it...

Please check `config/checkstyle/suppressions.xml`
I specified a suppression with an exact line number, to notify developers who modify the target source the problem.

miurahr:
The change here affects OmegaT UI faced by users. It is mandatory to update the manual page to explain marks and its meanings.

 @brandelune do you have any suggestions about the manual?

brandelune:
> The change here affects OmegaT UI faced by users. It is mandatory to update the manual page to explain marks and its meanings.
> 
> @brandelune do you have any suggestions about the manual?

@t-cordonnier could you please identify in the manual the parts that need updating and suggest a draft ?

t-cordonnier:
> > > Please adjust for master and update with feedbacks from #1220
> > 
> > 
> > Just done but each time I do a commit I receive alerts from checkstyle. Seems linked to the size of the method, but it was long even before I modified it...
> 
> Please check `config/checkstyle/suppressions.xml` I specified a suppression with an exact line number, to notify developers who modify the target source the problem.

I see nothing about FindMatches.java in my local copy. Are you saying it is intentional?

miurahr:
> > Please check `config/checkstyle/suppressions.xml` I specified a suppression with an exact line number, to notify developers who modify the target source the problem.
> 
> I see nothing about FindMatches.java in my local copy. Are you saying it is intentional? Do you expect us to try to make this method smaller?

If it is so, it means `FindMatches#search` was under 150 lines in master branch, and addition of your code brings the method over 150 lines.  You need to consider a way to reduce lines, because there are too match in the one method.

brandelune:
@t-cordonnier are you sure it is RFE 1251 ?
https://sourceforge.net/p/omegat/feature-requests/1251/

t-cordonnier:
> @t-cordonnier are you sure it is RFE 1251 ? https://sourceforge.net/p/omegat/feature-requests/1251/

Sorry I did not remember that it was registered as a bug, not as an RFE
https://sourceforge.net/p/omegat/bugs/1251/

brandelune:
We have https://github.com/omegat-org/omegat/pull/1220 merged. Does that change anything to this PR?

miurahr:
Please merge master to resolve conflict and update to solve checkstyle errors.

miurahr:
close because of stalled. please reopen when you resume the work.

PR #1227, opened 2024-12-19
feat: mark matches from source file translations as MATCH_SOURCE.FILES

miurahr:
## Pull request type

## Which ticket is resolved?

BUGS#1251

## What does this PR change?

- 
-
-


t-cordonnier:
I think this one seems to make consensus now and could be applied to all branches

PR #1234, opened 2024-12-27
remove dead main method of aligner

miurahr:
## Pull request type

- Other (describe below)

## What does this PR change?

- remove dead code


miurahr:
There is #1236 as successor

PR #1236, opened 2024-12-27
Have command line aligner correctly load plugins

t-cordonnier:
Command line aligner did not properly load filter plugins: you could open the UI but any file would be unsupported

## Pull request type

Bug

- Bug fix -> [bug]

### DIFF
--- a/aligner/src/main/java/org/omegat/gui/align/AlignFilePickerController.java
+++ b/aligner/src/main/java/org/omegat/gui/align/AlignFilePickerController.java
@@ -523,7 +523,9 @@ public class AlignFilePickerController {
         System.setProperty("apple.laf.useScreenMenuBar", "true");
 
         Preferences.init();
+        Core.initializeGUI(Collections.emptyMap());
         PluginUtils.loadPlugins(Collections.emptyMap());
+        FilterMaster.setFilterClasses(PluginUtils.getFilterClasses());
         Core.setFilterMaster(new FilterMaster(FilterMaster.createDefaultFiltersConfig()));
         Core.setSegmenter(new Segmenter(SRX.getDefault()));
         AlignFilePickerController picker = new AlignFilePickerController();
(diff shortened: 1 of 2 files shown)


miurahr:
## Summary of changes

This Pull Request introduces updates to improve the user interface and configuration management for file alignment functionalities in the codebase. It focuses on initializing design components and adjusting font scaling for better UI consistency.

## Main changes:

- AlignFilePickerController.java:
    - Added a call to UIDesignManager.initialize() for initializing the UI design manager.
    - Modified the filter initialization process via FilterMaster.setFilterClasses().
- AlignPanelController.java:
    - Updated the font configuration of the table component to use a scaled font via FontUtil.getScaledFont() for better adaptability.

brandelune:
Is it a problem that needs to be fixed in 6.0.2 too ?

miurahr:
> Is it a problem that needs to be fixed in 6.0.2 too ?

no.

aligner/src/main/java/org/omegat/gui/align/AlignFilePickerController.java
```diff
@@ -523,7 +524,9 @@ public static void main(String[] args) throws Exception {
         System.setProperty("apple.laf.useScreenMenuBar", "true");
 
         Preferences.init();
+        UIDesignManager.initialize();
```
miurahr, line 527: Current Main class load plugin before calling `Core.initializeGUI`.

Here is a copy from Main class
```
        Preferences.init();
        // broker should be loaded before module loading
        JLanguageTool.setClassBrokerBroker(new LanguageClassBroker());
        JLanguageTool.setDataBroker(new LanguageDataBroker());
        PluginUtils.loadPlugins(PARAMS);
        FilterMaster.setFilterClasses(PluginUtils.getFilterClasses());
        Preferences.initFilters();
        Preferences.initSegmentation();
        
        // then call Core.initializeGUI
       // InitializeGUI call UIDesignManager.initialize()
```

Call it in the timing before `loadPlugins`, the `UIDesignManager` does not find theme plugins, escpecially for default theme.

brandelune:
> In OmegaT 6.0, `aligner` is placed in `src/org/omegat/gui/align` and built into OmegaT.jar. In OmegaT 6.1, `aligner` is located in `aligner/` as a module.
> 
> This is why the command line issue has been raised in the master branch, and I think we do not need to backport it into 6.0 branch.

Can you two reach a conclusion on this item?

miurahr:
close because of stalled. And the issue seems already be resolved.

PR #1254, opened 2025-01-24
1277 wait scripts before compile

t-cordonnier:
## Pull request type

Bug

## Which ticket is resolved?

https://sourceforge.net/p/omegat/bugs/1277/

## What does this PR change?

Adds two waiters between the creation of the target files and the associated commit:
- one for the loop for COMPILE event (all associated handlers must be finished), which runs in the UI thread while compilation is in another thread;
- one for the scripts associated to COMPILE event, which are lauched by the previously mentioned loop but also run in a separate thread

I also add a last commit which answers to the question which was inserted as TODO, since part of my corrections are in this part.

### DIFF
--- a/src/org/omegat/core/CoreEvents.java
+++ b/src/org/omegat/core/CoreEvents.java
@@ -111,9 +111,12 @@ public final class CoreEvents {
     public static void unregisterEditorEventListener(final IEditorEventListener listener) {
         EDITOR_EVENT_LISTENERS.remove(listener);
     }
+    
+    public static boolean isInProjectEventLoop = false;
 
     /** Fire event. */
     public static void fireProjectChange(final IProjectEventListener.PROJECT_CHANGE_TYPE eventType) {
+        isInProjectEventLoop = true; // done by the calling thread, not by UI thread (inside invokeLater) !
         SwingUtilities.invokeLater(() -> {
             Log.logInfoRB("LOG_INFO_EVENT_PROJECT_CHANGE", eventType);
             for (IProjectEventListener listener : PROJECT_EVENT_LISTENERS) {
@@ -123,6 +126,7 @@ public final class CoreEvents {
                     log("ERROR_EVENT_PROJECT_CHANGE", t);
                 }
             }
+            isInProjectEventLoop = false;
         });
     }
 
--- a/src/org/omegat/core/data/RealProject.java
+++ b/src/org/omegat/core/data/RealProject.java
@@ -714,6 +714,9 @@ public class RealProject implements IProject {
         // them to
         // the repository (BUGS#1176)
         CoreEvents.fireProjectChange(IProjectEventListener.PROJECT_CHANGE_TYPE.COMPILE);
+        while (CoreEvents.isInProjectEventLoop) {
+            Thread.yield();
+        }
         while (org.omegat.gui.scripting.ScriptingWindow.window.inEventsLoop > 0) {
             Thread.yield();
         }
--- a/src/org/omegat/core/data/RealProject.java
+++ b/src/org/omegat/core/data/RealProject.java
@@ -714,6 +714,9 @@ public class RealProject implements IProject {
         // them to
         // the repository (BUGS#1176)
         CoreEvents.fireProjectChange(IProjectEventListener.PROJECT_CHANGE_TYPE.COMPILE);
+        while (org.omegat.gui.scripting.ScriptingWindow.window.inEventsLoop > 0) {
+            Thread.yield();
+        }
 
         if (remoteRepositoryProvider != null && config.getTargetDir().isUnderRoot() && commitTargetFiles
                 && isOnlineMode) {
--- a/src/org/omegat/gui/scripting/ScriptingWindow.java
+++ b/src/org/omegat/gui/scripting/ScriptingWindow.java
@@ -112,7 +112,7 @@ import org.omegat.util.gui.StaticUIUtils;
  */
 public class ScriptingWindow {
 
-    static ScriptingWindow window;
+    public static ScriptingWindow window;
 
     public static void loadPlugins() {
         CoreEvents.registerApplicationEventListener(new IApplicationEventListener() {
@@ -580,6 +580,7 @@ public class ScriptingWindow {
             } catch (ExecutionException e) {
                 logResultRB(e, "SCW_SCRIPT_ERROR");
             }
+            inEventsLoop--;
         }
     }
 
@@ -594,6 +595,8 @@ public class ScriptingWindow {
     public void executeScript(ScriptItem scriptItem, Map<String, Object> bindings, boolean cancelQueue) {
         executeScripts(Arrays.asList(scriptItem), bindings, cancelQueue);
     }
+    
+    public int inEventsLoop = 0;
 
     /**
      * Execute scripts sequentially to make sure they don't interrupt each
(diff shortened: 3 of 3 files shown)


miurahr (changes requested):

brandelune:
@t-cordonnier is there anything that blocks you here ?

t-cordonnier:
> @t-cordonnier is there anything that blocks you here ?

This code was tested by Manuel so for me all is OK
But requests from Hiroshi werent so clears so better ask him whenever something is still blocking

Note: I am in holidays now with limited access to computer so if I really need to do something this won't be during two next weeks

brandelune:
> > @t-cordonnier is there anything that blocks you here ?
> 
> This code was tested by Manuel so for me all is OK But requests from Hiroshi werent so clears so better ask him whenever something is still blocking
> 
> Note: I am in holidays now with limited access to computer so if I really need to do something this won't be during two next weeks

@miurahr Is there anything that needs to be done here?

src/org/omegat/gui/scripting/ScriptingWindow.java
```diff
@@ -116,7 +116,7 @@ public class ScriptingWindow {
 
     private static final Logger LOGGER = Logger.getLogger(ScriptingWindow.class.getName());
 
-    static ScriptingWindow window;
+    static public ScriptingWindow window;
```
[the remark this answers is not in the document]
miurahr: Please click "Run  CheckStyle" on GitHub CI result that is failed and understand what is wrong.
 
image

t-cordonnier:
I have corrected the first point (public after static) but there is still a problem, the test does not work.
The report says "Properties should be assigned using the 'propName = value' syntax. Setting a property via the Gradle-generated 'propName value' or 'propName(value)' syntax in Groovy DSL has been deprecated. (1)"
Problem : I did not touch gradle nor its configuration files.

So, just in case I did a checkout of your master branch and ran the tests, and obtained exactly the same result.
Conclusion: this error does not seem to be in my code. Can you explain?

src/org/omegat/core/CoreEvents.java
```diff
@@ -111,9 +111,12 @@ public static void registerEditorEventListener(final IEditorEventListener listen
     public static void unregisterEditorEventListener(final IEditorEventListener listener) {
         EDITOR_EVENT_LISTENERS.remove(listener);
     }
+    
+    public static boolean isInProjectEventLoop = false;
 
     /** Fire event. */
     public static void fireProjectChange(final IProjectEventListener.PROJECT_CHANGE_TYPE eventType) {
+        isInProjectEventLoop = true; // done by the calling thread, not by UI thread (inside invokeLater) !
```
miurahr, line 119: The `CoreEvent` class should not have a state in design. The class is entry point and handles it though `IProjectEventListener`. 
Please consider update the interface to support status.

src/org/omegat/core/CoreEvents.java
```diff
@@ -111,9 +111,12 @@ public static void registerEditorEventListener(final IEditorEventListener listen
     public static void unregisterEditorEventListener(final IEditorEventListener listener) {
         EDITOR_EVENT_LISTENERS.remove(listener);
     }
+    
+    public static boolean isInProjectEventLoop = false;
```
miurahr, line 115: The field should not expose as public but should be private and provide getter at least.
For thread safety, it should be `volative` at least.

Please consider avoiding status in the `CoreEvent` class.

src/org/omegat/core/CoreEvents.java
```diff
@@ -123,6 +126,7 @@ public static void fireProjectChange(final IProjectEventListener.PROJECT_CHANGE_
                     log("ERROR_EVENT_PROJECT_CHANGE", t);
                 }
             }
+            isInProjectEventLoop = false;
```
miurahr, line 129: In order to access to the status of listener, you can use `PropertyChangeListener` that is Java standard mechanism and many swing components implemented.

src/org/omegat/core/data/RealProject.java
```diff
@@ -714,6 +714,12 @@ public void compileProjectAndCommit(String sourcePattern, boolean doPostProcessi
         // them to
         // the repository (BUGS#1176)
         CoreEvents.fireProjectChange(IProjectEventListener.PROJECT_CHANGE_TYPE.COMPILE);
+        while (CoreEvents.isInProjectEventLoop) {
```
miurahr, line 717: This should not be a field access.

miurahr, line 717: Could you please implement a generic method that fire event and wait completion, and call the new method instead of calling `CoreEvent.fireProjectChange` function?

miurahr (changes requested):
We need a modern design for invoking events and waiting an execution. A mechanism should be a thread safe.

miurahr:
Here is an sample how to implement `involeAndWait` function

```java


    /**
     * Notifies all registered project change listeners about a change in the project state
     * and waits until all listeners have processed the event or the specified timeout has elapsed.
     *
     * @param eventType the type of project change event that has occurred. It must be one of the constants
     *                  defined in {@link IProjectEventListener.PROJECT_CHANGE_TYPE}.
     * @param wait      the maximum time to wait for all listeners to process the event, in seconds.
     */
    public static void fireProjectChangeAndWait(final IProjectEventListener.PROJECT_CHANGE_TYPE eventType, int wait) {
        synchronized (PROJECT_EVENT_LISTENERS) {
            CountDownLatch latch = new CountDownLatch(PROJECT_EVENT_LISTENERS.size());
            SwingUtilities.invokeLater(() -> {
                Log.logInfoRB("LOG_INFO_EVENT_PROJECT_CHANGE", eventType);
                PROJECT_EVENT_LISTENERS.parallelStream().forEach(listener -> {
                    try {
                        listener.onProjectChanged(eventType);
                        latch.countDown();
                    } catch (Exception e) {
                        log("ERROR_EVENT_PROJECT_CHANGE", e);
                    }
                });
            });
            try {
                latch.await(wait, TimeUnit.SECONDS);
            } catch (InterruptedException e) {
                log("ERROR_EVENT_PROJECT_CHANGE_WAIT", e);
                Thread.currentThread().interrupt();
            }
        }        
    }
```

src/org/omegat/core/data/RealProject.java
```diff
@@ -714,6 +714,12 @@ public void compileProjectAndCommit(String sourcePattern, boolean doPostProcessi
...
+            Thread.yield();
+        }
+        while (org.omegat.gui.scripting.ScriptingWindow.window.inEventsLoop > 0) {
```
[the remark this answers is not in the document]
miurahr: Because there is no timeout, when script become infinite loop, this blocks infinitely. We should consider timeout for waiting.

brandelune:
> I already answered (see last comment about "RealProject.java") To be short: even if I agree in the principle about the changes requested by Hiroshi I don't have actually the time neither to implement nor to test these changes (while the actual version has been tested by Manuel) If what I implemented is not acceptable, then you will have to wait.

@miurahr can you make a decision here please?

miurahr:
There is errors in CI. I don't merge.

brandelune:
> There is errors in CI. I don't merge.

Ok, so we can’t keep this with a 6.0.2 target.

@miurahr can you evaluate a proper milestone?

PR #1259, opened 2025-01-31
chore: use custom launch4j binaries

miurahr:
When building OmegaT from source on Linux ARM64 platform, it failed to run `launch4j`.
The `launch4j` somehow dead project, so I maintain the forked version of the `launch4j` and the topic branch here use the forked version of it. Because `gradle-launch4j` has a bug which configuration syntax `launch4jBin` does not support multi-platform project, I detect os and arch then conditional configuration is used.

## Pull request type
- Build and release changes 

## Which ticket is resolved?

-    Support Windows 11 and Linux on ARM64 for development environment
-    https://sourceforge.net/p/omegat/feature-requests/1773/

## What does this PR change?

- detect os and architecture
- use forked version of launch4j

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -37,8 +37,6 @@ tasks.named('updateDaemonJvm') {
 def os = osdetector.os
 def osArch = osdetector.classifier
 
-// Define target Java version to compatible.
-def javaVersion = 11;
 
 // OmegaT distribution package meta data.
 def shortDescription = 'The free translation memory tool'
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -33,6 +34,11 @@ def gradleOnJava17OrLater = JavaVersion.current().isCompatibleWith(JavaVersion.V
 tasks.named('updateDaemonJvm') {
     jvmVersion = JavaVersion.VERSION_17
 }
+def os = osdetector.os
+def osArch = osdetector.classifier
+
+// Define target Java version to compatible.
+def javaVersion = 11;
```
miurahr, line 41:
```suggestion
def javaVersion = 11
```

miurahr, line 41:
```suggestion
```

PR #1265, opened 2025-02-07
move junixsocket to 2.10.1 for macOS notarization

brandelune:
Current junixsocket is too old and block OmegaT 6.0 macOS notarization.

https://sourceforge.net/p/omegat/bugs/1283/

This fix upgrades the library to its latest version and allows 6.0 macOS notarization.


miurahr:
LGTM.

PR #1267, opened 2025-02-09
Change label for "break/exception" checkbox

brandelune:
Change UI label for breaks setting UI checkbox.
https://sourceforge.net/p/omegat/feature-requests/484/


miurahr (approved):

miurahr:
It is better to have UI test case, and it can be another PR.

brandelune:
> It is better to have UI test case, and it can be another PR.

I don’t know what UI test cases are or how to create them.

If you provide guidance I might be able to do that myself.

PR #1271, opened 2025-02-13
Add direct links to items in the HTML manual

brandelune:
This feature adds a small space in front of relevant headers to provide access to direct links to the related subjects.

## Which ticket is resolved?
https://sourceforge.net/p/omegat/documentation/410/

### DIFF
--- a/doc_src/en/First_Steps.xml
+++ b/doc_src/en/First_Steps.xml
@@ -5,7 +5,7 @@
 <article lang="en" id="first.steps">
   <title>OmegaT</title>
   <section>
-	<title id="first-steps">First steps</title>
+	<link linkend="first-steps"> </link><title id="first-steps">First steps</title>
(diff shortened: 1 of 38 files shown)


doc_src/en/First_Steps.xml
```diff
@@ -5,7 +5,7 @@
 <article lang="en" id="first.steps">
   <title>OmegaT</title>
   <section>
-	<title id="first-steps">First steps</title>
+	<link linkend="first-steps"> </link><title id="first-steps">First steps</title>
```
miurahr, line 8: First Steps is not a manual page but greeting contents.
It does not have a link

brandelune: Thank you for catching this one.

It was hard enough to batch process all that with all the IDs that were not standardized. I spent most of my time visually checking the result by using a different CSS that made the invisible space appear as a red dot...

PR #1277, opened 2025-02-23
feat: expand manual in the application data directory to allow users to keep bookmark on the browser

miurahr:
- allow users to make bookmark in the browser

## Pull request type

## Which ticket is resolved?

- https://sourceforge.net/p/omegat/feature-requests/1778/
- expand manual in the application data directory to allow users to keep bookmark on the browser

- related https://sourceforge.net/p/omegat/mailman/message/59152285/

## What does this PR change?

- Expand HTML manual under `~/Library/Application Support/OmegaT/manual/<version>/<lang>/` instead of temporary folder on macos. `%USERPROFILE%\\AppData\\Local\OmegaT` on Windows and `~/.local/share/OmegaT/` on Linux.

### DIFF
--- a/src/org/omegat/Bundle.properties
+++ b/src/org/omegat/Bundle.properties
@@ -1962,7 +1962,7 @@ SU_CONFIG_DIR_CREATE_ERROR=Your operating system refused to let OmegaT\n\
 SU_SCRIPT_DIR_CREATE_ERROR=It was not possible to create the script folder\n\
     inside the configuration folder. The configuration folder will be used\n\
     instead.
-SU_DATA_DIR_CREATE_ERROR=Your operationg system refused to let OmegaT create the application data folder.
+SU_DATA_DIR_CREATE_ERROR=Your operating system refused to let OmegaT create the application data folder.
 
 # HttpConnectionUtils
 
--- a/src/org/omegat/Bundle.properties
+++ b/src/org/omegat/Bundle.properties
@@ -1962,6 +1962,7 @@ SU_CONFIG_DIR_CREATE_ERROR=Your operating system refused to let OmegaT\n\
 SU_SCRIPT_DIR_CREATE_ERROR=It was not possible to create the script folder\n\
     inside the configuration folder. The configuration folder will be used\n\
     instead.
+SU_DATA_DIR_CREATE_ERROR=Your operationg system refused to let OmegaT create the application data folder.
 
 # HttpConnectionUtils
 
--- a/build.gradle
+++ b/build.gradle
@@ -1525,6 +1525,7 @@ tasks.jacocoTestCoverageVerification {
 
 tasks.getByName("run") {
     jvmArgs(["--add-opens", "java.desktop/sun.awt.X11=ALL-UNNAMED"])
+    dependsOn manualZips
 }
 
 tasks.register('debug', JavaExec) {
@@ -1545,6 +1546,7 @@ tasks.register('runOnJava17', JavaExec) {
     mainClass.set application.mainClass
     classpath = sourceSets.main.runtimeClasspath
     jvmArgs(["--add-opens", "java.desktop/sun.awt.X11=ALL-UNNAMED"])
+    dependsOn manualZips
     group = 'application'
 }
 
@@ -1557,6 +1559,7 @@ tasks.register('runOnJava21', JavaExec) {
     mainClass.set application.mainClass
     classpath = sourceSets.main.runtimeClasspath
     jvmArgs(["--add-opens", "java.desktop/sun.awt.X11=ALL-UNNAMED"])
+    dependsOn manualZips
     group = 'application'
 }
 
--- a/src/org/omegat/help/Help.java
+++ b/src/org/omegat/help/Help.java
@@ -8,7 +8,7 @@
                2007 Didier Briel
                2009 Alex Buloichik
                2015 Aaron Madlon-Kay
-               2023 Hiroshi Miura
+               2023-2025 Hiroshi Miura
                Home page: https://www.omegat.org/
                Support center: https://omegat.org/support
 
@@ -89,7 +89,8 @@ public final class Help {
     /**
      * Shows help in the system browser.
      *
-     * @throws IOException when URI creation failed.
+     * @throws IOException
+     *             when URI creation failed.
      */
     public static void showHelp() throws IOException {
         String lang = detectHelpLanguage();
--- a/src/org/omegat/help/Help.java
+++ b/src/org/omegat/help/Help.java
@@ -48,6 +48,7 @@ import java.util.zip.ZipEntry;
 import java.util.zip.ZipInputStream;
 
 import org.omegat.util.Language;
+import org.omegat.util.Log;
 import org.omegat.util.OConsts;
 import org.omegat.util.OStrings;
 import org.omegat.util.StaticUtils;
@@ -88,7 +89,7 @@ public final class Help {
     /**
      * Shows help in the system browser.
      *
-     * @throws IOException
+     * @throws IOException when URI creation failed.
      */
     public static void showHelp() throws IOException {
         String lang = detectHelpLanguage();
@@ -125,7 +126,11 @@ public final class Help {
             return null;
         }
         try {
-            Path destinationDir = Files.createTempDirectory("omegat-" + OStrings.VERSION + "-help-" + lang);
+            Path destinationDir = Paths.get(StaticUtils.getConfigDir(), "manual", OStrings.VERSION, lang);
+            if (destinationDir.resolve("index.html").toFile().exists()) {
+                // already have manual
+                return destinationDir.toUri();
+            }
             return extractZip(zipFile, destinationDir).toURI();
         } catch (IOException ignored) {
         }
@@ -148,7 +153,7 @@ public final class Help {
             try {
                 cleanUp(destinationDir);
             } catch (IOException e) {
-                e.printStackTrace();
+                Log.log(e);
             }
         }));
         return destinationDir.resolve(OConsts.HELP_HOME).toFile();
@@ -223,7 +228,7 @@ public final class Help {
 
     /**
      * Detects the documentation language to use.
-     *
+     * <p>
      * If the latest manual is not available in the system locale language, it
      * returns null, i.e. show a language selection screen.
      */
--- a/src/org/omegat/help/Help.java
+++ b/src/org/omegat/help/Help.java
@@ -40,15 +40,12 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.nio.file.StandardCopyOption;
-import java.util.Comparator;
 import java.util.Properties;
 import java.util.concurrent.ThreadLocalRandom;
-import java.util.stream.Stream;
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipInputStream;
 
 import org.omegat.util.Language;
-import org.omegat.util.Log;
 import org.omegat.util.OConsts;
 import org.omegat.util.OStrings;
 import org.omegat.util.StaticUtils;
--- a/src/org/omegat/help/Help.java
+++ b/src/org/omegat/help/Help.java
@@ -151,27 +151,7 @@ public final class Help {
                 zipInputStream.closeEntry();
             }
         }
-        Runtime.getRuntime().addShutdownHook(new Thread(() -> {
-            try {
-                cleanUp(destinationDir);
-            } catch (IOException e) {
-                Log.log(e);
-            }
-        }));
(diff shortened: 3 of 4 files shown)


brandelune:
This does not work. The expansion is temporary. The manual is only available when OmegaT is running.

We need the manual to be **always** available, from a specific place, even when OmegaT is not running.

We need to have the same situation as 6.0.

I don’t understand what you are trying to achieve with temporary expansion of the manual in 6.1. If you want a zip distribution and a location that’s more OS specific, it is OK. But we can’t have a manual whose availability depends on OmegaT running or not.

miurahr:
I use `StaticUtils.getConfigDir()` + `manual/<version>/<lang>/` for the expansion directory.  The function returns ` ~/Library/Application Support/OmegaT/` on macOS.

It looks no problem. Any issue on that?

miurahr:
The equivalent folder on Windows is `%APPDATA%\OmegaT`, and it seems `~/.local/share/OmegaT` in linux.

brandelune:
> I use `StaticUtils.getConfigDir()` + `manual/<version>/<lang>/` for the expansion directory. The function returns ` ~/Library/Preference/OmegaT/` on macOS.
> 
> You suggest `~/Library/Application Support/OmegaT/` which may not be detected by `StaticUtils`.

I understand.

Here is the suggestion by Apple:

https://developer.apple.com/library/archive/qa/qa1170/_index.html

For exemple, the BBEdit manual is here:
~/Library/Application Support/BBEdit/BBEdit User Manual (15.1.4).pdf

miurahr:
Implement `StaticUtils.getApplicationData` function and update the manual expand folder.

brandelune:
> Implement `StaticUtils.getApplicationData` function and update the manual expand folder.

Thank you. The location is better, in my opinion.

Now, the bookmark won’t work since the manual disappears when OmegaT is closed.

miurahr:
Is `C:\\Users\\<user name>\\AppData\\Roaming\\OmegaT\\` acceptable for Windows users?
I am wondering because the folder is a special folder. When searching for knowledge, I got

>  The "Roaming" folder, located at %AppData%\Roaming, is a special folder in the Windows operating system that stores user-specific data that should be synchronized across multiple devices. This folder is part of the User Profile, which is stored in the C:\Users\<username>\ directory.
> 
> The "Roaming" folder is used to store data that should "roam" with the user, meaning that it should be available on any device where the user logs in with their Windows account. This is particularly useful in scenarios where users have multiple devices, such as a desktop and a laptop, and want to access their data from any device.
> 
> The "Roaming" folder is suitable for storing application data that meets the following criteria:
> 
>     1. User-specific: The data is specific to the user and should be available only to them.
>     2. Small in size: The data should be relatively small, as it will be synchronized across devices, which can impact performance and bandwidth.
>     3. Non-machine-specific: The data should not be specific to a particular device or machine.
> 
> Examples of data that are well-suited for the "Roaming" folder include:
> 
>     - User preferences and settings
>     - Application configuration files
>     - Small databases or caches
>     - User-created content, such as documents or images
> 
> As for storing application help data, it might not be the best fit for the "Roaming" folder, depending on the type and size of the data. If the help data is:
> 
>     - Large in size (e.g., videos, large PDFs)
>     - Not user-specific (e.g., generic help files that don't change often)
>     - Not required to be synchronized across devices
> 
> you might consider storing it in a different location, such as:
> 
>     - The application's installation directory (e.g., C:\Program Files\<app_name>)
>     - A subfolder within the user's "Local" AppData folder (e.g., %AppData%\Local\<app_name>)
>     - A separate data storage location, such as a database or a cloud storage service
> 
> However, if your application help data is small, user-specific, and needs to be synchronized across devices, the "Roaming" folder could be a suitable location.
> 
> Keep in mind that the "Roaming" folder has some limitations and considerations, such as:
> 
>     - Limited storage space (typically 5 GB per user)
>     - Synchronization delays or conflicts between devices
>     - Potential security and privacy concerns if sensitive data is stored in the "Roaming" folder
>  

So I want to propose "local" folder instead of a roaming folder. Any opinions?

brandelune (approved):

brandelune:
> Is `C:\\Users\\<user name>\\AppData\\Roaming\\OmegaT\\` acceptable for Windows users? I am wondering because the folder is a special folder. 

> So I want to propose "local" folder instead of a roaming folder. Any opinions?

I think what matters is that it is not the configuration folder. We should ask the user list.

miurahr:
Currently on Windows platform, `StaticUtils.getConfigDir()` returns `$USERPROFILE%\\AppData\\Roaming`.
From a principle of the folder definitions, it is for the configuration which is portable among devices. It is not good for plugins, logs, dictionaries and manuals. These have a better location which is `%USERPROFILE%\\Local` folder that is for these type of files.

I the same sense,

brandelune:
> > > Is `C:\\Users\\<user name>\\AppData\\Roaming\\OmegaT\\` acceptable for Windows users? I am wondering because the folder is a special folder.
> > 
> > 
> > > So I want to propose "local" folder instead of a roaming folder. Any opinions?
> > 
> > 
> > I think what matters is that it is not the configuration folder. We should ask the user list.
> 
> This is purely technical question. Why ask user list?

Because I have no idea what is technically best for Windows.

brandelune:
> As you know, the directories `Local`, `LocalLow`, and `Roaming` under `\Users\<username>\AppData` are part of the Windows operating system's user data storage hierarchy. These directories serve a specific purpose:
> 
> **`Local (%USERPROFILE%\AppData\Local)`:**
> 
> * Stores application data that is specific to the current machine and user account
> * Data is not synced with other devices or the cloud
> * Typically used for storing:
>   
>   * Cache files
>   * Temporary data
>   * Application settings that are machine-specific
>   * Data that is too large to be synced (e.g., game saves, video editing projects)
> * Examples of data stored in Local include:
>   
>   * Browser cache and cookies
>   * Application logs
>   * Temporary files generated by applications
> 
> **`Roaming (%USERPROFILE%\AppData\Roaming)`:**
> 
> * Stores application data that is synced across devices using the same user account
> * Data is synced with other devices connected to the same Microsoft account or omain
> * Typically used for storing:
>   -  Application settings that are user-specific and should be synced (e.g., browser bookmarks, email client settings)
>   -  Small to medium-sized data files that are frequently accessed (e.g., documents, spreadsheets)
>   -  Data that is not machine-specific and can be safely synced
> * Examples of data stored in Roaming include:
>   -  Browser bookmarks and extensions
>   -  Email client settings and contacts
>   -  Office application settings and templates
> 
> In summary:
> 
> * Local stores machine-specific, non-synced data
> * LocalLow stores low-integrity, non-synced data with additional security restrictions
> * Roaming stores user-specific, synced data that is shared across devices
> 
> That is why I propose `Local` folder instead of `Roaming` folder. Manual page is not a device portable configuration so `Roaming` is not good place to expand.

But a manual is something you would want to have access to from a portable device so as to read it separately. I don’t understand why MS suggests that should not be the case.

miurahr:
> But a manual is something you would want to have access to from a portable device so as to read it separately. I don’t understand why MS suggests that should not be the case.

size

miurahr (commented):
Coding style can be improved

brandelune:
> > But a manual is something you would want to have access to from a portable device so as to read it separately. I don’t understand why MS suggests that should not be the case.
> 
> Its size and the manual data can be reproducible.

Obviously, MS does not care much about size and tries to sell OneDrive sharing device. But I am not in a position to have informed opinions on Windows, so please select what you think it best according to the documentation.

miurahr:
> > > But a manual is something you would want to have access to from a portable device so as to read it separately. I don’t understand why MS suggests that should not be the case.
> > 
> > 
> > Its size and the manual data can be reproducible.
> 
> Obviously, MS does not care much about size and tries to sell OneDrive sharing device. But I am not in a position to have informed opinions on Windows, so please select what you think it best according to the documentation.

The Roaming data is held by Microsoft Windows Server PDC as user profile or Microsoft account mechanism. It is not the OneDrive sharing storage.

miurahr:
> > > But a manual is something you would want to have access to from a portable device so as to read it separately. I don’t understand why MS suggests that should not be the case.
> > 
> > 
> > Its size and the manual data can be reproducible.
> 
> Obviously, MS does not care much about size and tries to sell OneDrive sharing device. But I am not in a position to have informed opinions on Windows, so please select what you think it best according to the documentation.

Exactly. The data is stored in OneDrive. I obtained an explanation how the data sync as follows.

> Microsoft Windows uses a feature called OneDrive to synchronize the Roaming folder (or Roaming Profile) across devices, primarily in enterprise or business environments. Here's how it works:
> 
> **Roaming Profile:** In a typical setup, a Roaming Profile is a user profile that includes user data such as desktop settings, documents, and other personalized elements. This allows the user’s settings and files to follow them across multiple devices in a network.
> 
> **OneDrive Integration:**
>        - With Windows 10/11 and OneDrive for business, OneDrive handles synchronization of user files across multiple devices. The Roaming folder in this case typically resides in OneDrive, allowing files and settings to be synchronized across devices.
>        - Files from the Roaming folder (like documents, pictures, and application settings) are automatically uploaded to the cloud and downloaded to other devices the user logs into.
> 
>  **Group Policy or Intune:** For enterprise environments, IT administrators may configure synchronization settings through Group Policy or Microsoft Intune to control what gets synchronized and manage it securely. These policies can enable or disable certain features like Known Folder Move, which moves important user folders (Documents, Pictures, etc.) to OneDrive, for seamless access across devices.
> 
>  **Active Directory:** In enterprise networks, Active Directory (AD) might be used in conjunction with Roaming Profiles to manage user settings. AD manages user permissions, and the Roaming Profile ensures that these settings follow the user across devices. OneDrive, in this case, could be used to store and sync data.
> 
>  **Roaming User Profile Folders Sync:** For enterprise solutions, the Roaming user profile ensures that if a user logs into a different computer, their personalized settings (themes, desktop shortcuts, etc.) are restored automatically.

In my old knowledge, Roaming data is stored in Active Directory. I remember a large size data in user profile make performance pressure for enterprise systems. So I said size.

src/org/omegat/Bundle.properties
```diff
@@ -1962,6 +1962,7 @@ SU_CONFIG_DIR_CREATE_ERROR=Your operating system refused to let OmegaT\n\
 SU_SCRIPT_DIR_CREATE_ERROR=It was not possible to create the script folder\n\
     inside the configuration folder. The configuration folder will be used\n\
     instead.
+SU_DATA_DIR_CREATE_ERROR=Your operating system refused to let OmegaT create the application data folder.
```
Kazephil, line 1965: Fixed a minor type, but looks OK otherwise.

Kazephil (approved):

Kazephil:
On a more general note about where to store the documentation..

As far as I can tell from what I see on my work computer, Windows. does not have a  standard location for documentation (and probably not even a recommended location).

If they do provide a local copy of their documentation, I think most programs use either a subfolder under the folder for the application in `C:\Program Files\<application>`, or under the (hidden by default, if I remember correctly) `C:\ProgramData\<application>` if it is meant to be availale to all users of the machine.

If documentation is only stored for the user that installed the program, I would expect to find it under `C:\Users\<user name>\AppData\Roaming\<application>\`.

The above is just what I've observed on the computer I use at work. It may not be correct as a generality.

On Linux, I believe the preferred location is `/usr/share/doc/<application>` for all users (requiring root privilege during installation, of course), and either `home/<username>/.config/<application>/<subdirectory>` or `home/<username>/.local/<application>/<subdirectory>` if the documentation is only put in userspace.

I would also note, however, that I find very few programs actually provide a local copy of their help or documentation by default, and simply access the website instead. Most software publishers seem to assume all users have a permanent, stable Internet connection these days…
(Some software—LibreOffice, for example—default to accessing the website, but also offer the option to download an offline copy of their documentation. That approach might be another option: give the user the choice to use the manual online or offline, and let them choose where they want to save it if they choose to use it offline.)

Just a few random thoughts.

miurahr:
> `C:\ProgramData\<application>`

Thank you. It seems the best place for OmegaT manual. The folder can be written by administrator or privileged application. The installer should expand manuals into the folder, I think.

miurahr:
> `C:\ProgramData\<application>`
Equivalent on Linux is `/usr/share/doc/<application>` for standard applications, and `/usr/local/share/doc/<application>` or `/opt/<application>/share/doc` for 3rd party applications.

miurahr:
> > `C:\ProgramData\<application>`
> 
> Thank you. It seems the best place for OmegaT manual. The folder can be written by administrator or privileged application. The installer should expand manuals into the folder, I think.

I put an RFE issue around the discussion.
https://sourceforge.net/p/omegat/feature-requests/1777/

It is ready to merge.

PR #1296, opened 2025-03-17
chore: adjust spotbugs error level and exclude configuration, with CI change of annotation configuration on GitHub Actions Spotbugs task

miurahr:
Update spotbugs staitc code analysis configuration.

1 . Not limit to HIGH severity but DEFAULT level 
2.  Update problem-matcher.json for GitHub Actions
3.  Improve SpotBugs configurations of Gradle plugin
4.  Exclude some errors which requires overhall of core application design
5.  Exclude generated code which is located in `src/gen/core` path.

## Pull request type

Please mark github LABEL of the type of change your PR introduces:

- Build and release changes -> [build/release]

## What does this PR change?

- GitHub workflow: Added a SpotBugs report step triggered on failures.
- SpotBugs configuration: Adjusted reporting levels and removed redundant tasks while optimizing configurations (reportLevel, onlyAnalyze, excludeFilter).
- New exclusion file: Introduced exclude.xml to handle specific SpotBugs warnings.
- Gradle upgrade: Updated Gradle Wrapper version to 8.12.1.
- Script adjustments: Minor changes to gradlew and gradlew.bat for better compliance and maintainability.
- Problem matcher: Improved SpotBugs regex patterns for better error capturing and added severity levels.

This PR is to make all SpotBugs error as build error.
The PR can be merged when All the fixes are merged because the CI build become failed.

## Fixes

- [x]  PR #1297
- [x]  PR #1298
- [x] PR #1299
- [x] PR #1300
- [x] PR #1301
- [x] PR #1305
- [x] PR #1307 
- [x] PR #1308
- [x] PR #1309
- [x] PR #1310 
- [x] PR #1311
- [x] PR #1312
- [x] PR #1313
- [x] PR #1315
- [x] PR #1316
- [x] PR #1317
- [x] PR #1328
- [x] PR #1360
- [x] PR #1416
- [x] PR #1431
- [x] PR #1453
- [x] PR #1456
- [x] PR #1457
- [x] PR #1458
- [x] PR #1464 
- [x] PR #1465
- [x] PR #1468 
- [ ]  PR #1469 
- [x] PR #1470
- [x] PR #1471
- [ ] PR #1472
- [x] PR #1473
- [x] PR #1474
- [x] PR #1476 
- [x] PR #1477 
- [x] PR #1478 
- [x] PR #1479 
- [x] PR #1480 
- [x] PR #1481 
- [x] PR #1482 
- [x] PR #1483 
- [x] PR #1484 
- [x] PR #1485 
[...]

PR #1298, opened 2025-03-17
Remove unnecessary exception and handle filter loading errors

miurahr:
Removed the "throws Exception" declaration from the constructor to simplify its signature. Added a try-catch block to gracefully handle potential IOException during `loadProjectFilters()` and ensure `projectFilters` is properly initialized.

## Pull request type

- Other (describe below)
refactoring to fix spotbugs error

## What does this PR change?

- Ctor should not throw exception for a security reason.
- Bug description: https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#ct-be-wary-of-letting-constructors-throw-exceptions-ct-constructor-throw
-
-

## Other information

> CT: Be wary of letting constructors throw exceptions. (CT_CONSTRUCTOR_THROW)[](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#ct-be-wary-of-letting-constructors-throw-exceptions-ct-constructor-throw)
> 
> Classes that throw exceptions in their constructors are vulnerable to Finalizer attacks
> 
> A finalizer attack can be prevented, by declaring the class final, using an empty finalizer declared as final, or by a clever use of a private constructor.
> 
> See [SEI CERT Rule OBJ-11](https://wiki.sei.cmu.edu/confluence/display/java/OBJ11-J.+Be+wary+of+letting+constructors+throw+exceptions) for more information.
> 

- pearent issue/pr #1296

### DIFF
--- a/src/org/omegat/core/data/ProjectProperties.java
+++ b/src/org/omegat/core/data/ProjectProperties.java
@@ -245,6 +245,7 @@ public class ProjectProperties {
         return tmDir.getAsString() + OConsts.PENALTY_TM + '/';
     }
 
+    @SuppressWarnings("unused")
     public ProjectPath getDictDir() {
         return dictDir;
     }
@@ -259,6 +260,7 @@ public class ProjectProperties {
         dictDir.setRelativeOrAbsolute(dictRoot);
     }
 
+    @SuppressWarnings("unused")
     public String getDictRootRelative() {
         return dictDir.getAsString();
     }
@@ -311,6 +313,7 @@ public class ProjectProperties {
         }
     }
 
+    @SuppressWarnings("unused")
     public void setSourceRootRelative(String sourceRootRelative) {
         if (!StringUtil.isEmpty(sourceRootRelative)) {
             sourceDir.setRelativeOrAbsolute(sourceRootRelative);
@@ -489,6 +492,14 @@ public class ProjectProperties {
         return repositories;
     }
 
+    /**
+     * Determines if the project is a team project. A project is considered a team project
+     * if it has repositories defined, and at least one repository mapping has either
+     * a local path set to an empty string or a single forward slash to specify OmegaT
+     * project root.
+     *
+     * @return true if the project is a team project, false otherwise
+     */
     public boolean isTeamProject() {
         if (hasRepositories()) {
             for (RepositoryDefinition repositoryDefinition : repositories) {
@@ -629,11 +640,10 @@ public class ProjectProperties {
         // creation.
         //
         File dict = new File(getDictRoot());
-        if (!dict.exists()) {
-            if (getDictRoot().equals(getProjectRoot() + OConsts.DEFAULT_DICT + '/')) {
-                dict.mkdirs();
-            }
+        if (!dict.exists() && getDictRoot().equals(getProjectRoot() + OConsts.DEFAULT_DICT + '/')) {
+            dict.mkdirs();
         }
+
     }
 
     public void autocreateDirectories() {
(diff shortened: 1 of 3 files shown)


brandelune (approved):

src/org/omegat/core/data/ProjectProperties.java
```diff
@@ -87,7 +87,7 @@ protected ProjectProperties() {
      * Default constructor to initialize fields (to get no NPEs). Real values
      * should be applied after creation.
      */
-    public ProjectProperties(File projectDir) throws Exception {
+    public ProjectProperties(File projectDir) {
```
miurahr, line 90: It should be checked that caller may have a try block, and now constructor does not throw exception.
We should update callers to remove a try block or a catch clause.

src/org/omegat/core/data/ProjectProperties.java
```diff
@@ -110,7 +110,11 @@ public ProjectProperties(File projectDir) throws Exception {
         setTargetLanguage("UK-UA");
 
         loadProjectSRX();
-        loadProjectFilters();
+        try {
+            loadProjectFilters();
+        } catch (IOException ignored) {
+            projectFilters = new Filters();
```
miurahr, line 116: This makes changing behavior of the error handling of project loading.
In a previous case, project loading become failed when filter configuration `filters.xml` corrupted.
This changes the behavior when the filter configuration is corrupted, it will be reset as default.
Is it intended?

src/org/omegat/core/data/ProjectProperties.java
```diff
@@ -490,13 +494,12 @@ public List<RepositoryDefinition> getRepositories() {
...
-        for (RepositoryDefinition repositoryDefinition : repositories) {
-            for (RepositoryMapping repositoryMapping : repositoryDefinition.getMapping()) {
-                if ("".equals(repositoryMapping.getLocal()) || "/".equals(repositoryMapping.getLocal())) {
-                    return true;
+        if (hasRepositories()) {
+            for (RepositoryDefinition repositoryDefinition : repositories) {
+                for (RepositoryMapping repositoryMapping : repositoryDefinition.getMapping()) {
+                    if ("".equals(repositoryMapping.getLocal()) || "/".equals(repositoryMapping.getLocal())) {
+                        return true;
+                    }
```
miurahr, line 502: Here we find a loop is too deep. Are there any way to simplify the logic?

miurahr (commented):
The changes are approved but there are several discussion points.

src/org/omegat/core/data/ProjectProperties.java
miurahr: The constructor here is called by 58 places. Many are located in language modules and spellchecker modules. Otherwise, there are 20 in the test code. Does the change affect 3rd party plugins?

src/org/omegat/core/data/ProjectProperties.java
miurahr: There is no way to detect it without traversing repository mapping definitions.

brandelune:
> The changes are approved but there are several discussion points.

I do not have the skills to understand what you do, but I can test a specified outcome.
Hence, anything that I cannot test, I trust you and I approve.
Anything that I can test, I test, comment and eventually approve.

Thank you for the message on the dev-list.

miurahr:
@brandelune  You are automatically registered as a reviewer because the change has a `Bundle.properties` message.

src/org/omegat/core/data/ProjectProperties.java
miurahr: I think it natural behavioir.

PR #1304, opened 2025-03-21
Fix: null check in saveWordLists method of SpellChecker

miurahr:
Add a safeguard to prevent potential NullPointerException in the saveWordLists method by ensuring file paths are not null before proceeding. This ensures stability when saving ignored and learned words.

## Pull request type

Please mark the GitHub LABEL of the change your PR introduces:

- Bug fix -> [bug]

## Which ticket is resolved?

- Spellchecker raised NPE when writing word list file
- https://sourceforge.net/p/omegat/bugs/1288/

## What does this PR change?

- Null check for the ignore words file and the learned file path on the saveWordLists method


miurahr:
The issue is essential and needs fixing early, but review feedback has never been received.
I want to merge it  w/o review.

PR #1307, opened 2025-03-22
refactor: adjust custom exception class not to expose internal object

miurahr:
This change make getparams method of KnownException class returns copy of a array of object. It reduces the risk of exposing internal fields.

## Pull request type

- Other (describe below)
refactor

## What does this PR change?

- Added a @Nullable annotation for better null-safety awareness.
- Modified the getParams() method to return a copy of the params array for immutability and preventing potential side effects.
- Introduced additional imports to support the changes (e.g., Arrays and Nullable).

## Other information

#1296

### DIFF
--- a/src/org/omegat/core/KnownException.java
+++ b/src/org/omegat/core/KnownException.java
@@ -24,8 +24,11 @@
  **************************************************************************/
 package org.omegat.core;
 
+import org.jspecify.annotations.Nullable;
 import org.omegat.util.OStrings;
 import org.omegat.util.StringUtil;
+
+import java.util.Arrays;
 
 /**
  * Know exception for some problems.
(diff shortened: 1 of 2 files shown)


brandelune (approved):

src/org/omegat/core/KnownException.java
```diff
@@ -47,7 +49,7 @@ public KnownException(Throwable ex, String errorCode, Object... params) {
     }
 
     public Object[] getParams() {
-        return params;
+        return Arrays.copyOf(params, params.length, Object[].class);
```
miurahr, line 52: This can be 
```suggestion
        return Arrays.copyOf(params, params.length);
```

miurahr: There is no usage of parameters for creation of the `KnownException` in the OmegaT source code.
It may be better to remove a field params and update constructor without params.

miurahr:
The PR takes the wrong approach to the issue.
There is no constructor call to give parameters so that we can remove `getParams` method.

PR #1308, opened 2025-03-22
Refactor: Main class to adjust visibility of fields

miurahr:
## Pull request type

- Other (describe below)
- refactor

## What does this PR change?

- Adjust for the following SpotBugs warnings.
```
M V MS_MUTABLE_COLLECTION_PKGPROTECT MS: org.omegat.Main.PARAMS is a mutable collection which should be package protected At Main.java:[line 136]	
M V MS_PKGPROTECT MS: org.omegat.Main.projectLocation should be package protected At Main.java:[line 130]	
M V MS_PKGPROTECT MS: org.omegat.Main.remoteProject should be package protected At Main.java:[line 133]	
M V MS_PKGPROTECT MS: org.omegat.Main.runMode should be package protected At Main.java:[line 139]	
M D NP_NULL_ON_SOME_PATH_FROM_RETURN_VALUE NP: Possible null pointer dereference in org.omegat.Main.restartGUI(String) due to return value of called method Dereferenced at Main.java:[line 267]	
M D REC_CATCH_EXCEPTION REC: Exception is caught when Exception is not thrown in org.omegat.Main.runGUI() At Main.java:[line 358]	
```

## Other information

#1296

### DIFF
--- a/src/org/omegat/Main.java
+++ b/src/org/omegat/Main.java
@@ -349,27 +349,12 @@ public final class Main {
         }
 
         Log.logInfoRB("STARTUP_GUI_DOCKING_FRAMEWORK", DockingDesktop.getDockingFrameworkVersion());
-
-        // Set X11 application class name to make some desktop user interfaces
-        // (like Gnome Shell) recognize OmegaT
-        Toolkit toolkit = Toolkit.getDefaultToolkit();
-        Class<?> cls = toolkit.getClass();
-        try {
-            if (cls.getName().equals("sun.awt.X11.XToolkit")) {
-                Field field = cls.getDeclaredField("awtAppClassName");
-                if (field.trySetAccessible()) {
-                    field.set(toolkit, "OmegaT");
-                }
-            }
-        } catch (Exception ignored) {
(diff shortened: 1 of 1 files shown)


brandelune (approved):

src/org/omegat/Main.java
```diff
@@ -264,14 +264,19 @@ public static void restartGUI(String projectDir) {
...
-            javaBin = Paths.get(StaticUtils.installDir()).getParent().resolve("bin/OmegaT");
-            if (!javaBin.toFile().exists()) {
-                // abort restart
-                Core.getMainWindow().displayWarningRB("LOG_RESTART_FAILED_NOT_FOUND");
+            var installDir = StaticUtils.installDir();
+            if (installDir == null) {
                 return;
+            } else {
+                javaBin = Paths.get(installDir).getParent().resolve("bin/OmegaT");
+                if (javaBin.toFile().exists()) {
```
miurahr, line 272: The logic become reverted. The change led the restart on jpackage distribution become failure.

PR #1314, opened 2025-03-23
refactor: define enum value by getString and avoid dynamic key for bundle lookup

miurahr:
- ExternalFinder construct bundle key from enum name. This is to avoid a dynamic key and use enum ctor to define value.
- Add javadoc for enum entries

## Pull request type

- Other (describe below)
refactor

## What does this PR change?

-
-
-


brandelune (approved):

Kazephil (approved):
Looks fine to me.

PR #1315, opened 2025-03-23
refactor: Color style enum name definition on ctor

miurahr:
Improve Color style definitions.
 
## Pull request type

- Other (describe below)
refactor

## What does this PR change?

- Update Styles class ctor to use explicit bundle key access
- Add javadoc for enum entries
- Define error message in Bundle.properties, which was defined in direct string in the source.
- Updated comments throughout the code for clarity and accuracy.

## Other information

#1296

### DIFF
--- a/src/org/omegat/Bundle.properties
+++ b/src/org/omegat/Bundle.properties
@@ -2482,8 +2482,6 @@ FILENAMEPATTERNS_ADD=&Add
 FILENAMEPATTERNS_REMOVE=&Remove
 
 # Colour names
-GUI_COLORS_SELECTION=Custom Colours Selection
-GUI_COLORS_TITLE=Custom Colours
 COLOR_BACKGROUND=Background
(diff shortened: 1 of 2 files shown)


src/org/omegat/Bundle.properties
```diff
@@ -2543,6 +2541,7 @@ COLOR_MARK_ALT_TRANSLATION=Alternative translation highlight
 GUI_COLORS_COLOR=Item:
 GUI_COLORS_RESET_COLOR=&Reset Colour
 PREFS_TITLE_COLORS=Colours
+PREFS_COLOR_VALUE_PARSE_ERROR=Cannot set custom color for {0}, default to {1}.
```
Kazephil, line 2545: I'm changing "color" to "colour" and proposing a tweak the phrasing.

miurahr: thank you.

PR #1372, opened 2025-05-03
Update 45.LocalizeApplicationAndManuals.md

brandelune:
Thorough l10n process update.

I may have forgotten something. Please check.


miurahr:
Thanks for addition.

PR #1376, opened 2025-05-04
Update 01.ContributingOmegaT.md

brandelune:
Wording, structure and links.

Also, I referred to a "Coding conventions" chapter that is also referred to in the introduction, but I could not find it. We need to fix that link before release.


brandelune:
> Wording, structure and links.
> 
> Also, I referred to a "Coding conventions" chapter that is also referred to in the introduction, but I could not find it. We need to fix that link before release.

I found the relevant files and added the link, to "30.CodingStyles.md"

PR #1382, opened 2025-05-04
Update Bundle.properties

brandelune:
Slightly change wording


miurahr:
Thank you.

I think you can directly commit to the topic branch because you are a repository owner. Or you can put it as a review comment in GitHub Pull-Reqeust, then I can merge it to the topic branch as reviewer feedback.

PR #1387, opened 2025-05-05
the 'latest' suffix is not used anymore

brandelune:
When using the online help, the master build points at "omegat-latest" on sourceforge.

The "latest" suffix is not used anymore and instead of pointing at a recent "weekly" build, it points at obsolete data.

This is a temporary fix until we implement a "weekly" or "dev" suffix for the online documentation system.

### DIFF
--- a/src/org/omegat/help/Help.java
+++ b/src/org/omegat/help/Help.java
@@ -74,7 +74,7 @@ public final class Help {
      */
     public static final String ONLINE_HELP_URL = OStrings.IS_BETA
             //? "https://omegat.sourceforge.io/manual-latest/"
-            ? "https://omegat.sourceforge.io/manual-standard/"
+            ? "https://omegat.sourceforge.io/manual-snapshot/"
             : "https://omegat.sourceforge.io/manual-standard/";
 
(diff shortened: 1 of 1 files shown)


src/org/omegat/help/Help.java
```diff
@@ -69,13 +69,17 @@ private Help() {
...
     /**
      * URL for the online manual.
+     * TODO implement a "manual-dev" and "javadoc-dev"
+     * since the "latest" suffix is not used anymore
      */
     public static final String ONLINE_HELP_URL = OStrings.IS_BETA
-            ? "https://omegat.sourceforge.io/manual-latest/"
+            //? "https://omegat.sourceforge.io/manual-latest/"
+            ? "https://omegat.sourceforge.io/manual-standard/"
             : "https://omegat.sourceforge.io/manual-standard/";
```
miurahr, line 78: Why not remove redundant conditional operator?

It can be `public static final String ONLINE_HELP_URL = "https://omegat.sourceforge.io/manual-standard/";`

brandelune:
> There is no javadoc publishing CI/CD configuration. Javadoc should be stable, so it is only for stable version.

Ok, thank you. So I’m merging.

PR #1388, opened 2025-05-05
[6.0] chore: check for the existence of the menu invocation action method

miurahr:
- Backport the check for BUGS#1290
- OmegaT 6.0.x does not have the same issue.
- The bug is not detected in master branch because there is missing the test, so backport the test case.

## Pull request type

## What does this PR change?

- Added testMenuActions_invokeActions test method to validate that all menu actions in the source match defined methods in MainWindowMenuHandler.
- Introduced processSourceContent utility method for processing .java files in the /src directory.
- Improved code readability by:
    -  Replacing generic type declarations with the diamond operator (<>).
    - Reformatting assertions for better clarity.

### DIFF
--- a/test/src/org/omegat/gui/main/MainWindowMenuTest.java
+++ b/test/src/org/omegat/gui/main/MainWindowMenuTest.java
@@ -28,16 +28,29 @@ package org.omegat.gui.main;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertTrue;
 
+import java.io.IOException;
 import java.lang.reflect.Field;
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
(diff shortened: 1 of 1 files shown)


test/src/org/omegat/gui/main/MainWindowMenuTest.java
```diff
@@ -84,7 +97,63 @@ public void testMenuActions() throws Exception {
...
+                    }
+                }
+            }
+        });
+    }
+
+    /**
+     * Process the text content of all .java files under /src. Also check
+     * DIRECTIONALITY_LEFT_TO_RIGHT_OVERRIDE for security reasons.
+     * And check invisible space character.
```
miurahr, line 131: Javadoc is not correct. I removed the feature which is new rule for master branch.

PR #1392, opened 2025-05-10
Do not eat exception in loadConfFile

t-cordonnier:
SRX.loadConfigFile: If we fail to read the file, throw the exception instead of silently ignore it In unit tests, that can show whenever we break compatiblityhat is normal. 

## Pull request type

- Bug fix -> [bug]
- 
## Which ticket is resolved?

None, it is done to make resolution of CVE-2024-51366 better


miurahr:
Please fix
> checkstyle
> Unused import - java.io.IOException.

PR #1394, opened 2025-05-11
fix: filters4: resource leaks in AbstractZipFilter class

miurahr:
Fix leaks in StAX filter class.

## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?

There is no issue to point

## What does this PR change?

- Ensure closure of XMLReader resource

### DIFF
--- a/src/org/omegat/filters4/AbstractZipFilter.java
+++ b/src/org/omegat/filters4/AbstractZipFilter.java
@@ -80,6 +80,7 @@ public abstract class AbstractZipFilter extends AbstractFilter {
                 }
             }
         } catch (IOException ignored) {
+            // always considered non-supported
         }
         return false;
     }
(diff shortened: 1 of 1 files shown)


brandelune (approved):
I don’t understand how that affects the code, so I trust that you are correct.

miurahr:
waiting @t-cordonnier comment.

miurahr:
The key of the refactor.

There was a private helper method `AbstractZipFilter#processFileImpl` and `translateEntry`.
There is a object creation in `translateEntry` 
```
 BufferedWriter writer = new BufferedWriter(
                        new OutputStreamWriter(zipout, xReader.getEncoding()));
```
in the loop and not closed properly.

The proposal moves `writer` object creation in `public void processFile` API method with another file I/O objects inside try-with-resources to guarantee the close of the resource.

```
        try (ZipFile zipFile = new ZipFile(inFile);
             ZipOutputStream zipOutputStream = createZipOutputStream(outFile);
             BufferedWriter writer = createWriter(zipOutputStream)) {
```

To realize the structure change, the helper method `createWriter` requires encoding of the internal XML file.
I introduce the `private Charset internalEncoding` field, and give it in the constructor. and also detect when `isFileSupported` called.

miurahr:
@t-cordonnier  could you please drop your comment for the PR?

t-cordonnier:
> @t-cordonnier could you please drop your comment for the PR?

Hi Hiroshi, sorry that I did not have the time to do it before, I have added a few comments but globally I agree with this change. I know that it is already merged but you can still do the modification about encoding. The other is not a mistake but simply something which must be noticed for the future.

PR #1396, opened 2025-05-13
fix: Update segmentation.conf loader to fix the vulnerability reported as CVE-2024-51366 with XSLT stylesheet to convert to SRX standard format.

t-cordonnier:
Corrects CVE 2024 51366 by replacing XMLDecoder with XSLT stylesheet

## Pull request type

- Bug fix -> [bug]

## What does this PR change?

- SRX Test : in existing migration tests, remove one rule to ensure the test fails if you accidentally fallback to default rules
- Add one test with a conf file made without any relation to default rules
- Reimplement loadConfigFile using XSL stylesheet, to solve vulnerability

## Other information

After conversion the method loadConfigFile reads the newly generated SRX file. So it can work only if OmegaT implements loading of SRX files, which is not the case for OmegaT 6.0
This PR is only for 6.1 for the moment


miurahr:
Do you forget to add xslt file to git repository?

t-cordonnier:
You are true, I had forgotten. Just pushed a new release where I also add a copy of your unit test for the vulnerability, showing that it is not here anymore

miurahr:
Please also use two malicious data sets for testing from https://github.com/omegat-org/segmentation-migrator/tree/main/src/test/resources/segmentation 
These should not be accepted by the converter and notify users of the issue with their input files.

miurahr:
I prefer to move migration logics and the XSLT file to a separate package `org.omegat.convert.segmentation`.
I prefer `SegmentationConfMigrator.java` or something similar to the converter's name.

I also propose the `segmentation.conf` validator as `org.omegat.convert.segmentation.SegmentationConfValidator.java` as a separate PR soon.

t-cordonnier:
Already added one example of malicious data (commit cbf4df4ec905099d0bc9e8e1dbadc5d7730b5180)
And nothing prevents you from accepting the PR and add more tests after.

test/src/org/omegat/core/segmentation/SRXTest.java
```diff
@@ -197,7 +220,40 @@ public static void testSrxMigration(File segmentConf, File configDir) throws Exc
...
+        public final TemporaryFolder folder = TemporaryFolder.builder().assureDeletion().build();
+
+        
+        @Test
+        public void testSRXLoaderSecureCVE_2024_51366() throws IOException {
+            File tmpDir = folder.newFolder();
+            Path segmentConf = tmpDir.toPath().resolve("segmentation.conf");
+            // prepare CVE-2024-51366 exploit code
+            String xmlContent = "<java>\n" +
+                    "    <object class=\"java.lang.ProcessBuilder\">\n" +
```
miurahr, line 239: I found it essential that the attack code is not well formatted to prevent pattern matching.

```suggestion
                    "    <object\nclass=\n\"java.lang.ProcessBuilder\">\n" +
```

miurahr (changes requested):
In my recent investigation, I found the attack code for testing should not be well-formated to prevent the pattern matching.

miurahr (changes requested):
Please adjust two checkstyle errors. 

```
[ERROR] /home/runner/work/omegat/omegat/src/org/omegat/core/segmentation/SRX.java:47:8: Unused import - javax.xml.transform.TransformerException. [UnusedImports]
[ERROR] /home/runner/work/omegat/omegat/src/org/omegat/core/segmentation/SRX.java:233:43: ',' is not followed by whitespace. [WhitespaceAfter]
```

test/src/org/omegat/core/segmentation/SRXTest.java
t-cordonnier: no problem for me

PR #1406, opened 2025-05-16
fix: Enhance error handling for HTTP response exceptions

miurahr:
## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?

- When mapping contents not exist, the error message does not provide enough information.
- https://sourceforge.net/p/omegat/bugs/1295/

## What does this PR change?

- Improve error handling of http remote repository access.
- use NetworkException instead of generic RuntimeException

### DIFF
--- a/src/org/omegat/Bundle.properties
+++ b/src/org/omegat/Bundle.properties
@@ -2342,6 +2342,11 @@ TEAM_OPEN=Opening team project
 TEAM_NETWORK_ERROR=Remote repository network communication error: {0}
 TEAM_UPDATE_REPO_ERROR=Error updating repository. Skipping. Error: {0}
 
+TEAM_HTTP_FORBIDDEN=Access to {0} is forbidden.
+TEAM_HTTP_UNAUTHORIZED=Access to {0} is rejected by authentication error.
+TEAM_HTTP_NOT_FOUND=Contents at {0} is not found.
+TEAM_HTTP_OTHER_ERRORS=HTTP connection {1} response with code: {2}
+
 # Save options
 SAVE_DIALOG_TITLE=Saving and Output Options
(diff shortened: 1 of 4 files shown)


src/org/omegat/core/team2/impl/HTTPRemoteRepository.java
```diff
@@ -217,7 +217,8 @@ protected void retrieve(Properties etags, String file, String url, final File ou
                 logger.atDebug().setMessage("Retrieve {0}: not modified").addArgument(url).log();
                 return;
             default:
```
miurahr, line 219: It is better to handle more precise detection of the reasons.
We can use different message for 404 and 431 error.

miurahr: I want to check forbidden, unauthorized and not_found errors.

src/org/omegat/Bundle.properties
```diff
@@ -2342,6 +2342,11 @@ TEAM_OPEN=Opening team project
 TEAM_NETWORK_ERROR=Remote repository network communication error: {0}
 TEAM_UPDATE_REPO_ERROR=Error updating repository. Skipping. Error: {0}
 
+TEAM_HTTP_FORBIDDEN=Access to {0} is forbidden.
+TEAM_HTTP_UNAUTHORIZED=Access to {0} is rejected by authentication error.
+TEAM_HTTP_NOT_FOUND=Contents at {0} is not found.
+TEAM_HTTP_OTHER_ERRORS=HTTP connection {1} response with code: {2}
+
```
Kazephil, line 2349: Proposing a few tweaks to the wording.

miurahr: Thank you for improvement.

PR #1422, opened 2025-05-26
fix: team/HTTPRemoteRepository: handle HTTP_NOT_MODIFIED response correctly

miurahr:
Fix the issue I entered in last change for HTTPRemoteRepository class.

## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?
There is no issue to point
## What does this PR change?

- Fix the handling of 304 http response:
    - Consolidated HTTP error response handling into the main method instead of a separate function.
    - Just return when 304 HTTP response from the main method.
- Simplification of logging and exception mechanisms:
    - Refactored logging for better readability and debugging support.
    - Improved consistency in method formatting and indentation.
- Improve error handling
    - Added a check for directory creation failure with an appropriate exception. 
- Enhancements in unit tests (HTTPRemoteRepositoryTest):
    - Added test cases to verify retrieve handling of "304 Not Modified" responses.
    - Modified existing test cases to improve structuring and maintain readability.

### DIFF
--- a/src/org/omegat/core/team2/impl/HTTPRemoteRepository.java
+++ b/src/org/omegat/core/team2/impl/HTTPRemoteRepository.java
@@ -198,10 +198,8 @@ public class HTTPRemoteRepository implements IRemoteRepository2 {
 
         String currentEtag = etags.getProperty(fileName);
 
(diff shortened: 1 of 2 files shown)


miurahr:
The fix here is mandatory for passing weekly release CI/CD.

PR #1451, opened 2025-06-10
Match origin string improvement

brandelune:
In the UI context, the strings don’t make much sense. This is an improvement proposal.

Screen Shot 2025-06-10 at 18 09 06


brandelune:
Ok. I seem to be getting there.

PR #1455, opened 2025-06-12
fix(spellchecker):  avoid NPE when exit/restart OmegaT

miurahr:
## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?
- NPE at program exit 
- https://sourceforge.net/p/omegat/bugs/1294/
## What does this PR change?

- Project load check before writing wordlists
- Remove write wordlists from project close event and write when project save 
-


brandelune:
No NPE anymore. Thank you.

PR #1459, opened 2025-06-13
Match origin string improvement

brandelune:
I reworded the strings and added a space for readability.


brandelune:
The original state was: 

Screen Shot 2025-06-10 at 18 09 06

Now it is like this:

Screen Shot 2025-06-10 at 18 26 25

There is a space at the beginning of the string so that the % are not stuck to it.

brandelune:
I’d like to have both your opinion on the content of the string.

I’m thinking using parenthesis could work too: `(from reference TM)` for ex.

brandelune:
> I would use either:
> 
> 1. `<90/90/90% Reference TM: ` (with a colon), or
> 2. <90/90/90% /path/to/tmx (reference TM)` if I wanted to use parentheses.
> 
> My first choice would still be `Reference TM: ` for both readability and simplicity.

The path that follows in the screen is not compulsory and the location of the template is decided by the user (here, we only have the default situation), so a colon would be weird.

So the string must be meaningful by itself. If you think that’s still the case with removing the "from" for the items, I’m fine with that.

Regarding the parenthesis, if we keep in mind that the template item can move around, I think isolating the string is a better way to not mix it with its arbitrary surrounding strings.

What do you think about that?

miurahr:
I have added an acceptance test to check the fuzzy match pane UI with external TM as #1461 
If you plan to update the template rather than the message, please update it also for the acceptance test.

brandelune:
> I have added an acceptance test to check the fuzzy match pane UI with external TM as #1461 If you plan to update the template rather than the message, please update it also for the acceptance test.

In fact, I’m wondering if, generally speaking, the spaces should be inside the template rather than automatically added when creating the output. What do you think about that? What is the best practice ?

miurahr:
> > I have added an acceptance test to check the fuzzy match pane UI with external TM as #1461 If you plan to update the template rather than the message, please update it also for the acceptance test.
> 
> In fact, I’m wondering if, generally speaking, the spaces should be inside the template rather than automatically added when creating the output. What do you think about that? What is the best practice ?

It is better to add space character especially non-breakable space character to `Bundle.properteis` instead of the code.

src/org/omegat/gui/matches/MatchesVarExpansion.java
```diff
@@ -227,14 +227,13 @@ private String getPropValue(List<TMXProp> props, String type) {
     private String expandMatchSource(String localTemplate, NearString.MATCH_SOURCE comesFrom) {
         switch (comesFrom) {
         case TM:
-            return localTemplate.replace(VAR_MATCH_SOURCE, OStrings.getString("MATCHES_VAR_EXPANSION_MATCH_COMES_FROM_TM") + " ");
+            return localTemplate.replace(VAR_MATCH_SOURCE, " " + OStrings.getString("MATCHES_VAR_EXPANSION_MATCH_COMES_FROM_TM") + " ");
```
miurahr, line 230:
```suggestion
            return localTemplate.replace(VAR_MATCH_SOURCE, OStrings.getString("MATCHES_VAR_EXPANSION_MATCH_COMES_FROM_TM"));
```

src/org/omegat/gui/matches/MatchesVarExpansion.java
```diff
@@ -227,14 +227,13 @@ private String getPropValue(List<TMXProp> props, String type) {
...
         case FILES:
-            return localTemplate.replace(VAR_MATCH_SOURCE, OStrings.getString("MATCHES_VAR_EXPANSION_MATCH_COMES_FROM_FILES"));
+            return localTemplate.replace(VAR_MATCH_SOURCE, " " + OStrings.getString("MATCHES_VAR_EXPANSION_MATCH_COMES_FROM_FILES"));
```
miurahr, line 232:
```suggestion
            return localTemplate.replace(VAR_MATCH_SOURCE, OStrings.getString("MATCHES_VAR_EXPANSION_MATCH_COMES_FROM_FILES"));
```

src/org/omegat/gui/matches/MatchesVarExpansion.java
```diff
@@ -227,14 +227,13 @@ private String getPropValue(List<TMXProp> props, String type) {
...
         case MEMORY:
-            return localTemplate.replace(VAR_MATCH_SOURCE, OStrings.getString("MATCHES_VAR_EXPANSION_MATCH_COMES_FROM_MEMORY"));
+            return localTemplate.replace(VAR_MATCH_SOURCE, " " + OStrings.getString("MATCHES_VAR_EXPANSION_MATCH_COMES_FROM_MEMORY"));
```
miurahr, line 234:
```suggestion
            return localTemplate.replace(VAR_MATCH_SOURCE, OStrings.getString("MATCHES_VAR_EXPANSION_MATCH_COMES_FROM_MEMORY"));
```

src/org/omegat/gui/matches/MatchesVarExpansion.java
```diff
@@ -227,14 +227,13 @@ private String getPropValue(List<TMXProp> props, String type) {
...
         case SUBSEGMENTS:
-            return localTemplate.replace(VAR_MATCH_SOURCE, OStrings.getString(
-                    "MATCHES_VAR_EXPANSION_MATCH_COMES_FROM_SUBSEGMENTS") + " ");
+            return localTemplate.replace(VAR_MATCH_SOURCE, " " + OStrings.getString("MATCHES_VAR_EXPANSION_MATCH_COMES_FROM_SUBSEGMENTS") + " ");
```
miurahr, line 236:
```suggestion
            return localTemplate.replace(VAR_MATCH_SOURCE, OStrings.getString("MATCHES_VAR_EXPANSION_MATCH_COMES_FROM_SUBSEGMENTS"));
```

src/org/omegat/Bundle.properties
```diff
@@ -2967,7 +2967,7 @@ DICTIONARY_LOAD_ERROR=Error load dictionary from '{0}': {1}
...
 EDITOR_CONTROLLER_EXCEPTION=bad location exception when changing case
 
-MATCHES_VAR_EXPANSION_MATCH_COMES_FROM_TM=External TM
-MATCHES_VAR_EXPANSION_MATCH_COMES_FROM_FILES=Source Files
-MATCHES_VAR_EXPANSION_MATCH_COMES_FROM_MEMORY=Project
-MATCHES_VAR_EXPANSION_MATCH_COMES_FROM_SUBSEGMENTS=Sub-segmented match
+MATCHES_VAR_EXPANSION_MATCH_COMES_FROM_TM=From reference TM
+MATCHES_VAR_EXPANSION_MATCH_COMES_FROM_FILES=From source files
+MATCHES_VAR_EXPANSION_MATCH_COMES_FROM_MEMORY=Internal match
+MATCHES_VAR_EXPANSION_MATCH_COMES_FROM_SUBSEGMENTS=From merged segments
```
miurahr, line 2973:
```suggestion
MATCHES_VAR_EXPANSION_MATCH_COMES_FROM_SUBSEGMENTS=\u00A0From merged segments\u00A0
```

src/org/omegat/Bundle.properties
```diff
@@ -2967,7 +2967,7 @@ DICTIONARY_LOAD_ERROR=Error load dictionary from '{0}': {1}
 DICTIONARY_MANAGER_ERROR_SAVE_IGNORE=Error saving ignore words"
 EDITOR_CONTROLLER_EXCEPTION=bad location exception when changing case
 
-MATCHES_VAR_EXPANSION_MATCH_COMES_FROM_TM=External TM
-MATCHES_VAR_EXPANSION_MATCH_COMES_FROM_FILES=Source Files
-MATCHES_VAR_EXPANSION_MATCH_COMES_FROM_MEMORY=Project
-MATCHES_VAR_EXPANSION_MATCH_COMES_FROM_SUBSEGMENTS=Sub-segmented match
+MATCHES_VAR_EXPANSION_MATCH_COMES_FROM_TM=From reference TM
```
miurahr, line 2970:
```suggestion
MATCHES_VAR_EXPANSION_MATCH_COMES_FROM_TM=\u00A0Reference TM\u00A0
```

miurahr (commented):
This is an example how to insert NBSP.

brandelune:
> I'd forgotten this was part of the template. In that case, you're right that the colon might be odd if the user changes the template. Wording wise, I think we can dispense with the "from".
> 
> * Reference TM
> * Source file
> * Internal match
> * Merged segments
> 
> should all be clear enough.

Ok. I’m reverting. Thank you.
> 
> For the space, I agree the bundle is probably not the best choice.
> 
> This is strictly a matter of opinion, but:
> 
> * if we want to account for UI localization in languages that don't use spaces, or if we think it should be left to user choice, then the template would be the best place for them,

In fact we have the solution. The default template is:
```
${id}. ${fuzzyFlag}${sourceText}
${targetText}
<${score}/${noStemScore}/${adjustedScore}%${matchSource}${filePath}>
```
There is an obvious lack of spaces here. Our default should be:
```
${id}. ${fuzzyFlag}${sourceText}
${targetText}
<${score}/${noStemScore}/${adjustedScore}% ${matchSource} ${filePath}>
```
Et voilà ! No need to add spaces in the bundle or in the expansion code.

@miurahr what do you think of that?

> Random thought: maybe we should consider a different (visually stronger?) separator? One that comes to mind would be " | " (a pipe with a space on either side).

That’s something that can be added manually if necessary:
```
${id}. ${fuzzyFlag}${sourceText}
${targetText}
<${score}/${noStemScore}/${adjustedScore}% | ${matchSource} | ${filePath}>
```

brandelune:
With the latest modifications I have:

Screen Shot 2025-06-14 at 17 44 36

Screen Shot 2025-06-14 at 17 44 22

So, we’re good. @Kazephil What about the strings? Are they OK with you?

@miurahr are you OK with the code modifications?

Kazephil (approved):
The strings look fine to me.

PR #1472, opened 2025-06-18
fix(filters4/xliff1):  fix the anti-pattern that catch the null pointer exception and refactor complex switch-case blocks

miurahr:
## Pull request type
- Refactor

## What does this PR change?

- Added helper methods:
    - getRequiredAttribute(): Centralizes and simplifies attribute retrieval, with error handling for mandatory attributes, and replace try-catch-null-exception anti-pattern.
    - handleMrkElement(), handleDefaultElement(), and others: Modularized the processing of XML elements for better readability.
    - Dedicated methods for handling specific end elements (handleFileEndElement, handleGroupEndElement, etc.).
- Improved code clarity:
    - Removed inline verbose switch case logic by delegating to helper methods.
    - Replaced repetitive patterns with reusable functions.
- Minor code cleanup:
    - Removed outdated comments and cleaned unused code.
     - Simplified generics in target initialization (new LinkedList<>()).
- Errors centralized: Introduced better exception handling for mandatory attributes like id and original in XML elements.

### DIFF
--- a/src/org/omegat/filters4/xml/xliff/Xliff1Filter.java
+++ b/src/org/omegat/filters4/xml/xliff/Xliff1Filter.java
@@ -78,6 +78,8 @@ public class Xliff1Filter extends AbstractXliffFilter {
     private static final String SOURCE_ELEMENT = "source";
     private static final String TARGET_ELEMENT = "target";
     private static final String NOTE_ELEMENT = "note";
+    private static final String TRANS_UNIT_ELEMENT = "trans-unit";
+    private static final String ID_ATTRIBUTE = "id";
     private static final String STATE_ATTRIBUTE = "state";
     private static final String TRANSLATED_STATE = "translated";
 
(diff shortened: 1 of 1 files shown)


miurahr:
There is no meaningful Javadoc or comment that justifies anti-patterns.

miurahr:
There is no comment here.
image

Did you save your comment?

t-cordonnier:
Screenshot_20250701_065302

miurahr:
> The most important one: let's not forget that Xliff1Filter is not only a concrete class for standard XLIFF, it is also overridden by some subclasses for editor-specific variants of XLIFF 1. Actually in the code this concerns only SDLXLIFF but I have also an implementation of MQXliff not yet published because of some known bugs. You should at least check for compatibility with SDLXLIFF.

The proposal does not change anything `protected` method signatures and functions. A purpose of the change is to avoid anti-patterns, catching NPE, and fall-through of the switch-case.

PR #1482, opened 2025-06-20
refactor(core): unify `OSXEventListener` and `IOpenFilesHandler` into single class

miurahr:
There is an anti-pattern code in the class

## Pull request type

- bugs
- refactor
- 
## Which ticket is resolved?

There is no issue to point

## What does this PR change?
- Combined `APP_LISTENER` and `OPEN_FILES_HANDLER` into a new `OSXEventListener` class to simplify logic and reduce redundancy.
- Updated registration logic in `init` to use the new unified event listener.
- Avoid updating static field from instance


miurahr:
This chunk of code is for mac.
It passes CI on mac but I am wondering the current coverage is enough for the OSXIntegration class.

PR #1488, opened 2025-06-20
refactor(core): improve error handling and simplify directory and path logic in `RemoteRepositoryProvider`

miurahr:
## Pull request type

refactor
## Which ticket is resolved?
there is no issue to point
## What does this PR change?

-
-
-

## Other information
#1296


miurahr:
FYI, the current OmegaT creates the prepare subdirectory, and when it fails, OmegaT ignores the status.

`public static final String REPO_PREPARE_SUBDIR = ".repositories/prep/";`

The patch changes the behavior. When OmegaT detects the failure of the creation of the folder, it records the log message.

Kazephil:
OK. So the message is about stuff in the `.repositories' set of subdirectories.

Let me think about it a little more.

Is the message shown to users, or only stored in the log for developers?

miurahr:
Only in log file.

Kazephil (approved):
OK. I just pushed a proposed rewording.

PR #1525, opened 2025-06-30
chore(ci): Improve CI(azure and github actions) execution detection logic

miurahr:
Currently, azure-pipelines are not running when document only change.
This PR improve the logic.

## Pull request type

## What does this PR change?

- Detect document change and run document build task
- Detect non-document change and run test
- When only CI configuration changes, all task skipped.


miurahr:
I think we should also update GitHub Actions configuration to skip manual build when no document change.

PR #1531, opened 2025-07-02
feat(gradle): implement dependency caching mechanism in integration tests

miurahr:
Add gradle cache for integration test container

## Pull request type

- Other (describe below)
- Improve test
## What does this PR change?

- Added a new volume, gradle-cache, in compose.yml for Gradle dependency caching.
- Updated the entrypoint.sh script to handle Gradle dependency caching:
    - Set up a GRADLE_USER_HOME environment variable pointing to the cache.
    - Computed a hash based on build configuration files to determine cache validity.
    - Implemented logic to download dependencies if the cache is outdated or missing.
    - Cleaned up old cache markers keeping the most recent three for flexibility.
    - Displayed cache statistics like size and configuration count.

### DIFF
--- a/compose.yml
+++ b/compose.yml
@@ -27,6 +27,9 @@ services:
       - type: volume
         source: keygen
         target: /keys
+      - type: volume
+        source: gradle-cache
+        target: /gradle-cache
     tmpfs:
       - /tmp
       - /workdir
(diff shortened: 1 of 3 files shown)


miurahr:
By enabling cache, it will consume about 800MB for caching libraries.

miurahr:
Here is a maintenance command to clear the cache.

```bash
# Clear cache (if needed)
docker compose down
docker volume rm $(docker compose config --volumes | grep gradle-cache)
```

Please use `nerdcgtl` if you use containerd.

PR #1570, opened 2025-07-19
refactor(core): add `TMXLSResourceResolver` and improve TMX validation in `TMXReader2`

miurahr:
## Pull request type

## What does this PR change?

- add a validator TMXReader2 based on XSD


miurahr:
This should be failed because many TMX test data are invalid as TMX standard for both 1.1 and 1.4.
PR #1569 is the fix of the test data.

@t-cordonnier could you check the change for TMXReader2.java that implement validator with dynamic TMX version detection.

PR #1577, opened 2025-07-23
Project tmx level2

t-cordonnier:
Use TMX 1.4 level 2 for project-save.tmx

concerning team work, I implemented also the algorithm which I had suggested a few days ago:
- when you create a new project, or as long as there is not yet a project-save.tmx, the first one is saved as 1.4 Level 2 with correct tags;
- when you open an existing project, if it contains reference to DTD 1.1,
--- you load it using the old format, taking tags as they are in OmegaT < 6.1
--- when you save, you continue with the old format if you are in a team project, to avoid conflicts in Git
--- if the project is not a team one, you definitively switch to new format
- if you open a project which is already 1.4 Level 2, you continue with it
Note that I migrated also to 1.4, because this time I use the DTD version number to check whenever I am reading an old-style project (in which case I will continue with this format for better compatibility) or not.
Older versions of OmegaT will correctly read the 1.4 file, they will only continue to write it in old format but I reduce the potential impact on team working by using the just described algorithm: logically after 2-3 iterations everybody will use 1.1 in this case.

## Pull request type

feature

## Which ticket is resolved?

Discussion https://sourceforge.net/p/omegat/mailman/message/59208059/ (will be completed later)


src/org/omegat/util/TMXReader2.java
```diff
@@ -158,6 +159,10 @@ public void readTMX(File file, final Language sourceLanguage, final Language tar
             while (xml.hasNext()) {
                 XMLEvent e = xml.nextEvent();
                 switch (e.getEventType()) {
+                case XMLEvent.DTD:
+                    DTD ed = (DTD) e;
+                    callback.setOldOmegaTFormat(ed.getDocumentTypeDeclaration().contains("11") || ed.getDocumentTypeDeclaration().contains("1.1"));
+                    break;
                 case XMLEvent.START_ELEMENT:
                     StartElement eStart = (StartElement) e;
                     if ("tu".equals(eStart.getName().getLocalPart())) {
```
miurahr, line 168: Can we add a callback when we get the start element of "tmx"?

src/org/omegat/core/team2/TeamTool.java
```diff
@@ -106,7 +106,7 @@ public static void initTeamProject(File dir, String srcLang, String trgLang) thr
         // Create empty project TM
         ProjectTMX projectTMX = new ProjectTMX();
         projectTMX.load(props.getSourceLanguage(), props.getTargetLanguage(), props.isSentenceSegmentingEnabled(), null, Core.getSegmenter());
-        projectTMX.save(props, new File(props.getProjectInternal(), OConsts.TMX_EXTENSION).getPath(), false);
+        projectTMX.save(props, new File(props.getProjectInternal(), OConsts.TMX_EXTENSION).getPath(), false, true);
```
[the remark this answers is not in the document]
t-cordonnier: Previous line is even longer and you did not complain...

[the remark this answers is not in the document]
miurahr: You have a chance to improve code, and I am not complaining but I suggest for you to notify the chance.

miurahr (approved):

src/org/omegat/core/data/RealProject.java
```diff
@@ -834,9 +834,10 @@ public synchronized void saveProject(boolean doTeamSync) {
                     saveProjectProperties();
                     synchronized (projectTMX) {
                         projectTMX.save(config, config.getProjectInternal() + OConsts.STATUS_EXTENSION,
-                                isProjectModified());
+                                isProjectModified(), remoteRepositoryProvider != null);
                     }
-                    if (remoteRepositoryProvider.isManaged() && doTeamSync) {
+
+                    if (remoteRepositoryProvider != null && doTeamSync) {
```
miurahr: `remoteRepositoryProvider` has never been null, so it is always true. It changes the correct `remoteRepositoryProvider.isManaged()` to be degraded.

src/org/omegat/util/TMXReader2.java
```diff
@@ -41,11 +41,7 @@
 import javax.xml.stream.XMLInputFactory;
 import javax.xml.stream.XMLResolver;
 import javax.xml.stream.XMLStreamException;
-import javax.xml.stream.events.Attribute;
-import javax.xml.stream.events.Characters;
-import javax.xml.stream.events.EndElement;
-import javax.xml.stream.events.StartElement;
-import javax.xml.stream.events.XMLEvent;
+import javax.xml.stream.events.*;
```
miurahr: it should be individual imports.

src/org/omegat/core/data/ProjectTMX.java
```diff
@@ -142,8 +147,8 @@ public boolean isEmpty() {
     /**
      * It saves current translation into file.
      */
-    public void save(ProjectProperties props, String translationFile, boolean translationUpdatedByUser)
-            throws Exception {
+    public void save(ProjectProperties props, String translationFile, boolean translationUpdatedByUser,
```
miurahr, line 157: We should have the `save` with the old signature with `@deprecated`

miurahr (changes requested):

miurahr (approved):

PR #1597, opened 2025-08-09
[6.0][BUGS#1302] XML filters does not handle DTD path with CJK characters properly 

miurahr:
OmegaT 6.0.x does not handle path with CJK character properly.

## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?

- Unable to import a team project to a path containing Chinese characters(
- https://sourceforge.net/p/omegat/bugs/1302/

## What does this PR change?

- Add reproducible as functional test
- `XMLFilter#processFile` to set system ID with URL encoded string.

### DIFF
--- a/release/changes.txt
+++ b/release/changes.txt
@@ -3,7 +3,7 @@
 ----------------------------------------------------------------------
 
   2 Enhancement
-  7 Bug fixes
+  8 Bug fixes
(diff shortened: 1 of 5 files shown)


miurahr:
The test should be failed because of reproducible

PR #1600, opened 2025-08-10
fix(LT): always load LT English module

miurahr:
When loading project without English, OmegaT throws exception in log file.

## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?

-  LanguageTool failed to load en-US when project is other than English
- https://sourceforge.net/p/omegat/bugs/1303/

## What does this PR change?

- Change `LanguageToolNativeBridge` always load English LT module


brandelune (approved):

brandelune:
The build works and the error is not generated anymore. Thank you.

PR #1601, opened 2025-08-10
[6.0][BUGS#1276] fix: OmegaT failed to start when malformed jar in plugin folder

miurahr:
Backport of the security change from master branch.
This Pull Request enhances the plugin management system by introducing better handling for plugin version conflicts. It adds logic to exclude outdated or similar versions of plugins, modifies string formatting for plugin-related logs, and updates how class names are processed in one utility.
## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?

- OmegaT failed to start if there are 'jar' file w/o manifest in plugin folder
- https://sourceforge.net/p/omegat/bugs/1276/
- 
## What does this PR change?

- Backport PR #1196
- Update `Bundle.properties` for new messages
- Backport a fix for PluginInformation class.
    -  Safeguarded against potential null values when processing class names in PluginInformation.
- Backport a fix of `PluginUtils#loadPlugins`
    -  Added logic to detect and exclude old or duplicate versions of OmegaT plugins by comparing versions.
    -  Introduced new log messages to notify users about excluded plugins and their reasons.
    -  Included utility classes (VersionChecker, JarInputStream) for version comparison and handling plugin manifests.
    -  Added constants and string enhancements in PluginUtils for better comprehension and reduced error risk.

### DIFF
--- a/src/org/omegat/Bundle.properties
+++ b/src/org/omegat/Bundle.properties
@@ -1230,6 +1230,9 @@ PLUGIN_JAVA_VERSION_ERROR=Error loading : {0}
 PLUGIN_LOAD_JAR=Jar used for plugins: {0}
 PLUGIN_LOAD_OK=Plugin successfully loaded from class '{0}'
 PLUGIN_SKIP_PREVIOUSLY_LOADED=Skipping previously loaded plugin from class '{0}'
+PLUGIN_EXCLUDE_OLD_VERSION=Excluding old version of '{0}' ({1}), keeping {2}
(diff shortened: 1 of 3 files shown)


miurahr:
Does this really happened in verison 6.0? 
A BUGS#1276 stack trace tell me that OmegaT 6.1.0 has `PluginUtils#populatePluginUrlList` 

```
org.omegat.filters2.master.PluginUtils.populatePluginUrlList(PluginUtils.java:331)
```

There is a place to detect previous versions.
OmegaT don't have a feature, then the patch has a previous version detection logic.

I think we don't need to port it version 6.0.

I want to close here as invalid

PR #1603, opened 2025-08-11
fix(gui/main): reset status bar when reload/close project

miurahr:
OmegaT do nothing for the status bar when project closed in previous versions.
Now we reset status bar when project close/reload.
If reloading empty project, status bar remains previous state but now it is fixed to reset.

## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?

- Quick stats don't reset if there are no more source files
- https://sourceforge.net/p/omegat/bugs/491/

## What does this PR change?

- Refactor MainWindow and MainWindowProgressBar classes to have MainWindowProgressBarController for modular and clarifications.
- Add project event listener to catch close event and reset status bar


brandelune (approved):
Thank you. The reset works now when the project is emptied.

PR #1604, opened 2025-08-11
fix(core): improve project file validation and error messages

miurahr:
Improve project loader to validate `omegat.project` strictly.

## Pull request type

- Bug fix -> [bug]
- Enhancement

## Which ticket is resolved?

- Lack of language codes does not trigger error 
- https://sourceforge.net/p/omegat/bugs/465/

## What does this PR change?

- `o.o.u.ProjectFileStorage#parseProjectFile` to check mandatory property should not be null or empty

### DIFF
--- a/src/org/omegat/Bundle.properties
+++ b/src/org/omegat/Bundle.properties
@@ -1168,6 +1168,13 @@ PM_ERROR_BACKING_UP_PREFS_FILE=Failed to back up preferences file.
 
 # {0} - the version, which we don't support
 PFR_ERROR_UNSUPPORTED_PROJECT_VERSION=Unsupported project file version ({0})!
+PFR_ERROR_MISSING_PROJECT_TAG=omegat.project don't have a project tag
+PFR_ERROR_MISSING_PROJECT_VERSION_TAG=omegat.project don't have version tag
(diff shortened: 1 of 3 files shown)


brandelune:
The current master replaces the language codes with "und" (presumably for "undefined"). The TMX is thus created with the two xml:lang set to "und".

Since OmegaT generally does not enforce the validity of the xml:lang value, the situation that we currently have is an evolution compared to the bug report and is not really a bug.

brandelune:
Your fix now generates two error messages that are not helpful:

```
"Unable to read project file! 

The specified folder is not an OmegaT project, or the project file is broken, or there is no valid backup.
```

and

```
Unable to read project file! Trying with backup file

omegat.project has an empty version tag
```

The pre-fix behavior is better : the project opens, OmegaT adds "und" language codes.

What we need to have is a warning that the codes are the same. Because *that* is what will be a problem later.

I don’t know when OmegaT checks for language codes but we should have : 

- For the lack of language codes:
```
Source and target language codes were not defined. OmegaT will replace the codes with "undefined". Please change the codes in Project Properties.
```

Also, we should change "und" to "undefined".

- When two language codes are identical:
```
Source and target language codes are identical. OmegaT will not be able to extract information from your project memory. Please change the codes in Project Properties.
```

PR #1610, opened 2025-08-18
feature(tokenizer): add full stemming support for glossary and fuzzy match and implemented in Latin languages

miurahr:
## Pull request type

## Which ticket is resolved?
- Tokenizer to support the snowball generated stemming mode 
-  https://sourceforge.net/p/omegat/feature-requests/1785/

## What does this PR change?

- Introduced `GLOSSARY_FULL` and `MATCHING_FULL` stemming modes to enhance snowball stemming functionality.
- Updated Italian tokenizer to support new stemming modes.
- Refactored tokenization logic to accommodate additional stemming configurations.
- Added new preferences for enabling/disabling full stemming (`TOKENIZER_STEMMING_FULL`).
- Improved glossary searcher and statistics matching to utilize extended stemming modes.
- Added tests for Italian full stemming and glossary functionality to ensure correctness.

### DIFF
--- a/src_docs/manual/en/OmegaT_Preferences.xml
+++ b/src_docs/manual/en/OmegaT_Preferences.xml
@@ -42,7 +42,7 @@
           <keycap>Tab</keycap> instead.</para>
 
 		  <para>This option is useful with some Chinese,
-          Japanese or Korean character input systems.</para>
+          Japanese, or Korean character input systems.</para>
         </listitem>
       </varlistentry>
 
@@ -284,7 +284,7 @@
        <varlistentry id="dialogs.preferences.glossary.use.full.stemming">
            <term id="dialogs.preferences.glossary.use.full.stemming.title"><option>Use full stemming</option></term>
            <listitem>
-               <para>Some tokenizers offer finer word stemming for a limited number of languages. Currently, Italian and French are implemented.</para>
+               <para>Some tokenizers offer finer word stemming for a limited number of languages. Currently, English, French, and Italian are implemented.</para>
            </listitem>
        </varlistentry> 
 	</variablelist>
--- a/src/org/omegat/tokenizer/LuceneEnglishTokenizer.java
+++ b/src/org/omegat/tokenizer/LuceneEnglishTokenizer.java
@@ -30,7 +30,9 @@ import java.io.StringReader;
 import org.apache.lucene.analysis.CharArraySet;
 import org.apache.lucene.analysis.TokenStream;
 import org.apache.lucene.analysis.en.EnglishAnalyzer;
+import org.apache.lucene.analysis.snowball.SnowballFilter;
 import org.omegat.core.Core;
+import org.tartarus.snowball.ext.EnglishStemmer;
 
 /**
  * The LuceneEnglishTokenizer class provides a tokenization implementation for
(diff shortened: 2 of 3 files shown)


src/org/omegat/gui/preferences/view/GlossaryPreferencesController.java
```diff
@@ -67,6 +68,7 @@ protected String getDisplayText(IGlossaryRenderer value) {
                 return value.getName();
             }
         });
+        Mnemonics.setLocalizedText(panel.useFullStemmingCheckBox, OStrings.getString( "PREFS_GLOSSARY_STEMMING_FULL"));
```
miurahr, line 71:
```suggestion
        Mnemonics.setLocalizedText(panel.useFullStemmingCheckBox, OStrings.getString("PREFS_GLOSSARY_STEMMING_FULL"));
```

src/org/omegat/Bundle.properties
```diff
@@ -2751,6 +2751,7 @@ PREFS_TITLE_GLOSSARY=Glossaries
 PREFS_GLOSSARY_TBX_DISPLAY_CONTEXT=Display TBX &glossaries context description
 PREFS_GLOSSARY_EXACT_MATCH=Match term groups even if the terms appear separately
 PREFS_GLOSSARY_STEMMING=Use &stemming
+PREFS_GLOSSARY_STEMMING_FULL=Use &full stemming
```
brandelune, line 2754: I think we’ll need a good explanation of what "full stemming" is vs "normal" stemming, and why/in which cases is the difference relevant.

If "full stemming" is always better than "normal" stemming, we need to think about a better way to propose the choice.

miurahr: I don't know "normal" stemming. It is a fact that the Italian language module of LanguageTool is called a "light" stemming algorithm. The "full stemming" is always a heavier processing load than default of LanguageTool.

From my short research

> Key Difference
> 
> Snowball Italian stemmer (full): aggressive, designed to reduce words to morphological stems as much as possible.
> 
> Lucene default Italian stemmer (Light Italian Stemmer): less aggressive, optimized for information retrieval where precision matters.
> 

> Why the Snowball Stemmer is Called "Full"
> 
> When people contrast “Snowball stemmer” with “Lucene Light stemmer”, they often shorthand it as “full vs. light stemming”. This isn’t an official name from Martin Porter, but rather a convention in information retrieval and NLP communities.
> 
> Snowball (Full Stemmer):
> 
> Implements the complete set of linguistic rules for a given language.
> 
> Aggressively reduces words down to a core stem (sometimes very short).
> 
> Designed with the intention of normalizing most inflected forms.
> 
> Example (Italian):
> 
> amavano → am
> 
> ragazze → ragazz
> 
> Light Stemmer (Lucene Light):
> 
> Called light because it applies only minimal suffix-stripping.
> 
> Keeps more of the word intact to avoid conflating unrelated terms.
> 
> Prioritizes search precision over maximum normalization.
> 
> Example (Italian):
> 
> amavano → amav
> 
> ragazze → ragazz (same as Snowball here, but not always)

miurahr:
> Considering that this option is limited to Italian but that we do not want to have information specific to Italian users, I think it would be beneficial to have that option enabled by default.
> 
> Are there drawbacks in terms of performance if we do so?

There is a feature request for Italian, then I update for Italian. If there are other languages request and language developer implement it, it can be. As you know, snowball stemmer supports  15-17  languages.
The core European languages covered include:

- Germanic languages: English, German, Dutch, Swedish,  Norwegian, Danish
- Romance languages: French, Spanish, Portuguese, Italian, Romanian
- Other Indo-European: Russian, Hungarian
- Semitic: Arabic
- Finno-Ugric: Finnish
- Celtic: Irish (contributed)
- Slavic: Czech (contributed), Russian

It is why Italian is an example, so use "such as".

There are many drawbacks for full stemmer, and I am not native speaker of these languages, I cannot explain it. The performance drawback is obvious.
It is a reason why LanguageTool takes "light" stemmer in default not Snowball stemmer.

src_docs/manual/en/OmegaT_Preferences.xml
```diff
@@ -280,6 +280,13 @@
 		  <para>If a glossary item has multiple definitions, sort by a length of definition text.</para>
 		</listitem>
 	  </varlistentry>
+        
+       <varlistentry id="dialogs.preferences.glossary.use.full.stemming">
+           <term id="dialogs.preferences.glossary.use.full.stemming.title"><option>Use full stemming</option></term>
+           <listitem>
+               <para>Some tokenizers offer finer word stemming for a limited number of languages. Currently, only Italian is supported.</para>
```
brandelune, line 287: ... is implemented.

src/org/omegat/Bundle.properties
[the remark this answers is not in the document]
brandelune: Ok, I understand now.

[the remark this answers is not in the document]
brandelune: So what keeps us from supporting the other possible languages now ?

src_docs/manual/en/OmegaT_Preferences.xml
miurahr:
```suggestion
               <para>Some tokenizers offer finer word stemming for a limited number of languages. Currently, only Italian is implemented.</para>
```

brandelune (approved):

PR #1618, opened 2025-08-21
fix(pofilter): support multi-line msgctxt 

miurahr:
This PR is intended to reproduce BUGS#1305 and set proper target behavior for the multi-line continuation of the PO file format.

## Pull request type

- Bug fix -> [bug]
- Refactor

## Which ticket is resolved?
- .PO filter bug: display issue when a msgctxt spans several lines
- https://sourceforge.net/p/omegat/bugs/1305/
## What does this PR change?

- refactor `POoFilter#processPoFile` private method to split multiple internal methods with the single purpose method principle.
- Add test data and reproducible of the BUGS#1305
- The test case also includes a check for multiline msgid
- Add behavior check for '\n' in multiline messages

## Other information

- The second commit should be failed.
- https://www.gnu.org/software/gettext/manual/html_node/More-Details.html

### DIFF
--- a/src/org/omegat/filters2/po/PoFilter.java
+++ b/src/org/omegat/filters2/po/PoFilter.java
@@ -696,7 +696,7 @@ public class PoFilter extends AbstractFilter {
         }
         if (references.length() > 0) {
             sb.append(OStrings.getString("POFILTER_REFERENCES")).append("\n").append(unescape(references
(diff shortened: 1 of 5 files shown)


miurahr:
Let's merge it and ask the user community for further feedbacks.

PR #1630, opened 2025-08-28
chore(build): migrate to real `OmegatModulePlugin` from conventions plugin

miurahr:
## Pull request type

## What does this PR change?

-
-
-


miurahr:
It is very strange that the CI is failed only Hunspell on Windows for Arabic.
There is no change in the code itself. There is no meaningful log message from test log on Azure Pipeline CI plataform.

Because Hunspell library only support Intel 64bit Windows only, I need to prepare Intel Windows test environment to reproduce it, it is now impossible investigate it.

miurahr:
#1633 is a working branch and PR that try to test with Lucene Hunspell checker for AR on Windows. If #1633 passed, the issue here will be a bug in Hunspell Library.

miurahr:
It also failed on Windows for Arabic. It is the issue in the #1630 change or AR language module.

miurahr:
I give up supporting the Arabic dictionary for hunspell/morfologik. Other languages are working fine. It is too weird.

PR #1668, opened 2025-09-11
fix(glossary): skip whitespace tokens in `GlossarySearcher` and extend unit test cases

miurahr:
`GlossarySearcher#tokenize` produce a list of tokens with space characters when user configure glossary without stemming.

## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?

- Space character considered as individual term
- https://sourceforge.net/p/omegat/bugs/1198/

## What does this PR change?

- Skip when token string is a white space in `GlossarySearcher#tokenize`.
- refactor for code styling
- Add regression tests
- Also add a test for `BaseTokenizer#tokenizeVerbatim` to clarify behavior.


miurahr:
I have reproduced the bug as a unit test and solved the issue. I want to merge it.

PR #1691, opened 2025-09-20
feature(scripting): introduce `ScriptingModule` and modularize scripting functionality

miurahr:
## Pull request type

Please mark github LABEL of the type of change your PR introduces:

- Bug fix -> [bug]
- Feature enhancement -> [enhancement]
- Documentation -> [documentation]
- Build and release changes -> [build/release]
- Other (describe below)

## Which ticket is resolved?

- Title ex. fix...
  * URL ex. https://...
 
- Title
  * URL

## What does this PR change?

-
-
-

PR #1703, opened 2025-09-23
refactor(scripting): introduce scripting function as a module

miurahr:
This pull request introduces a scripting module feature to OmegaT, with various modifications and additions to support scripting functionality, localized language files, and relevant dependencies.
## Pull request type

-refactor

## What does this PR change?

- Project Structure: Introduced a new scripting submodule.
- Code Cleanup: Removed scripts from main to avoid clutter via standalone modularization.
- Dependencies: Updated project dependencies (`build.gradle`) by adding required libraries for scripting.
- Localization: Added script-related entries in various `Bundle` property files for language support, and add `ScriptingUtil#getBundleString` for replacement of the `OStrings.getString`.
- Refactor: Integrated the scripting feature while removing redundant/legacy script references and logic from the existing system (`Main.java`).
- Tests: Adjusted ScriptItemTest.java for new scenarios.

## Other information
 rework of #1691

There are so many files are touched because some L10N messages are moved from core to the module. It means 36 languages x 2 = 72 files are modifed.
It is also the tests are moved from core to the module.


scripting/src/main/java/org/omegat/gui/scripting/ScriptingModule.java
```diff
@@ -0,0 +1,105 @@
...
+     * Private constructor for the class that contains only static methods.
+     */
+    private ScriptingModule() {
+    }
+
+    /**
+     * Execute a script as PROJECT_CHANGE events.
+     */
+    @SuppressWarnings("unused")
+    public static void executeConsoleScript(IProjectEventListener.PROJECT_CHANGE_TYPE eventType,
```
miurahr, line 69: With this static method, we keep the console-mode function. I checked it is working on my local.

src/org/omegat/Main.java
```diff
@@ -697,19 +695,19 @@ private static void executeConsoleScript(IProjectEventListener.PROJECT_CHANGE_TY
...
-                    Log.log(ex);
+                    ClassLoader cl = PluginUtils.getClassLoader(PluginUtils.PluginType.MISCELLANEOUS);
+                    if (cl == null) {
+                        Log.logErrorRB("SCW_SCRIPT_LOAD_ERROR", "the plugin classloader is null");
+                        return;
+                    }
+                    Class<?> scriptingClass = cl.loadClass("org.omegat.gui.scripting.ScriptingModule");
+                    Method method = scriptingClass.getMethod("executeConsoleScript",
+                            IProjectEventListener.PROJECT_CHANGE_TYPE.class, File.class);
+                    method.invoke(null, eventType, script);
```
miurahr, line 707: Based on ADR #1710 we should replace it with new API defined in the ADR

brandelune (approved):

PR #1726, opened 2025-10-12
docs: ADR: Refactoring EntryInfoPane threading model

miurahr:
## Pull request type

## What does this PR change?

- Add discussion document as ADR
    - Migration to a Service-Controller-View (SCV) pattern for all five entry info panes to improve testability, resource management, and separation of concerns.
    - Replacement of unmanaged thread creation with a shared thread pool managed by `ExecutorService`, ensuring bounded thread usage and proper cleanup.
    - Introduction of `EntryInfoService`, `EntryInfoController`, and `EntryInfoView` interfaces to clearly separate business logic, event handling, and UI responsibilities.
    - Comprehensive refactoring of `MachineTranslateTextArea` as the reference implementation for the new architecture.
    - Acceptance test overhaul with deterministic and reliable testing mechanisms for asynchronous code.


miurahr:
> However, as you mention, if there were no discussions on the list about the flaws that you discovered, it is mostly because they were not noticed.

This analysis serves as the basis for my future discussion on the list. Today is a day before I notice.

brandelune:
Ok, thank you for removing the pseudo historical data.

So what you propose is that we use constructs that exist in Java 8 but did not exist in Java 7, along with modern coding practices ?

I understand about technical debt, but how do you estimate the practical benefit to end users ?

miurahr:
This PR contains a draft ADR intended to prepare a structured discussion on the dev list.

Questions such as how to evaluate end-user impact versus technical debt are exactly the points I’d like to discuss on the mailing list.

For this PR, I’d prefer to keep comments focused on whether the ADR clearly frames the problem, context, and decision options, rather than debating the decision itself. I’ll start a dedicated thread on the dev list to discuss the benefits and trade-offs.

PR #1896, opened 2026-02-10
Segments states (RFE [#628])

miurahr:
In 2019, Oct

> There is a very tentative implementation of this feature on the branch topic/briacp/segments-states where the state of a segment can be changed by right-clicking on it. The segment states are currently stored as props (<prop type="x-omegat-state">DRAFT</prop>) in the TMX file.

## Pull request type

## Which ticket is resolved?

- Manage different states for segments
- https://sourceforge.net/p/omegat/feature-requests/628/

## What does this PR change?

-
-
-


src/org/omegat/core/states/SegmentState.java
zs-stpa, line 44: not sure an enum is the way to go. yes, this would make new states. but rigidly so. while there are also these open tickets:
Custom segment states: https://sourceforge.net/p/omegat/feature-requests/628/
Proofreading status: https://sourceforge.net/p/omegat/feature-requests/948/

🤦 that is what RFE stands for and is even linked above … good night, it is late.

zs-stpa: which i understood as the desire for an arbitrary number of generic user-addable segment states, but having reread the tickets, i think they enums would be good enough.

considring that, with the sortability possibly coming up soon, users could just use the notes-field and the apply sorting order on the notes field which allows arbitrary states but not coloring them. which you wouldn't really need that much when you have grouped same note values together already…

miurahr: Thank you for looking it.
This PR is just for archive, I only pick from the forgotten stale branch before clean.

There are two type of RFEs for the translation statuses;   LOCK(proofed) and UNDER-REVIEW.
There is a position that OmegaT is toThanks for taking a look.

To set expectations: this PR is mainly for archival. I picked it up from a forgotten, stale branch before cleaning up old branches, so it is not a proposal to merge as-is.

On the design question, some background on where the project stands:

There are two kinds of RFEs around translation states: a LOCK / proofed state and an UNDER-REVIEW state. Whether to add either one is contested. One position is that OmegaT is a tool for translators, not for project administrators, and that workflow states don't belong in it. Another position is that OmegaT should support XLIFF 2 fully, including the workflow that XLIFF 2 assumes; from that point of view, segment states are important and must be preserved through a round trip.

There is also a practical driver: several web translation platforms export XLIFF and expect a proper state (e.g. translated) on the target when the file is re-imported. OmegaT already handles this in the legacy XLIFF 1 filter with a preference flag that lets the user mark all target outputs as "translated".

The deeper issue is storage. OmegaT uses TMX as its core project format, and TMX has no notion of the states discussed above, which is why the branch stored them as prop entries. There was a long discussion on the dev list about whether to stay with TMX or not, and that question is still open. So whether states end up as an enum, free-form values, or notes-based sorting as you suggest, the real constraint is how they survive in the project files.

PR #1914, opened 2026-03-01
fix(ExternalFinder): remove detection of GUI environment

miurahr:
ExternalFinder does not add menu.
Because it detects GUI in very early stage of OmegaT start, the module skip to add menu item of defined items from finder.xml.

## Pull request type

- Bug fix -> [bug]
- Add test

## Which ticket is resolved?

- External search not working in v 6.1.0
- https://sourceforge.net/p/omegat/bugs/1318/

## What does this PR change?

- Extend ExternalFinderItemMenuGenerator to add name of the menu item for test.
- Add acceptance test case to check menu item additions.
-


miurahr:
Acceptance test is failed as expected because the case reproduces the bug and second commit is to fix the bug.

PR #1936, opened 2026-03-07
fix(core): improve HiDPI support

miurahr:
Oracle and OpenJDK recommend enabled a JVM managed scaling.

## Pull request type

## What does this PR change?

- Improve HiDPI detection by adding `sun.java2d.uiScale.enabled` to be `true`


miurahr:
image

Here is comparison between 6.1.0 (left) and 6.0.2 (right)

HiDPI font scale feature is implemented in 6.1.0.

PR #1978, opened 2026-04-09
Update 2026001.GlossarySearchAndSort.md

brandelune:
"Priority" assignation is only implemented for the writable glossary.

### DIFF
--- a/src_docs/developer/adr/2026001.GlossarySearchAndSort.md
+++ b/src_docs/developer/adr/2026001.GlossarySearchAndSort.md
@@ -59,7 +59,7 @@ Results are sorted in `sortGlossaryEntries` using `compareGlossaryEntries` with 
   - The term length option was added in **OmegaT 6.0.1**.
 
 - **Note (4): Glossary priority implementation**
-  - The glossary priority was implemented in **OmegaT 3.0.5**.
+  - The glossary priority was implemented in **OmegaT 3.0.5**. As of 6.1, **priority** is only implemented for the writable glossary and cannot be assigned to another glossary.
 
 ### 3. Filtering and Deduplication
 The `filterGlossary` method processes the sorted list:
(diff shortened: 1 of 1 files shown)


src_docs/developer/adr/2026001.GlossarySearchAndSort.md
```diff
@@ -59,7 +59,7 @@ Results are sorted in `sortGlossaryEntries` using `compareGlossaryEntries` with
   - The term length option was added in **OmegaT 6.0.1**.
 
 - **Note (4): Glossary priority implementation**
-  - The glossary priority was implemented in **OmegaT 3.0.5**.
+  - The glossary priority was implemented in **OmegaT 3.0.5**. As of 6.1, **priority** is only implemented for the writable glossary and cannot be assigned to another glossary.
```
miurahr, line 62:
```suggestion
  - The glossary priority was implemented in **OmegaT 3.0.5**. As of 6.1, **priority** is only implemented for the writable glossary by default; other glossaries can be loaded as prioritized via a custom plugin, but this is not supported out of the box.
```

miurahr (changes requested):

src_docs/developer/adr/2026001.GlossarySearchAndSort.md
brandelune: Do we have custom plugins that assign priority ?

brandelune: I think a better modification would be `and cannot be assigned to another glossary through the user interface`. That leaves the field open to programmatic solutions. What do you think ?

miurahr: It "can" be when someone create it. It should not be negated.

brandelune: Ok. Just to stick to what exists, what about:

`As of 6.1, **priority** is only implemented for the writable glossary.`

miurahr: Please update the topic branch and push.

brandelune:
Fixes applied.

miurahr (approved):

PR #1998, opened 2026-05-02
[6.0] chore(ci): update JRE download URLs and build tasks in Azure Pipelines

miurahr:
## Pull request type

## What does this PR change?

- Update bundle JRE to up-to-date 11.0.31
- Supress mac-zip distribution from release
-


brandelune (approved):

brandelune:
I think we need to have the mac JREs included in the local dependencies so that I don’t have to manually update them on my side when I build OmegaT.app. We need to be able to build OmegaT.app without any manual download.

PR #2014, opened 2026-05-06
chore(mac): switch appbundler with upstream new version

miurahr:
We have used the forked version of appbundler because we want to use multi-architecture launcher.
Now the upstream project accept our proposal and released a new version.

We should now use the new upstream version 1.4.1 from forked version of 1.2.1.

## Pull request type

## What does this PR change?

- Switch to upstream appbunlder 1.4.1
- Update mac-conventions to use localization from our repository.
- FIx task dependency


brandelune (approved):
Excellent news.

PR #2016, opened 2026-05-08
[6.0] update UI  in pl

miurahr:
Update from L10N/pl that has been updated from IJS project.

## Pull request type
- UI

## What does this PR change?

- Update polish messages


miurahr:
@brandelune 
A contributed localization is incomplete. Is it valuable to merge?

brandelune:
> @brandelune A contributed localization is incomplete. Is it valuable to merge?

Plus it comes with an error in Didier’s name. So maybe we can check the current status with the contributors before merging.

miurahr:
@brandelune there is no update from May in L10N/pl project.
Could you confirm that we merge it in a status of partial translated.

PR #2023, opened 2026-05-18
fix(issues): skip glossary entries with empty target in terminology c…

fung911:
Fixes BUGS#1315: When a glossary entry has an empty target term (e.g. monolingual glossaries or entries with only a comment column), the terminology checker incorrectly raises a warning expecting an empty string in the translation.

Fix: filter out entries where all target terms are empty before running the target match check in TerminologyIssueProvider.

## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?

- Title: Terminology issue raised when there's no target term in glossary
  * URL: https://sourceforge.net/p/omegat/bugs/1315/

## What does this PR change?

- In TerminologyIssueProvider.getIssues(), add a filter step that skips
  glossary entries where all target terms are empty strings.
- Prevents false-positive terminology warnings for entries where the target column is empty but a comment column
  is present (e.g. "source\t\tcomment" format).

### DIFF
--- a/src/org/omegat/gui/issues/TerminologyIssueProvider.java
+++ b/src/org/omegat/gui/issues/TerminologyIssueProvider.java
@@ -78,11 +78,21 @@ class TerminologyIssueProvider implements IIssueProvider {
         if (glossaryEntries.isEmpty()) {
             return Collections.emptyList();
         }
-        return glossaryEntries.stream().map(glossaryEntry -> {
-            List<String> trgTerms = Core.getGlossaryManager().searchTargetMatches(tmxEntry.translation,
-                    sourceEntry.getProtectedParts(), glossaryEntry);
-            return trgTerms.isEmpty() ? new TerminologyIssue(sourceEntry, tmxEntry, glossaryEntry) : null;
-        }).filter(Objects::nonNull).collect(Collectors.toList());
+        return glossaryEntries.stream()
+                .filter(glossaryEntry -> {
+                    // Bug #1315: skip monolingual glossary entries where all target
+                    // terms are empty. An empty target term is not a real term to
+                    // check against, so raising a warning for it is a false positive.
+                    String[] locTerms = glossaryEntry.getLocTerms(false);
+                    return Arrays.stream(locTerms).anyMatch(t -> !t.isEmpty());
+                })
+                .map(glossaryEntry -> {
+                    List<String> trgTerms = Core.getGlossaryManager().searchTargetMatches(tmxEntry.translation,
+                            sourceEntry.getProtectedParts(), glossaryEntry);
+                    return trgTerms.isEmpty() ? new TerminologyIssue(sourceEntry, tmxEntry, glossaryEntry) : null;
+                })
+                .filter(Objects::nonNull)
+                .collect(Collectors.toList());
     }
 
     /**
--- a/src/org/omegat/gui/issues/TerminologyIssueProvider.java
+++ b/src/org/omegat/gui/issues/TerminologyIssueProvider.java
@@ -79,13 +79,7 @@ class TerminologyIssueProvider implements IIssueProvider {
             return Collections.emptyList();
         }
         return glossaryEntries.stream()
-                .filter(glossaryEntry -> {
-                    // Bug #1315: skip monolingual glossary entries where all target
-                    // terms are empty. An empty target term is not a real term to
-                    // check against, so raising a warning for it is a false positive.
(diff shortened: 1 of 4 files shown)


brandelune:
Thank you for the fix. It looks like Manuel has tested it and confirmed the fix.
You’ll need to check the documentation to update it where necessary.

src/org/omegat/gui/issues/TerminologyIssueProvider.java
```diff
@@ -78,11 +78,21 @@ public List<IIssue> getIssues(SourceTextEntry sourceEntry, TMXEntry tmxEntry) {
...
             return Collections.emptyList();
         }
-        return glossaryEntries.stream().map(glossaryEntry -> {
-            List<String> trgTerms = Core.getGlossaryManager().searchTargetMatches(tmxEntry.translation,
-                    sourceEntry.getProtectedParts(), glossaryEntry);
-            return trgTerms.isEmpty() ? new TerminologyIssue(sourceEntry, tmxEntry, glossaryEntry) : null;
-        }).filter(Objects::nonNull).collect(Collectors.toList());
+        return glossaryEntries.stream()
+                .filter(glossaryEntry -> {
+                    // Bug #1315: skip monolingual glossary entries where all target
```
miurahr: No need to reference of the issue in a source code.
We can see a git blame and see GitHub pull request discussion when necessary.

src/org/omegat/gui/issues/TerminologyIssueProvider.java
```diff
@@ -78,11 +78,21 @@ public List<IIssue> getIssues(SourceTextEntry sourceEntry, TMXEntry tmxEntry) {
...
+                    // terms are empty. An empty target term is not a real term to
+                    // check against, so raising a warning for it is a false positive.
+                    String[] locTerms = glossaryEntry.getLocTerms(false);
+                    return Arrays.stream(locTerms).anyMatch(t -> !t.isEmpty());
```
miurahr: It is better to make this part as private method and use method reference in the stream pipeline. When the private method name is self explaining a purpose, it is best.

miurahr:
Thank you for the proposal. This is a good catch.
I have small code style comments, and I will merge when solved.

src/org/omegat/gui/issues/TerminologyIssueProvider.java
fung911: Thank you for the review! Please check my latest update :

1. Removed the issue reference from the inline comment — the method name hasNonEmptyTargetTerms is self-explanatory.
2. Extracted the filter logic into a private static method hasNonEmptyTargetTerms and used a method reference in the stream pipeline.
3. Also updated the documentation in Menus_Tools.xml and App_Glossaries.xml to clarify that entries with an empty target term are excluded from terminology checks.
4. Added unit tests for hasNonEmptyTargetTerms covering empty, non-empty, all-empty, and partially-empty target term cases.

miurahr: looks perfect for me. @brandelune could you review manual updates?

src_docs/manual/en/App_Glossaries.xml
```diff
@@ -105,6 +105,13 @@
          three-column lists, with the source term in the first column, an optional
          target term in the second column, and an optional comment in the third
          column.</para>
+      <note>
```
brandelune: This note section is not necessary since the target part is indicated in the sentence before as being optional.

src_docs/manual/en/Menus_Tools.xml
```diff
@@ -72,9 +72,15 @@
                   <para>
                      <guilabel>Terminology Issues</guilabel> (optional): detects
                      all the glossary items that are not properly translated. See the
-                     
+
                      <link linkend="dialogs.preferences.glossary" endterm="dialogs.preferences.glossary.title"></link> preferences for
                      details.</para>
+                  <note>
+                     <para>Glossary entries with an empty target term are excluded
+                        from terminology checks. Such entries — for example, those in
```
brandelune: Only the first sentence is necessary.

brandelune (changes requested):

src_docs/manual/en/Menus_Tools.xml
fung911: Thank you for the update! I delete the unnecessary content. Please check.

brandelune (approved):
Excellent ! Thank you !

PR #2025, opened 2026-05-18
feat: add bilingual export and import feature (CSV/XLSX) [RFE#1800]

SakthiRJ:
Allows exporting unique project segments to a bilingual CSV or Excel file for external translation, and importing the translated file back to automatically update project segments.

# feat: add Bilingual Export and Import feature (CSV/XLSX)

## Pull request type

## Which ticket is resolved?

- Title: Add Bilingual Export and Import feature (CSV/XLSX)
  * URL: https://sourceforge.net/p/omegat/feature-requests/ (New Feature Contribution)

## What does this PR change?

- **Core Export/Import Logic**: Creates `ExportBilingualHandler.java` in `src/org/omegat/core/export/` to handle collecting unique project segments, formatting them into two-column (`Source`, `Target`) bilingual files, and parsing imported translations back into the active project.
- **Spreadsheet & CSV Support**: Supports standard `.csv` (with robust RFC-4180 quoting and multi-line segment escaping) as well as `.xlsx` Excel spreadsheets.
- **Decoupled Excel Integration**: Creates `ExcelHelper.java` using reflection to interact with Apache POI, ensuring the core OmegaT application remains cleanly decoupled and can compile regardless of POI dependency configuration.
- **Menu Integration**: Wires two new menu actions into `BaseMainWindowMenu.java` and `MainWindowMenuHandler.java` under the **Project** menu (`Project > Bilingual Export...` and `Project > Import Translations from Bilingual File...`).
- **Localization**: Adds localized UI titles, dialog headers, file filters, and success/error messages to `Bundle.properties`.
- **Build & Dependencies**: Adds `org.apache.poi:poi-ooxml:5.2.3` to `gradle/libs.versions.toml` and `build.gradle` for reliable Excel file generation and reading.
- **Unit Testing**: Introduces a comprehensive test suite `ExportBilingualHandlerTest.java` covering CSV quoting rules, multi-line values, BOM handling, and translation application logic.

#### 1. Deduplication & Segment Scope
- **Export**: Iterates over `project.getAllEntries()` and automatically deduplicates repeated source texts. Each unique segment appears exactly once in the exported file. If a segment already has a translation in OmegaT, it is pre-populated in the `Target` column for context; untranslated segments have a blank `Target` column.
- **Import**: Reads the bilingual file and matches segments by their `Source` string. Non-empty `Target` strings are ingested via `project.setTranslation(...)` as project-wide default translations, instantly updating all instances of those segments across all project documents.

#### 2. Robust CSV Handling (No External CSV Library Required)
- Implements strict RFC-4180 CSV parsing and escaping natively to avoid adding unnecessary third-party CSV dependencies.
- Safely handles embedded double-quotes (escaped as `""`), commas, and multi-line segment strings with line breaks. Automatically detects and handles UTF-8 Byte Order Marks (BOM) on import.

#### 3. Apache POI Reflection Design
- To prevent hard dependency locking in core classes, `ExcelHelper.java` loads Apache POI classes (`XSSFWorkbook`, `XSSFSheet`, `XSSFRow`) dynamically via reflection. If maintainers decide to make Apache POI an optional plugin or separate module in the future, the core bilingual handler will continue to compile and function seamlessly for CSVs.


brandelune:
Thank you for this feature.

Do you have a specific use case in mind?
Why prefer XSLX when ODS is also available and is supported by Excel?

We have a plugin that reads in memory an ODT file table after external review of an OmegaT translation.

It would be nice if we could discuss what you have in mind.

SakthiRJ:
> discuss
Hi @brandelune 
As both a translator and a software engineer, I built this feature to solve a frequent workflow challenge. I often collaborate with direct clients and external subject matter experts who don't use CAT tools and simply want a two-column spreadsheet to review or translate. This feature automates that process by exporting only the unique segments across the entire project so reviewers don't waste time on repetitions, while pre-filling existing translations for context. When the spreadsheet is imported back, OmegaT automatically applies the updates globally across all project files.

Regarding the file format, corporate clients and freelance reviewers work almost exclusively in Microsoft Excel or Google Sheets. While Excel can open ODS files, it frequently displays compatibility warning prompts that confuse non-technical clients, so XLSX provides them with a completely friction-free experience. 

To ensure a fully open standard as well, I also built native CSV support that requires no external libraries and works perfectly in Calc, Excel, and Sheets. To keep OmegaT's core clean, the Excel logic uses reflection so the Apache POI library remains cleanly decoupled. Because the underlying code is modular, adding ODS support alongside XLSX and CSV would be very simple, and I would be more than happy to collaborate on adding that or adjusting the implementation to fit your vision for the project.

brandelune:
Here is the plugin that I mentioned earlier:
https://github.com/briacp/plugin-odt-review

brandelune:
CSV never works perfectly. There is no notion of encoding in CSV and Excel/LO have different ideas of how to import the format. Also, it is notoriously painful as a translation text format if you just want to open it in a text editor.

brandelune:
As far as the review side of your feature is concerned, we have that plugin that is only waiting to be properly integrated. If you want to add the XSLX format as a possibility, go ahead.

Regarding the translation side of your feature, what do you do with references TMs? glossary matches?

SakthiRJ:
Thank you, you’re absolutely right about CSV. The lack of a standard encoding definition, different import behavior between Excel and Calc, and the difficulty of handling multi-line segments in a text editor are all major pain points.

Those issues are exactly why I added robust `.xlsx` support to this feature. With `.xlsx`, encoding is consistently handled through XML/UTF-8, multi-line segments and cell boundaries are preserved correctly, and both Excel and Calc open the file directly without import prompts or encoding confusion. CSV was only included as a fallback for users without office suites, so I’m perfectly fine removing it entirely if preferred.

I’m also glad to hear the review plugin is ready for integration.

Regarding reference TMs and glossaries, that’s a very good suggestion. At the moment, the export only includes exact active translations (for example from `/tm/auto/` or existing project memory). Untranslated segments are exported with an empty target field, so external translators currently don’t receive TM or glossary context.

Since the export is spreadsheet-based, a clean solution would be to add two additional columns:

Column C: best fuzzy match from reference TMs
Column D: glossary terms detected in the source text

This would allow external translators to access TM and glossary context directly within the spreadsheet.

brandelune:
> Thank you, you’re absolutely right about CSV. The lack of a standard encoding definition, different import behavior between Excel and Calc, and the difficulty of handling multi-line segments in a text editor are all major pain points.

If you consider a plain text format for a fallback export, TSV is probably better. Segments that contain TAB should be extremely rare (and even then, TAB can be escaped) and the format is actually the same as the glossary format. So I guess you can reuse some machinery we have to work on our writable glossary.
 
> Those issues are exactly why I added robust `.xlsx` support to this feature. With `.xlsx`, encoding is consistently handled through XML/UTF-8, multi-line segments and cell boundaries are preserved correctly, and both Excel and Calc open the file directly without import prompts or encoding confusion.

The reason being that Microsoft prefers to spread FUD on its users rather than properly handling the files. When LO just does what needs to be done.

Since OmegaT is free software and aims at freeing users from non free software, it is difficult to accept that a non-free format is chosen as the default, instead of the free format that is ODS.

> I’m also glad to hear the review plugin is ready for integration.

I don’t remember why Briac chose a table in an ODT file, instead of an ODS file, but that code is intended to be included in OmegaT and as far as the data import side is concerned I don’t see why we’d have two different mechanisms for that.
 
> Regarding reference TMs and glossaries, that’s a very good suggestion. At the moment, the export only includes exact active translations (for example from `/tm/auto/` or existing project memory). Untranslated segments are exported with an empty target field, so external translators currently don’t receive TM or glossary context.

Nice.
 
> Since the export is spreadsheet-based, a clean solution would be to add two additional columns:
> Column C: best fuzzy match from reference TMs Column D: glossary terms detected in the source text

I guess you’ll have to put some thought into the contents of column D.

Also, as you must be aware, there is great interest in having a mechanism to handle "game localization Excel files" as source files in OmegaT.

Such files come as spreadsheets with labeled columns (source / translation / references / comments / whanot) and the translator is supposed to check all the data and write the translation into the "translation" column.

You know that our PO support reads already translated segments and "PO fuzzies" and puts all that into TM. Such "Spreadsheet" file could work the same. And your proposal sounds like one big step in that direction. 

> This would allow external translators to access TM and glossary context directly within the spreadsheet.

Indeed. And that’s what they have in those game-l10n spreadsheet files.

miurahr (changes requested):
Please fix the checkstyle errors.

src/org/omegat/Bundle.properties
```diff
@@ -2959,8 +2959,12 @@ TF_BILINGUAL_EXPORT_TITLE=Export Bilingual Segments
 TF_BILINGUAL_IMPORT_TITLE=Import Bilingual Translations
 TF_BILINGUAL_EXPORT_SUCCESS=Bilingual file exported successfully: {0}
 TF_BILINGUAL_IMPORT_SUCCESS=Imported {0} translation(s) successfully.
+TF_BILINGUAL_IMPORT_NONE=No new translations were imported.
 TF_BILINGUAL_IMPORT_ERROR=Failed to import bilingual file: {0}
 TF_BILINGUAL_EXPORT_ERROR=Failed to export bilingual file: {0}
+TF_BILINGUAL_FILTER_CSV=CSV Files (*.csv)
+TF_BILINGUAL_FILTER_XLSX=Excel Files (*.xlsx)
```
brandelune: I mentioned that a proprietary format was not to be considered as a default option in OmegaT. Please consider ODS instead.

SakthiRJ:
@miurahr  The checkstyle issues are fixed

src/org/omegat/core/export/ExcelHelper.java
```diff
@@ -0,0 +1,166 @@
+/**************************************************************************
+ OmegaT - Computer Assisted Translation (CAT) tool
+          with fuzzy matching, translation memory, keyword search,
+          glossaries, and translation leveraging into updated projects.
+
+ Copyright (C) 2026 OmegaT contributors
```
miurahr, line 6: Here should be your name or nickname.

miurahr (changes requested):

miurahr:
Please raise RFE ticket before PR raised.
https://sourceforge.net/p/omegat/feature-requests/ 

I don't merge it without issue ticket. See
https://omegat.readthedocs.io/en/latest/04.ContributingOmegaT.html

brandelune:
> Please raise RFE ticket before PR raised. https://sourceforge.net/p/omegat/feature-requests/
> 
> I don't merge it without issue ticket. See https://omegat.readthedocs.io/en/latest/04.ContributingOmegaT.html

And there is not merging before the feature corresponds to something that we need and that is not redundant with existing features. Hence my comments.

SakthiRJ:
@miurahr  I have raised the RFE ticket on SourceForge: https://sourceforge.net/p/omegat/feature-requests/1800/

brandelune:
> @miurahr I have raised the RFE ticket on SourceForge: https://sourceforge.net/p/omegat/feature-requests/1800/

I closed the PR as a duplicate of Briac’s plugin. Please review the comments made above.

miurahr:
PLease discuss a direction on dev-list of OmegaT. After reaching consencus of the direction, I resume to come here.

zs-stpa:
possible duplicate ticket: https://sourceforge.net/p/omegat/feature-requests/366/

i independently reimplemented csv export for my own use case in #2336 because what i needed was to share from one project a filtered sorted list of segments with their comments to someone who has no need for a full omegaT or the full project.

i am intrigued by news of this review plugin - where can i find that and does that really solve my export-csv-problem and how can i help it along maybe to get its integration finished?

zs-stpa:
i needed it at a certain point of time in the past. hence i implemented a solution and tried to share it so others can reuse it in the future. how is the planned timeline on this, if i may ask? i see a draft here, mention of a stalled plugin that is not yet ready, and a closed PR that did indeed solve my problem but was rejected. clarification please? and yes, i will send a copy on the dev-list a.s.a.p.

PR #2026, opened 2026-05-19
Add documentation about "orphan segments" and fix related UI label

brandelune:
A Telegram channel support request hinted that _orphan segments_ were not properly documented in the manual.

This PR adds a paragraph to the manual and corrects the UI label for orphan segments when they appear in matches or searches (the label should be singular, not plural).


miurahr (approved):

brandelune:
What is the best approach to merge that into master?
I locally tried a cherry-pick, but since the XML file has been displaced it does not look like an optimal choice.

What do you advise @miurahr?

miurahr:
Because a change is small, you can cherry-pick and work with a Git conflict resolution or edit it by hand as you like.

PR #2029, opened 2026-05-22
Apply orphan segment 6.0 patch to 6.1

brandelune:
The patch was applied to 6.0 to add a short description of what orphan segments were and to clarify the UI wording (no plural to "orphan segment" in the match and search UI).

### DIFF
--- a/src/org/omegat/Bundle.properties
+++ b/src/org/omegat/Bundle.properties
@@ -720,7 +720,7 @@ CT_ERROR_ACCESS_GLOSSARY_DIR=Cannot access glossary folder!
 
 CT_ERROR_CREATING_TARGET_DIR=Cannot create folder for translated files:
 
-CT_ORPHAN_STRINGS=Orphan segments
+CT_ORPHAN_STRINGS=Orphan segment
 
 CT_START_EXTERNAL_CMD=Executing post-processing commands...
 
--- a/src_docs/manual/en/OmegaT_EditorsPanes.xml
+++ b/src_docs/manual/en/OmegaT_EditorsPanes.xml
@@ -748,8 +748,25 @@ Dossier de configuration
          parts in 
          <emphasis role="bold">green</emphasis> (when the adjacent part is a
          space, as above, the part is left uncolored).</para>
-      <para>The match also displays three matching percentages.</para>
-      <para>These percentages have the following meaning, in order:</para>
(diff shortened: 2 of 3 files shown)


miurahr:
The failures was caused by your change.  Please fix the test expectation.

miurahr (changes requested):
Please fix the test case that was failed by your UI change.

brandelune:
Is it related to this line : 
https://github.com/omegat-org/omegat/blob/3d889d635268fd492aa884a6d15ad024afb92607/test/src/org/omegat/core/search/SearcherTest.java#L352

miurahr:
> ❌ Quality checks **failed**.
> 
> Please look a Gradle Scan page for details: https://gradle.com/s/3bsjye3uwuzsm

You can see the failed test details on the Gradle Scan screen.

brandelune:
> > ❌ Quality checks **failed**.
> > Please look a Gradle Scan page for details: https://gradle.com/s/3bsjye3uwuzsm
> 
> You can see the failed test details on the Gradle Scan screen.

I checked the page and it is not useful.

`at org.omegat.core.search.SearcherTest.testGetSearchResultsExactMatch(SearcherTest.java:352)`

I checked that line and I have no idea why we have a static UI string in English in a test case and I also have no idea what the test case is testing.

So I am not going to change anything there.

If you want to explain to me why we have that string is a test case and how that relates to the bundle, I’ll be able to make an informed decision.

Also, I really don’t think it is a good idea to use UI strings in test cases.

miurahr:
There are several similar test cases that is

https://github.com/omegat-org/omegat/blob/2b45c4768b66c2f564fad5ebcdd9f1a2e184a175/test/src/org/omegat/core/search/SearcherTest.java#L367

It indicates us a way to fix.

brandelune:
> There are several similar test cases that is
> 
> https://github.com/omegat-org/omegat/blob/2b45c4768b66c2f564fad5ebcdd9f1a2e184a175/test/src/org/omegat/core/search/SearcherTest.java#L367
> 
> It indicates us a way to fix.

How can I infer from the original error message how I can fix the test case this way ?

And how is it that the test case uses a static string and not a bundle ID ?

brandelune (commented):
It would be nice to check that we don’t have other static UI strings in the test cases. Unless they are logically required.

PR #2043, opened 2026-05-26
test: add coverage for space-delimited language handling

FanShuai2022:
## Pull request type

- Other -> test

## Which ticket is resolved?

- None

## What does this PR change?

- Adds a unit test for `Language.isSpaceDelimited()`.
- Verifies that English and French are treated as space-delimited languages.
- Verifies that Chinese, Japanese, and Tibetan are treated as non-space-delimited languages.

## Other information

This test is useful because OmegaT supports many languages, and not all languages separate words using spaces. The test helps protect the expected behavior for languages such as Chinese, Japanese, and Tibetan, where normal space-based word separation should not be assumed.

Verification performed:

- `.\gradlew.bat :test --tests org.omegat.util.LanguageTest.testIsSpaceDelimited`

### DIFF
--- a/test/src/org/omegat/util/LanguageTest.java
+++ b/test/src/org/omegat/util/LanguageTest.java
@@ -116,6 +116,19 @@ public class LanguageTest {
         assertTrue(Language.verifySingleLangCode("ar-AE-x-dubai"));
     }
 
+    /**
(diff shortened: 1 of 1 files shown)


miurahr:
NIce to be a first PR!

miurahr (approved):

PR #2064, opened 2026-06-02
docs(devel): add offline/enterprise build instructions and update ADR#2025001

miurahr:
We recently add CI test to check source distribution consistency.
We have unique policy in source.zip distribution that should have all dependency library jars to allow future developers to reproduce old version of the OmegaT even if the dependency project had removed.

This PR add an explanation of the offline build and source distribution policy, and amend explanation and PR links to ADR#2025001.

## Pull request type

- Documentation -> [documentation]

## Which ticket is resolved?
none
## What does this PR change?

- Update ADR#2025001
- Add 09.OfflineBuild.md
- Update links in index.md and 08.HowToBuild.md

PR #2066, opened 2026-06-03
chore(build): add offline dependency support and integrate into source distribution build

miurahr:
Implement a way to bundle build-logic dependencies into source package.

## Pull request type

## Which ticket is resolved?
none
## What does this PR change?

- add new `build-logic` task to download dependencies
- configure source distribution bundles `build-logic` dependencies in mavenLocal format.

### DIFF
--- a/build-logic/src/main/groovy/org.omegat.java-conventions.gradle
+++ b/build-logic/src/main/groovy/org.omegat.java-conventions.gradle
@@ -20,6 +20,7 @@ ext {
     }
     providedCoreLibsPath = 'lib/provided/core'
     providedModuleLibsPath = 'lib/provided/module'
+    providedBuildLogicPath = 'lib/provided/build-logic'
     providedCoreLibsDir = layout.settingsDirectory.dir(providedCoreLibsPath).asFile
     providedModuleLibsDir = layout.settingsDirectory.dir(providedModuleLibsPath).asFile
 }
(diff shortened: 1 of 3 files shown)


miurahr:
Related to #2064

build-logic/src/main/groovy/org.omegat.java-conventions.gradle
```diff
@@ -20,6 +20,7 @@ ext {
     }
     providedCoreLibsPath = 'lib/provided/core'
     providedModuleLibsPath = 'lib/provided/module'
+    providedBuildLogicPath = 'lib/provided/build-logic'
```
miurahr: The definition is used only in `<root>/build.gradle#439`.
So we should move it into `build.gradle`

build.gradle
```diff
@@ -332,6 +332,8 @@ def moduleLibs = subprojects
         }
         .flatten()
 
+def buildLogicLibsDir = layout.settingsDirectory.file('build-logic/build/libs/build-logic')
+
```
miurahr, line 337:
```suggestion
def providedBuildLogicPath = 'lib/provided/build-logic'

```

build-logic/src/main/groovy/org.omegat.java-conventions.gradle
miurahr:
```suggestion
```

PR #2073, opened 2026-06-08
feat:Add optional progress display to Source Files window

burgger:
## Pull request type

Feature enhancement -> [enhancement]

## Which ticket is resolved?

- Display translated percentage on Project files dialog
  * URL: https://sourceforge.net/p/omegat/feature-requests/1775/

## What does this PR change?

- Adds an optional per-file Progress column to the Source Files window.
- Adds a Preferences > Appearance > Source Files page for enabling per-file progress and optional progress color bars.
- Keeps per-file progress disabled by default to avoid extra file-entry traversal during normal use.
- Shows the overall translated unique segment percentage in the Source Files totals table.
- Refreshes per-file progress only when the Source Files window is visible and the progress option is enabled.

## Other information

Tests run:

- `./gradlew :test --tests org.omegat.gui.filelist.ProjectFilesListControllerTest`
- `./gradlew changeOnBranch`
- `./gradlew spotlessChangedApply`
- `git diff --check`
Source Files page
Preference Page


brandelune:
Nice. Thank you.

I am not sure the preference belongs to "Appearance". For lack of a better location I’d put it in General (note to self: we need to sort all the preferences, because it’s becoming a mess).

Also, why 2 preferences ?

burgger:
> Nice. Thank you.
> 
> I am not sure the preference belongs to "Appearance". For lack of a better location I’d put it in General (note to self: we need to sort all the preferences, because it’s becoming a mess).
> 
> Also, why 2 preferences ?

Thank you for the feedback.

That makes sense. I will move the option to General and replace the two checkboxes with a single setting, e.g. Off / Percentage / Percentage with color bars.

I will also fix the SpotBugs warning from CI.

brandelune:
> > Nice. Thank you.
> > I am not sure the preference belongs to "Appearance". For lack of a better location I’d put it in General (note to self: we need to sort all the preferences, because it’s becoming a mess).
> > Also, why 2 preferences ?
> 
> Thank you for the feedback.
> 
> That makes sense. I will move the option to General and replace the two checkboxes with a single setting, e.g. Off / Percentage / Percentage with color bars.
> 
> I will also fix the SpotBugs warning from CI.

Could you explicit the setting?
The "Percentage" is the number shown at the bottom, and the "color bar" is the, well, color bar shown at the top right?

If that is the case, I don’t think we need that much fine grained preferences. Either we have a "show progress" or we don’t. Adding the extra option does not add much value to the UI.

What do you think?

burgger:
> > > Nice. Thank you.
> > > I am not sure the preference belongs to "Appearance". For lack of a better location I’d put it in General (note to self: we need to sort all the preferences, because it’s becoming a mess).
> > > Also, why 2 preferences ?
> > 
> > 
> > Thank you for the feedback.
> > That makes sense. I will move the option to General and replace the two checkboxes with a single setting, e.g. Off / Percentage / Percentage with color bars.
> > I will also fix the SpotBugs warning from CI.
> 
> Could you explicit the setting? The "Percentage" is the number shown at the bottom, and the "color bar" is the, well, color bar shown at the top right?
> 
> If that is the case, I don’t think we need that much fine grained preferences. Either we have a "show progress" or we don’t. Adding the extra option does not add much value to the UI.
> 
> What do you think?

I agree that a single preference is enough.

I will simplify this to one “Show Source Files progress” option. When enabled, it will show the per-file progress(percentage and color bar) and the project total progress(only percentage); when disabled, the progress calculation/display stays off(both total and per-file) by default.

brandelune:
> > > > Nice. Thank you.
> > > > I am not sure the preference belongs to "Appearance". For lack of a better location I’d put it in General (note to self: we need to sort all the preferences, because it’s becoming a mess).
> > > > Also, why 2 preferences ?
> > > 
> > > 
> > > Thank you for the feedback.
> > > That makes sense. I will move the option to General and replace the two checkboxes with a single setting, e.g. Off / Percentage / Percentage with color bars.
> > > I will also fix the SpotBugs warning from CI.
> > 
> > 
> > Could you explicit the setting? The "Percentage" is the number shown at the bottom, and the "color bar" is the, well, color bar shown at the top right?
> > If that is the case, I don’t think we need that much fine grained preferences. Either we have a "show progress" or we don’t. Adding the extra option does not add much value to the UI.
> > What do you think?
> 
> I agree that a single preference is enough.
> 
> I will simplify this to one “Show Source Files progress” option. When enabled, it will show the per-file progress(percentage and color bar) and the project total progress(only percentage); when disabled, the progress calculation/display stays off(both total and per-file) by default.

Thank you for your understanding regarding the preferences.

Now there is the visibility issue. A number with a thin color bar in its background not super visible. Either the color bar takes the full height of the field (like the background color for the selected file), with characters displayed in it (the percentage), or the percentage should be removed and a different indication for progress is given (the RFE suggests colors).

Also, I’m not sure why we can’t align the color bars at the top and the percentage at the bottom. In both cases, the data (without the progress widget) is right aligned. It seems nicer to just add a column on the right to have the progress data also right aligned. And the overall progress data at the bottom could also be a progress bar.

(I don’t want to overburden you with silly requests, but once your code is in OmegaT, it’ll stay for the foreseeable future.)

burgger:
> > > > > Nice. Thank you.
> > > > > I am not sure the preference belongs to "Appearance". For lack of a better location I’d put it in General (note to self: we need to sort all the preferences, because it’s becoming a mess).
> > > > > Also, why 2 preferences ?
> > > > 
> > > > 
> > > > Thank you for the feedback.
> > > > That makes sense. I will move the option to General and replace the two checkboxes with a single setting, e.g. Off / Percentage / Percentage with color bars.
> > > > I will also fix the SpotBugs warning from CI.
> > > 
> > > 
> > > Could you explicit the setting? The "Percentage" is the number shown at the bottom, and the "color bar" is the, well, color bar shown at the top right?
> > > If that is the case, I don’t think we need that much fine grained preferences. Either we have a "show progress" or we don’t. Adding the extra option does not add much value to the UI.
> > > What do you think?
> > 
> > 
> > I agree that a single preference is enough.
> > I will simplify this to one “Show Source Files progress” option. When enabled, it will show the per-file progress(percentage and color bar) and the project total progress(only percentage); when disabled, the progress calculation/display stays off(both total and per-file) by default.
> 
> Thank you for your understanding regarding the preferences.
> 
> Now there is the visibility issue. A number with a thin color bar in its background not super visible. Either the color bar takes the full height of the field (like the background color for the selected file), with characters displayed in it (the percentage), or the percentage should be removed and a different indication for progress is given (the RFE suggests colors).
> 
> Also, I’m not sure why we can’t align the color bars at the top and the percentage at the bottom. In both cases, the data (without the progress widget) is right aligned. It seems nicer to just add a column on the right to have the progress data also right aligned. And the overall progress data at the bottom could also be a progress bar.
> 
> (I don’t want to overburden you with silly requests, but once your code is in OmegaT, it’ll stay for the foreseeable future.)

Thank you for the feedback. I think this makes a lot of sense.

I agree that the current progress bar is not visible enough, and your suggestion to make the color bar take the full height of the cell is a good improvement. I am working on changing the renderer in that direction so the progress indication is clearer.

I will also clean up the layout/formatting issues so the per-file progress column and the total progress at the bottom are aligned more consistently.

brandelune:
> Thank you for the feedback. I think this makes a lot of sense.
> 
> I agree that the current progress bar is not visible enough, and your suggestion to make the color bar take the full height of the cell is a good improvement. I am working on changing the renderer in that direction so the progress indication is clearer.
> 
> I will also clean up the layout/formatting issues so the per-file progress column and the total progress at the bottom are aligned more consistently.

Thank you for your understanding.

I think that when you have done that, we’ll have a proper prototype that the community will be able to test and discuss.

burgger:
> Thank you for your understanding.
> 
> I think that when you have done that, we’ll have a proper prototype that the community will be able to test and discuss.

How do you feel about this preference for the progress?

brandelune:
Additional items will be:
- add relevant color settings in the color preferences for this feature
- add relevant sentences in the manual

src/org/omegat/gui/preferences/view/SourceFilesViewOptionsController.java
```diff
@@ -0,0 +1,101 @@
...
+
+import java.awt.BorderLayout;
+import java.util.Objects;
+
+import javax.swing.JCheckBox;
+import javax.swing.JComponent;
+import javax.swing.JPanel;
+import javax.swing.border.EmptyBorder;
+
+import org.jetbrains.annotations.Nullable;
```
miurahr: Please use  `org.jspecify.annotaitons.Nullable`

miurahr (changes requested):
We now migrate to the modern Java standard [ jspecify ](https://jspecify.dev/) 

FYI: see https://dev.to/headf1rst/solving-the-billion-dollar-mistake-modern-java-null-safety-with-jspecify-and-nullaway-2ie7

burgger:
I updated the PR to add the Source Files progress colors to the color preferences, added the related manual text, and replaced the JetBrains Nullable with jspecify Nullable.

I also noticed that the color chooser has transparency controls, but custom colors are currently persisted as RGB only, so alpha would not survive an application restart. I think that is a separate color preferences issue rather than part of this Source Files progress feature. For this PR, I kept the new progress colors as regular RGB values with readable defaults.

src/org/omegat/gui/filelist/ProjectFilesListController.java
```diff
@@ -138,10 +149,15 @@
 @SuppressWarnings("serial")
```
miurahr, line 150: It is better to have `@NullMarked` JSpecify annotation that indicate all fields, arguments are not null.

```suggestion
@SuppressWarnings("serial")
@NullMarked
```

There are two ways to anotate non-nullity; one is annotation on class another is annotation on package-info.java

miurahr (commented):
It is not mandatory but better to have `@NullMarked`

miurahr (approved):

brandelune:
@burgger using master on the current l10n, I see weird things.
`greeting/philosophy` is supposed to be already translated from the website strings, but I get 10 segments, 0 unique and 0% translated.

It seems to me that files with 0 unique segments should be considered 100% translated.

brandelune:
Screen Shot 2026-08-28 at 12 52 34

PR #2077, opened 2026-06-15
feat: sign macOS dylibs in jar dependencies for improved compatibility

miurahr:
We have many dependencies in our subprojects/modules.
There are several dependnecies includes native libraries for its feature such as

- `FlatLaf LookAndFeel library` in `theme` subproject
- `Dumont Hunspell` in `spellchecker/hunspell`  subproject

These native libraries should be singned for notarization of apple macOS platform to run on.
This adds a logic to sign the nativef libraries which has `.dylib` extension in jar files in subproject/module dependencies when creating FarJar jar files.
 
## Pull request type

## Which ticket is resolved?

none

## What does this PR change?
- Add logic to detect and sign *.dylib files in jar dependencies using Apple developer identity
- Update fat-jar process to include signed dylibs instead of original zipTree
- Inject `ExecOperations` and `FileSystemOperations` for file manipulation and command execution


miurahr:
Build cleanly with the change. There is no change in app feature and build process when there is no signing property. So we can merge it.

PR #2083, opened 2026-06-19
docs: update releases/readme.txt

miurahr:
## Pull request type
- Docs

## Which ticket is resolved?
none

## What does this PR change?

- Sync with current status


miurahr:
We should update l10n-ed readme in all languages after merged.

PR #2110, opened 2026-07-11
refactor(preferences): install the preferences store through a testable seam

zs-stpa:
Preferences is a static facade over a single store built once per JVM behind a
didInit guard. In the test suite that let the first test to touch preferences
fix the shared store for every later test, so mutated values leaked and produced
order-dependent failures - most visibly EditorControllerTest, whose offset
assertions depend on MARK_PARA_DELIMITATIONS (left false by the console tests;
the fragile assertions came from #1326).

Instead of a bespoke test-only reset of private state, add a single explicit
installation point setPreferences(IPreferences), annotated @VisibleForTesting,
mirroring the CoreState and RuntimePreferenceStore seams. init() now routes
through it and stays idempotent. Tests install a fresh, in-memory TestPreferences
per test (via TestPreferencesInitializer), so no preference state leaks across
the suite; the resetPreferencesForTest hook is removed.

Production behavior is unchanged: setPreferences is package-private and reached
only through init() outside tests.

The mechanism and this decision are documented in ADR 2026002.

Addresses the order-dependent failures caused by the assertions introduced in
898c07f666de89db3e8f8e6c7d1a0a2c78349c01 (#1326).

### DIFF
--- a/src/org/omegat/util/Preferences.java
+++ b/src/org/omegat/util/Preferences.java
@@ -41,6 +41,7 @@ import java.io.IOException;
 import java.util.Objects;
 
 import org.jetbrains.annotations.Nullable;
+import org.jetbrains.annotations.VisibleForTesting;
 import org.omegat.core.segmentation.SRX;
 import org.omegat.core.segmentation.SRXManager;
 import org.omegat.filters2.master.FilterMaster;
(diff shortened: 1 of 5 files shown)


miurahr (changes requested):
There are two commits and the first commit is removed by the second commit.
This means the first commit is meanless.
Please `git rebase -i` to squash commits and make the Pull-Request clean.
The commit message is good and a enough explaination.

miurahr:
For reference;
We have `org.omegat.core.data.CoreStats` class that is a singleton but it has a `CoreStats#setInstance` which is out-of-theory for test purpose. I put `@VisibleForTesting` annotation fot the method.
We also have `TestCoreStats` inherit from `CoreStats`.  it has `TestCoreStats#resetStats` method for test isolation.

We can take a similar strategy for the `Preferences` mechanism in the OmegaT core.
You propose a hack of the Preferences class -- it seems minimum change for the purpose -- but we have a chance to refactor the preferences architecture.

Unfortunately, we don't have an architecture decision record and technical specification document about the Preferences. I think it is good chance for us to prepare the specification document and update it with the change.

zs-stpa:
> For reference; We have `org.omegat.core.data.CoreStats` class that is a singleton but it has a `CoreStats#setInstance` which is out-of-theory for test purpose. I put `@VisibleForTesting` annotation fot the method. We also have `TestCoreStats` inherit from `CoreStats`. it has `TestCoreStats#resetStats` method for test isolation.
> 
> We can take a similar strategy for the `Preferences` mechanism in the OmegaT core. You propose a hack of the Preferences class -- it seems minimum change for the purpose -- but we have a chance to refactor the preferences architecture.
> 
> Unfortunately, we don't have an architecture decision record and technical specification document about the Preferences. I think it is good chance for us to prepare the specification document and update it with the change.

okay, it is done.

zs-stpa:
i have updated the title and description text of this PR.

miurahr:
Now your provide a perfect ADR and the implementation. This is super improvment!
Thank you for the contribution.

miurahr (approved):

brandelune (approved):

PR #2111, opened 2026-07-12
fix(matches): treat full-width digits as equivalent when substituting…

zs-stpa:
… fuzzy-match numbers

Full-width Arabic numerals (U+FF10-U+FF19), common in Japanese source text, were not recognized as equivalent to their half-width counterparts when OmegaT substitutes numbers into an inserted fuzzy match. As a result the number-equivalence check could fail (blocking a valid substitution), and a substituted number was inserted with the source's width instead of the target's, producing for example "sentence ９" instead of "sentence 9".

Number comparison and index mapping in MatchesTextArea.substituteNumbers now operate on the width-normalized (StringUtil.normalizeWidth) form, and each substituted number adopts the digit width of the target token it replaces. Behaviour stays gated by the existing CONVERT_NUMBERS preference; no new option is introduced.

Adds tests for the reported case, the equivalence guard, both width directions, multiple and mixed-width numbers, and the real Japanese (Kuromoji) tokenizer path, and pins the scope limits (full-width decimals and non-full-width numeral systems such as Arabic-Indic are out of scope).

Addresses feature request #1193.
- Convert full-width Arabic numbers to half-width when inserting a fuzzy match
- https://sourceforge.net/p/omegat/feature-requests/1193/

### DIFF
--- a/src/org/omegat/gui/matches/MatchesTextArea.java
+++ b/src/org/omegat/gui/matches/MatchesTextArea.java
@@ -301,20 +301,29 @@ public class MatchesTextArea extends EntryInfoThreadPane<List<NearString>> imple
         List<String> sourceNumbers = Stream.of(sourceTok.tokenizeVerbatimToStrings(source))
                 .filter(MatchesTextArea::isNumber).collect(Collectors.toList());
 
+        // Compare and map numbers on their width-normalized form so that
+        // full-width (ASCII) digits (U+FF10-U+FF19, common in Japanese) are
+        // treated as equivalent to their half-width counterparts. See feature
+        // request #1193.
+        List<String> normSourceMatchNumbers = normalizeDigitWidth(sourceMatchNumbers);
+        List<String> normTargetMatchNumbers = normalizeDigitWidth(targetMatchNumbers);
+
         if (sourceMatchNumbers.size() != sourceNumbers.size()
                 || sourceMatchNumbers.size() != targetMatchNumbers.size()
-                || !new HashSet<>(sourceMatchNumbers).equals(new HashSet<>(targetMatchNumbers))) {
+                || !new HashSet<>(normSourceMatchNumbers).equals(new HashSet<>(normTargetMatchNumbers))) {
             return targetMatch;
         }
 
-        Map<Integer, Integer> locationMap = mapIndices(sourceMatchNumbers, targetMatchNumbers);
+        Map<Integer, Integer> locationMap = mapIndices(normSourceMatchNumbers, normTargetMatchNumbers);
 
-        // Substitute new numbers in the target match
+        // Substitute new numbers in the target match. The inserted number
+        // adopts the digit width of the target token it replaces, so the
+        // target match's digit convention is preserved.
         StringBuilder result = new StringBuilder();
         int i = 0;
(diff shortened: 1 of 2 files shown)


test/src/org/omegat/gui/matches/MatchesTextAreaTest.java
```diff
@@ -100,4 +101,112 @@ public void testReplaceNumbers() {
...
+     * adopt the digit width used by the target match. Feature request #1193.
+     */
+    @Test
+    public void testReplaceNumbersFullwidth() {
+        ITokenizer tok = new DefaultTokenizer();
+
+        // Reported case: full-width source number, half-width (Latin) target.
+        // The inserted number must be converted to half-width to match the target.
+        String source = "これは例文９です"; // これは例文９です
+        String srcMatch = "これは例文8です";    // これは例文8です
```
miurahr: Please don't use non-English comment.

The project coding style says comments should be in English:
src_docs/developer/30.CodingStyles.md “You should use English for comment language.”

The string literals themselves are fine because UTF-8 string literals are allowed, and these test values are necessary. However, the trailing Japanese comments do not add useful information and conflict with the English-comment guideline. They also make the tests noisier.
Suggested cleanup: keep the Unicode literals, but remove or replace the duplicated Japanese comments with English descriptions only

src/org/omegat/gui/matches/MatchesTextArea.java
```diff
@@ -301,20 +301,29 @@ static String substituteNumbers(String source, String sourceMatch, String target
         List<String> sourceNumbers = Stream.of(sourceTok.tokenizeVerbatimToStrings(source))
                 .filter(MatchesTextArea::isNumber).collect(Collectors.toList());
 
+        // Compare and map numbers on their width-normalized form so that
+        // full-width Arabic numerals (U+FF10-U+FF19, common in Japanese) are
+        // treated as equivalent to their half-width counterparts. See feature
+        // request #1193.
```
miurahr, line 307: ### Potentially misleading terminology in comments

The new code and tests repeatedly use “full-width Arabic numerals” and “Arabic-Indic digits” in close proximity. Technically this is understandable, but it may be confusing for future readers because:
０–９ are usually called full-width digits or full-width ASCII digits.
٠–٩ / ۰–۹ are commonly referred to as Arabic-Indic / Extended Arabic-Indic digits.
The commit message is clear enough, but in code comments it may be better to use the less ambiguous phrase full-width ASCII digits or full-width digits U+FF10–U+FF19.

miurahr (changes requested):
The code is good enough but there are several minor issue in comments.
Please clean-up the comments?

test/src/org/omegat/gui/matches/MatchesTextAreaTest.java
zs-stpa: ok

src/org/omegat/gui/matches/MatchesTextArea.java
zs-stpa: ok

zs-stpa:
ammending the commit per force push with the suggested clean up changes.

miurahr (approved):

PR #2114, opened 2026-07-13
fix(test): stop BundleTest from leaking the default locale

zs-stpa:
testUndefinedString did set Locale.ENGLISH without restoring it, making later tests in the same JVM (e.g. ConsoleTranslateTest) order-dependent. Using the existing LocaleRule fixture instead.


miurahr:
No need to do it. Bundle has a mechanism to fall back to English.
We use the expectation with `OStrings.getString()` not raw string.

miurahr:
It is better to do it. Thanks.

PR #2120, opened 2026-07-16
[6.1] docs: update L10N(ca,fr,nl,ja)

miurahr:
## Pull request type

## What does this PR change?

- Update UI/manual l10n in ca, fr, nl and ja
- Update readme_*.txt to replace Java 1.8 to Java 11
- Remove too old readme


miurahr:
Translations in sat is very limited and it seems almost untranslated status. I would like to drop sat from PR.

brandelune:
> Translations in sat is very limited and it seems almost untranslated status. I would like to drop sat from PR.

Ok. I’m fine with that. The group will have more time to complete the translation.

miurahr:
@brandelune is it ok for other languages?

miurahr:
I have updated Japenese translations.
Any other tasks?

PR #2121, opened 2026-07-18
fix(ui): correct misleading labels (SF-1516, SF-1265, SF-1731)

zs-stpa:
## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?

- rename 'Search in comments', it is actually searching in all properties
  * https://sourceforge.net/p/omegat/feature-requests/1516/
- Rename "Save" to "Export" in Autotext Configuration dialog
  * https://sourceforge.net/p/omegat/feature-requests/1265/
- Change label "Project translation" to "Committed target files"
  * https://sourceforge.net/p/omegat/feature-requests/1731/

## What does this PR change?

- Search window: renames the "In comments" option to "In properties"; the option actually searches segment properties (see `Searcher.searchProject()`), not comments.
- Auto-text configuration: renames the "Save..." button to "Export..."; it exports the entries to a file and does not save the settings. Mnemonics were chosen per language to stay conflict-free within the dialog.
- Team projects: uses "Committed target files" as the VCS commit message for target files instead of "Project translation".

## Other information

Both labels were corrected in the `en`, `de`, `es`, `fr`, `it`, `nl`, `pt`, `pt_BR`, `ru` and `us` bundles. The existing translations in the remaining languages are now factually wrong (they still say "In comments" / "Save..."); they are deliberately left to the respective l10n teams.


miurahr:
Please follow the PR description guide.

 Please provide a general summary of your changes in the title above -->

 Note: The OmegaT project uses GitHub pull requests for code review only.
The "real" code base is hosted on SourceForge; the GitHub project is a mirror.

If your PR is accepted it will be applied to the SourceForge repository by a
core contributor and the PR ticket will be closed. It will appear to have been
"closed without merging" but that is normal. 

## Pull request type

 Please try to limit your pull request to one type; submit multiple pull
requests if needed. 

- Bug fix -> [bug]
- Other (describe below)

## Which ticket is resolved?

Please refer to a relevant SourceForge ticket

Feature requests: https://sourceforge.net/p/omegat/feature-requests/

Bugs: https://sourceforge.net/p/omegat/bugs/

Documentation: https://sourceforge.net/p/omegat/documentation/ 

 Please paste a ticket title and URL  


- Title ex. fix...
  * URL ex. https://...
 
 Above block is used for changelog. 

## What does this PR change?

-
-
-

## Other information

 Any other information that is important to this PR, such as
before-and-after screenshots

miurahr:
@brandelune could you check a proposal of the UI labels?

brandelune:
> ## Pull request type
> * Bug fix -> [bug]
> 
> * rename 'Search in comments', it is actually searching in all properties
>   
>   * https://sourceforge.net/p/omegat/feature-requests/1516/
> * Rename "Save" to "Export" in Autotext Configuration dialog
>   
>   * https://sourceforge.net/p/omegat/feature-requests/1265/
> * Change label "Project translation" to "Committed target files"
>   
>   * https://sourceforge.net/p/omegat/feature-requests/1731/
> 
> * Search window: renames the "In comments" option to "In properties"; the option actually searches segment properties (see `Searcher.searchProject()`), not comments.
> * Auto-text configuration: renames the "Save..." button to "Export..."; it exports the entries to a file and does not save the settings. Mnemonics were chosen per language to stay conflict-free within the dialog.
> * Team projects: uses "Committed target files" as the VCS commit message for target files instead of "Project translation".
> 
> ## Other information
> Both labels were corrected in the `en`, `de`, `es`, `fr`, `it`, `nl`, `pt`, `pt_BR`, `ru` and `us` bundles. The existing translations in the remaining languages are now factually wrong (they still say "In comments" / "Save..."); they are deliberately left to the respective l10n teams.
> 

src/docs/manual/en/Windows_TextSearch.xml
```diff
@@ -221,9 +221,11 @@
             </listitem>
          </varlistentry>
          <varlistentry>
-            <term>In comments</term>
+            <term>In properties</term>
             <listitem>
-               <para>Search in comments attached to segments.</para>
+               <para>Search in the properties attached to segments, as displayed
+                  		  in the
+                  <link linkend="panes.segment.properties" endterm="panes.segment.properties.title"></link> pane.</para>
```
miurahr, line 228: OmegaT does not search in ID property which shown in properties pane, see RFE discussion. @brandelune could you help us to give better phrase?

miurahr: We can provide a condition like

> Search in the properties attached to segments (other than id),  as displayed  in the properties pane.

zs-stpa: i changed the label accordingly

miurahr (approved):

src/docs/manual/en/Windows_TextSearch.xml
brandelune: The problem stems from the fact that the ID property is not covered by the search, and we’re trying to find a short wording around that.

That’s not the proper way to handle the issue.

Either we remove the ID field from this box, or we add it to the search options. Anything else is a ugly hack.

brandelune (changes requested):

miurahr:
I do a fact check again.  `IParseCallback` has `addEntry` which is used by File Filters to register segment to core. 
```java
    void addEntryWithProperties(@Nullable String id, String source, @Nullable String translation, boolean isFuzzy,
                                String[] props, @Nullable String path, IFilter filter,
                                @Nullable List<ProtectedPart> protectedParts);

```
It handles `id` as a separated value from `props`.

In the `Searcher` class,
```java
            if (searchExpression.searchComments && properties != null) {
                // loop over values only, not keys.
                for (int i = 1; i <= properties.length; i = i + 2) {
                    if (searchString(properties[i], true)) {
                        propertyMatches = foundMatches.toArray(new SearchMatch[0]);
                        firstMatchedProperty = properties[i];
                        break;
                    }
                }
            }
```
When check the `Search in comments` ( now proposed a change to `In properties` ) check all the each properties.
We don't have `searchPath` or `searchID` or similar method/function/feature in the `Searcher`. 

in `SegmentPropertiesArea` class;
```java
    private void setKeyProperties(EntryKey key) {
        setProperty(KEY_FILE, key.file);
        setProperty(KEY_ID, key.id);
        setProperty(KEY_PATH, key.path);
    }
```

We handles `KEY_FILE`, `KEY_ID` and `KEY_PATH` as a special keys and values because these are a part of `EntryKey`.
It is an evidence `Searcher` does not look `id` for search target when check `search in comments/properties`.

src/docs/manual/en/Windows_TextSearch.xml
zs-stpa: i will add it to the search so the previous label text can stay.

zs-stpa: pushed. The ID is now included in the property search (it is appended to the searched properties exactly as the properties pane displays it), so the option label and the manual no longer need to talk around it. Manual updated accordingly, with a regression test.

zs-stpa:
You're right, thanks for the fact check. The pane shows file, id and path (the EntryKey group) and the property search covered none of them, so singling out id was inconsistent. I've extended it to all three via the existing value-only property loop, and moved the key names into `SegmentProperties` as shared constants — they were duplicated as private literals in `SegmentPropertiesArea` — so the pane and the search share one source. Manual and a test for the three fields updated. A field-specific search API in `Searcher` would be a separate, larger change if it's ever wanted.

zs-stpa:
visually inspected and tested search for properties with and without checkbox checked in a small sample project, looked okay.

Bildschirmfoto 2026-07-28 um 15 25 14

miurahr:
I think we need to take care other checkboxes of the Search Window.
We have "m_fileNamesCB" checkbox which has a label `SW_FILE_NAMES=file names`
that match a filename of external TM, or a memory entry with a matched source filename.

Another thing is uniquness of the ID and Path. Both are set by filter but we do not guarantee an uniqueness of the id.
It is why id and path can be searched, but not be as a JUMP indicator.

PR #2122, opened 2026-07-18
fix(filelist): release the closed project's file list

zs-stpa:
## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?

- No tracker ticket. The leak was found while testing repeated project reloads: after a few open/close cycles of large projects, OmegaT exhausts the default heap and shuts down with an out-of-memory error.

## What does this PR change?

- Closing a project only replaced the files table model of the Project Files window; the row sorter, the model field and the progress column's renderer kept referencing the `FileInfo` list — and through it the whole segment graph of the closed project. On reopening, the old and the new project were in memory at the same time.
- The close handler now drops the sorter, the model and the progress column, and the row context menu guards against the cleared state.
- `DirectoryMonitor.fin()` additionally interrupts the polling sleep (and the stop flag is now volatile), so the monitor threads — whose callbacks also reference the closing project — end immediately instead of up to a second later.
- Two `WeakReference`-based regression tests pin that closed projects become unreachable after close: one for the core data-engine cycle (`ProjectReloadLeakTest`), one for the GUI cycle with the editor and the Project Files window attached (`EditorProjectReloadLeakTest`).

## Other information

Review feedback: `modelFiles` and `currentSorter` are now annotated `@Nullable` and every reference site either guards against null or documents why it is safe; the getter that can return null is annotated as well.

### DIFF
--- a/src/main/java/org/omegat/gui/filelist/ProjectFilesListController.java
+++ b/src/main/java/org/omegat/gui/filelist/ProjectFilesListController.java
@@ -155,9 +155,9 @@ public class ProjectFilesListController implements IProjectFilesList {
     private static final int MIN_PROGRESS_WIDTH = 3;
 
     private ProjectFilesList list;
-    private FileInfoModel modelFiles;
+    private @Nullable FileInfoModel modelFiles;
     private AbstractTableModel modelTotal;
-    private Sorter currentSorter;
+    private @Nullable Sorter currentSorter;
     private @Nullable TableColumn filesProgressColumn;
     private @Nullable TableColumn totalsProgressColumn;
 
@@ -247,6 +247,14 @@ public class ProjectFilesListController implements IProjectFilesList {
         CoreEvents.registerProjectChangeListener(eventType -> {
             switch (eventType) {
             case CLOSE:
+                // Drop every reference to the closed project's file list: the
+                // sorter, the model and the progress column each hold the
+                // FileInfo list and with it the project's whole segment
+                // graph, which otherwise stays in memory until the next load.
+                list.tableFiles.setRowSorter(null);
+                currentSorter = null;
+                modelFiles = null;
+                filesProgressColumn = null;
                 list.tableFiles.setModel(new DefaultTableModel());
                 list.tableFiles.repaint();
                 modelTotal.fireTableDataChanged();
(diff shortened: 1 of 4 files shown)


miurahr:
Please follow the description guide.

    type of PR
    Fixed issue title and URL in raw
    explain the changes in detail

miurahr:
Good catch. I would like to review thoroughly.

src/main/java/org/omegat/gui/filelist/ProjectFilesListController.java
```diff
@@ -247,6 +247,14 @@ public void actionPerformed(ActionEvent e) {
         CoreEvents.registerProjectChangeListener(eventType -> {
             switch (eventType) {
             case CLOSE:
+                // Drop every reference to the closed project's file list: the
+                // sorter, the model and the progress column each hold the
+                // FileInfo list and with it the project's whole segment
+                // graph, which otherwise stays in memory until the next load.
+                list.tableFiles.setRowSorter(null);
+                currentSorter = null;
```
miurahr, line 255: You set `null` to `currentSorter`  and `modelFiles` so please add `@Nullable` at
```
    private @Nullable FileInfoModel modelFiles;
    private AbstractTableModel modelTotal;
    private @Nullable Sorter currentSorter;
```
and  guard the 21 places to reference the fields to check nullity or leave comments when it is safe.

src/main/java/org/omegat/gui/filelist/ProjectFilesListController.java
```diff
@@ -406,6 +414,10 @@ private void updateButtonState() {
     }
 
     private JPopupMenu createContextMenuForRow(int row) {
```
miurahr: This can return null, please add `@Nullable`

miurahr (changes requested):
It is good direction for storing `null` to "eliminate obsolete reference" as Effective Java (Josh Bloch) suggested. I worndering whether the properly check nullity in the other places in the class.

src/main/java/org/omegat/gui/filelist/ProjectFilesListController.java
zs-stpa: done.

miurahr (approved):

PR #2124, opened 2026-07-18
feat(gui): add open-folder buttons to project properties dialog (SF-1535)

zs-stpa:
## Pull request type

## Which ticket is resolved?

- Add "open folder/file" button for paths in the "File locations" part of the "Edit Project" window
  * https://sourceforge.net/p/omegat/feature-requests/1535/

## What does this PR change?

- Adds an Open button next to each of the seven resource fields in the Edit Project dialog (source, target, translation memory, glossary, writable glossary, dictionaries, exported TM); it shows the folder in the system file manager via the existing `ProjectUICommands.openFile` mechanism. For the writable glossary, which is a file path, the containing folder is opened.
- The buttons track the entered path live through a `DocumentListener` and are disabled while the field does not point to an existing directory.
- Because all seven buttons share the same short caption, each carries an explanatory tooltip and a distinguishable accessible name (built from the row label), so screen reader users can tell them apart.
- The English user manual (project properties dialog chapter) documents the new buttons.

### Screenshots

before, without the change:
unchanged-noOpen

after, including the change:
withChange-withOpen

### DIFF
--- a/src/docs/manual/en/Dialogs_ProjectProperties.xml
+++ b/src/docs/manual/en/Dialogs_ProjectProperties.xml
@@ -347,6 +347,20 @@
                   		  instead of the default resource folder.</para>
             </listitem>
          </varlistentry>
+         <varlistentry xml:id="dialogs.project.properties.file.locations.open">
+            <term xml:id="dialogs.project.properties.file.locations.open.title">
+               <guibutton>Open</guibutton>
+            </term>
+            <listitem>
+               <para>The
+                  <guibutton>Open</guibutton> button next to each
+                  <guibutton>Browse</guibutton> button opens that resource folder in
+                  		  the system file manager. For the writable glossary, it opens the
+                  		  folder that contains the glossary file.</para>
+               <para>The button is only available while the folder entered in the
+                  		  corresponding field exists.</para>
+            </listitem>
+         </varlistentry>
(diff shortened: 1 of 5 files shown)


brandelune:
> ## Feature
> Add "open folder/file" button for paths in the "File locations" part of the "Edit Project" window
> 
> ## Ticket
> https://sourceforge.net/p/omegat/feature-requests/1535/
> 
> ### Changes
> Each of the seven directory fields in the project properties dialog (source, target, translation memory, glossary, writeable glossary, dictionaries, exported TM) gets an **Open** button that shows the folder in the system file manager, using the existing `ProjectUICommands.openFile` mechanism.
> 
> The buttons track the entered path live via a `DocumentListener` and are disabled while the field points to a directory that does not exist, so they can never fail on a stale path.
> 
> New bundle keys are provided in English and German; other languages fall back to English until their l10n teams pick them up.
> 
> ### Testing
> Full test suite passes, including `BundleTest` for the new keys. Visual check done in a live session: all seven buttons present, they open the correct folders, and the enable/disable logic follows edits to the path fields immediately.
> 

miurahr (approved):

brandelune (approved):

miurahr:
Thank you for your valuable contribution! I would like to add your name in the releases/contributors.txt later.

PR #2127, opened 2026-07-18
feat(aligner): remember window, parameters, languages and directories

zs-stpa:
## Pull request type

## Which ticket is resolved?

- Aligner to remember window position, size, and values of parameters.
  * https://sourceforge.net/p/omegat/feature-requests/1456/
- Remember aligned TMX save location
  * https://sourceforge.net/p/omegat/feature-requests/1358/

## What does this PR change?

- The aligner window remembers its position and size between sessions (via the existing `StaticUIUtils.persistGeometry` mechanism); it is centered on the parent window only on the very first run.
- The align parameters (algorithm, calculator, counter type, segmenting, remove tags) are persisted and restored. The comparison mode is deliberately not persisted because the available modes depend on the files being aligned.
- The languages chosen in the file picker are persisted; they take precedence over the locales of the last created project when no project is open (those remain the first-use fallback).
- The source and target file choosers reopen in the directories last used.
- The save dialog offers the directory of the last saved TMX.
- Hardens the `Aligner` constructor against an empty non-null filters configuration from the preferences store, which previously left the `FilterMaster` unable to parse any file (a latent bug that surfaced as an order-dependent test interaction).
- Covered by eight test cases in the new `AlignSettingsPersistenceTest`.

### DIFF
--- a/src/docs/manual/en/Windows_Aligner.xml
+++ b/src/docs/manual/en/Windows_Aligner.xml
@@ -42,6 +42,11 @@
          use the languages from that project, as well the project-specific
          segmentation rules, if any.</para>
    </note>
+   <para>The aligner remembers its window position and size, the align
+      parameters, the selected languages, and the folders last used for the
+      input files and the saved TMX file between sessions. When a translation
+      project is open, its languages take precedence over the remembered
+      ones.</para>
    <para>After selecting the language and files, click 
       <guibutton>OK</guibutton>
       button to bring up the 
--- a/src/docs/manual/en/Windows_Aligner.xml
+++ b/src/docs/manual/en/Windows_Aligner.xml
@@ -42,7 +42,7 @@
          use the languages from that project, as well the project-specific
          segmentation rules, if any.</para>
    </note>
-   <para>The aligner remembers its window position and size, the align
+   <para>The aligner remembers its window position and size, the alignment
       parameters, the selected languages, and the folders last used for the
       input files and the saved TMX file between sessions. When a translation
       project is open, its languages take precedence over the remembered
--- a/aligner/src/main/java/org/omegat/gui/align/AlignFilePickerController.java
+++ b/aligner/src/main/java/org/omegat/gui/align/AlignFilePickerController.java
@@ -81,9 +81,84 @@ public class AlignFilePickerController {
     private static final ResourceBundle BUNDLE = ResourceBundle.getBundle("org.omegat.gui.align.Bundle");
 
     public AlignFilePickerController() {
+        // Preselect the languages of the previous aligner session (feature
+        // request #1456). Callers overwrite them through the setters when a
+        // project is open.
         List<Language> allLangs = Language.getLanguages();
-        sourceLanguage = allLangs.get(0);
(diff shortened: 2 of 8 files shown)


miurahr:
It is ok to use Prerefence system in Aligner module, but defines constants in Preferneces is not good because Aligner is separete module sub project.

brandelune:

As mentioned in earlier exchanges. Either you intellectually own your code, you are able to understand it, explain it, mend it, and we don’t care which tool you used, or you can’t, and we won’t accept your code.

We don’t require people to write whether they used a semi-automated IDE with powerful code completion or whatever feature, and we don’t require either that they display approved qualifications in Java or whatnot.

For that matter, you could write code with assistance from a friend and that would be fine with us *as long as you understand and can explain and mend it* when required. And we would not need to know which path you used to write that code.

So, if the following sentence:

means that you don’t want to take intellectual responsibility for your code, regardless of how trivial the code is I will refuse it.

If it means the same thing as “I used VS Code to produce this code”, then it is irrelevant to the PR and I’m asking you to remove this sentence.

Either you represent yourself as somebody who can be trusted and we trust you, or you don’t, and you can go work on your fork.

brandelune:
> the controversial phrase has been removed from the commit messages. it was originally intended for transparency. the phrase remains in the descriptions of the PR's and can be removed from there as well if so desired,

Please remove it.

zs-stpa:
> > the controversial phrase has been removed from the commit messages. it was originally intended for transparency. the phrase remains in the descriptions of the PR's and can be removed from there as well if so desired,
> 
> Please remove it.

aye, on it. although, a few commits with this message have already been accepted, too late to do anything about them.

miurahr (approved):

src/docs/manual/en/Windows_Aligner.xml
```diff
@@ -42,6 +42,11 @@
          use the languages from that project, as well the project-specific
          segmentation rules, if any.</para>
    </note>
+   <para>The aligner remembers its window position and size, the align
```
brandelune: → the alignment
parameters

brandelune (changes requested):

src/docs/manual/en/Windows_Aligner.xml
zs-stpa: yes ok and sry i overlooked your remark, feel free to do such trivial edits yourself. [exkurs: der berühmte Modellkollaps … we (humanity) now still have living memory of the time before AI but the generation that grows up now (your students) will be exposed to such an enormous stream of AI generated things not only text that I wonder how that will affect spoken languages and patterns of thinking in general and that affects the available training material for the next wave of LLM tribes. Yes sure off topic. but when the majority of speechsamples in the world will soon be machine generated not human generated, then that will certainly change .. well it will change everything, we are living the change right now, life is change by definition ....... damnable philisophical rabit holes truly sorry you have to read this but here it is.   Lange Rede kurzer Sinn (long speech, short point) – it would in the long run be much mor efficient to learn how to correct the models themselbers instead of correcting single mistakes in the output of the models but that is a collective task that humanity is quite obiously declining and leaving to the few like us …

PR #2128, opened 2026-07-18
feat(search): add source concordance search to Edit menu (SF-1162)

zs-stpa:
## Pull request type

## Which ticket is resolved?

- Shortcut for direct search for source text
  * https://sourceforge.net/p/omegat/feature-requests/1162/

## What does this PR change?

- Adds an Edit menu entry "Search for Source Segment" (Ctrl/Cmd+Shift+E) that opens the Search window pre-filled with the source text of the current segment and runs the search immediately.
- The search scope is forced to source text within the whole project, so the concordance results are shown without any further clicks.
- An already open search window is reused instead of opening a new one; the menu entry is disabled while no project is loaded.


brandelune:
> ## Pull request type
> 
> * Shortcut for direct search for source text
>   
>   * https://sourceforge.net/p/omegat/feature-requests/1162/
> 
> * Adds an Edit menu entry "Search for Source Segment" (Ctrl/Cmd+Shift+E) that opens the Search window pre-filled with the source text of the current segment and runs the search immediately.
> * The search scope is forced to source text within the whole project, so the concordance results are shown without any further clicks.
> * An already open search window is reused instead of opening a new one; the menu entry is disabled while no project is loaded.
> * New bundle keys are provided in English and German; other languages fall back to English.
> 
> ## Other information

brandelune:
The comments in the RFE were pretty interesting. Especially the ones related to using the same shortcut and autofilling with Source when nothing was selected.

zs-stpa:
> The comments in the RFE were pretty interesting. Especially the ones related to using the same shortcut and autofilling with Source when nothing was selected.

yes i scrolled through. possible follow-ups later.

miurahr (changes requested):
RFE#1162 discussion has reached a rough concensus to "Default-to-off" and add preference to option to enalbe the feature among lead developer and project manager at that time.

Please implement preference to enable the feature and preference GUI to select the option at least.

RFE#1162 discussion has a concern about the preference options already so busy. When you want to merge the #2128, it is required to propose a solution of the user preference UI improvement.

zs-stpa:
rather than solving only this one single user preference ui/ux problem, my response to your request would be something along the following lines:

it was mentioned on the ticket also by other devs, that it is not an easy task, to present options settings in an easy to use easy to understand simple intuitive way while at the same time having many many many different options to enable every possible specialized use case – this is a task i can not imagine solving in this PR alone, and a task that grows with every added new optional setting.
i'd like to just add the option anywhere it seems fit for now, like maybe preferences -> editor -> new checkbox or radio items to determine default search behaviour. but since this is a global setting, i would also like to add a project override. the gui for that would be logical to put into project settings.
which is another big problem: project settings should in my opinion mirror global settings with an added project-override checkbox for every setting, effectively i think we need triple the global settings: one set is global settings for one computer user home directory and default for new projects. second set is per project per setting a boolean checkbox whether to use project override for this setting. third set is the mirrored first set of settings for one project. that third set travels with team projects also and is potentially distributed to different released version numbers with different capabilities/features/bugs which needs to be accounted for. different settings of a single team member needs to be negotiated per setting (keep local setting VS. use team setting VS. share local setting with team) from which follows user permissions requirements who is allowed to set project settings etc. a cascade of new settings…
so as you see this all leads very far away from the initial ticket, and has lots of other relevant tickets asking for things like configurable button bar or rearrangable menus or other generic UI/UX improvements.

zs-stpa:
> RFE#1162 discussion has reached a rough concensus to "Default-to-off" and add preference to option to enalbe the feature among lead developer and project manager at that time.
> 
> Please implement preference to enable the feature and preference GUI to select the option at least.
> 
> RFE#1162 discussion has a concern about the preference options already so busy. When you want to merge the #2128, it is required to propose a solution of the user preference UI improvement.

ok adding a preference will be easily done. the hard part of preference UI improvements, i would say is out of scope in this PR, there are other tickets and other generic ways of improving UI, to follow as separate PRs.

brandelune:
> it was mentioned on the ticket also by other devs, that it is not an easy task, to present options settings in an easy to use easy to understand simple intuitive way while at the same time having many many many different options to enable every possible specialized use case

We don’t need to *present option settings* in an easy to understand [etc.] way.

We need to have an elegant way to set parameters. Buttons and checkboxes all over the place are not an elegant way to solve this issue.

And we need to document that way so that is *feels* easy.

> – this is a task i can not imagine solving in this PR alone, and a task that grows with every added new optional setting.

It certainly should not.

> i'd like to just add the option anywhere it seems fit for now, like maybe preferences -> editor -> new checkbox or radio items to determine default search behaviour. but since this is a global setting, i would also like to add a project override.

This setting is *not* an important setting. We do not need to provide a GUI for each and every setting. We need reasonable defaults based on what experienced (not "experts") users need and an elegant way to tweak the system.

> the gui for that would be logical to put into project settings. which is another big problem: [...]

Based on what exists and what users are used to (some settings in the menus, most in the preference dialogs, files in `omegat`, separate configuration folders), we need to carefully think of an elegant way to cascade all those settings.

brandelune:
I’ve read the RFE comments again, and I don’t see why we need to have a different setting for local vs general configurations.

Aaron’s opinion re. the user who’d be surprised of having a default value, is fair, but if we do as Yu-Tang suggests : the field is selected so any user input overrides the default, then we’re good.

All the comments in that thread come from very experienced users, two were OmegaT project managers, one wrote a book about OmegaT and Kos is a plugin author.

My conclusion is the following : we create a new default with the current source as content. This default is properly documented as "concordance search" (whatever that means) and gets a checkbox in the Search/Replace dialogs (the current layout of preferences in those dialogs is ugly, we need to improve that).

That preference is a global preference. Just like every search preference, once (dis)abled they apply to the next search. There is absolutely no practical need to split search preferences between local/general/team scopes.

zs-stpa:
okay i am amending the preselection of the inserted text thing because i forgot that earlier and reworked it before finding new response here.

no project setting. i have added a checkbox in global preferences > editor > (see screenshot, the lowest checkbox at the bottom)
Bildschirmfoto 2026-08-31 um 22 34 00

and also an unobtrusive gear button top right corner of the search window
Bildschirmfoto 2026-08-31 um 22 33 24

where i have put the the same checkbox again for easier access from the search window itself
Bildschirmfoto 2026-08-31 um 22 33 40

and copied some preexisting menu entries there as well.

as for cascading settings with project and team - overkill in this case, it stays a general user preference.

as for elegance in presentation - yeah, i'd like to see that as well and open to suggestions ^^ i think i'll look at some button bar or other ui-improvement tickets for ideas next

brandelune:
> okay i am amending the preselection of the inserted text thing because i forgot that earlier and reworked it before finding new response here.

Have you read what I wrote just 8 hours ago ?

**My conclusion is the following : we create a new default with the current source as content. This default is properly documented as "concordance search" (whatever that means) and gets a checkbox in the Search/Replace dialogs (the current layout of preferences in those dialogs is ugly, we need to improve that).**

**That preference is a global preference. Just like every search preference, once (dis)abled they apply to the next search. There is absolutely no practical need to split search preferences between local/general/team scopes.**

zs-stpa:
yes i have read it just now when committing what i worked over the day, sorry for asynchronous communication, is what i tried to say with the above "reworked it before finding new response here."

and i'll reread it again tomorrow fresh and clear, alright, good night ^.^

zs-stpa:
> > okay i am amending the preselection of the inserted text thing because i forgot that earlier and reworked it before finding new response here.
> 
> Have you read what I wrote just 8 hours ago ?
> 
> **My conclusion is the following : we create a new default with the current source as content. This default is properly documented as "concordance search" (whatever that means) and gets a checkbox in the Search/Replace dialogs (the current layout of preferences in those dialogs is ugly, we need to improve that).**
> 
> **That preference is a global preference. Just like every search preference, once (dis)abled they apply to the next search. There is absolutely no practical need to split search preferences between local/general/team scopes.**

and seems so be exactly what i did there, so .. unless i misinterpreted something, we seem to be in agreement maybe?

brandelune:
> and seems so be exactly what i did there, so .. unless i misinterpreted something, we seem to be in agreement maybe?

You have implemented what you describe in your screenshots, not what I wrote.

- we need **one** checkbox in the Search (and Replace) dialogs to set the search field default to "empty" when there is no selection (the default is now "source segment")
- we need to explain the new default in the manual
- we don’t need any other settings, submenus or whatnot.

I don’t see how _that_ reads as what you implemented.

zs-stpa:
okay there is one new checkbox. it is right below the improvement to the layout which is the gear button top right corner of the search window where it was unobtrusive still enough space for a tiny little gearbutton. no new menu, just copy of preexisting menu items. no project setting, no team negotiation, the default documents itself as the clearly visible new checkbox in the settings … and i can add a line to the manual, when all is finalized.

brandelune:
We don’t need this :
Screen Shot 2026-08-31 at 23 36 42

We don’t need this :
Screen Shot 2026-08-31 at 23 37 18

We don’t need this :
Screen Shot 2026-08-31 at 23 37 59

We only need a checkbox in the Search diialog and the same in the Replace dialog that says something like "Search for source segment", that is unchecked (to keep backward compatibility, for now) and a line in the manual that says something like : 

"When no text is selected, the search field is automatically filled with the source segment contents."

That’s what the RFE was about.

zs-stpa:
Oh Kay - now i see, that is also one way to interprete the ticket.

instead of two similar but slightly different functions existing in parallel (one legacy cmd+f as always plus one new cmd+shift+e), you prefer to make one function switchable with a new checkbox, that needs to be added to an already cluttered window, and at the same time you want me to improve the layout and reduce messiness of settings by adding it where the mess is.

well it is easy to change again, all the building blocks are there, just need to reposition checkbox and rewire the source-insertion+selection-behaviour to cmd+F, can do, will do, easy.

my demessing strategy would have been to move more of the settings / options to gear like auxilliary menus and they should get hover effects instead of clicks - like hovering for tooltips but getting and expanded set of clickable options and settings. this would free up screen space and it is not forbidden to repeat copies of same setting at different places in the gui whereever that setting applies so that settings can be managed directly where they apply - not instead of dedicated settings page, but in addition, hence copies of same settings. this aids discoverability. think of a light switch. you do not put one single light switch in a big room, you put them at every door that enters the room, and they all control the same light setting.

and i see lots of negatives - not like this, not like that, dont need this, dont need that, and that is totally fine and okay, and if you know positively where we want that checkbox exactly then i am sure you would tell.

brandelune:
> Oh Kay - now i see, that is also one way to interprete the ticket.

That’s something we could have found out by discussing the implementation before you started coding.
 
> instead of two similar but slightly different functions existing in parallel (one legacy cmd+f as always plus one new cmd+shift+e), you prefer to make one function switchable with a new checkbox,

That’s correct.

> that needs to be added to an already cluttered window, and at the same time you want me to improve the layout and reduce messiness of settings by adding it where the mess is.

No. I suggested that it’s something that we need to do.
 
> well it is easy to change again, all the building blocks are there, just need to reposition checkbox and rewire the source-insertion+selection-behaviour to cmd+F, can do, will do, easy.

Thank you.
 
> my demessing strategy would have been to move more of the settings / options to gear like auxilliary menus and they should get hover effects instead of clicks

We don’t need to add a new GUI paradigm now. We can discuss that on the dev list, see where and how the/a new paradigm could be implemented, etc. but that can’t happen in the implementation of a simple RFE. I think that’s pretty obvious. And this comment goes for the color/template modification. 

> this would free up screen space

We can use columns:

Type  | Location | Status | Display

That will reduce the mess by adding structure.

> and i see lots of negatives - not like this, not like that, dont need this, dont need that, and that is totally fine and okay, and if you know positively where we want that checkbox exactly then i am sure you would tell.

I’ve used OmegaT as my only work tool for close to 25 years now. I have a pretty good understanding of our history (I was there at the beginning) and of what we had in mind then and later on, until I became project manager.

It is not easy to share 25 years of experience, but our principle remains : the less clutter possible, the fewer buttons possible, everything accessible through the keyboard (as far as possible), a clear and homogenous (G)UI that requires the less documentation possible. You get the drift.

Also, I _work_ with the master branch that I regularly build. I’m ok with bugs in new features, but regressions are not acceptable. Developing OmegaT means two things : improving the performance of users by improving its "under the hood" performance, making features discoverable, simplifying the workflow, the labels and the documentation, and drastically filtering RFEs by carefully considering their impact on the overall experience and the way they fit in the current (G)UI paradigms.

OmegaT has a reputation for being simple, efficient and robust. We need to stick to that.

brandelune:
The table I suggested above could be structured this way:

```
        Location                   |   Type       Options           Scope          Status          Display     |
 ○ Project     ○ Files             | ○ Exact    □ Whole words     □ Source       ○ Translated    □ All matches |
 □ Memory      Path [   ] [Select] | ○ Keyword  □ Case sensitive  □ Translation  ○ Untranslated  □ File paths  |
 □ References  □ Recursive         | ○ Regex    □ [Space]         □ Notes        ○ Both                        |
 □ Glossaries                      |            □ Source          □ Properties                                 |
```

That’s 5 lines for ~100 characters. That fits in the current window.

There are probably better ways to handle this. I’m not a UI magician. But most of those options are not modified over searches, so we could even have a drawer like the current [TM Search Options], to hide them when they’re not meant to be modified:
```
[search field ............................................] [Search]
[Search Options]
[TM parameters]
```

Or even separate drawers:

```
 [search field ............................................] [Search]
 [ Location]  [Type]  [Options]  [Scope]  [Status]  [Display]
 [TM parameters]
```

zs-stpa:
> The table I suggested above could be structured this way

this some good looking ascii art. maybe you can propose actual gui mockup as a separate PR.

PR #2129, opened 2026-07-18
feat(search): add find previous/next buttons to the Search window

zs-stpa:
## Pull request type

## Which ticket is resolved?

- Add a "Find Previous" button to the Search window
  * https://sourceforge.net/p/omegat/feature-requests/1380/
- Find Next button on Search window
  * https://sourceforge.net/p/omegat/feature-requests/1125/

## What does this PR change?

- Adds Find Previous and Find Next buttons below the search results that walk through the results, wrapping around at both ends. F3 and Shift+F3 trigger the buttons from anywhere in the Search window; the buttons are disabled while there are no results.
- The buttons always show the reached result in the editor, regardless of the auto-sync option: like the existing double-click on a result, an explicit navigation command overrides the passive-sync setting. Unlike double-click they do not bring the main window to the front, so the results can be walked through without losing focus.
- The keyboard navigation inside the results list keeps its previous behavior (stops at the ends, follows the auto-sync option).

### Screenshots

before without the change:
unchanged-noPrevNext

after, including the proposed change:
withChange-withPrevNext

### DIFF
--- a/src/docs/manual/en/Windows_TextSearch.xml
+++ b/src/docs/manual/en/Windows_TextSearch.xml
@@ -382,6 +382,21 @@
             </listitem>
          </varlistentry>
       </variablelist>
+      <para>The
+         <guibutton>Find Previous</guibutton> and
+         <guibutton>Find Next</guibutton> buttons below the results, as well as
+         <keycombo>
+            <keycap>S</keycap>
+            <keycap>F3</keycap>
+         </keycombo> and
+         <keycap>F3</keycap> from anywhere in the window, move the selection one
+         	result backward or forward, wrapping around at both ends of the list.
+         	The reached segment is always displayed in the
+         <link linkend="panes.editor" endterm="panes.editor.title"></link> pane,
+         	regardless of the
+         <guilabel>Auto-sync with editor</guilabel> option, without moving the
+         	focus away from the Search window. When the navigation wraps around,
+         	a short note appears next to the number of results.</para>
       <para>The selected segment is highlighted in green:</para>
       <example xml:id="select.second.match">
          <title xml:id="select.second.match.title">Select the second match</title>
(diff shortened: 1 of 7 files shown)


miurahr (approved):
It looks good that the logic in the `SearchWindowController` simply perform the introduced `EntryListPane#selectNextEntry` or `EntryListPane#selectPreviousEntry`.
Even better, introduce keyboard shotcuts (F3 and shift plus F3) -- @brandelune could you check a consistency among the application?
New feature to show status in `resultsLabel`
I want to wait a check by other reviewer for the UI and message.

brandelune:
> ## Pull request type
> 
> * Add a "Find Previous" button to the Search window
>   
>   * https://sourceforge.net/p/omegat/feature-requests/1380/
> * Find Next button on Search window
>   
>   * https://sourceforge.net/p/omegat/feature-requests/1125/
> 
> * Adds Find Previous and Find Next buttons below the search results that walk through the results, wrapping around at both ends. F3 and Shift+F3 trigger the buttons from anywhere in the Search window; the buttons are disabled while there are no results.
> * The buttons always show the reached result in the editor, regardless of the auto-sync option: like the existing double-click on a result, an explicit navigation command overrides the passive-sync setting. Unlike double-click they do not bring the main window to the front, so the results can be walked through without losing focus.
> * The keyboard navigation inside the results list keeps its previous behavior (stops at the ends, follows the auto-sync option).
> * New bundle keys are provided in English and German; other languages fall back to English.
> 
> ## Other information

brandelune:
**This feature is the wrong feature.**

https://sourceforge.net/p/omegat/feature-requests/1380/#b1e4

It was not requested in the Search window where we already have a C+N/P to move around. It was requested in the Replace window.

We eventually need a navigation menu with indicated shortcuts, but please no additional buttons.

zs-stpa:
oops sorry, seems i did not read that clearly, we could add the same to the replace-window, but okay no more extra buttons.

brandelune:
https://github.com/omegat-org/omegat/pull/2303#issuecomment-5474187573

brandelune:
This UX feature was never approved by me and I explicitly requested modifications.

Please remove it from `master`.

miurahr:
@brandelune I just raised PR to revert the merge of the PR#2129.

PR #2130, opened 2026-07-18
feat(search): add whole words only option to the Search window (SF-849)

zs-stpa:
## Pull request type

## Which ticket is resolved?

- Add "whole words only" option to Find dialog
  * https://sourceforge.net/p/omegat/feature-requests/849/

## What does this PR change?

- Adds a "Whole words only" checkbox to the search options of the Search window.
- The option applies to exact and keyword searches by anchoring the search terms with lookarounds on Unicode word boundaries (`UNICODE_CHARACTER_CLASS`, so it works for non-ASCII scripts as well).
- In a regular expression search the user controls word boundaries in the expression itself, so the checkbox is disabled there.
- The checkbox state is remembered between sessions.
- New bundle keys are provided in English and German; other languages fall back to English. The German label carries no mnemonic because all its letters are already taken by other options in the Search window.
- Covered by new test cases in `SearcherTest`.

### Screenshots

before, without the change:
unchanged-withoutGanzeWorte

after, including the proposed new checkbox:
withChange-nurGanzeWorte

edit: made the change more visable

### DIFF
--- a/src/main/java/org/omegat/core/search/Searcher.java
+++ b/src/main/java/org/omegat/core/search/Searcher.java
@@ -274,9 +274,6 @@ public class Searcher {
         entryMap.clear();
         matchers.clear();
 
-        // determine pattern matching flags
-        int flags = searchExpression.caseSensitive ? 0 : Pattern.CASE_INSENSITIVE | Pattern.UNICODE_CASE;
-
         // Normalize width of search string if width insensitivity is requested.
         // Then, instead of modifying the regex, we also normalize the
         // comparison strings later on.
@@ -294,16 +291,23 @@ public class Searcher {
             // expression
             textSearchExpression = StaticUtils.globToRegex(textSearchExpression,
                     searchExpression.spaceMatchNbsp);
+            if (searchExpression.wholeWordsOnly) {
+                textSearchExpression = anchorWholeWords(textSearchExpression);
+            }
 
             // create a matcher for the search string
-            matchers.add(Pattern.compile(textSearchExpression, flags).matcher(""));
+            matchers.add(Pattern.compile(textSearchExpression, getPatternFlags()).matcher(""));
             break;
         case KEYWORD:
             // break the search string into keywords,
             // each of which is a separate search string
+            final int flags = getPatternFlags();
             Pattern.compile(" ").splitAsStream(textSearchExpression.trim()).filter(word -> !word.isEmpty())
                     .map(word -> {
                         String glob = StaticUtils.globToRegex(word, false);
+                        if (searchExpression.wholeWordsOnly) {
+                            glob = anchorWholeWords(glob);
+                        }
                         return Pattern.compile(glob, flags).matcher("");
                     }).forEach(matchers::add);
             break;
@@ -315,7 +319,7 @@ public class Searcher {
             }
 
             // create a matcher for the search string
-            matchers.add(Pattern.compile(textSearchExpression, flags).matcher(""));
+            matchers.add(Pattern.compile(textSearchExpression, getPatternFlags()).matcher(""));
             break;
         default:
             throw new IllegalStateException("Unknown search expression type");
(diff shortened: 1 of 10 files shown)


miurahr:
Thank you for valueable contribution.

src/main/java/org/omegat/core/search/Searcher.java
```diff
@@ -274,7 +274,7 @@ public void search() throws IOException {
         matchers.clear();
 
         // determine pattern matching flags
-        int flags = searchExpression.caseSensitive ? 0 : Pattern.CASE_INSENSITIVE | Pattern.UNICODE_CASE;
+        final int flags = getPatternFlags();
```
miurahr: There seems no reason to add `final` to the `flags` auto variable.
If you are guantee no change during streams processing in the case KEYWORD, it is better to put `final int flags` just before the stream processing, and replace flags in case EXACT with a method call.

miurahr (commented):
There is a small place to improve a readability-of-the-code.
In a same sense, you are recommended to execute gradle task `spotlessChangedApply`

brandelune:
> ## Pull request type
> 
> * Add "whole words only" option to Find dialog
>   
>   * https://sourceforge.net/p/omegat/feature-requests/849/
> 
> * Adds a "Whole words only" checkbox to the search options of the Search window.
> * The option applies to exact and keyword searches by anchoring the search terms with lookarounds on Unicode word boundaries (`UNICODE_CHARACTER_CLASS`, so it works for non-ASCII scripts as well).
> * In a regular expression search the user controls word boundaries in the expression itself, so the checkbox is disabled there.
> * The checkbox state is remembered between sessions.
> * New bundle keys are provided in English and German; other languages fall back to English. The German label carries no mnemonic because all its letters are already taken by other options in the Search window.
> * Covered by new test cases in `SearcherTest`.
> 
> ## Other information

src/main/java/org/omegat/core/search/Searcher.java
zs-stpa: okay i have changed it accordingly.

zs-stpa:
Spotless overhead (all 8 branches)
```
┌─────────┬───────┬─────────────────────────────┐
│ Branch  │ Files │         Line churn          │
├─────────┼───────┼─────────────────────────────┤
│ labels  │ 0     │ —                           │
│ 1684    │ 1     │ +5/−4                       │
│ 1535    │ 2     │ +24/−26                     │
│ aligner │ 6     │ +29/−26                     │
│ 1162    │ 4     │ +30/−27                     │
│ 1380    │ 3     │ +181/−116                   │
│ 849     │ 5     │ +228/−149                   │
│ leak    │ 4     │ +26/−29                     │
├─────────┼───────┼─────────────────────────────┤
│ Total   │ 25    │ +523/−377 (about 900 lines) │
└─────────┴───────┴─────────────────────────────┘
```
Why so much: The task does not format "only your changes", as the docs claim. It reformats every file the branch touched, in full — including upstream code never edited PR-side. Example: added one line to Preferences.java, and Spotless rewrapped javadoc comments all over that file.

Problem: In the NetBeans-generated SearchWindowForm.java files (branches 1380 and 849), Spotless wraps the marker comment // <editor-fold ... //GEN-BEGIN:initComponents onto two lines. That breaks the GEN-BEGIN marker that the NetBeans Form Editor needs — the same marker the project's own coding-style docs tell us never to touch. So upstream's own formatting task damages upstream's own generated-code convention.

Two upstream bugs found along the way:
1. spotlessChangedApply is completely broken with the configuration cache that upstream enabled recently (#2054). It only runs with --no-configuration-cache.
2. The GEN-BEGIN damage described above.

zs-stpa:
i will handle the bugs with separate PR's but not apply spotless to the existing ones if that is ok?
edit: see https://github.com/omegat-org/omegat/pull/2152

zs-stpa:
the change in the force push was made from filler31 to filler33 so as to minimize mergeconflicts. noticed when integrating all my open PRs locally.

brandelune:
The UI changes are not very clear in the screenshots, since I don’t read German.

zs-stpa:
i have updated the sample screenshot accordingly.

miurahr (approved):

PR #2133, opened 2026-07-19
chore(gradle): enhance `genMac` task with refined inputs/outputs and …

miurahr:
…safer file operations

This is a fix for latest weekly build error.

## Pull request type

## Which ticket is resolved?
none
## What does this PR change?

- Introduce `InjectedOps` interface for cleaner dependency injection and copy localized resources in the task
- Enhance the task with mac-specific JVM property and version
- Enhance the task to tweak executable permission
- Simplify InstallTask

### DIFF
--- a/build-logic/src/main/groovy/org.omegat.mac-conventions.gradle
+++ b/build-logic/src/main/groovy/org.omegat.mac-conventions.gradle
@@ -1,4 +1,7 @@
+import java.nio.file.Files
 import java.util.zip.ZipFile
+
+import org.omegat.gradle.InjectedOps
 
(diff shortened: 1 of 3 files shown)


build-logic/src/main/groovy/org.omegat.mac-conventions.gradle
```diff
@@ -1,5 +1,8 @@
+import java.nio.file.Files
```
miurahr: Unused import
```suggestion
```

miurahr (commented):
There is a small fix.

PR #2140, opened 2026-07-20
chore(core): add `@Nullable` annotations for improved null-safety in core data APIs

miurahr:
## Pull request type
refactor
## Which ticket is resolved?
none
## What does this PR change?

- Add @Nullable annotations
-
-


miurahr:
I revisit here and review by myself. This should not lead any behavior change.

PR #2152, opened 2026-07-21
fix(build): branch workflow task compatibility with configuration cache (SF-1801)

zs-stpa:
## Pull request type

- Bug fix -> [bug] (build tooling only, so [build/release] may fit as well)

## Which ticket is resolved?

- changedOnBranch and spotlessChangedApply fail with the configuration cache; git remote detection is broken
  * https://sourceforge.net/p/omegat/bugs/1801/

## What does this PR change?

- Moves the git queries of `changedOnBranch` out of the task actions to configuration time, guarded so they only run when `changedOnBranch` or `spotlessChangedApply` was actually requested. Their results become configuration inputs, which makes both tasks compatible with the configuration cache (enabled by default in gradle.properties).
- Fixes the git remote detection: the previous code compared an `ExecResult` object with the integer `0`, which is always false, so `branch.<name>.pushRemote` / `branch.<name>.remote` were never honored and the task always fell back to the remote name `origin`. The comparison now uses `exitValue`.
- The changed-files list is computed once and shared by `changedOnBranch` (printing) and the spotless java target, instead of being recomputed and assigned in a `doFirst` block.

## Other information

Manual verification on this branch (no automated tests exist for build-logic):

- `./gradlew changedOnBranch` — BUILD SUCCESSFUL with the configuration cache enabled; prints the files changed on the branch.
- Second invocation: "Reusing configuration cache." with correct output.
- After committing a new file: "Calculating task graph as configuration cache cannot be reused because output of the external process 'git' has changed." — the new file is listed. So the cached git results are correctly invalidated when the branch content changes (Gradle treats `providers.exec` output at configuration time as a configuration input).
- `./gradlew spotlessChangedApply` with an unformatted probe `.java` file committed on the branch: the probe file is reformatted, files not changed on the branch are untouched.
- Remote detection: on a branch whose `branch.<name>.remote` is not `origin`, the configured remote is now used and the "Assuming 'origin'" warning no longer appears.

build-logic currently has no test infrastructure at all. If the project is interested, I would be happy to contribute a Gradle TestKit based functional test setup for these tasks (temporary git repository fixture, assertions on the behaviour above) as a follow-up PR.

### DIFF
--- a/build-logic/src/main/groovy/org.omegat.spotless-conventions.gradle
+++ b/build-logic/src/main/groovy/org.omegat.spotless-conventions.gradle
@@ -5,35 +5,47 @@ plugins {
     id 'org.omegat.common-utilities'
 }
 
-tasks.register('changedOnBranch') {
-    description = 'Lists files that have been modified on this git branch.'
-    doLast {
-        def branch = providers.exec {commandLine("git", "name-rev", "--name-only", "HEAD")}
-                .standardOutput.asText.get().trim()
+// The git queries run at configuration time, and only when one of the branch
+// workflow tasks was requested: their results become configuration inputs,
+// which keeps the tasks compatible with the configuration cache.
+def branchWorkflowRequested = gradle.startParameter.taskNames.any {
+    def taskName = it.tokenize(':').last()
+    taskName == 'changedOnBranch' || taskName == 'spotlessChangedApply'
+}
+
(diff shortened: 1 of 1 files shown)


miurahr (approved):
Thank you for capturing this and propose a fix.

PR #2155, opened 2026-07-21
feat: match fuzzy numbers

zs-stpa:
## Pull request type

## Which ticket is resolved?

- Consider numbers for fuzzy matches

## What does this PR change?

- Adds the opt-in project option `match_numbers` (Project properties). When enabled, fuzzy matching takes the numeric value of number tokens into account:
  - the verbatim similarity pass compares number tokens by value, so Roman numerals, Han numerals, Arabic-Indic and full-width digits count as equal to their ASCII counterpart; letter-only tokens are read as numerals only in unambiguous forms;
  - two number-only segments are scored by a placeholder comparison (all numbers as one token class, minus a fixed 10% penalty) instead of leaving the word-token passes at 0, so segments differing only in dates or bare integers surface as fuzzy matches at all.
- The flag is only written to `omegat.project` when enabled, so default project files stay byte-identical to what older OmegaT versions write.
- Adds a per-match tooltip in the Matches pane showing the three similarity values with plain-language labels, the penalties applied, and a warning when the matched numbers differ by value.
- New GUI strings in the English and German bundles.
- Tests: characterization of the previous behaviour, option-enabled expectations, project-file persistence round trip, tooltip rendering, and a load comparison (slow-down factor about 1.3 at 10,000 TM entries, bounded at 3 in the test).

## Other information

Builds on #2154 (`topic/stpa/util/numeral-parser`): the three foundation commits of that PR appear in this diff as well; only the top commit is new here. Once #2154 is applied upstream, this branch rebases to a single commit.

(edit: actually two commits because a ToTD was added separately)

Verification: full test suite and checkstyleMain pass on this branch, rebased onto current master. A demo video for the number-feature series is linked in the announcement thread on the omegat-development mailing list.

### Let fuzzy matching understand numbers

Manuals, price lists and reports repeat the same sentence with different numbers: "Tighten to 12 Nm" and "Tighten to 8 Nm", or segments that are nothing but a date or an amount. Normally such number-only segments find no fuzzy match at all.

Enable **Consider numbers in fuzzy matching** in the project properties, and the Fuzzy Matches pane starts comparing numbers by their value:

- Number-only segments (dates, quantities, prices) are offered as matches for each other, with a small penalty.
- Numbers count as equal across writing systems when their value is the same: XIV, 十四, ١٤ and the full-width １４ all equal 14.
- A tooltip on every match shows how its score was computed — and warns you when the numbers of the match differ in value from your segment, so a wrong number never slips into your translation unnoticed.

The option is stored in the project, so the whole team benefits from it.


brandelune:
You’ll need to update the manual as well.

zs-stpa:
manual has been updated accordingly.

brandelune:
> manual has been updated accordingly.

Please consider my remark regarding the option label.

zs-stpa:
> > manual has been updated accordingly.
> 
> Please consider my remark regarding the option label.

okay done. also added a tip of the day.

brandelune:
Could you resolve the comments I wrote please?

zs-stpa:
what/how exactly? i do not see a 'resolve' button on those comment-threads and i thought i had adressed all the issues. please clarify.

miurahr:
@zs-stpa could you solve the conflict?

zs-stpa:
> @zs-stpa could you solve the conflict?

which one? i do not see conflict in this particular PR - there is no 'resolve' button on any comments, merge is clear, CI is green, i've shortened the text of the tip of the day as requested elsewhere, i am not sure what is missing now.

miurahr (approved):

zs-stpa:
i did change the way the new setting is stored in the project file After This Was Merges, sorry there will be mergeconflicts as follow-up, but with hopefully good reason: downward-compatibility (i let the bot explain):

Heads-up on storage of `match_numbers`, follow-up to #2155:

- #2155 as merged stores `match_numbers` in `omegat.project`. Released OmegaT versions reject unknown elements there and can no longer open such projects ("Unable to read project file").
- This PR therefore moves setting into new sidecar `omegat/project_settings.properties`, ignored by released versions, and removes element from `omegat.project`/schema again — rebase onto current master shows up as revert of those #2155 hunks plus expected merge conflicts in `Bundle*`/`FindMatches`.
- #2274 additionally makes parser tolerate unknown elements for future schema growth; #2247 stores its `check_numbers` option in same sidecar.

No behaviour change otherwise; team negotiation dialogs and tests were reworked against sidecar storage.

PR #2161, opened 2026-07-22
feat(editor): decouple segment sorting from filtering (SF-1094)

zs-stpa:
## Pull request type

- Feature enhancement -> [enhancement] (groundwork; no user-visible change on its own)

## Which ticket is resolved?

No ticket is resolved by this PR alone. It is the first, minimal step towards:

- Sorting segments
  * https://sourceforge.net/p/omegat/feature-requests/1094/

## What does this PR change?

- Separates two concerns in the editor that were previously fused: filtering decides which segments are shown, a new `IEditorSorter` decides their order. The sorter is applied in `loadDocument()` after filtering; with no sorter set, the document order is byte-identical to before.
- `setSort` / `getSort` / `removeSort` on `IEditor` mirror the existing filter API, so plugins and scripts (via `ConsoleBindings`) can reorder the editor without touching filtering.
- Entry navigation now locates entries by their exact entry number (with a nearest-entry fallback) instead of assuming ascending positional order, so navigation keeps working under any custom order.
- Tests: `EditorControllerTest` covers the sorter contract, the unchanged-order guarantee and the navigation fallback.

## Other information

Second groundwork PR of the segment-sorting series (see #2154 for the numeral foundation): the sorter model and the sort bar UI follow as separate pull requests, so each step stays small and independently reviewable. Making sortability of segments available as a plugin or in any other way depends on this.

Verification: full test suite and checkstyleMain pass on this branch, rebased onto current master.

### DIFF
--- a/src/main/java/org/omegat/gui/editor/EditorController.java
+++ b/src/main/java/org/omegat/gui/editor/EditorController.java
@@ -256,6 +256,8 @@ public class EditorController implements IEditor {
     volatile @Nullable IEditorFilter entriesFilter;
     private @Nullable Component entriesFilterControlComponent;
 
+    volatile @Nullable IEditorSorter entriesSorter;
+
     private final SegmentExportImport segmentExportImport;
 
     protected @Nullable String currentEntryOrigin;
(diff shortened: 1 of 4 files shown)


miurahr (commented):
Great groundwork PR! Since this introduces a significant architectural change to `IEditor`, I'd like to propose the following next steps:

1. **Raise a separate PR with an ADR** documenting the sorting/filtering separation. Include:
   - problem statement a reason why these concerns needed to be decoupled
   - Decision of the design especially for the `IEditorSorter` interface and its role
   - Consequences
   - Referance of the PRs.

2. **Post to the dev mailing list** to start a discussion about the ADR. Many community members follow the dev list and may not actively monitor GitHub, so a discussion thread there will help ensure the design is understood and discussed widely.

This approach makes sure the architectural intent is documented and that the broader community can participate in shaping how the feature develops through the follow-up PRs.

zs-stpa:
see https://github.com/omegat-org/omegat/compare/master...zs-stpa:omegat:topic/stpa/docs/adr-editor-sorting, forewarding to dev-list a.s.a.p.

zs-stpa:
> This approach makes sure the architectural intent is documented and that the broader community can participate in shaping how the feature develops through the follow-up PRs.

okay, i did that, but did not see reactions, yet. unsure: is this a good sign - nobody against? or neutral sign - nobody read it? or bad sign - i do not know?

miurahr (approved):
LGTM.

PR #2163, opened 2026-07-22
feat(editor): sort segments by multiple criteria and numeric value (SF-1094)

zs-stpa:
## Pull request type

## Which ticket is resolved?

- Sorting segments
  * https://sourceforge.net/p/omegat/feature-requests/1094/

## What does this PR change?

- Adds a collapsible sort bar above the editor's filter bar (shown only with a project open and more than one filtered segment; collapsed it costs one line showing a summary of the active sort).
- Up to four sort criteria with +/- controls, each ascending or descending: source/target/note/comment (alphabetical, reverse or length order), translation status, change/creation date, changer/creator. Sorting is stable, with the entry number as final tiebreaker; a locale-aware Collator compares text.
- Text criteria additionally offer numeric ascending/descending: segments are ordered by the value of the number they contain ("Step 2" before "Step 10"), using the exact rational comparison of the `NumeralValueParser` foundation — all writing systems, fractions and signs included; number-less segments fall back to text order.
- Sort changes are staged and applied with apply/discard instead of resorting live on every control change. A notice above the controls warns that segment numbers are no longer sequential while a sort is active.
- Sorting cooperates with filtering instead of replacing it (see #2161): the set of visible segments never changes, and navigation keeps working under any order.
- Performance: numeric sort keys are cached per string during a sort pass; a progress indicator is shown while a sort is prepared; two performance tests guard the resize cost of the (expanded) sort bar, and the developer guide documents the policy for wall-clock assertions.
- A mixed-script demonstration XLIFF (shipped as a test resource) serves as an in-repo manual test document for the numeric sort.
- German GUI strings, and two tips of the day (English and German) introducing the sort bar and the numeric sort.

## Other information

Final PR of the segment-sorting series. It builds on the three groundwork PRs #2154 (numeral parser foundation), #2161 (sorting decoupled from filtering) and #2162 (collapsible bar base); their commits appear in this diff as well. The twelve commits on top are new here; once the groundwork PRs are applied upstream, this branch rebases to those twelve.

Verification: full test suite (including the new performance tests) and checkstyleMain pass on this branch, rebased onto current master.

### Screenshots

collapsed sortbar by default unsorted:
Bildschirmfoto 2026-07-22 um 19 03 29

expand and select options, with tooltips on the options:
Bildschirmfoto 2026-07-22 um 19 04 23

lots and lots of sorting options, anything and everything i could think of, really:
Bildschirmfoto 2026-07-22 um 19 05 38

up to four sorting criteria can be combined:
Bildschirmfoto 2026-07-22 um 19 06 49

applying the options collapses the sortbar again so the editor space loss is minimal:
Bildschirmfoto 2026-07-22 um 19 07 13

options even include sorting by numeric value (any script and any numbering system)
Bildschirmfoto 2026-07-22 um 19 08 06

### Tip of the Day's contents:

sort_bar_reorder.html — "Reorder segments with the sort bar"

<h3 class="title">Reorder segments with the sort bar</h3>
<p>The editor has a collapsible Sort bar above the segment list. When it is
collapsed it shows a one-line summary such as "Sort: File order (unsorted)";
click the header to expand it.</p>
<p>After "Sort by:" pick a criterion &mdash; grouped into Order &amp; structure,
Source text, Translation, Notes &amp; comments, Repetitions &amp; tags and History
&mdash; and a direction. Add "then by:" criteria to break ties, and move or remove
criteria to change their priority.</p>
<p>Sorting only changes how the segments are displayed; it does not change your
project or the segment order in the files. A short warning reminds you that, while
sorted, the segment numbers are no longer sequential.</p>
<p>To return to the original order, choose "File order (unsorted)" again.</p>

sort_bar_numeric.html — "Sort segments by numeric value"

<h3 class="title">Sort segments by numeric value</h3>
<p>When you sort by a text-based criterion in the sort bar, the direction
selector offers <b>Numeric ascending</b> and <b>Numeric descending</b> next to
the plain alphabetical order.</p>
<p>Numeric sorting orders segments by the actual value of the number they
contain, so "Step 2" comes before "Step 10" &mdash; where a plain alphabetical
sort would place "10" before "2".</p>
<p>It recognizes numbers written in many ways: a leading sign, decimals and
fractions (for example -5, 1.5, 3/2, &frac12;), and numerals from many writing
systems &mdash; Western and other decimal digits, Roman numerals (XII), CJK
(&#21313;&#20108;) and further systems supported by
<a class="ulink" href="https://icu4c-demos.unicode.org/icu-bin/locexp" target="_blank">ICU</a>, as well as enclosed or
circled numbers (&#9321;).</p>
<p>Because ordinary words such as "mix" or "XL" are valid Roman numerals too,
the selector also offers <b>Numeric asc./desc. no-Roman</b>: the same
value-based order, but Roman numerals count as plain text. The collapsed sort bar shows this mode as 1&uarr;/1&darr; instead of
#&uarr;/#&darr;.</p>
<p>Values are compared exactly, so 1/3 orders correctly against any decimal and
very large numbers never overflow. Segments with no recognizable number, or with
equal values, fall back to the normal alphabetical order for a stable result.</p>


zs-stpa:
switched off the new load tests so that they do not make the CI testing fail again

brandelune:
> the only thing i forgot (just had the idea this very moment) is to add a pseudo-random sorting order with a random seed as parameter so the pseudo-random sequence becomes reproducable… sort of an "i'm feeling lucky" button, if that makes sense? that would be the cherry on top, in my humble opinion.

How useful would that be?

zs-stpa:
> > the only thing i forgot (just had the idea this very moment) is to add a pseudo-random sorting order with a random seed as parameter so the pseudo-random sequence becomes reproducable… sort of an "i'm feeling lucky" button, if that makes sense? that would be the cherry on top, in my humble opinion.
> 
> How useful would that be?

controlled randomness was the only thing i could think of that was missing from the sort ordering options after all the even remotely sensible options are already there. "missing" in the sense of philosophical completeness, not necessarily practical value. never mind, jokes do not work well in this context here.

zs-stpa:
another thought: sorting by numeric value still needs the option to disregard roman numbers so that strings containing words like 'civ' or 'mix' or 'i' or 'v' or 'xl' or 'cm' are not automatically treated like numbers. hence more sorting direction options plus tooltips on the options:

Bildschirmfoto 2026-07-24 um 10 22 47


Bildschirmfoto 2026-07-24 um 10 22 54


Bildschirmfoto 2026-07-24 um 10 41 26

zs-stpa:
another possibility would be to make the collation locale optional, so you could reorder by the rules of an arbitrary locale instead of (as is now) always the source or target locale. for example, in a english->swedish project you could then decide you want the sorting order like a german dictionary.

brandelune:
> another possibility would be to make the collation locale optional, so you could reorder by the rules of an arbitrary locale instead of (as is now) always the source or target locale. for example, in a english->swedish project you could then decide you want the sorting order like a german dictionary.

It could be useful if it were based on the "bridging" language, otherwise I’d like to see how that could be useful.
https://omegat.sourceforge.io/manual-standard/en/chapter.how.to.html#how.to.tm.bridge.two.languages

brandelune:
> another thought: sorting by numeric value still needs the option to disregard roman numbers so that strings containing words like 'civ' or 'mix' or 'i' or 'v' or 'xl' or 'cm' are not automatically treated like numbers. hence more sorting direction options plus tooltips on the options:

I’m pretty sure "sort" is not the appropriate verb to put in the tooltip.

brandelune:
What does "rime" sorting refer to ?

It’s ok to have plenty of options, but keeping them to a list of realistically useful ones is better.

zs-stpa:
> What does "rime" sorting refer to ?
> 
> It’s ok to have plenty of options, but keeping them to a list of realistically useful ones is better.

text is reversed before sorting. in left-to-right languages like english, this groups the endings together instead of the beginnings. a normal dictionary does alphabetic from the beginning of a word. this rhyme thing does it reverse, starting from the last letter but then alphabetically ascending. this will put all segments ending with an exclamation mark together for example instead of those that start with one.

edit: what futurama referred to as a backtionary

zs-stpa:
> What does "rime" sorting refer to ?
> 
> It’s ok to have plenty of options, but keeping them to a list of realistically useful ones is better.

could reorder them so the 'most useful' are top and less often used are below? but i would rather like to exhaust all possibilities from the beginning instead of later adding endlessly, and i am sure i could come up with imagined use cases for any combination of the criteria if i had to.

edit: they are thematically ordered right now:

File order (unsorted)
── Order & structure ──
Paragraph start (Groups segments that begin a paragraph apart from those that do not.)
Source file (By the name of the source file the segment comes from.)
Path (By the path/identifier inside the source file (e.g. the XLIFF group or key path).)
Segment ID (By the segment id assigned by the source format (e.g. XLIFF or PO id).)
── Source text ──
Source alphabetical
Source backwards (Sorts by the source text read backwards, grouping segments with similar endings.)
Source length
── Translation ──
Target alphabetical
Target backwards (Sorts by the target text read backwards, grouping translations with similar endings.)
Target length
Translation status (Untranslated segments first, translated segments last.)
Origin (MT/source) (By the origin of the translation, e.g. the machine-translation engine that produced it.)
Match/leverage status (By match/leverage status (e.g. ICE, 100%, auto, enforced).)
Source translation fuzzy (Groups segments whose source-provided translation is marked fuzzy.)
── Notes & comments ──
Note alphabetical
Note backwards (Sorts by the note text read backwards, grouping notes with similar endings.)
Note length
Has note (Groups segments that have a note apart from those that do not.)
Comment alphabetical
Comment backwards (Sorts by the comment text read backwards, grouping comments with similar endings.)
Comment length
── Repetitions & tags ──
Duplicate status (By repetition role: unique, first occurrence, or a repetition of an earlier segment.)
Duplicate count (By how many times the same source segment occurs in the project.)
Tag count (By the number of inline tags (protected parts) in the source segment.)
── History ──
Change date
Creation date
Modification author
Creation author

zs-stpa:
edit: since the intended target audience for omegaT are (professional) translators, i think we could give them the option and those who need it would know how to use it in their own special cases. think of multilingual documents with more than just a few languages in them. i don't know how it will be useful does not equal that i know that it can never be useful.

edited again: the ticket asked for lots of sorting ways and "etc." and this should cover the etc.

zs-stpa:
ok here comes the shuffle … hope that does not disqualify the PR? the use case i can see is that you would want to still group same values together (e.g. Author or Comment) but shuffle the order so you do not always get the same Alfred's text first everytime for example.

Bildschirmfoto 2026-07-25 um 02 21 01

Bildschirmfoto 2026-07-25 um 02 25 46

and i thought some more about the optional locale, but that contradicts the strict source to target philosophy, simpler just to use source locale on source and comment and note and author and the target locale only for target text. which seems i ommitted previously and fixed it now, i.e. sorting by target alphabetically now really uses the target locale.

so it is now possible to sort segments by

File order (unsorted) (—)
Paragraph start (↕, ~°)
Source file (↕, #↕, 1↕, ~°)
Path (↕, #↕, 1↕, ~°)
Segment ID (↕, #↕, 1↕, ~°)
Source alphabetical (↕, #↕, 1↕, ~°)
Source backwards (↕, ~°)
Source length (↕, ~°)
Target alphabetical (↕, #↕, 1↕, ~°)
Target backwards (↕, ~°)
Target length (↕, ~°)
Translation status (↕, ~°)
Origin (MT/source) (↕, #↕, 1↕, ~°)
Match/leverage status (↕, ~°)
Source translation fuzzy (↕, ~°)
Note alphabetical (↕, #↕, 1↕, ~°)
Note backwards (↕, ~°)
Note length (↕, ~°)
Has note (↕, ~°)
Comment alphabetical (↕, #↕, 1↕, ~°)
Comment backwards (↕, ~°)
Comment length (↕, ~°)
Duplicate status (↕, ~°)
Duplicate count (↕, ~°)
Tag count (↕, ~°)
Change date (↕, ~°)
Creation date (↕, ~°)
Modification author (↕, ~°)
Creation author (↕, ~°)

Total sorting direction options: 144 = 8 text keys × 8 (↕, #↕, 1↕, rnd+seed) + 20 keys × 4 (↕, rnd+seed); And they can be combined.

1) alphabotic acsending+ 2) descending, 3) numeric asc. 4) +desc., 5) numeric w/o roman asc+ 6) desc, 7) random, 8) random w. seed

brandelune:
> > What does "rime" sorting refer to ?
> > It’s ok to have plenty of options, but keeping them to a list of realistically useful ones is better.
> 
> text is reversed before sorting. in left-to-right languages like english, this groups the endings together instead of the beginnings.

"rime" does not mean what you say it does.

brandelune:
> edited again: to be honest, most of the imaginary use cases will involve such a big software product 🫣 and the criteria are essentially every possible way to look at every possible property of a segment

I am totally unconvinced. How many "big software product" did you have to handle where one of your translator team required such a feature?

zs-stpa:
> > > What does "rime" sorting refer to ?
> > > It’s ok to have plenty of options, but keeping them to a list of realistically useful ones is better.
> > 
> > 
> > text is reversed before sorting. in left-to-right languages like english, this groups the endings together instead of the beginnings.
> 
> "rime" does not mean what you say it does.

true. see also https://en.wikipedia.org/wiki/Rhyming_dictionary quote "A simple [reverse dictionary](https://en.wikipedia.org/wiki/Reverse_dictionary), which collates words starting from the end, provides a rough rhyming dictionary to the extent that spelling follows pronunciation. However, a precise rhyming dictionary reflects pronunciation, not spelling." see also Futurama Episode where the robot Bender refers to it as Backtionary https://www.theinfosphere.org/Webster%27s_New_Abridged_Becktionary

zs-stpa:
> > edited again: to be honest, most of the imaginary use cases will involve such a big software product 🫣 and the criteria are essentially every possible way to look at every possible property of a segment
> 
> I am totally unconvinced. How many "big software product" did you have to handle where one of your translator team required such a feature?

i said in the commit message "Reviewers can now shuffle the segment order to avoid position bias". this adds the freedom to shuffle in dozens of different ways, users are free not to use shuffle, the default remains as it always has been the unsorted natural file order and the shuffle option is hidden away in a dropdown.

edit: counter question: why would anyone want to shuffle anything? there were shuffle buttons on cd players. why was Apple able to ever sell an iPod Shuffle?

zs-stpa:
> > edited again: to be honest, most of the imaginary use cases will involve such a big software product 🫣 and the criteria are essentially every possible way to look at every possible property of a segment
> 
> I am totally unconvinced. How many "big software product" did you have to handle where one of your translator team required such a feature?

edit: not so many, true.

zs-stpa:
> > > What does "rime" sorting refer to ?
> > > It’s ok to have plenty of options, but keeping them to a list of realistically useful ones is better.
> > 
> > 
> > text is reversed before sorting. in left-to-right languages like english, this groups the endings together instead of the beginnings.
> 
> "rime" does not mean what you say it does.

true. but writing only 'reverse' as in
"Comment reverse"
seemed not intuitive enough and there may be users who rather know what a rhyming dictionary is so i added the (truely not exact) rhyme
"Comment reverse (rhyme)"
and because that still did not seem clear enough, i added these tooltips to the reverse entries:
"Sorts by the comment text read backwards, grouping comments with similar endings."
see also above the complete list of entries including tooltips.

brandelune:
> > > > What does "rime" sorting refer to ?
> > > > It’s ok to have plenty of options, but keeping them to a list of realistically useful ones is better.
> > > 
> > > 
> > > text is reversed before sorting. in left-to-right languages like english, this groups the endings together instead of the beginnings.
> > 
> > 
> > "rime" does not mean what you say it does.
> 
> at least some people in the texting business will have heard of this expression. see also https://en.wikipedia.org/wiki/Rhyming_dictionary quote "A simple [reverse dictionary](https://en.wikipedia.org/wiki/Reverse_dictionary), which collates words starting from the end, provides a rough rhyming dictionary to the extent that spelling follows pronunciation. However, a precise rhyming dictionary reflects pronunciation, not spelling." see also Futurama Episode where the robot Bender refers to it as Backtionary https://www.theinfosphere.org/Webster%27s_New_Abridged_Becktionary

That’s exactly what I am saying : 
> > "rime" does not mean what you say it does.

brandelune:
> > > edited again: to be honest, most of the imaginary use cases will involve such a big software product 🫣 and the criteria are essentially every possible way to look at every possible property of a segment
> > 
> > 
> > I am totally unconvinced. How many "big software product" did you have to handle where one of your translator team required such a feature?
> 
> yes you are not completely wrong here and i agree, maybe this was a little much over the top … but on the other hand, this gives new degrees of freedom to the users and preemptively anticipates followup wishes. the end-solution of the sortability-question so to say.

You seem to confuse OmegaT with your personal playground.

OmegaT is a tool for professional translators and we try to base its development from requests emanating from real translation issues. Not from what a developper thinks of what translators need.

zs-stpa:
consider. the default is unchanged: natural segment order same as the input file as it has always been. no click, no change, just that harmles bar 'unsorted' over the editor. if a user should require any sorting, just click expand, rest is self explanatory with tooltips and in depth exhaustive options leaving the freedom Not to use the options to the user themselves.

zs-stpa:
removed "rhyme" from tooltips and descriptions, it now says 'backwards' instead of reverse
edit: and rebased on current master
edit again: this makes the screenshots and demo video obsolete but they still serve to get the general idea so i will not update them all

zs-stpa:
#2163 was amended to reduce the merge-conflict surface: the sort hook (getSort/setSort/removeSort) was declared abstract on IEditor, which forced a no-op stub in every implementer (ConsoleBindings, the test EditorStub, MatchesVarExpansionTest). Because the sort-hook commit also exists as a divergent copy on this branch, those identical stubs collided when integrated alongside #2161 — duplicate method definitions, and without conflict markers, so they slipped past a clean merge. Converting the methods to `default` (returning null / no-op) drops the per-implementer stubs entirely, so implementers no longer touch these files and the recurring conflict disappears.

zs-stpa:
> > > > edited again: to be honest, most of the imaginary use cases will involve such a big software product 🫣 and the criteria are essentially every possible way to look at every possible property of a segment
> > > 
> > > 
> > > I am totally unconvinced. How many "big software product" did you have to handle where one of your translator team required such a feature?
> > 
> > 
> > yes you are not completely wrong here and i agree, maybe this was a little much over the top … but on the other hand, this gives new degrees of freedom to the users and preemptively anticipates followup wishes. the end-solution of the sortability-question so to say.
> 
> You seem to confuse OmegaT with your personal playground.
> 
> OmegaT is a tool for professional translators and we try to base its development from requests emanating from real translation issues. Not from what a developper thinks of what translators need.


sorry. was just trying to solve tickets from users. claude-code overwhelmed me too, this was a bit much, i realize that, and i am switching back to traditional IDE for sanity's sake.

edited: this ticket asks to decouple sorting from filtering -> done. asks for multiple criteria -> done. have you seen the screenshots at the top? typical usecase. order by translated/untranslated. then by author. group same authors together. other tickets (whole cluster of them) asks for better support for reviewers. as stated before, no problem for me to set this up in-house for my payed daytime job as an add-on on top of the official repo. but i thought, sharing was the idea of the whole thing here.

zs-stpa:
should i squash these many evolutionary commits down to 1–2 logical commits for easier reviewability?

src/main/java/org/omegat/gui/editor/sort/package-info.java
```diff
@@ -0,0 +1,29 @@
+/**************************************************************************
```
miurahr:
```suggestion
/*
```

Javadoc task will be failed when copyright header claims its javadoc.

miurahr:
@brandelune  do you confimred this feature and labels?

You comment about "sorting" on the tooltips.
`Bundle.properties` has
```
SORT_DIR_NUM_TOOLTIP=Compares by numeric value: "Kap.10" sorts after "Kap.2".
```
that use "sorts". Is it ok?

The code looks good to me, but discussion thread looks not finished.

brandelune:
@zs-stpa after a quick search on the dev list, it occurs to me that this proposal has not been formally introduced to the dev list.

I would appreciate that you do so by indicating:
- the sorting criteria that you propose
- the general UI paradigm
- all the other information that you deem relevant

Please do not copypaste this PR to the dev list, instead make a short and to the point informative mail.

brandelune:
> @brandelune do you confimred this feature and labels?
> 
> You comment about "sorting" on the tooltips. `Bundle.properties` has
> 
> ```
> SORT_DIR_NUM_TOOLTIP=Compares by numeric value: "Kap.10" sorts after "Kap.2".
> ```
> 
> that use "sorts". Is it ok?
> 
> The code looks good to me, but discussion thread looks not finished.

Thank you for the reminder. That use of the verb "sort" does not seem correct.

I’ll be waiting for Stephan to send all this to dev before commenting further.

PR #2172, opened 2026-07-24
chore(core): improve error handling and optimize file processing logic

miurahr:
Improve error handler for the external command execution

## Pull request type

refactor

## Which ticket is resolved?
no issue ticket, the fix for the corner case when external command failed silently

## What does this PR change?

-
-
-

### DIFF
--- a/src/main/java/org/omegat/core/data/RealProject.java
+++ b/src/main/java/org/omegat/core/data/RealProject.java
@@ -797,9 +797,22 @@ public class RealProject implements IProject {
             stdout.start();
             stderr.start();
         } catch (IOException e) {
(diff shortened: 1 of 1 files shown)


zs-stpa:
i am looking into the 'help wanted'.

The same pattern — passing a possibly-blank getLocalizedMessage()/getMessage() straight into a dialog — recurs at ~13 UI call sites. I am handling those in a separate PR: a shared StringUtil.describeException(Throwable) helper (localized message → message → class name, null-safe) plus a unit test, swapped in at each site. That keeps this PR focused on RealProject; the follow-up will reference #2172.

miurahr:
I want to know a real case of exception on external command execution.

PR #2177, opened 2026-07-25
fix(core): return empty collections instead null from NotLoadedProject

zs-stpa:
## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?

no ticket exists, coincidental finding.
- NotLoadedProject returns null from collection getters despite the IProject contract

## What does this PR change?

- `NotLoadedProject` now returns empty collections instead of `null` from its six collection getters (`getAllEntries`, `getUniqueEntries`, `getTransMemories`, `getOtherTargetLanguageTMs`, `getProjectFiles`, `getSourceFilesOrder`).
- The `IProject` javadoc promises an unmodifiable list, so callers do not expect `null` and code iterating over these results could throw a `NullPointerException` while no project is loaded.
- The only existing null check (`EditorController`) remains correct, since it also accepts an empty list.
- Adds a regression test asserting all six getters return non-null, empty collections.

## Other information

No behavioural change for loaded projects; `NotLoadedProject` is only active while no project is open.

### DIFF
--- a/src/main/java/org/omegat/core/data/NotLoadedProject.java
+++ b/src/main/java/org/omegat/core/data/NotLoadedProject.java
@@ -27,6 +27,7 @@
 package org.omegat.core.data;
 
 import java.io.IOException;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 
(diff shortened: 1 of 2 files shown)


miurahr:
Good direction!

PR #2178, opened 2026-07-25
fix(filelist): let ProjectFilesListController unregister its listeners

zs-stpa:
## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?

no bug ticket, accidental finding.
- ProjectFilesListController never unregisters its CoreEvents and preference listeners

## What does this PR change?

- Adds `ProjectFilesListController.dispose()`, which unregisters the project, entry and font-changed listeners from `CoreEvents`, removes the show-progress listener from `Preferences`, and disposes the window.
- Keeps the listeners in fields so they can be unregistered again; previously they were anonymous registrations with no way to remove them, so every controller instance stayed strongly reachable through the static `CoreEvents` registries for the lifetime of the JVM.
- Adds the missing `Preferences.removePropertyChangeListener` overloads (global and per-property) as counterparts to the existing add methods.
- `EditorProjectReloadLeakTest` now disposes the controller it creates instead of leaving its listeners registered for subsequent tests.
- Adds `ProjectFilesListControllerDisposeTest`, which proves via a weak reference under GC pressure that a disposed controller becomes unreachable.

## Other information

In the running application the controller is created once per main window, so the practical impact is on tests and any embedding code that creates the window repeatedly; the missing `Preferences` remove API affects all per-property listeners.

### DIFF
--- a/src/main/java/org/omegat/gui/filelist/ProjectFilesListController.java
+++ b/src/main/java/org/omegat/gui/filelist/ProjectFilesListController.java
@@ -51,6 +51,7 @@ import java.awt.event.MouseAdapter;
 import java.awt.event.MouseEvent;
 import java.awt.event.WindowAdapter;
 import java.awt.event.WindowEvent;
+import java.beans.PropertyChangeListener;
 import java.io.File;
 import java.io.IOException;
 import java.math.RoundingMode;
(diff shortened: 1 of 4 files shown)


zs-stpa:
#2181 + rerunning CI probably makes it green all the way

miurahr (approved):

PR #2179, opened 2026-07-25
fix(build): copy tips into the distribution from src/docs/tips

zs-stpa:
## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?

- No SourceForge ticket: incidental finding — installed builds ship without `docs/tips`, so Tip of the Day fails at startup with a Jackson exception.

## What does this PR change?

- The distribution copy spec in `build.gradle` and the greeting-page path in the document conventions still pointed at `src_docs`, which no longer exists since the source tree reorganization (#2117). Both now point at `src/docs`, so `docs/tips` is included in distributions again.
- Updates the two developer documentation pages that still referred to `src_docs`.
- The tip loader (`TipOfTheDayUtils.hasIndex`, `OmegaTTipOfTheDayModel.initTips`) now tolerates a missing tips index instead of throwing, so a broken or trimmed installation degrades to simply not showing tips.

## Other information

Verified with `installDist`: `build/install/OmegaT/docs/tips/en/tips.yaml` is present again and Tip of the Day opens in the installed layout.


miurahr:
Thank you for finding my mistake for the migration.

miurahr (approved):

PR #2159, opened 2026-07-22
feat(autoconvertnumbersonlysegments): new numerological tool window (RFE-794)

zs-stpa:
## Pull request type

- Feature enhancement -> [enhancement]

## Which ticket is resolved?

- Auto fill "number only" segments
  * https://sourceforge.net/p/omegat/feature-requests/794/

## What does this PR change?

- Adds Tools -> "Auto-convert Number Segments...": a window (modeled on the Issues window: minimizable frame, asynchronous scan, jump to segment) that scans the project for segments whose source is only a number, proposes a locale-aware conversion for each, and applies the accepted ones as translations.
  - Checklist of convertible types (integer, decimal, percent, currency, date, time, ordinal) next to the target-format options, with connector lines showing which option affects which type; scope selector (all segments / current file / current filter).
  - Results table with type, original, target preview, confidence and a per-row accept checkbox; a minimum-confidence threshold and bulk apply.
- The conversion engine `org.omegat.util.NumberAutoConverter` is stateless and GUI-free: it parses the value in the source locale (all writing and numbering systems the `NumeralValueParser` foundation handles — Roman, CJK, Arabic-Indic, Devanagari, full-width, ...) and re-renders it in the target locale. Confidence is lowered for separator usage foreign to the source locale; the grouping option also applies to currency and percent rendering.
- Applied conversions are marked with a new `TMXEntry.ExternalLinked` value `xNUMBER`, persisted as an `x-number` TMX property gated by the existing "save auto-populated status" preference (like `x-auto`). They are therefore recognizable as OmegaT's own automatic conversions, get their own editor marker color (light and dark scheme), and Ctrl+U ("skip to next untranslated") passes over them with no further code.
- New GUI strings in the English and German bundles.
- Developer documentation: ADR 2026003 (design and decisions of this feature) and a new "common extension recipes" chapter.
- User manual (English): a short chapter for the window and a Tools menu entry.
- Tip of the Day (German and English): with some example conversions.
- Tests: converter engine (about 520 lines), scanner, and persistence round trip of the `xNUMBER` marking.

## Other information

Builds on #2154 (`topic/stpa/util/numeral-parser`), like #2155: the three foundation commits of that PR appear in this diff as well; the eight commits on top are new here. Once #2154 is applied upstream, this branch rebases to those eight.

Verification: full test suite and checkstyleMain pass on this branch, rebased onto current master. A demo video for the number-feature series is linked in the announcement thread on the omegat-development mailing list.

### Screenshot

new tool window:
[...]

PR #2181, opened 2026-07-25
test(core): make the reachability verdict of leak tests deterministic

zs-stpa:
## Pull request type

- Other: test robustness (flaky CI test)

## Which ticket is resolved?

- No SourceForge ticket: incidental finding — `EditorProjectReloadLeakTest` fails spuriously on the Windows CI runners ("closed project ... is still strongly reachable") while macOS and Linux pass; seen on unrelated PRs (#2159, #2163, #2178).

## What does this PR change?

- `ProjectReloadLeakTest` and `EditorProjectReloadLeakTest` decided reachability with advisory `System.gc()` rounds plus allocation pressure, which the loaded Windows runners ignore often enough to fail spuriously.
- `becomesUnreachable()` keeps a short advisory-GC fast path but now falls back to filling the heap in 4 MB blocks until either the `WeakReference` clears or an `OutOfMemoryError` is caught. The JVM must run a full collection that clears weakly reachable objects before it throws, so a still-set reference proves an actual leak instead of guessing.

## Other information

Both tests ran five consecutive times locally with the CI flags (`-PenvIsCi -PtestMaxHeapSize=2048M`, `--rerun-tasks`) without failure; a genuine leak still fails deterministically because the reference survives even heap exhaustion.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -305,6 +305,31 @@ jar {
 test {
     // some test case depends on modules from subproject
     dependsOn subprojects.collect {it.tasks.withType(Jar)}
+    useJUnit {
+        // Destructive memory-stress tests deliberately exhaust the heap;
+        // they only run on explicit request via the memoryStressTest task.
+        excludeCategories 'org.omegat.core.MemoryStressTests'
+    }
+}
+
+tasks.register('memoryStressTest', Test) {
+    group = 'verification'
+    description = 'Runs the destructive memory-stress tests, which deliberately exhaust the heap.'
+    // some test case depends on modules from subproject
+    dependsOn subprojects.collect {it.tasks.withType(Jar)}
+    testClassesDirs = sourceSets.test.output.classesDirs
+    classpath = sourceSets.test.runtimeClasspath
+    workingDir = projectDir
+    // Test in headless mode with ./gradlew memoryStressTest -Pheadless
+    if (project.hasProperty('headless')) {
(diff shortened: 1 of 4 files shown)


miurahr:
I agree with the intent of this regression test, but the implementation is destructive because it deliberately exhausts the heap to force GC. I don’t think this should run by default. Please add a dedicated marker/category for destructive or memory-stress tests and configure it so the test only runs when that marker is explicitly requested.

zs-stpa:
done, the heap exhaustion should not be part of the default runs.

- New JUnit category marker `org.omegat.core.MemoryStressTests` for destructive memory-stress tests.
- Both leak tests are annotated with `@Category(MemoryStressTests.class)`.
- The default `test` task now excludes that category, so the leak tests no longer run (and no longer flake) in regular and CI builds.
- A dedicated `memoryStressTest` task runs exactly this category on explicit request: `./gradlew memoryStressTest` (it honours the same `-PtestMaxHeapSize`, `-Pheadless` and logging settings as `test`).

Verified locally: the default suite is green with the two tests absent from its results, and `memoryStressTest` runs exactly the two classes, green. The branch is rebased onto current master.

As a follow-up once this is in, the performance tests proposed in #2163 (currently skipping on CI via `CI`/`TF_BUILD` environment detection) could migrate to the same mechanism with their own category, which would be cleaner than environment sniffing.

miurahr (approved):

PR #2183, opened 2026-07-25
feat(gui): apply color changes w/o restart and add Apply button to prefs

zs-stpa:
## Pull request type

## Which ticket is resolved?

- Preferences/CutomColorSelector don't show notification immediately when modifying color
  * https://sourceforge.net/p/omegat/bugs/1273/

## What does this PR change?

- Colour changes in Preferences > Colours no longer require an application
  restart: after saving, the open document is redisplayed with the new
  colours. The workflow becomes WYSIWYG instead of
  change-restart-inspect-repeat.
- Adds an **Apply** button to the Preferences dialog: save and see the effect
  without closing the dialog, for every preferences view. A short "Changes
  applied" notice appears in the message area (restart/reload warnings take
  precedence), and reload-requiring changes prompt for a project reload just
  like OK.
- "Reset colour" and "Restore defaults" are explicit restore actions and take
  effect immediately.
- Markers and some windows cached their painters/attributes at construction
  time, freezing the colours of the first start; painters are now created
  when marks are (re)calculated. A regression test
  (`MarkerColorFreshnessTest`) asserts that a colour change is reflected in
  the very next mark calculation.

## Other information

The discussion in bug SF-1273 ended with an invitation to implement a live
update of colours; this PR implements it. Performance note: mark
recalculation already creates one `Mark` object per match, while painters are
created at most a handful of times per entry and marker; the Swing
highlighter removes old highlights before re-adding, so nothing accumulates.

### DIFF
--- a/config/checkstyle/suppressions.xml
+++ b/config/checkstyle/suppressions.xml
@@ -64,7 +64,7 @@
               checks="(DesignForExtension|MagicNumber)"/>
     <suppress checks="WhitespaceAround"
             files="(AlignFilePicker|InstanceEditor|CreateGlossaryEntry|NewTeamProject|FilenamePatternsEditor|FilterEditor|SegmentPropertiesListCell|FilterBarReplace|ProjectFilesList)\.java"/>
-    <suppress checks="ParameterNumber" files="EntryListPane\.java" lines="304"/>
+    <suppress checks="ParameterNumber" files="EntryListPane\.java" lines="302"/>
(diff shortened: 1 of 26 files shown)


miurahr:
LGTM for logic. This PR to move color object creation from ctor/static to a time to use. Good direction. 
Please check manual change by reviewers.

brandelune (approved):

miurahr (approved):

zs-stpa:
i have noticed that the editor background color is not applied and am currently working on it and on other improvements regarding user defined color schemes, like import/export. more PR's to come soon.

edit: i did hope i could fold it into existing work, but turns out this is a whole new kettle of worms … the 'apply' does lie, other places need live refresh with a deeper change, PR coming soon.

PR #2185, opened 2026-07-26
docs(tips): add a German "Machine Translation 4tw" comic tip

zs-stpa:
## Pull request type

- Documentation contents changes -> [docs]

## Which ticket is resolved?

No SourceForge ticket: new tip-of-the-day content addressing the recurring confusion between machine translation and translation memory.

## What does this PR change?

- Adds a German tip of the day "Machine Translation 4tw": a three-page comic explaining why machine-translation output that changes between runs needs a translation memory for stable terminology, and which part of a localisation pipeline (XLIFF in, XLIFF out) OmegaT covers.
- Ships the hand-editable SVG source of the comic next to the tip (`docs/tips/de/machine_translation_4tw.svg` in the installation); the tip text points users to it so they can adapt the presentation for their own audience.
- Embeds the comic as a compressed PNG export (pngquant, 116 KB, visually lossless) because the CSSBox-based tip pane cannot render SVG.
- Registers the tip in the German `tips.yaml` only; other languages are unaffected until translations exist.

## Other information

The confusion the comic addresses is well documented. Corinne McKay (translator and translator trainer) in "How to choose a translation memory tool" (Training for Translators, 2024-10-15, https://www.trainingfortranslators.com/2024/10/15/how-to-choose-a-translation-memory-tool/):

> "When I first started translating, I didn't even understand the difference between TM/translation memory, and MT/machine translation. The first time I used a translation memory tool, I wondered why it wasn't doing the translation for me!"

Notes on the SVG source: it deliberately avoids `feDropShadow` (unsupported by Inkscape, which would drop the page backgrounds and rewrite the filter region on save) and documents its page grid and editing conventions in comments, so it stays hand-editable in a text editor and survives an Inkscape round-trip. The OmegaT logo paths are inlined from `images/OmegaT.svg` because external `<use>` references do not work in GitHub rendering; a comment marks the copy for re-synchronisation should the logo ever change.


zs-stpa:
text written myself and only in german. here are screenshots of page one:

Bildschirmfoto 2026-07-26 um 17 35 49

and

Bildschirmfoto 2026-07-26 um 17 36 32

zs-stpa:
and page 2

Bildschirmfoto 2026-07-26 um 17 42 22

and page 3

Bildschirmfoto 2026-07-26 um 17 42 36

and that is it there is only 3 pages.

zs-stpa:
strong reference cycle test failing again on a documentation-only-PR, is another indication for #2181

zs-stpa:
wrote only in german
 - because it contains linguistic puns which are hard to translate.
 - some of those puns are pretty funny, if i may do say so myself…
 - they convey (especially to translators and linguaphiles and polyglots and similir clientele) some 'hidden' messages between the lines
 - there are no watermarks or steganography involved, please do not take the previous remark too literal. feel free to check.
 - it is a widely known fact that jokes become more funny the longer the explanation gets.
 - i did some research on 'who are users of omegaT and what do they want'
 - this is my way of giving them that and having fun at the same time
 - please do not try to disallow me from having fun while doing useful work at the same time, that will not work, or i will find ways to have fun while doing useful stuff elsewhere.

have a nice evening. peace. omnia sol temperat. (latin, the sun is shining for all of us.) well not at the same time. whatever. ciao amici

zs-stpa:
oh shoot, sorry i only just noticed that translating the one pun to english makes it even better but only when you know the origin ... see, there was a german TV ad long ago but well known about menstruation products 'Die Geschichte der Menstruation ist eine Geschichte voller Mißverständnisse' and i made it i18n instead of a bodily function … and in english it becomes 'The story of i18n is a story full of misconceptions' which is absolutely hillarious because it illustrates the point that the 3slide presentation is also trying to make between the lines: there are cringy mistranslations and horrible errors waiting to happen when you let ambiguous unchecked stuff loose without context knowledge
edit: sorry i let out the obvious conclusion because it seems so obvious: the presentation answers a need of users to better understand (and explain to their clients) why they are using omegaT and not just hit the magic autotranslate button. it still lets open how the two can come together in omegaT.

miurahr:
Thanks for this, I really enjoyed the comic and the writeup, including the pun explanation at the end.

One note on the tip pane rendering. The pane does have vertical scrolling, but no horizontal scroll bar, and content wider than 900px gets cut off rather than reflowed. There is a separate ongoing discussion on the dev list about this reflow problem, and I would like to see a horizontal scroll bar enabled there as a general fix for the tip pane.

Even with that fix though, I think this particular comic would work better split into three separate tips rather than one long vertical scroll. Each page is already designed as a self contained beat (setup, payoff, punchline), so splitting it into three sequential tips.yaml entries would let each one display without scrolling at all, at roughly 495px tall for a 900px wide pane, and would fit the day to day tip format more naturally.

zs-stpa:
i'd like to keep the SVG together. others are free to edit how ever they like, this is how i like it so i keep it like that. i'll expand on the text before the comic a little to make it fill a page and then the three split PNGs, makes four new tips instead of one, coming soon.

zs-stpa:
done.

more text explaining some of the puns and how to edit the slides for users own use.
(edit: leaving the unsayd conclusion to be made by the user themselves, that they can even use omegaT to translate the svg into their own private languages…)

Bildschirmfoto 2026-07-29 um 00 36 33

(cropping is a known issue, i did not bother to reformat the html, please say so if i should)

and then three slides.

Bildschirmfoto 2026-07-29 um 00 36 55
(see also https://www.merriam-webster.com/dictionary/free-for-all) 
only one shown. others see above, only a little smaller.

edit: brief design study for how to write the project name … but overkill so decided against.
Bildschirmfoto 2026-07-29 um 21 22 09

miurahr (approved):

PR #2188, opened 2026-07-27
fix(core): keep enforced translations highlighted after a team save (SF-RFE-1171)

zs-stpa:
## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?

Enforced translations lose background color - https://sourceforge.net/p/omegat/bugs/1171/

## What does this PR change?

- Keeps the highlighting of enforced translations after a team-project save. The enforced status (`xENFORCED`) previously lived only in memory: the project TMX export never wrote it and `calcExternalLinkedMode` never read it back, so a team rebase - which replaces the project TMX content with the entries parsed from the merged file - silently downgraded enforced entries to plain ones and their colour disappeared until the next full project load. `RealProject` now re-applies the loaded `tm/enforce` memories directly after the TMX rebase.
- Persists the enforced status as an `x-enforced` TMX property when the user opted into saving auto states (preference `save_auto_status`), symmetrically to the existing `x-auto`/`x-ice`/`x-100pc` handling, so the status also round-trips through the saved file.
- Stops the spurious "reverting enforced segment" warning: deactivating an enforced default segment raised the warning unconditionally, so merely navigating away - or the editor view refresh triggered by a preferences/colour change (which commits via both `gotoFile` and `gotoEntry`) - popped it, and via the refresh even twice. The warning now only appears when the translation or note was actually changed.

## Other information

Tests: `AutoTmxTest` gains coverage for the save/reload round-trip of the enforced status (with and without the opt-in) and for the re-apply after a content replacement. A new `EditorEnforcedSegmentTest` verifies that a plain view refresh of an enforced segment stays silent while an actual change still warns exactly once (headless-skipped, like the sibling editor tests).

The spurious-warning fix is a closely related incidental finding: the two symptoms share the same enforced-segment code path and both surface when a colour change is applied while sitting on an enforced segment. The persistence half is only reachable in team projects (the content replacement happens during the rebase-on-save), so a full round-trip reproduction needs a git-backed team project; the unit tests exercise the underlying TMX and marker mechanics directly.

### DIFF
--- a/src/test/java/org/omegat/gui/editor/EditorEnforcedSegmentTest.java
+++ b/src/test/java/org/omegat/gui/editor/EditorEnforcedSegmentTest.java
@@ -1,0 +1,316 @@
+/**************************************************************************
+ OmegaT - Computer Assisted Translation (CAT) tool
+          with fuzzy matching, translation memory, keyword search,
+          glossaries, and translation leveraging into updated projects.
+
+ Copyright (C) 2026 Stephan Pakebusch
+               Home page: https://www.omegat.org/
+               Support center: https://omegat.org/support
+
+ This file is part of OmegaT.
+
+ OmegaT is free software: you can redistribute it and/or modify
+ it under the terms of the GNU General Public License as published by
+ the Free Software Foundation, either version 3 of the License, or
+ (at your option) any later version.
+
+ OmegaT is distributed in the hope that it will be useful,
+ but WITHOUT ANY WARRANTY; without even the implied warranty of
+ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ GNU General Public License for more details.
+
+ You should have received a copy of the GNU General Public License
+ along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ **************************************************************************/
+package org.omegat.gui.editor;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assume.assumeFalse;
+
+import java.awt.Font;
+import java.awt.GraphicsEnvironment;
+import java.io.File;
+import java.io.IOException;
+import java.lang.reflect.InvocationTargetException;
+import java.nio.file.Files;
+import java.util.ArrayList;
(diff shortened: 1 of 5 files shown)


src/test/java/org/omegat/gui/editor/EditorEnforcedSegmentTest.java
```diff
@@ -0,0 +1,316 @@
...
+        Core.setFilterMaster(new FilterMaster(FilterMaster.createDefaultFiltersConfig()));
+
+        TestCoreInitializer.initNotes(new RecordingNotes());
+        File projectRootDir = Files.createTempDirectory("omegat-enforce").toFile();
+        EnforcedProjectProperties props = new EnforcedProjectProperties();
+        props.setProjectRoot(projectRootDir.getAbsolutePath());
+        props.setSourceTokenizer(LuceneEnglishTokenizer.class);
+        props.setTargetTokenizer(DefaultTokenizer.class);
+        EnforcedProject project = new EnforcedProject(props);
+        Core.setProject(project);
```
miurahr: Please find a `TestCoreState` class to set-up project.
```
    TestCoreState.getInstance().setProject(...);
``` 
is better for test isolation.

As you know, `Core.setProject` is facade to `CoreState#setProject` that is equivalent.

miurahr (changes requested):
Looks good but there is a place to improve, not mandatory.

src/test/java/org/omegat/gui/editor/EditorEnforcedSegmentTest.java
zs-stpa: feedback on original ticket also positive: https://sourceforge.net/p/omegat/bugs/1171/#fdfb/9b5d

i'll put these micromanagement-adjustments back, sorry. but please feel totally free to edit it in yourself maybe? just goes a lot faster than with the days between messsages … (sorry if i'm still too used to agility)

zs-stpa: i have adressed the issue and will amend per force push with lease

miurahr (approved):

PR #2189, opened 2026-07-27
feat(preferences): rework the editor colours table

zs-stpa:
**Pull request type**

**Which ticket is resolved?**

This is a usability follow-up to the recent colour work and does not close a dedicated ticket. Related tickets and pull requests:
- Colour codes for source-file pre-translation — SF-1433 (#2182)
- Apply colour changes without a restart — SF-1273 (#2183)
- Enforced-match colour persistence — SF-1171 (#2188)

**What does this PR change?**
- Replaces the flat list in the Colours preference pane with a sortable table ~~with Item, Colour and Internal name columns~~ whose first column is a running number (#), followed by Item, Colour and Internal name.
- Adds a search field that filters on either the display name or the internal identifier, backed by a live hit counter and a clear button.
- Keeps each column at its natural width and lets the long internal-name column overflow into a horizontal scroll bar, so the layout no longer depends on the viewport size and the pane renders correctly with no project open. Columns are pinned (they cannot be dragged out of order) but stay resizable; the # column has 8 px of horizontal padding.
- ~~Sorts by the internal identifier on first display; header clicks re-sort each column (the colour column by hex value).~~ Sorts by the # column on first display, i.e. in the natural EditorColor enum order. Header clicks re-sort each column; the colour column is special — repeated clicks rotate its ordering through hex, HSB (colour wheel: hue -> saturation -> brightness) and luminance (perceived brightness), each ascending then descending, with a multi-line header tooltip that explains this.

**Other information**
- ~~A short screen recording of the result is included under `src/docs/developer/media/`.~~ A short screen recording is included under `src/docs/developer/media/`, encoded as H.264 without an audio track (the earlier HEVC clip had playback-compatibility issues on some platforms).
- ~~A unit test covers the colour sort key.~~ Unit tests cover the three colour sort orders (hex, HSB, luminance) and the position of the # column.
- Follow-up: a short-lived status message announcing the current colour sort order after each re-sort would aid discoverability; it depends on the transient-message plumbing from the colour import/export PR (#2195), so it fits once both land.


brandelune:
>  2026-07-27-colors-preferences-table-demo.mp4

The movie does not seem to work.

zs-stpa:
movie works for me, i will try another codec next time. also, i've noticed that i've inadvertently taken away the previous enum-order-based sorting so i am at this moment inserting a # column before the translated name column populated with i++ and default sorted asc so that the visible ordering sequence stays the same with the change.

edit: i'll try and make another demo video tonight with all the integrated PR's together and maybe upload different encodings of it, and that would then be my finish line for a while until the PR board clears itself up a little …

edit again: before the question comes, yes it is not nice to show internal untranslated names of enums in the gui, but it makes total sense with the other PR to export/edit/import the colors .. otherwise we could export the localized names but this is off topic here i should wight it there …

ahhh alll the writininnnnngs chears 🍷 🍻

zs-stpa:
update: well you know how it goes. sorting color by hex may have its use … but by hsb or luminance is also valid, so i am now cycling the sorting order between the six (each asc/desc) and tried to make transient messages … but those are in another PR ………… this is really fun

zs-stpa:
updated short color sorting demo: https://github.com/zs-stpa/omegat/blob/integration/all-open/src/docs/developer/media/2026-07-28-colors-preferences-table-sorting-demo.mp4

to be committed here any moment. the full integration demo will have to wait for another day.


https://github.com/user-attachments/assets/c80d1a9e-0db5-480f-ad74-29497cbe2e9d


edit: you can see it is getting late .. me thinking 'how can i order colors' and then after the fact i see in the demo video 'Swatch hsv hsl rgb cmyk' .. oh kay now i _could_ implement more sorting directions for colours but i will not shuffle them. promise. 🫣 ok jokes aside, no more sorting directions for colors, those have to be enough for now or the near future or whenever the next build comes out.

edit again: in the full integration the count is not 59 like in this singular branch but now at 62 colors because some other tickets asked for added ways of coloring segments.

aand sry but no integration demo vid, it would just take too long to cram it all into one session, and be outdated soon again anyway

miurahr:
No need mp4 blob.

miurahr (changes requested):
We don't need mp4 blob.

zs-stpa:
> We don't need mp4 blob.

the mp4 is no longer part of this PR, only linked in the comment from another branch on my fork as an example what it would look like.

src/main/java/org/omegat/gui/preferences/view/CustomColorSelectionController.java
```diff
@@ -221,19 +422,56 @@ public void persist() {
...
             return values()[index];
         }
     }
 
+    /**
+     * Renders a cell as plain text or, in swatch mode, as a colour sample.
+     */
+    @SuppressWarnings("serial")
+    static class ColorCellRenderer extends DefaultTableCellRenderer {
+        private final ColorIcon swatch;
```
miurahr:
```suggestion
        private final @Nullable ColorIcon swatch;
```

src/main/java/org/omegat/gui/preferences/view/CustomColorSelectionController.java
```diff
@@ -221,19 +422,56 @@ public void persist() {
...
+
+        ColorCellRenderer(int swatchSize) {
+            this.swatch = swatchSize > 0 ? new ColorIcon(swatchSize) : null;
+        }
+
+        @Override
+        protected void setValue(Object value) {
```
miurahr:
```suggestion
        protected void setValue(@Nullable Object value) {
```

miurahr (changes requested):
For null-safety, please add `@Nullable` annotation for fields which can be null.

src/main/java/org/omegat/gui/preferences/view/CustomColorSelectionController.java
zs-stpa: done

miurahr (approved):

miurahr:
Thank you for great contribution!

PR #2193, opened 2026-07-28
feat(editor): mark and navigate segments where source=target (SF-RFE-1051)

zs-stpa:
## Pull request type

- Feature -> [feature]

## Which ticket is resolved?

Mark and go to segments where source = target - https://sourceforge.net/p/omegat/feature-requests/1051/

## What does this PR change?

- Adds a View menu option "Highlight Identical Segments (source = target)" that colours, in a dedicated editor colour, segments whose translation is identical to the source. Off by default.
- Adds Go To > Next / Previous Identical Segment navigation to jump between those segments. "Next" is bound to Ctrl+Alt+I (Cmd+Opt+I on macOS); "Previous" ships without a default shortcut.
- The identical verdict is taken from the stored translation (getTranslationInfo), not from the displayed strings, so untranslated segments that merely echo the source in the target area are never marked.
- New editor colour COLOR_MARK_IDENTICAL (default #c8e6c9, in both light and dark colour schemes); it shows up in Preferences > Colours and, since #2183, applies without a restart.

## Other information

Implementation notes for review:

- The new IEditor / IEditorSettings members (nextIdenticalEntry, prevIdenticalEntry, isMarkIdentical, setMarkIdentical) are declared as default methods, so only the real editor and settings implementations override them; the console and test stubs need no change. This keeps the change surface small and avoids touching every IEditor implementation.
- The navigation reuses the existing iterateToEntry plumbing that already powers Next/Previous Untranslated, Unique, Note and tm/auto|enforce.
- The ParameterNumber checkstyle suppression for EditorSettings (its getAttributeSet has 8 parameters) is changed from a line-pinned entry to a file-wide one, matching the many sibling entries (RealProject, Searcher, NearString, ...). The line-pinned form silently breaks whenever unrelated edits shift the method's line number.
- "Previous" navigation is deliberately shipped without a default shortcut: it avoids a four-key combo, and once the segment sort bar (SF-1094) lands with reversible ordering, the "previous" direction of these navigation commands becomes largely redundant anyway - one can reverse the order and keep using "next".

Tests: IdenticalSegmentMarkerTest covers the enabled/disabled, identical, different-translation and untranslated-echo cases (the existing autotmx/auto1.tmx fixture gained an identical translation unit). Verified visually with a small EN->DE project mixing identical, differing and untranslated segments; the highlight and both navigation directions behave as expected.

### DIFF
--- a/config/checkstyle/suppressions.xml
+++ b/config/checkstyle/suppressions.xml
@@ -99,7 +99,7 @@
     <suppress files="(BidiUtils|Java8Compat|MagicComment|JnaLoader)\.java" checks="HideUtilityClassConstructor"/>
     <suppress files="JTextPaneLinkifier\.java" checks="EmptyBlock"/>
     <suppress files="RepositoriesMappingController\.java:" checks="MethodLength"/>
-    <suppress files="EditorSettings\.java" checks="ParameterNumber" lines="525"/>
+    <suppress files="EditorSettings\.java" checks="ParameterNumber"/>
     <suppress files="PreferencesWindowController\.java" checks="MethodLength"/>
     <suppress files="MatchesVarExpansion\.java" checks="InnerAssignment"/>
     <suppress files="EditorTextArea3\.java" checks="MethodLength"/>
--- a/src/main/java/org/omegat/gui/editor/EditorController.java
+++ b/src/main/java/org/omegat/gui/editor/EditorController.java
@@ -1596,6 +1596,29 @@ public class EditorController implements IEditor {
         iterateToEntry(true, ste -> ste.getDuplicate() != DUPLICATE.NEXT);
     }
 
+    private void identicalEntry(boolean forward) {
+        iterateToEntry(forward, ste -> {
+            TMXEntry info = Core.getProject().getTranslationInfo(ste);
+            return info.isTranslated() && info.translation.equals(ste.getSrcText());
+        });
+    }
+
+    /**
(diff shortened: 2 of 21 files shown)


zs-stpa:
ongoing discussion on the dev-list so i am converting this to a draft

zs-stpa:
discussion has ended, reopening.

brandelune:
> discussion has ended, reopening.

Have you asked whether the discussion had ended ? Is there a consensus that was reached ? If yes can you report it here ?

zs-stpa:
last active messages were end of July, no i have not asked whether it has ended, because that seemed long enough, consensus as i understood it at the time was

- all source=target segments are marked when isMarkIdentical = true
- literal equality without trimming or character folding, but only exact byte identical equality including tags and all
- nothing is changed in project.tmx or other output, the mark is calculated at runtime only

brandelune:
> last active messages were end of July, no i have not asked whether it has ended, because that seemed long enough, consensus as i understood it at the time was
> 
> * all source=target segments are marked when isMarkIdentical = true
> * literal equality without trimming or character folding, but only exact byte identical equality including tags and all
> * nothing is changed in project.tmx or other output, the mark is calculated at runtime only

Ok, that’s the proposed conclusion mail that you should send to the list, just to confirm that we have indeed reached an agreement and not forgotten about the issue.

miurahr (approved):
LGTM. especially for code and tests.

PR #2195, opened 2026-07-28
feat(gui): export and import a colour scheme (RFE-1399)

zs-stpa:
## Pull request type

- Feature -> [feature]

## Which ticket is resolved?

Importing/exporting custom colour themes - https://sourceforge.net/p/omegat/feature-requests/1399/

## What does this PR change?

- Adds **Export Scheme...** and **Import Scheme...** buttons to the Colours preferences panel. A colour scheme is a plain, human-readable Java `.properties` file with one `EditorColorName = #rrggbb` line per colour, in enum order — exactly the shape of the bundled `ColorScheme_light.properties` / `ColorScheme_dark.properties` and of the per-colour preference keys, so nothing new has to be learned or parsed.
- **Export** writes the currently effective colours, including edits that are still staged in the panel but not yet saved. It refuses to clobber an existing file without a yes/no confirmation, and remembers the folder for next time.
- **Import** reads such a file back, stages the colours into the table (they appear immediately) and applies them when you confirm with **Apply** or **OK**, so **Cancel** discards them. Keys that are not known editor colours are ignored, so schemes stay forward-compatible across versions; malformed values are skipped.
- **Restore defaults** in this panel now asks for confirmation, but only when the colours actually differ from their defaults.
- Export, import and reset each show a short confirmation with the number of colours involved, in the preferences window's message area — the same place the restart/reload notices appear. This goes through a new, small transient-message channel added to the preferences controller interface as a `default` method, so only the window implements it and no other controller has to change.
- Documentation: a paragraph in the Colours section of the manual and a "Share your colour scheme" tip of the day (en + de). A `*.hprof` rule was added to `.gitignore` (heap dumps from GUI runs must never be committed).
- The second commit is a small, related fix (see Other information): the Appearance panel no longer wrongly demands a restart after applying colours.

## Other information

**Scope / terminology.** This request is specifically about the *editor colours* — the marks and highlights configured under Preferences > Colours (the `Styles.EditorColor` values). It is deliberately **not** about the overall *look and feel* (the Swing Look-and-Feel / theme), which is chosen separately under Appearance and is provided by theme plugins. The two are easy to conflate, so the wording throughout (buttons, manual, tip) says "scheme" and keeps the look-and-feel out of it. The look-and-feel / theme side is already covered by earlier, closed tickets — e.g. SF-1567 (support theme plugins, choose theme in preferences), SF-1580 (theme plugin adjusts colour scheme/style), SF-1540 (add FlatLaf look and feel), SF-1513 (custom look and feel), SF-1653 (UI FlatLaf, wont-fix). SF-1514 (externalise default colour definitions) is what produced the `ColorScheme_*.properties` files this exchange format mirrors.

**Context — this is the logical continuation of recent colour work.** It builds directly on #2183 (apply colour changes without a restart, plus the Apply button this feature reuses), and follows #2182 (mark segments pre-translated in the source file), #2188 (keep enforced translations highlighted, SF-1171), #2190 (colour tags for right-to-left targets, SF-1122) and #2193 (mark segments where source = target, SF-1051). Import/export is the natural next step once colours are customisable and apply live: it lets users keep, share and version their colours.

**Only partly applied live so far — a dedicated follow-up PR is prepared.** #2183 made *marks* apply live, and this PR relies on that. However, several side panes (fuzzy matches, dictionary, machine translation, comments, notes, segment properties, the project files window, search results, glossary rendering, the issue providers) still read their colours once at construction, and `MatchesTextArea` caches some as static attribute sets. So a freshly imported colour that only those panes use will show fully only after a restart. Rather than sprinkle partial fixes here, the complete solution — a `CoreEvents.fireColorsChanged()` broadcast modelled on the existing `fireFontChanged()`, with every colour-consuming component re-applying on the event — is being done as its own focused PR that completes #2183 across the whole UI. Happy to reorder or fold that in if you prefer.

**The Appearance restart fix (second commit) — happy to split out.** While testing, applying a colour change raised "Restart OmegaT to apply these changes" on a fresh configuration, even though the change applies live. Root cause: the Colours panel is a child node of Appearance, and `AppearanceController` reported a spurious restart-required — its menu-style combo listener was registered before `initFromPrefs`, so populating the combo fired `setRestartRequired(isModified())` before the theme radios were initialised, and `isModified()` compared the menu style against the raw (empty) preference instead of its effective default. The fix registers the listener after initialisation (consistent with the theme listeners already there) and compares against the default the combo is filled with; a real theme/menu change still requests a restart. It is an independent, pre-existing bug and can move to its own PR if that is cleaner.

**How this was reviewed / the feedback rounds.** The feature went through several rounds of hands-on visual checks and adjustments before this PR:
1. Format decision: a plain `.properties` file was chosen over JSON/XML because it matches the existing colour-scheme files and the preference storage and needs almost no new code.
2. Naming: switched from "theme" to "scheme" (English) / "Farbschema" (German) to keep it distinct from the look-and-feel theme.
3. Export contents: it exports the 42 colours that have a concrete value; the 16 that have no fixed value follow the active look and feel and are intentionally omitted, with a short note in the file header (which also carries a shortened, GPL/free-and-open project header).
4. Import semantics: reworked from "apply immediately" to "stage in the table, apply only on Apply/OK, discard on Cancel", per request.
5. Usability: the buttons remember the last-used folder; a short confirmation is shown, and it was moved from an ad-hoc label into the window's message area to match the restart/reload notices; the confirmation now reports the number of colours; export asks before overwriting; reset asks for confirmation and also reports success.
6. A final, deliberately adversarial code review (fresh reviewer) flagged: exporting over an existing file without confirmation, the mnemonic `&` leaking into dialog titles, and the confirmation always claiming success — all fixed here. It also confirmed that Cancel truly discards (a fresh controller is created each time the dialog opens) and that the Appearance fix does not suppress any legitimate restart prompt. Lower-severity notes (OOM on a deliberately huge file, freezing a look-and-feel-derived colour via a hand-edited file, no auto-added extension, the programmatic button placement relying on the generated form's container) were considered and left as-is.

Tested manually with small EN->DE / EN->AR / EN->HE projects: export produces a readable file, import stages and applies as described, Cancel discards, overwrite and reset prompts behave, and the counts and confirmations appear in the right place. Full test suite and checkstyle are green.

### screenshots

export …
Bildschirmfoto 2026-07-28 um 17 57 00

edit ...
Bildschirmfoto 2026-07-28 um 17 57 54

import ...
Bildschirmfoto 2026-07-28 um 17 58 20

integrating my/your/our/these (take a pick) PR's locally …

Bildschirmfoto 2026-07-28 um 17 47 45

noticed merge conflict, amending cause, less conflict surface, blah blah i could produce walls of text faster than you could read and i am sorry for that and this will not go on forever for sure but right now i'm having a productive phase, so, here it is. next time less text, promised.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -46,3 +46,6 @@ GEMINI.md
 .github/copilot-instructions.md
 .supermaven/
 .idx/
+
+# heap dumps from GUI/OOM test runs
+*.hprof
--- a/src/docs/manual/en/OmegaT_Preferences.xml
+++ b/src/docs/manual/en/OmegaT_Preferences.xml
@@ -399,10 +399,21 @@
             <link linkend="windows.source.files.list" endterm="windows.source.files.list.title"></link>
             window can also be changed here.</para>
          <para>Scripts can be used to set predefined themes. OmegaT is bundled with
-            a script called 
+            a script called
             <link linkend="windows.scripts.distribution.switch.colour.theme" endterm="windows.scripts.distribution.switch.colour.theme.title"></link> that
-            provides a default dark theme. See the 
+            provides a default dark theme. See the
             <link linkend="windows.scripts" endterm="windows.scripts.title"></link> window for details.</para>
+         <para><guibutton>Export Scheme...</guibutton> writes your current colours
+            to a plain, human-readable <literal>.properties</literal> file (one
+            colour per line), which you can keep, share or edit in any text
+            editor. <guibutton>Import Scheme...</guibutton> reads such a file
+            back: the imported colours appear in the table at once and are
+            applied when you confirm with <guibutton>Apply</guibutton> or
+            <guibutton>OK</guibutton>, so <guibutton>Cancel</guibutton> discards
+            them. Unknown entries in an imported file are ignored, so schemes
+            stay compatible across versions. This exchanges only the editor
+            colours listed here, not the overall look and feel, which is chosen
+            separately under <guilabel>Appearance</guilabel>.</para>
          <note>
             <para>Automatic dark theme detection is available on Linux systems
                		with the Gnome/GTK look and feel for users who run OmegaT on OpenJDK.
--- a/src/docs/tips/de/share_colour_scheme.html
+++ b/src/docs/tips/de/share_colour_scheme.html
@@ -1,0 +1,44 @@
+<html lang="de">
+<!--
+  ~  OmegaT - Computer Assisted Translation (CAT) tool
+  ~           with fuzzy matching, translation memory, keyword search,
+  ~           glossaries, and translation leveraging into updated projects.
+  ~
+  ~  Copyright (C) 2026 Stephan Pakebusch
+  ~                Home page: https://www.omegat.org/
+  ~                Support center: https://omegat.org/support
+  ~
+  ~  This file is part of OmegaT.
+  ~
+  ~  OmegaT is free software: you can redistribute it and/or modify
+  ~  it under the terms of the GNU General Public License as published by
+  ~  the Free Software Foundation, either version 3 of the License, or
+  ~  (at your option) any later version.
+  ~
+  ~  OmegaT is distributed in the hope that it will be useful,
+  ~  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  ~  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  ~  GNU General Public License for more details.
+  ~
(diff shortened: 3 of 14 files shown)


zs-stpa:
french spell check test failing on windows machine? je ne comprois pas, pas de toute … da bin ich mit meinem latein leider am ende - looks greek to me as the british would say and probably unrelated to the actual content of this PR

zs-stpa:
did the force push to retrigger CI testing, all green now

zs-stpa:
rebased on current master

PR #2196, opened 2026-07-28
feat(gui): apply colour changes live across panels (follow-up to #2183)

zs-stpa:
## Pull request type

- Feature -> [feature]

## Which ticket is resolved?

No SourceForge ticket: follow-up to #2183, making colour changes apply live everywhere.

## What does this PR change?

- New broadcast `CoreEvents.fireColorsChanged()` + `IColorsChangedEventListener`, mirroring the existing `fireFontChanged`.
- Fired on colour Apply / reset / restore-defaults in the Colours preferences.
- Consumers re-apply on the event, no restart: editor pane; the `EntryInfoPane`-based panes (fuzzy matches, dictionary, machine translation, comments, notes); fuzzy match diff colours (static caches -> instance); segment properties; glossary; search results; and the project files progress column.

## Other information

Completes #2183 for panes that read their colours once at construction; marks were already live. A few edge consumers (Document3, the linkifier, docking UI, flash interpolator, issue providers) still to follow.

Maintainer decision requested: this PR again hit a **line-pinned checkstyle suppression** (`ParameterNumber` with `lines="N"` for `EntryListPane`) that broke only because unrelated edits shifted the method's line number; I switched it to a file-wide suppression, matching its siblings. These line-pinned entries recur as a maintenance/CI nuisance. Please decide the policy: (a) drop line pinning entirely (file- or method-scoped) — never breaks on line shifts, at the cost of slightly less precision (a future extra >7-parameter method in the same file would be silently allowed); or (b) keep line pinning for precision and accept the churn, e.g. only (re)pin shortly before releases.
whichever decision you will reach, i will try and comply, it just seems a little inconsistant at the moment.


zs-stpa:
Update: extended the live colour refresh to the views that were still stale or unthemed — the segment-properties **table**, the **Tip of the Day** dialog (restyles while it is open behind Preferences), and the **first-steps / empty-project** pages (background, text colour and font). Added a small `Styles.applyColors(Component)` helper so a view reacts with a single call, and folded everything into one commit. Colour consumption is still opt-in per view for now; a self-unregistering `bindColors(...)` primitive and a plugin colour-contribution SPI are planned as a follow-up.
correction: un-schemed not unthemed

zs-stpa:
**Update (2026-08-04):** second commit puts colour initialisation on solid ground and implements ADR 2026004 (#2206) decisions A + B.

- Every `Styles.EditorColor` entry now uses one canonical theme-backed constructor (UIManager key + fallback mirroring the bundled light scheme, test-enforced) and resolves its default on every read — no more class-load freezing, theme switches apply without restart. The one- and three-argument constructors are gone.
- Both bundled themes define every `OmegaT.*` key and reference the keys through `EditorColor`, so a typo no longer compiles. This fixes three dead dark-theme keys (`markComesFromTm`, `markComesFromTxXice`, `markComesFromTmXendorced`) whose dark values never reached the editor.
- The dark theme lightens its pen colours (spell check, terminology, search/replace, match diff, bidi, search-field error) for readability; background fills untouched.
- Former inherit entries default to the theme's own base fg/bg and render as before, with two disclosed exceptions: repeated segments are now tinted with the alternating hilite by default (replacing the translated/untranslated fill there), and a customised base fg/bg is no longer inherited into those entries.
- New: `docs/developer/media/editor-colors/` — colour overview SVG plus micro-generator-skript including GPL headers.

Editor colours overview

Porting note: a new colour is one line, `COLOR_FOO("OmegaT.foo", "#rrggbb")`, plus Bundle key and values in both themes and both ColorScheme files (a test reminds you).

Open question: before/after GUI screenshot tests per corrected colour would be possible via AssertJ-Swing, but with #2204 still moving they may freeze an intermediate state — opinions welcome.

zs-stpa:
all green, CI is passing now but not before, that looked like an unrelated fluke

zs-stpa:
moved the SVG out of this PR into another branch on my fork

zs-stpa:
superseeded by #2239 which is already merged.

PR #2175, opened 2026-07-25
docs(adr): documenting editor segment sorting/filtering separation (SF-RFE-1094)

zs-stpa:
## Pull request type

- Documentation -> [documentation]

## Which ticket is resolved?

- Sorting segments
  * URL https://sourceforge.net/p/omegat/feature-requests/1094/

## What does this PR change?

- Adds ADR 2026003 "Separation of Segment Sorting and Filtering in the Editor" plus its entry in the ADR index (documentation-only change).
- Documents the design decision requested in the review of #2161: sorting is decoupled from filtering through a dedicated minimal core hook (`IEditorSorter`, the counterpart of `IEditorFilter`), with the sorter model and UI kept outside the core.
- Covers why sorting inside `IEditorFilter` was rejected, the hook's contract (stable total order, natural-order fallback, applied after filtering, unchanged code path when no sorter is set), the `IEditor` API symmetry (`setSort`/`getSort`/`removeSort`), the per-file scope, and navigation by exact entry number.

## Other information

Related PRs implementing the design: #2161 (hook), #2162 (collapsible bar), #2163 (sort bar and model), #2154 (numeral parsing foundation).

PR #2206, opened 2026-08-02
docs(devel/adr): add ADR for theme, color, and presentation architecture

miurahr:
ADR#2026003 proposed an issue around UI/UX for accessibility in the PR #2204.
This is additional ADR that focus in the architecture in the OmegaT core how to handle colours of theme/markers/preferences.

Unfortunately, OmegaT core has an inconsistent in the architecture. This ADR propose a design and direction for the fix.

## Pull request type

## What does this PR change?

-
-
-

## Other information

Ref #2204

### DIFF
--- a/src/docs/developer/adr/2026004.EditorColorRegistryConsistency.md
+++ b/src/docs/developer/adr/2026004.EditorColorRegistryConsistency.md
@@ -56,17 +56,22 @@ cheaper to unify the registry first.
 
 ## Decision
 
-**Decision A.** Every `EditorColor` entry will have a real, theme-resolvable default. Concretely:
+**Decision A.** Every `EditorColor` entry will have a real, theme-resolvable default,
+reached via the standard two-argument, theme-backed constructor:
 
 - The three Source Files progress colours gain corresponding `OmegaT.*` keys in both
-  `DefaultFlatLightTheme` and `DefaultFlatDarkTheme` (using the current hardcoded values as
-  the starting colour choice, subject to normal visual/contrast review), and move from the
-  three-argument constructor to the standard two-argument, theme-backed constructor.
+  `DefaultFlatLightTheme` and `DefaultFlatDarkTheme`, and move off the three-argument
+  constructor.
 - The ~15 theme-less entries gain corresponding `OmegaT.*` keys in both bundled themes as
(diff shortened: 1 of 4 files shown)


brandelune (approved):

zs-stpa:
i just noticed, that maybe we can change the naming convention of the ADR's so they do not as easily conflict between branches=?

Note: three open PRs currently claim ADR number 2026003, so some will need to renumber or rename…

- https://github.com/omegat-org/omegat/pull/2175 — 2026003.EditorSortingFilteringSeparation.md
- https://github.com/omegat-org/omegat/pull/2159 — 2026003.NumberOnlyAutofill.md
- https://github.com/omegat-org/omegat/pull/2204 — 2026003.PresentationAndAccessibility.md

src/docs/developer/adr/2026004.EditorColorRegistryConsistency.md
```diff
@@ -0,0 +1,122 @@
...
+## Decision
+
+**Decision A.** Every `EditorColor` entry will have a real, theme-resolvable default,
+reached via the standard two-argument, theme-backed constructor:
+
+- The three Source Files progress colours gain corresponding `OmegaT.*` keys in both
+  `DefaultFlatLightTheme` and `DefaultFlatDarkTheme`, and move off the three-argument
+  constructor.
+- The ~15 theme-less entries gain corresponding `OmegaT.*` keys in both bundled themes as
+  well, and move off the one-argument constructor.
```
zs-stpa, line 66: ok doing it now, finishing the final touches, generating a generator-skript to generate svg preview of 'new' old color schemes as #2196 follow-up

edit: old in the sense that, the a11y roadmap is not as utopian as it may ~sound~ _appear_ 🤷, and not much use putting energy into user documentation and manuals with every step of the way, would have to rewrite half of that with every following step anyway, and would be outdated within days …

miurahr:
> i just noticed, that maybe we can change the naming convention of the ADR's so they do not as easily conflict between branches=?

we can reserve the number in the dev list discussion. ADR is a basically wrap up of the dev list discussion. In recent days, I wrote ADR draft and start discussion on dev list. This way help us to manage a focused discussion among members.

PR #2208, opened 2026-08-02
feat(gui): marker styles

zs-stpa:
## Pull request type

## Which ticket is resolved?

- "Match display template" including ${diff} (styling part)
  * https://sourceforge.net/p/omegat/feature-requests/1793/

Also accessibility groundwork ("information never colour-only") in direction of draft ADR on presentation and accessibility.

## What does this PR change?

Configurable text style per marking, first stage:

- Text-marking colour entries get bold/italic/strikethrough/underline flags, edited via four checkbox columns with header tooltips; only entries with wired call sites (15: editor segment states, match pane) show checkboxes, cells stay empty elsewhere.
- Entries carry intrinsic defaults, e.g. match diff: struck-through deleted, underlined inserted, active variants additionally bold. Table shows shipped rendering; Restore Defaults returns to it. Storage: `<NAME>_TEXT_STYLE_<FLAG>` preferences with `__DEFAULT__` sentinel, so untouched installations follow later default changes; key scheme stays forward-compatible with richer line styles (dotted/dashed/wave underline) planned as follow-up.
- `Styles.overlayTextStyle()` applies flags additively, existing view options (e.g. bold source text) keep working; style changes refresh built segments, colour-only changes stay refresh-free.

Remaining marker call sites (glossary hits, MT origin, removed tags, panes) and inline style preview cell follow in next round.

## Other information

Full test suite and checkstyle pass; verified visually on macOS.

Follow-up PRs stacked on this one: #2322 (red deletion tones), #2323 (default template with ${diff}). Known limitations, follow-up material: active fuzzy match selection marked by bold (not accessible, collides with bold flags); flags attach to segment-state rows, not *_FG text rows.

### short sample screen recording

will not be updated if details change in this PR
https://github.com/user-attachments/assets/86eb7a4b-eaab-4196-9653-ba21c4933ddf


zs-stpa:
empty force push to retrigger CI

PR #2214, opened 2026-08-03
fix(teamsync): robustness

zs-stpa:
## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?

- (no ticket yet; found in the field while migrating a 3.x-era team project to OmegaT 6.2 — can file a SourceForge bug ticket on request)

## Motivation

There are still users out there who seem to have a need for the long-term stable 3.x version; I suspect this might be due to transitioning pains. See also
https://sourceforge.net/projects/omegat/files/OmegaT%20-%20Standard/OmegaT%203.6.0%20update%2010/stats/timeline?dates=2020-05-05%20to%202026-08-03

## What does this PR change?

Four robustness fixes for team project synchronization (`org.omegat.core.team2`), each as its own commit, plus generated developer diagrams:

1. **Keep path and versions in the prepared team sync info.** The asynchronous save path returned a `PreparedFileInfo` without `path`/`versionHead`, so every later `commitPrepared()` failed with an NPE and autosaves never uploaded on that route.
2. **Fail fast when a repository cannot be initialized.** A swallowed init failure (e.g. a stale `config.lock`) left the repositories list out of sync with the definitions and surfaced later as an unrelated `IndexOutOfBoundsException`.
3. **Explain failed version checkouts in user terms.** Raw backend errors such as JGit's `Missing unknown <hash>` are wrapped with the file, the version, the repository URL and why that version was needed.
4. **Recover from a stale saved base version.** If the stored sync marker provably points at history that no longer exists (rebuilt repository/local copy), local and remote are rebased as a three-way merge against an empty base and the marker is healed, instead of the project becoming unloadable. Transient failures (network, locks) still fail loudly and keep the marker.

New `RebaseAndCommitTest` (19 tests) covers the async prepare/rebase/commit cycle, stale-marker recovery, transient-error classification and re-entry with the returned info.

## Other

edit: moved the SVG diagrams (with their Python generator, all incl. omegaT GPL header) out of this PR into another branch on my fork

Team sync architecture
Team sync flow
Team sync failure chain
Repository migration


miurahr:
It is important to run [an Integration Test](https://omegat.readthedocs.io/en/latest/34.IntegrationTest.html) with enough long duration ex. >30min when we touch code related to team feature.
Could you run the integ test?

zs-stpa:
10min integration run done with speed-up acceleration configurable (17cd4edb7)  `-Domegat.test.delay=2` — maximum random delay between simulated translator's edits, default unchanged at 15s. --> ~1h of normal-speed sync cycles: 693 saves, 469 complete rebase cycles, 486 remote commits, 274 conflict events (one node alone hit "Push failed. Will be synchronized next time." 103 times). All three nodes finished OK and final revision check reported "All commits look good". Target was local bare git repository over file:// with gc disabled, as recommended in javadoc.

also fixed what kept the documented invocation from starting: gradle task forwarded unset `-Domegat.test.*` properties as empty strings, so TestTeamIntegration mistook empty map.repo for a real mapping and died with "There is no repository url". Only set properties are forwarded now, including `omegat.test.map.type`, which was missing from the list.

zs-stpa:
i moved the slides to another branch on my fork, out of this PR

src/main/java/org/omegat/core/team2/RebaseAndCommit.java
```diff
@@ -71,7 +76,20 @@ public static PreparedFileInfo prepare(RemoteRepositoryProvider provider, File p
...
+        try {
+            baseFile = provider.switchToVersion(path, currentBaseVersion);
+        } catch (Exception ex) {
+            if (!isMissingVersion(ex)) {
+                throw ex;
+            }
+            // The saved base version no longer exists in the repository, e.g.
+            // after the repository or the local copy was rebuilt. Skip the
+            // preparation; the synchronous rebase recovers via an empty base.
+            LOGGER.atWarn().logRB("TEAM_BASE_VERSION_LOST", path, currentBaseVersion);
```
miurahr: We now have functions in `Log` utility class to support warn and debug.

miurahr:
@brandelune  could you lead to check UI message changes?

src/main/java/org/omegat/core/team2/RebaseAndCommit.java
miurahr: For example,
```Log.logWarningRB("TEAM_BASE_VERION_LOST", path, currentBaseVersion);```

for debug, there is `Log.logDebug()`

miurahr (approved):
Good direction to have a single purpose `RebaseAndCommit.resolveBase` function. LGTM against integration test improvement. Thanks.

src/main/java/org/omegat/core/team2/RebaseAndCommit.java
zs-stpa: done

PR #2216, opened 2026-08-04
fix(build): reduces distribution size by deduplication of shipped libraries

zs-stpa:
## Pull request type

- Bug fix

## Which ticket is resolved?

No ticket; follow-up to #2215 (based on its two commits).

## What does this PR change?

Module jars no longer bundle third-party jars that the application already ships in lib/: modules load through a parent-first class loader, so the bundled copies are unreachable dead weight. Dependency jars whose file name matches an artifact on the root runtime classpath are skipped during the merge - conservative on purpose, any version drift bundles the artifact again. The modules/ folder shrinks from 253.7 MB to 200.3 MB with byte-identical results across build invocations; the largest wins are the German, English, Dutch and Russian language modules (about 11 MB each (edit: please excuse the mixup between compressed size and unzipped size)).

A new `distributionSmokeTest` task (excluded from the default test task, same pattern as `memoryStressTest`) proves plugin loading, the bundled Morfologik spelling dictionary and LanguageTool checks for English and German still work against the installed distribution.

Treemap - what is inside the language modules

## Other information

Full test suite, checkstyle and the new smoke tests pass; the distribution keeps the same 44 module files.

### DIFF
--- a/build-logic/src/main/groovy/org/omegat/module/OmegatModulePlugin.groovy
+++ b/build-logic/src/main/groovy/org/omegat/module/OmegatModulePlugin.groovy
@@ -76,10 +76,21 @@ class OmegatModulePlugin implements Plugin<Project> {
 
         Map<String, Object> manifestAttrs = buildManifestAttributes(project)
 
+        // Jars the application itself ships in lib/ must not be merged into the
+        // module fat jar: modules load through a parent-first class loader, so the
+        // application's copy always wins and a bundled duplicate is unreachable dead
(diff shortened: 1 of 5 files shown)


miurahr:
This smoke test is valuable for running as CI.

A CI "Source Distribution test" try to find a duplication of dependencies among core and modules in the source distribution. It also try to find a missing dependencies for module buid.
It seems a similar purpose to detect the dependencies issue.

I would like to take an approch in the other topic branchy/PR.

miurahr (approved):

PR #1321, opened 2025-04-04
feat: Improve CLI command line parser

miurahr:
Improve CLI command line parser with picoli library.
It also propose the improvement for syntax of CLI.

## Pull request type

- Feature enhancement -> [enhancement]

## What does this PR change?

- PicoCLI Integration:
    -  Replaces the custom CLIParameters class with new PicoCLI-based classes: Parameters, Command, and various subcommands like AlignCommand, StatsCommand, StartCommand, and TranslateCommand.
- Commands are now modular and support direct execution logic, enabling easier readability and lifecycle management.
- Refactor Core and Project Handling:
    -  CLIParameters references are removed throughout the codebase.
    -  New runtime preferences for CLI parameters like "no-team" and tokenizer options are added in RuntimePreferences.
- Remove Old CLI Logic:
    -  Legacy CLI-related code is cleaned up from multiple classes, including Main, Core, ProjectPropertiesDialog, and others.
    -  Deprecated test classes related to CLIParameters are fully removed.
- New CLI Feature Implementations:
    -  New commands such as team, stats, translate, align, and start are introduced, each handling a distinct aspect of the application's functionality.
- Testing Impact:
    -  Updates and additions in the test cases to reflect the enhanced console mode structure and behavior, while legacy tests are replaced or removed.

## Other information

- related#1506 the converter become dead code 
- [ ] functional tests for all sub commands

PR #2219, opened 2026-08-04
fix(cli): apply --config-dir again and restore legacy option handling

zs-stpa:
## Pull request type

- Bug fix

## Which ticket is resolved?

No ticket; regression from the command line parser rework (#1321).

## What does this PR change?

`--config-dir` was still declared but no longer applied anywhere. It is now applied in `Main` before the plugins load - user plugins live below the configuration folder, so it must take effect before picocli parsing (same pre-scan approach as `--dev-manifests`) - and in `LegacyParameters.initialize()` for the picocli routes.

Two more losses from the same rework are restored: the legacy syntax without a sub-command skipped `initialize()` entirely, so `--config-file` only worked with sub-commands, and `--no-team` was ignored because `legacyParams` was never handed to `StartCommand`. The GUI restart command line now forwards the configuration folder as well.

New tests cover the argument pre-scan and the option application including tilde expansion.

## Other information

Full test suite and checkstyle pass. Verified against the installed distribution: `OmegaT --config-dir <dir>` writes omegat.prefs, uiLayout.xml and logs below the given folder again.

### DIFF
--- a/src/main/java/org/omegat/Main.java
+++ b/src/main/java/org/omegat/Main.java
@@ -10,6 +10,7 @@
                2014 Alex Buloichik
                2018 Enrique Estevez Fernandez
                2022-2025 Hiroshi Miura
+               2026 Stephan Pakebusch
                Home page: https://www.omegat.org/
                Support center: https://omegat.org/support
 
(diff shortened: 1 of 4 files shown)


miurahr (approved):
Good catch, thank you. LGTM.

PR #2220, opened 2026-08-04
fix(core): create the user scripts folder under its documented name

zs-stpa:
## Pull request type

- Bug fix

## Which ticket is resolved?

No ticket; follow-up to the user scripts initialization introduced for RFE#1781 (#1869).

## What does this PR change?

`determineUserScriptsPath()` mixed up two folder names: it created the user scripts folder as `script` - the name of the unrelated segment-export folder - although its own javadoc, `ensureUserScriptsDir()` and the distribution all use `scripts`. It also tried to seed the default scripts from the non-existent install folder `script`, which logged "It was not possible to create the script folder" on every first start (`Files.copy` cannot copy a directory's contents anyway, and `ensureUserScriptsDir()` does the real seeding right afterwards). The fallback in `ensureUserScriptsDir()` additionally joined the install path without a separator.

The folder is now created under `scripts`, a `script` folder left behind by earlier builds is renamed to preserve user scripts, the broken seeding block is gone and the path join is fixed. Unit tests cover first start, migration, a legacy file of the same name and pre-existing folders.

## Other information

Full test suite and checkstyle pass. Verified against the installed distribution: a first start with a fresh home no longer logs the error and seeds the default scripts into `scripts`.

### DIFF
--- a/src/main/java/org/omegat/util/StaticUtils.java
+++ b/src/main/java/org/omegat/util/StaticUtils.java
@@ -11,6 +11,7 @@
                2014 Aaron Madlon-Kay, Alex Buloichik
                2015 Aaron Madlon-Kay
                2023 Hiroshi Miura
+               2026 Stephan Pakebusch
                Home page: https://www.omegat.org/
                Support center: https://omegat.org/support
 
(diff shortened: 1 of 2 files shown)


miurahr:
Thank you. This is my typo. It should be fixed in 6.1.2 maintain release, too.

miurahr (approved):

PR #2223, opened 2026-08-04
fix(usabilty): user distinguishable plugin names in the preferences->plugins table

zs-stpa:
## Pull request type

- Bug fix

## Which ticket is resolved?

No ticket; based on #2216 (shares the module build-logic file).

## What does this PR change?

The plugins preferences table listed dozens of identical rows (21x "File filters[bundle]" etc.). Names shared via package-level manifest sections or multi-plugin jars now get the class appended - "File filters: PoFilter" - and the "[bundle]" suffixes are gone. Module manifests repaired along the way: real Plugin-Version instead of "unspecified" (a duplicate write clobbered it), Plugin-Bundled added, and the Built-Date attribute dropped (made every module jar out of date on each configuration; nothing consumes it).

## Other information

all 147 plugin rows are distinguishable in the installed distribution.

Bildschirmfoto 2026-08-05 um 00 04 09

### DIFF
--- a/build-logic/src/main/groovy/org/omegat/module/OmegatModulePlugin.groovy
+++ b/build-logic/src/main/groovy/org/omegat/module/OmegatModulePlugin.groovy
@@ -76,10 +76,21 @@ class OmegatModulePlugin implements Plugin<Project> {
 
         Map<String, Object> manifestAttrs = buildManifestAttributes(project)
 
+        // Jars the application itself ships in lib/ must not be merged into the
+        // module fat jar: modules load through a parent-first class loader, so the
+        // application's copy always wins and a bundled duplicate is unreachable dead
+        // weight. Matching by archive file name keeps the filter conservative - as
+        // soon as either side changes the version, the artifact is bundled again.
+        // Reaching into the root project's configurations will need rework once
+        // Gradle enforces isolated projects, like the plugin's other rootProject uses.
+        def coreRuntimeClasspath = project.rootProject.configurations.getByName("runtimeClasspath")
+
         project.tasks.named("jar", Jar).configure { Jar jarTask ->
             from({
+                Set<String> coreJarNames = coreRuntimeClasspath.files.collect { it.name } as Set
                 project.configurations.getByName("runtimeClasspath")
                         .resolve()
+                        .findAll { dep -> !(dep.name in coreJarNames) }
                         .collect { dep ->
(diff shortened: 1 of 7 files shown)


miurahr:
Please separate PR for the developer media addtion.

zs-stpa:
will do. that implies some reshuffeling of branches and PR's though, sorry for the mess …

miurahr:
> will do. that implies some reshuffeling of branches and PR's though, sorry for the mess …

No problem. Other than infographics, here is LGTM.

miurahr:
There is a root cause that the filters are not 1st citizen of the plugins, but core bundled pseudo plugins.
I think filters3, filters4 can be  submodules of the project and can be loaded as module.
This has a benefit that we can provide hot-fix or override of the filter without releasing new core version.
Let's discuss and update the developer documentation in another PR.

zs-stpa:
> There is a root cause that the filters are not 1st citizen of the plugins, but core bundled pseudo plugins. I think filters3, filters4 can be submodules of the project and can be loaded as module. This has a benefit that we can provide hot-fix or override of the filter without releasing new core version. Let's discuss and update the developer documentation in another PR.

i am looking under the 'macroscope' (for lack of a better word) at the plugins and language modules and how to reduce redundancies and add gooddies and find hidden treasures and be happy to leave the maintainening to y'all … and please do not be alarmed by the following wave of force pushes in my open PR's - is to try and reduce merge conflict surface during subsequent integration and should hopefully not interfer with the whole reviewing process

miurahr:
Why this PR includes [module-size-treemap.svg](https://github.com/omegat-org/omegat/pull/2223/changes#diff-2e146fd39393b48afacb36a5cb01cd36c394f101ce524f8ae7d78f624b2fc652) ?

zs-stpa:
> Why this PR includes [module-size-treemap.svg](https://github.com/omegat-org/omegat/pull/2223/changes#diff-2e146fd39393b48afacb36a5cb01cd36c394f101ce524f8ae7d78f624b2fc652) ?

good question, seems it got into the wrong branch by mistake, i'll move it elsewhere.
edit: i have a documentation branch where i keep a few demo video clips, that's where i'll put it.
edited again: 🤦 contains the treemap because this PR is stacked on top of #2216 which contains it. still, can be moved..

miurahr (approved):
LGTM.

PR #2224, opened 2026-08-05
docs(developer): add module distribution wiring slides

zs-stpa:
## Pull request type

## Which ticket is resolved?

- (none — developer documentation follow-up to the module distribution fixes in #2215/#2216)

## What does this PR change?

- Adds three SVG slides to `src/docs/developer/media` documenting how module jars silently vanished from the distribution under configure-on-demand and the dependency-based wiring that fixes it.
- Docs-only, split out of the build-fix PRs; those drop the duplicated commit on their next rebase.

## Other information

The slides render inline below.

slide 1
slide 2
slide 3


zs-stpa:
🫣 so many red CI from those puny slides?… good night, this will have to wait

(edit: see also #2225)

PR #2228, opened 2026-08-06
fix(fuzzy): repair fuzzy matching with numbers

zs-stpa:
## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?

- Consider numbers for fuzzy matches

## What does this PR change?

- Fuzzy matching compares number tokens by value when `match_numbers` project option is enabled: same value in different notation counts as equal, number-only segments score through placeholder class instead of 0%.
- Inserting a match substitutes source segment's numbers into match target by value, rendered in target's own notation; exotic numeral systems (sign-written, ideographic) read and written by value.
- Acceptance fixture `project_numerals` + `NumeralMatchesSubstitutionTest` pin covered systems and limits.
- Stacked on #2319 (the last three commits belong to this PR).
- Scope reduced per review: tips and team distribution of the option split off; team distribution follows separately after dev list discussion. Design recorded in ADR PR #2317.

### DIFF
--- a/src/main/java/org/omegat/core/data/IProject.java
+++ b/src/main/java/org/omegat/core/data/IProject.java
@@ -29,6 +29,7 @@
 package org.omegat.core.data;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
@@ -106,6 +107,51 @@ public interface IProject {
      * @throws Exception if any error occurs during the commit process
      */
     void commitSourceFiles() throws Exception;
+
+    /**
+     * Sets a registered team project setting and commits the project
+     * settings file (omegat/project_settings.properties) to the team
+     * repository, so the whole team gets the value. The setting lives in
+     * its own file because older OmegaT versions reject unknown elements in
+     * omegat.project but ignore extra files. No operation for non-team
+     * projects. Must run through Core.executeExclusively, like saveProject.
+     *
+     * @param key
+     *            key of a setting registered in {@link TeamSettingsRegistry}
+     * @param value
+     *            raw value to distribute, null for the default
+     * @throws Exception
+     *             if writing or committing the file fails
+     */
+    default void publishProjectSetting(String key, @Nullable String value) throws Exception {
+    }
+
+    /**
+     * Sets a registered team project setting for this session and writes it
+     * to the local project settings file only: the next open of the team
+     * project supersedes it with the remote value again. No operation for
+     * non-team projects. Must run through Core.executeExclusively, like
+     * saveProject.
+     *
+     * @param key
+     *            key of a setting registered in {@link TeamSettingsRegistry}
+     * @param value
+     *            raw value to restore, null for the default
+     * @throws Exception
+     *             if writing the file fails
+     */
+    default void useLocalProjectSetting(String key, @Nullable String value) throws Exception {
+    }
+
+    /**
+     * The local raw values that the team values superseded during the last
+     * load of a team project, keyed by setting key; an entry with null value
+     * means the local side had the default. Empty when everything agreed,
+     * always empty for non-team projects.
+     */
+    default Map<String, @Nullable String> getSupersededLocalSettings() {
+        return Collections.emptyMap();
+    }
 
     /**
      * Get project properties.
--- a/src/main/java/org/omegat/core/data/ProjectSettingsStorage.java
+++ b/src/main/java/org/omegat/core/data/ProjectSettingsStorage.java
@@ -1,0 +1,168 @@
+/**************************************************************************
+ OmegaT - Computer Assisted Translation (CAT) tool
+          with fuzzy matching, translation memory, keyword search,
+          glossaries, and translation leveraging into updated projects.
+
+ Copyright (C) 2026 Stephan Pakebusch
+               Home page: https://www.omegat.org/
+               Support center: https://omegat.org/support
+
+ This file is part of OmegaT.
+
+ OmegaT is free software: you can redistribute it and/or modify
+ it under the terms of the GNU General Public License as published by
+ the Free Software Foundation, either version 3 of the License, or
+ (at your option) any later version.
+
+ OmegaT is distributed in the hope that it will be useful,
+ but WITHOUT ANY WARRANTY; without even the implied warranty of
+ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ GNU General Public License for more details.
+
+ You should have received a copy of the GNU General Public License
+ along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ **************************************************************************/
+
+package org.omegat.core.data;
+
+import java.io.File;
+import java.io.IOException;
+import java.io.Reader;
+import java.io.Writer;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.util.Map;
+import java.util.Properties;
+import java.util.TreeMap;
+
+import org.jspecify.annotations.Nullable;
+
+import org.omegat.util.Log;
+
+/**
+ * Storage of optional per-project settings in
+ * {@code omegat/project_settings.properties}. The settings live in their own
+ * file instead of {@code omegat.project} on purpose: the project file parser
+ * of released OmegaT versions rejects unknown elements, so a new element
+ * there would make the project unreadable for every team member on an older
+ * version, while an extra file in the {@code omegat} folder is ignored.
+ *
+ * The file is written deterministically (sorted keys, no timestamp comment),
+ * so distributing an unchanged file to a team repository stays recognisable
+ * as a byte-identical no-op.
+ *
+ * @author stephan.pakebusch at zollsoft.de
+ */
+public final class ProjectSettingsStorage {
+
+    public static final String FILE_PROJECT_SETTINGS = "project_settings.properties";
+
+    private ProjectSettingsStorage() {
+    }
+
+    /** The settings file of the given project. */
+    public static File getFile(ProjectProperties config) {
+        return new File(config.getProjectInternal(), FILE_PROJECT_SETTINGS);
+    }
+
+    /**
+     * The stored raw value of given key: null when the file or the key is
+     * absent or the file is unreadable, so callers can distinguish "not
+     * configured" from an explicit value.
+     */
+    public static @Nullable String load(ProjectProperties config, String key) {
+        File file = getFile(config);
+        if (!file.isFile()) {
+            return null;
+        }
+        Properties props = new Properties();
+        try (Reader in = Files.newBufferedReader(file.toPath(), StandardCharsets.UTF_8)) {
+            props.load(in);
+        } catch (IOException ex) {
+            Log.logErrorRB(ex, "TEAM_SETTING_APPLY_ERROR");
+            return null;
+        }
+        String value = props.getProperty(key);
+        return value == null ? null : value.trim();
+    }
+
+    /**
+     * Store the raw value under given key (null removes the key), keeping
+     * any other keys of the file so settings can share it. Removing a key
+     * from an absent file is a no-op, so it never materialises the file.
+     */
+    public static void save(ProjectProperties config, String key, @Nullable String value)
+            throws IOException {
+        File file = getFile(config);
+        if (value == null && !file.isFile()) {
+            return;
+        }
+        Map<String, String> entries = new TreeMap<>();
+        if (file.isFile()) {
+            Properties existing = new Properties();
+            try (Reader in = Files.newBufferedReader(file.toPath(), StandardCharsets.UTF_8)) {
+                existing.load(in);
+            }
+            existing.stringPropertyNames().forEach(k -> entries.put(k, existing.getProperty(k)));
+        }
+        if (value == null) {
+            entries.remove(key);
+        } else {
+            entries.put(key, value);
+        }
+        File dir = file.getParentFile();
+        if (dir != null && !dir.isDirectory() && !dir.mkdirs()) {
+            throw new IOException("Cannot create " + dir);
+        }
+        try (Writer out = Files.newBufferedWriter(file.toPath(), StandardCharsets.UTF_8)) {
+            for (Map.Entry<String, String> e : entries.entrySet()) {
+                out.write(escape(e.getKey(), true) + "=" + escape(e.getValue(), false) + "\n");
+            }
+        }
+    }
+
+    /**
+     * Escape a key or value the way {@code Properties.store} would, minus
+     * the Latin-1 escaping (the file is read and written as UTF-8), so
+     * foreign keys of other settings survive the load/save round trip
+     * unchanged.
+     */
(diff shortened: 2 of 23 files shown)


zs-stpa:
and again, pls excuse the following wave of force pushes, which is to remove the unnecessary unicode signs from comment lines, as requested earlier

zs-stpa:
the old default at preferences > tag processing > Custom tags (regex) was "\d+" which covers only [0-9]+, i am proposing to change the default to "\p{Nd}+|[\p{Nl}\p{No}]+" so that all unicode numerals get the protected tag marker and the tooltip converts non [0-9] numbers and the issues window is aware of exotic number conversions as well.

the new default is only applied once if the old default was in effect (so stays silent if user had changed their default before or after)

Informing user about new default

with that new pattern, all the exotic numerals get the protected tag status as well and the issues window can warn about errors in conversions between any of them. without the new default pattern, the number-aware fuzzy matching still does the new number-trick but the issues window stays unaware and may produce false warnings or falsely produce no warnings concerning the non [0-9] numerals.

Value mismatch flagged

yes sure, the new default regex _could_* be improved to also cover fractions with the protection tag / marker .…

Compatibility form accepted

here is a very short demo video:
https://github.com/zs-stpa/omegat/raw/media/demo-videos/videos/2026-08-10-numeral-protected-tags-demo.mp4

the demo project is part of this PR.
feedback and suggestions welcome.

*: edited: expanded the regex some more and added a 'restore default' button next to the input field so users can reset on demand, the new default would be:
\p{Nd}+|[\p{Nl}\p{No}&&[^¹²³\p{InSUPERSCRIPTS_AND_SUBSCRIPTS}\p{InENCLOSED_ALPHANUMERICS}\p{InENCLOSED_ALPHANUMERIC_SUPPLEMENT}\p{InENCLOSED_CJK_LETTERS_AND_MONTHS}\p{InENCLOSED_IDEOGRAPHIC_SUPPLEMENT}\p{InDINGBATS}]]+

which should then cover most of the numbers currently displayable with unicode.

brandelune:
> the old default at preferences > tag processing > Custom tags (regex) was "\d+" which covers only [0-9]+, i am proposing to change the default to "\p{Nd}+|[\p{Nl}\p{No}]+" so that all unicode numerals get the protected tag marker and the tooltip converts non [0-9] numbers and the issues window is aware of exotic number conversions as well.

I don’t think this is an improvement.

The `\d+` custom tag was set there as an example, because the feature was not easy to discover. Also, the `\d+` is easy to learn and understand and serves as an introduction to regex use in OmegaT. Last, it is very unlikely that non ascii numerals are required in the same script in target, unlike ascii numerals. The numerals pattern that you propose seems to me of little use in general as a user setting since most translators only work from one script to another script. And the message that you propose is not necessary. We don’t need to use messages to introduce new features.

zs-stpa:
> > the old default at preferences > tag processing > Custom tags (regex) was "\d+" which covers only [0-9]+, i am proposing to change the default to "\p{Nd}+|[\p{Nl}\p{No}]+" so that all unicode numerals get the protected tag marker and the tooltip converts non [0-9] numbers and the issues window is aware of exotic number conversions as well.
> 
> I don’t think this is an improvement.
> 
> The `\d+` custom tag was set there as an example, because the feature was not easy to discover. Also, the `\d+` is easy to learn and understand and serves as an introduction to regex use in OmegaT. Last, it is very unlikely that non ascii numerals are required in the same script in target, unlike ascii numerals. The numerals pattern that you propose seems to me of little use in general as a user setting since most translators only work from one script to another script. And the message that you propose is not necessary. We don’t need to use messages to introduce new features.

ok i will remove the message. and i understand the argument about regex-learning. and yes, most translators have only one script and another, and the proposed change would cover any combination of source and target scripts with the number-is-accurate-or-not-comparison for the issues window including scripts like arabic and all other unicode scripts (with very few exceptions), whereas at the moment it only works with 0123456789 (and maybe full-width). possibly make it a new marker instead where the user does not get to see the complicated regex and it simply says 'safeguard number translations' or something like that?

brandelune:
> ok i will remove the message. and i understand the argument about regex-learning. and yes, most translators have only one script and another, and the proposed change would cover any combination of source and target scripts with the number-is-accurate-or-not-comparison for the issues window including scripts like arabic and all other unicode scripts (with very few exceptions), whereas at the moment it only works with 0123456789 (and maybe full-width).

That’s the point. This feature is about "customized tags" not about "number validation". People have issues understanding/learning/using regex, so this location was the right place to introduce them.

> possibly make it a new marker instead where the user does not get to see the complicated regex and it simply says 'safeguard number translations' or something like that?

No. because the location is not about number validation. It is about "custom tags". And it must be a regex. So please remove the message, the complex regex and instead, if you want to add a section on how to do complex numerals validation then you can point at a special section in the manual.

zs-stpa:
> > ok i will remove the message. and i understand the argument about regex-learning. and yes, most translators have only one script and another, and the proposed change would cover any combination of source and target scripts with the number-is-accurate-or-not-comparison for the issues window including scripts like arabic and all other unicode scripts (with very few exceptions), whereas at the moment it only works with 0123456789 (and maybe full-width).
> 
> That’s the point. This feature is about "customized tags" not about "number validation". People have issues understanding/learning/using regex, so this location was the right place to introduce them.
> 
> > possibly make it a new marker instead where the user does not get to see the complicated regex and it simply says 'safeguard number translations' or something like that?
> 
> No. because the location is not about number validation. It is about "custom tags". And it must be a regex. So please remove the message, the complex regex and instead, if you want to add a section on how to do complex numerals validation then you can point at a special section in the manual.

exactly, the custom tag default \d+ stays as is, and the manual mentions somewhere that either the preexisting checkbox in preferences > editor > 'attempt to replace fuzzy match numbers' or the recently new but preexisting project property 'consider numbers in fuzzy matching' does then also take care of the validation side, so that the new numbering-marker would not interfere graphically with the custom tag. that marker would not need to change the segment visibly, just take care that the issues window can warn about inaccurate number translations

src/docs/manual/en/Menus_Tools.xml
```diff
@@ -52,7 +52,9 @@
                   <para>
                      <guilabel>Tag Issues</guilabel> (always selected): detects
                      missing or misplaced tags, including custom tags and flagged
-                     text. See the 
+                     text. Numbers protected by the custom-tag expression are
+                     compared by their numeric value, whatever numeral system the
```
brandelune: The custom tag expression has nothing to do with numbers validation. Please remove this sentence.

brandelune: Custom tags are strings that MUST be reproduced in target. They should not be interpreted for "target value". What you propose here is equivalent to glossary checks, but for numbers. Do not confuse the two functions please.

src/docs/manual/en/OmegaT_Preferences.xml
```diff
@@ -1092,10 +1095,17 @@
                   		  with 
                   <literal>|</literal> (equivalent to "OR" in regular
                   		  expressions).</para>
-               <para>For example, use 
-                  <userinput>\d+</userinput> to treat all numbers
-                  		  as tags, enabling you to check that numbers have not been accidentally
-                  		  changed in the translation.</para>
+               <para>By default, this expression protects numbers, enabling you to
```
brandelune: Please remove this part. This feature only has "\d+" has an exemple and is not intended to be a value validation tool.

brandelune (changes requested):

src/docs/manual/en/Menus_Tools.xml
zs-stpa: ok i'll try

src/docs/manual/en/OmegaT_Preferences.xml
zs-stpa: okay i redid it and the \d+ stays and i still add a new Restore Default button there which asks if different and puts the same \d+ back there and there is now a new number underline marker color which is by default the same as the background color in both light and dark themes. this then protects number translations by number value instead of by string value so that different string representations in different writing systems still can count as the same value and so we have it not visibly interfering with the custom tags, see screenshot attached where i set the new marker to a bright color. and the \d+ still does its thing as before.

Bildschirmfoto 2026-08-11 um 23 25 21

brandelune: > okay i redid it and the \d+ stays and i still add a new Restore Default button there which asks if different and puts the same \d+ back there

There is no need for a "restore the default" button since this value is just an example. We already have a restore defaults on this dialog.

And I have no idea what you’re talking about here: 
> and there is now a new number underline marker color which is by default the same as the background color in both light and dark themes.

> this then protects number translations by number value instead of by string value so that different string representations in different writing systems still can count as the same value and so we have it not visibly interfering with the custom tags, see screenshot attached where i set the new marker to a bright color. and the \d+ still does its thing as before.

How is that delated to the custom tag feature?

zs-stpa: first i thought the custom tag feature could be used to safeguard number value translations so translating a 3 in one writing system to a 4 in another would raise an alarm in the issues window by simply expanding the example regex. but the custom tag only does literal string comparison by design and as you say has another learning purpose. but i still wanted to make sure that 3.141 does not accidentily get translated to 2,718 so i made a new number marker. the number marker now does that which i previously tried to let the custom tag example do, so the custom tag thing can keep doing what it does, and they do not interfere. in the screenshot you can see the 0-9 numerals colored differently by the default custom tag, whereas the new number marker stays invisible (no color set, not drawn) until a user changes its default color to an actual color in this case pink.

brandelune: > first i thought the custom tag feature could be used to safeguard number value translations so translating a 3 in one writing system to a 4 in another would raise an alarm in the issues window by simply expanding the example regex.

Thank you. But that is something that should be taken care of in the validation system, as you already did.

> but the custom tag only does literal string comparison by design and as you say has another learning purpose.

That’s correct. The "\d+" expression was the easiest to understand that I could find out that could have some use (I thought about using \[fuzzy\] but the marker is localized to it would not have worked from languages other than English).

> but i still wanted to make sure that 3.141 does not accidentily get translated to 2,718 so i made a new number marker. the number marker now does that which i previously tried to let the custom tag example do, so the custom tag thing can keep doing what it does, and they do not interfere.

Ok. But you’re adding a UI that we do not already have : underlining validation issues in the Editor. That’s something that we need to discuss. Why would only numerals be validated in the Editor and not the other issues ?

> in the screenshot you can see the 0-9 numerals colored differently by the default custom tag, whereas the new number marker stays invisible (no color set, not drawn) until a user changes its default color to an actual color in this case pink.

Good, but you do see that from that idea of "matching numerals" you extend what is "possible/nice to have/etc." into directions that need serious discussions.

What I am asking is that you stay on focus with one feature (matching numerals) and that we progressively discuss all the related "extensions" of that feature. For ex. you added tooltips to the numerals validation, but that’s also something that we do not have for other validation items. Why would that be?

Same when you add messages here and there. We do not have that for other features. Maybe we should. But that’s the kind of UI choices that we need to discuss and eventually generalize (or not, if we have good reasons to do so).

I hope that you understand my position.

zs-stpa: yes. smaller pieces, easier to review, stay on topic. i can split it up into separate PR's but they would have to be stacked on top of each other.

here another few screenshots. first in default configuration light theme english as shipped without any changes to preferences.

Bildschirmfoto 2026-08-12 um 11 42 39
this was the actual topic of the ticket, to insert fuzzy match numbers from any source script to any target script, which was covered in the first rounds.

but this then led to follow-ups. like my misunderstanding with the custom tag default. and now this new number marker, which is not an issue marker itself, it just marks numbers and comes with no color by default so is not visibly drawn unless user changes the setting.

Bildschirmfoto 2026-08-12 um 11 42 59

Bildschirmfoto 2026-08-12 um 11 43 44

resetting the color sets it to nil so it is an innvisible mark.

and the new project setting 'Check numbers by value' that controls whether or not the conversion aware value equality check is done for the issues window, this is what would be out of scope for this PR too, i gues…

Bildschirmfoto 2026-08-12 um 11 44 25

zs-stpa: now the failing CI test says

loadComparisonOldVersusNumberAwareMatching
java.lang.AssertionError: number-aware matching must stay within factor 3 of plain matching (plain=1511 ms, numbers=8586 ms)

yet another indication i should split this up, coming soon. so this PR only handles the fuzzy insertion part and another on top of it will handle the issues/marker which still needs some optimization it seems

zs-stpa:
this PR now only about what it says in the description at the top, the follow-up is #2247

zs-stpa:
added runtime checks and dialog for team projects.

Bildschirmfoto 2026-08-21 um 12 42 47

zs-stpa:
reworked it again, this time so that it can stay backwards compatible: older omegaT-versions should now ignore the unknown new project setting and thus be able to open a project saved with a newer version containing this PR.

miurahr (changes requested):
- This PR includes multiple related but diffrent scoped changes. Please split or separate the concerns; tips, team project extensions, and fuzzy match numbers
- Please remove develper/media files.
- Please raise discussion about extension of the team synchronization feature and summarize developer ML discussion as ADR after thorough discussion especially for backward compatibility and multiple configuration of the team projects such as SVN/GitHub/GitLab.
- Please raise new PR to enhance team project with ProjectSettingsStorage and extension of the integration test to test the feature; which can be configurable with/without the feature.
-

zs-stpa:
ok done. changing to new PR text:

### Pull request type

### Which ticket is resolved?

- Consider numbers for fuzzy matches

### What does this PR change?

- Fuzzy matching compares number tokens by value when the match_numbers project option is enabled: same value in different notation counts as equal, and number-only segments score through a placeholder class instead of 0%.
- Inserting a match substitutes the source segment's numbers into the match target by value, rendered in the target's own notation: digit scripts, sign-written systems (Ethiopic, cuneiform, Aegean, Mayan, Kaktovik, Greek acrophonic, Ottoman Siyaq, Meroitic, Pahawh Hmong, Medefaidrin, CJK, Roman via ICU rule sets) and vulgar fractions; only valid notation is composed, otherwise plain digits.
- Acceptance fixture project_numerals (one segment per numeral system) and NumeralMatchesSubstitutionTest pin the behaviour end to end.

### Other information

Scope reduced as requested in review: tips and team distribution of the option were removed from this PR. Team distribution follows separately after a developer ML discussion. Design is recorded in a stacked ADR PR.

zs-stpa:
converted to draft. i followed the previous suggestions - documentation is split off, generic project settings negotiation split off and discussion started on dev list and stacked this PR on top of that, and since that is still not settled, hence converting to draft here.

and now with the ongoing discussion on character equivalence classes in #2303 came another new idea, to put the number matching checkbox with the other char eq classes and also make lenient number matching available in the search and replace windows. -> there will be refactoring again.

and there are ongoing discussions about gui demessing / uncluttering / cleaning-upping, so possibly there will be more refactoring later from that direction.

PR #2233, opened 2026-08-06
fix(cli): apply remaining runtime options and repair GUI restart command line

zs-stpa:
## Pull request type

- Bug fix

## Which ticket is resolved?

No ticket. Follow-up of #2219; stragglers of the picocli migration (#1321).

## What does this PR change?

Several declared options never took effect: --resource-bundle was completely dead, --disable-project-locking, --disable-location-save and --no-team were ignored on the GUI start path, and the sub-command form --no-team did nothing anywhere because a negatable option declared as --team with default true inverts the intent in picocli (--team even switched team support off). LegacyParameters.initialize() now applies all top-level options centrally, the previously uncalled parseCommonParams() replaces the duplicated blocks in the sub-commands, and the team option is declared through its negative form.

The GUI restart (language switcher) rebuilt a command line that picocli rejected once --config-dir or the alternate filenames were set: top-level options after the start token, plus a misspelled --alternate-filenames-to. The reconstruction now emits the options in parseable order and also forwards --config-file, --resource-bundle, the disable switches and the tokenizer overrides.

## Other information

Fourteen unit tests; the six behaviour tests fail on master and pass with the fix. Out of scope, unchanged: -D and --verbose are not forwarded on restart, and log messages keep the bundle captured during logger setup. Full test suite and checkstyle pass.

### DIFF
--- a/src/main/java/org/omegat/Main.java
+++ b/src/main/java/org/omegat/Main.java
@@ -46,6 +46,7 @@ import java.util.ResourceBundle;
 import org.jetbrains.annotations.VisibleForTesting;
 import org.jspecify.annotations.Nullable;
 
+import org.omegat.cli.CommonParameters;
 import org.omegat.cli.LegacyParameters;
 import org.omegat.cli.SubCommands;
 import org.omegat.core.Core;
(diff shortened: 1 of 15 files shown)


miurahr (approved):
Thank you for finding and reparing broken command options.

PR #2239, opened 2026-08-09
feat(editor): instantanious color refresh in the viewport instead of rerendering whole document

zs-stpa:
## Pull request type

## Which ticket is resolved?

- No ticket; incidental finding while working on the colour series (#2183/#2195/#2196). Groundwork for event flashes (relates to RFE discussions around presentation/accessibility).

## What does this PR change?

- Applying editor colours no longer rebuilds the editor document (a seconds-long freeze on large files): spans stay bound to their palette entry and views resolve the colour while painting — a repaint plus one marker pass, caret and scroll stay put.
- The preferences window rebuilds the editor after saving only when a visited panel needs it (`IPreferencesController.requiresEditorRefresh()`, default `true`); the Colours panel opts out.
- Four new test classes; full suite and checkstyle green. Stacked on #2196.

## Other information

(svg is not part of this PR but externally linked)
Paint-time colour resolution


miurahr (changes requested):
It may be valueable to add explanation of the color scheme with chart that is produced from source values, but the PR is not it.
Please remove infographics from the feature PR, or write explanation by your own and add diagrams with explanation which is directly connected to the feature changes.

zs-stpa:
ok i think you are referring to the previous PR #2196, which this one here is stacked on top of. and the colors in the SVG light and dark tables there, those are the proposed new default colors after contrast audit, so that we do no longer ship color combinations that are measurably poorly visible like dark blue on black background for example. this is another small accessibility step for users with less-than-perfect eyesight.

zs-stpa:
made the colors for ProtectedPartsMarker and RemoveTagMarker also live updatable (previously missed) by using createBoundAttributeSet, and rebased on current master

zs-stpa:
ok i have removed the svg and only linked it externally from my fork

src/test/java/org/omegat/util/gui/StylesTest.java
```diff
@@ -51,4 +66,103 @@ public void testColorStrings() throws Exception {
...
+     * falls back to exactly that value, and colors absent from the scheme
+     * have no default (consumers treat {@code null} as "inherit"). Also
+     * catches a renamed or misspelled key on either side, which would
+     * otherwise silently disconnect the theme from the color — the failure
+     * mode behind the dead {@code markComesFromTmXendorced}-style keys.
+     * Other tests may leave OmegaT keys in the UIManager, so each key is
+     * cleared and restored around the fallback check.
+     */
+    @Test
+    public void testFallbacksMatchBundledColorScheme() throws Exception {
```
miurahr, line 156: This test may never fail when ColorScheme_light.properties and ColorScheme_dark.properties have a same extra key but EditorColor does not reference.

src/test/java/org/omegat/util/gui/StylesBoundAttributesTest.java
```diff
@@ -0,0 +1,123 @@
...
+        assertEquals(bakedBackground, StyleConstants.getBackground(attrs));
+    }
+
+    @Test
+    public void testBoundAttributesBakeCurrentColorsForPlainConsumers() {
+        AttributeSet attrs = Styles.createBoundAttributeSet(EditorColor.COLOR_UNTRANSLATED_FG,
+                EditorColor.COLOR_UNTRANSLATED, true, false);
+        assertEquals(EditorColor.COLOR_UNTRANSLATED_FG.getColor(), StyleConstants.getForeground(attrs));
+        assertEquals(EditorColor.COLOR_UNTRANSLATED.getColor(), StyleConstants.getBackground(attrs));
+        assertEquals(true, StyleConstants.isBold(attrs));
```
miurahr: I prefer `assertTrue`

src/main/java/org/omegat/core/CoreEvents.java
```diff
@@ -104,6 +107,16 @@ public static void unregisterFontChangedEventListener(final IFontChangedEventLis
         FONT_CHANGED_EVENT_LISTENERS.remove(listener);
     }
 
+    /** Register listener. */
+    public static void registerColorsChangedEventListener(final IColorsChangedEventListener listener) {
+        COLORS_CHANGED_EVENT_LISTENERS.add(listener);
```
miurahr: `COLORS_CHANGED_EVENT_LISTENERS.addIfAbsent(Objects.requireNonNull(listener));` is safer.
Other register functions have non-safer `add` which will be refactored in another PR.

src/test/java/org/omegat/gui/editor/mark/MarkerBoundColorsTest.java
```diff
@@ -0,0 +1,111 @@
...
+        String translationText = "target %remove text";
+        EntryKey key = new EntryKey("file", sourceText, "id", "prev", "next", "path");
+        SourceTextEntry ste = new SourceTextEntry(key, 1, new String[0], sourceText,
+                Collections.emptyList());
+
+        List<Mark> result = marker.getMarksForEntry(ste, sourceText, translationText, true);
+
+        assertNotNull(result);
+        assertEquals(1, result.size());
+        assertEquals(Mark.ENTRY_PART.TRANSLATION, result.get(0).entryPart);
```
miurahr: we can use `getFirst()`

miurahr (approved):
The Proposal is good enough but there are several place to be able to improved, or where is new findings to improve.

src/test/java/org/omegat/gui/editor/mark/MarkerBoundColorsTest.java
zs-stpa: ok

src/main/java/org/omegat/core/CoreEvents.java
zs-stpa: ok

src/test/java/org/omegat/util/gui/StylesBoundAttributesTest.java
zs-stpa: ok

src/test/java/org/omegat/util/gui/StylesTest.java
zs-stpa: ok, making sure there are no unreferenced keys.

src/test/java/org/omegat/gui/editor/mark/MarkerBoundColorsTest.java
zs-stpa: Follow-up plan after #2268 is merged:

1. **Main follow-up (delayed on purpose):** four consumers still register colour listeners manually and should move to `Styles.bindColors` — `EditorController`, `EditorTextArea3`, `MatchesTextArea`, `BaseStatisticsPanel`. Delayed until branches touching those files are merged (editor cluster: #2196, #2239, editor-views stack; `MatchesTextArea` → #2228), otherwise conflict cascade across open PRs. Sequence: merges first, then small conversion PR.
2. **Stack hygiene before review:** #2268 sits on #2189 + #2195 — after those are merged, rebases onto master so PR shows only its one own commit.
3. **Cleanup afterwards:** once all consumers use `bindColors`, check whether manual `registerColorsChangedEventListener` pattern still has remaining users — fits announced refactoring PR unifying register functions.
4. **Regression watch after conversion:** two historically fragile spots. (a) `bindColors` re-applies colours on hierarchy change (dock/undock); SegmentProperties table loses row selection on re-dock — accepted trade-off, but conversion must not spread similar state loss to editor, matches or statistics panes (caret, scroll position). (b) windows open while colours change: Tip of the Day formerly kept stale colours until restart; same bug class can reappear in each newly converted consumer, so smoke test colour change with panes open.

PR #2241, opened 2026-08-09
fix(search): read selected radio button again when detecting the search type

zs-stpa:
## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?

- Regression from #2235 (BUGS-1327 follow-up), no separate ticket: since cc9c7ffc3 every search runs as an exact search.

## What does this PR change?

- The search-type detection introduced in #2235 always looked up the model of the *Exact* radio button, so regular expressions were quoted into literals (`a|b` found nothing) and keyword queries silently ran as phrases; replace mode additionally read the search-mode radios.
- The lookup now uses the actual button group selection, and the replace-mode radios are part of the model map.
- Two regression tests assert that the detected type follows the selected radio in both modes (verified red on the buggy lookup); the two detection methods became package-private with `@VisibleForTesting` for that.

## Other information

Found while testing a local integration build; root cause confirmed against a pre-#2235 build (regex search there returns the expected matches).


miurahr:
Thank you for fixing the regression caused by my change.

PR #2247, opened 2026-08-12
feat(issues): new project option to validate number translations by value instead of by string plus optional color marker for numbers

zs-stpa:
## Pull request type

## Which ticket is resolved?

- Accept alternatives to Arabic Numbers

## What does this PR change?

- Stacked on #2228 — only the two newest commits belong to this PR.
- The tag check compares numbers of source and translation by numeric value, whatever the writing system and whatever the custom-tag pattern; a changed or missing value is reported in the Issues window, while another script's digits, sign-written systems, vulgar fractions and decimal or grouped spellings count as the same number.
- A per-project option (on by default, written to omegat.project only when disabled) turns the check off for language pairs that rewrite numbers beyond recognition, such as Japanese ten-thousands grouping.
- A new editor marker underlines the checked numerals; it ships without a colour and paints nothing until one is assigned under Options > Preferences > Colours > Numbers (project setting).
- Manual and tip of the day describe the behaviour.

## Other information

see screenshots and discussion in #2228

### DIFF
--- a/src/docs/manual/en/Dialogs_ProjectProperties.xml
+++ b/src/docs/manual/en/Dialogs_ProjectProperties.xml
@@ -198,7 +198,7 @@
          </varlistentry>
          <varlistentry xml:id="dialogs.project.properties.match.numbers">
             <term xml:id="dialogs.project.properties.match.numbers.title">
-               <option>Consider numbers in fuzzy matching</option>
+               <option>Consider numbers in matches</option>
             </term>
             <listitem>
                <para>By default, the word-based match percentages ignore numbers, and
@@ -209,7 +209,11 @@
                   		  tokens by their numeric value: numbers written in different
                   		  numbering systems (Roman numerals, ideographic numbers,
                   		  Arabic-Indic or full-width digits, and so on) count as equal to
-                  		  their Western counterpart of the same value. Two segments that
+                  		  their Western counterpart of the same value. The sign-written
+                  		  systems (Ethiopic, cuneiform, Mayan and others) and the
+                  		  precomposed fraction glyphs are read the same way, and an
+                  		  inserted match rewrites each number in the notation the match
+                  		  itself uses. Two segments that
                   		  consist only of numbers are scored against each other as number
                   		  sequences, with a small fixed penalty, so differing dates or bare
                   		  numbers surface as fuzzy matches at all.</para>
@@ -218,6 +222,21 @@
                   <link linkend="panes.fuzzy.matches" endterm="panes.fuzzy.matches.title"></link> pane to see how its score was
                   		  computed; a warning is shown there when the numbers of the match
                   		  differ in value from those of the segment.</para>
+            </listitem>
+         </varlistentry>
+         <varlistentry xml:id="dialogs.project.properties.check.numbers">
+            <term xml:id="dialogs.project.properties.check.numbers.title">
+               <option>Check numerals by value</option>
+            </term>
+            <listitem>
+               <para>Makes the tag checks compare the numerals of the source text
+                  		  and the translation by their numeric value, whatever the numeral
+                  		  system, and report a changed or missing value in the
+                  <guilabel>Issues</guilabel> window. A value rewritten in the
+                  		  notation of the target language raises no issue.</para>
+               <para>Turn the option off for language pairs where numbers are
+                  		  legitimately rewritten beyond recognition, such as a target
+                  		  language that spells numbers out in words.</para>
             </listitem>
          </varlistentry>
          <varlistentry xml:id="dialogs.project.properties.external.processing.command">
--- a/src/docs/manual/en/Menus_Tools.xml
+++ b/src/docs/manual/en/Menus_Tools.xml
@@ -52,7 +52,10 @@
                   <para>
                      <guilabel>Tag Issues</guilabel> (always selected): detects
                      missing or misplaced tags, including custom tags and flagged
-                     text. See the 
+                     text. If the
+                     <link linkend="dialogs.project.properties.check.numbers" endterm="dialogs.project.properties.check.numbers.title"></link> project
+                     property is enabled, numerals are also compared by their
+                     numeric value, whatever the numeral system. See the
                      <link linkend="dialogs.preferences.tag.processing" endterm="dialogs.preferences.tag.processing.title"></link> preferences for
                      details.</para>
                   <note>
--- a/src/docs/manual/en/OmegaT_EditorsPanes.xml
+++ b/src/docs/manual/en/OmegaT_EditorsPanes.xml
@@ -472,6 +472,13 @@ When enabled, all formatting in the source segments are hidden.</programlisting>
                   <link endterm="dialogs.preferences.tag.processing.regular.expressions.for.fragments.that.should.be.removed.from.translation.title" linkend="dialogs.preferences.tag.processing.regular.expressions.for.fragments.that.should.be.removed.from.translation"></link>
                   		  preference.</para>
             </listitem>
+            <listitem>
+               <para>Underlining numerals by assigning the
+                  <guilabel>Numerals</guilabel> colour in the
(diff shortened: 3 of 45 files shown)


src/main/resources/org/omegat/Bundle.properties
```diff
@@ -2477,6 +2485,7 @@ COLOR_MOD_INFO_FG=Modification info text
 COLOR_PLACEHOLDER=Tags
 COLOR_REMOVETEXT_TARGET=Flagged text
 COLOR_NBSP=Non-breakable spaces
+COLOR_NUMERALS=Numbers (project setting)
```
brandelune: What does "(project setting)" mean ?

src/main/resources/org/omegat/Bundle.properties
```diff
@@ -789,6 +789,9 @@ PP_LOC_TOK=Target language tokenizer:
 PP_ALLOW_DEFAULTS=&Auto-propagation of translations
 
 PP_REMOVE_TAGS=Remove ta&gs
+PP_MATCH_NUMBERS=Consider &numbers in fuzzy matching
+
+PP_CHECK_NUMBERS=Check num&bers by value (all numeral systems)
```
brandelune: It seems to me that this should be: "Check numerals by value"

src/main/resources/org/omegat/Bundle.properties
```diff
@@ -789,6 +789,9 @@ PP_LOC_TOK=Target language tokenizer:
 PP_ALLOW_DEFAULTS=&Auto-propagation of translations
 
 PP_REMOVE_TAGS=Remove ta&gs
+PP_MATCH_NUMBERS=Consider &numbers in fuzzy matching
```
brandelune: Simplify: "Consider numbers in matches"

src/docs/tips/en/fuzzy_match_numbers.html
```diff
@@ -0,0 +1,51 @@
+<html lang="en">
```
brandelune, line 1: It seems to me that the 2 tips that you propose are way longer than what we already have. Tips should be just that. Not feature explications.

src/docs/manual/en/Dialogs_ProjectProperties.xml
```diff
@@ -196,6 +196,55 @@
                </note>
             </listitem>
          </varlistentry>
+         <varlistentry xml:id="dialogs.project.properties.match.numbers">
+            <term xml:id="dialogs.project.properties.match.numbers.title">
+               <option>Consider numbers in fuzzy matching</option>
+            </term>
+            <listitem>
+               <para>By default, the word-based match percentages ignore numbers, and
```
brandelune, line 204: We don’t need to explain the defaults. Only what the feature offers.

src/docs/manual/en/Dialogs_ProjectProperties.xml
```diff
@@ -196,6 +196,55 @@
                </note>
             </listitem>
          </varlistentry>
+         <varlistentry xml:id="dialogs.project.properties.match.numbers">
+            <term xml:id="dialogs.project.properties.match.numbers.title">
+               <option>Consider numbers in fuzzy matching</option>
```
brandelune: This needs validation in the Bundle file.

src/docs/manual/en/Dialogs_ProjectProperties.xml
```diff
@@ -196,6 +196,55 @@
...
+            <listitem>
+               <para>By default, the word-based match percentages ignore numbers, and
+                  		  a segment that consists only of numbers (a date, a bare quantity)
+                  		  scores too low against a segment with different numbers to be
+                  		  offered as a match.</para>
+               <para>Activating this option makes fuzzy matching compare number
+                  		  tokens by their numeric value: numbers written in different
+                  		  numbering systems (Roman numerals, ideographic numbers,
+                  		  Arabic-Indic or full-width digits, and so on) count as equal to
+                  		  their Western counterpart of the same value. The sign-written
```
brandelune, line 212: There is no "Western" counterpart. No need to details all the systems. Keep the explanation short and to the point please.

src/docs/manual/en/Dialogs_ProjectProperties.xml
```diff
@@ -196,6 +196,55 @@
...
+                  		  computed; a warning is shown there when the numbers of the match
+                  		  differ in value from those of the segment.</para>
+            </listitem>
+         </varlistentry>
+         <varlistentry xml:id="dialogs.project.properties.check.numbers">
+            <term xml:id="dialogs.project.properties.check.numbers.title">
+               <option>Check numbers by value (all numeral systems)</option>
+            </term>
+            <listitem>
+               <para>On by default. The tag checks compare the numbers of the source
```
brandelune: Same comments as above.

src/docs/manual/en/Menus_Tools.xml
```diff
@@ -52,7 +52,11 @@
                   <para>
                      <guilabel>Tag Issues</guilabel> (always selected): detects
                      missing or misplaced tags, including custom tags and flagged
-                     text. See the 
+                     text. Numbers are compared by their numeric value, whatever
```
brandelune: Numerals are compared ...

src/docs/manual/en/Menus_Tools.xml
```diff
@@ -52,7 +52,11 @@
...
+                     numeral system the translation writes, so a changed or
```
brandelune: Keep the description the shortest possible.

src/docs/manual/en/OmegaT_EditorsPanes.xml
```diff
@@ -793,7 +793,16 @@ Dossier de configuration
             </note>
          </listitem>
       </itemizedlist>
-      <para>You can change how matches are displayed or sorted. See the 
+      <para>Hover over a match to see a tooltip that presents the three
+         		percentages with plain-language labels, together with any penalties that
+         		were applied to the match.</para>
+      <para>If the project option
```
brandelune, line 806: If the (link) option is enabled, numerals are compared ...

src/docs/manual/en/OmegaT_Preferences.xml
```diff
@@ -395,6 +395,10 @@
             with
             <guibutton>OK</guibutton> or
             <guibutton>Apply</guibutton>.</para>
+         <para>The <guilabel>Numbers (project setting)</guilabel> colour drives
```
brandelune: This is not necessary. I have no idea why you’d make a feature not discoverable. That’s the opposite that we’re trying to do generally.

src/docs/manual/en/OmegaT_Preferences.xml
```diff
@@ -1092,10 +1096,16 @@
                   		  with 
                   <literal>|</literal> (equivalent to "OR" in regular
                   		  expressions).</para>
-               <para>For example, use 
+               <para>For example, use
                   <userinput>\d+</userinput> to treat all numbers
                   		  as tags, enabling you to check that numbers have not been accidentally
                   		  changed in the translation.</para>
+               <para>Independently of this expression, numbers of every numeral
```
brandelune: This has nothing to do here since the features are unrelated.

brandelune (changes requested):

zs-stpa:
ok, i have reworked it in light of these comments. just 'Numerals' and less verbose.

src/docs/tips/en/fuzzy_match_numbers.html
zs-stpa: i am shortening new tips in other open PRs as well.

zs-stpa:
reworked it again, this time so that it can stay backwards compatible: older omegaT-versions should now ignore the unknown new project setting and thus be able to open a project saved with a newer version containing this PR.

zs-stpa:
converting to draft. because: this is stacked on top of #2228 which is stacked on top of #2319 and both are not yet finalized.

also, new discussion requires to restack it also on top of #2303 with additional refactoring

PR #2248, opened 2026-08-14
feat(project): per-project custom tag and removed-text expressions (SF-RFE-926)

zs-stpa:
## Pull request type

edit: stacked on top of #2335, will be reopened once that is through

## Which ticket is resolved?

- Make custom tag definition project specific
  * https://sourceforge.net/p/omegat/feature-requests/926/

- Save custom tags with project properties
  * https://sourceforge.net/p/omegat/feature-requests/1427/

- Project-specific "Remove text" pattern
  * https://sourceforge.net/p/omegat/feature-requests/824/

## What does this PR change?

- New **Tag definitions** section in the project properties dialog: the project can override the global custom-tag and removed-text expressions; fields are prefilled with the global values, an empty field switches the expression off, invalid input is rejected.
- The overrides are stored in `omegat/tag_patterns.xml` (like `filters.xml`), so they travel with team projects automatically while `omegat.project` stays untouched - up and downward compatible.
- Applied before the first source parse, cleared on close; documented in the manual and covered by unit tests.

## Other information

i only discovered the tickets after the fact.

### screenshot

Bildschirmfoto 2026-08-14 um 01 01 20

### DIFF
--- a/src/docs/manual/en/Dialogs_ProjectProperties.xml
+++ b/src/docs/manual/en/Dialogs_ProjectProperties.xml
@@ -332,6 +332,30 @@
                </note>
             </listitem>
          </varlistentry>
+         <varlistentry xml:id="dialogs.project.properties.tag.definitions">
+            <term xml:id="dialogs.project.properties.tag.definitions.title">
+               <guibutton>Local Tag Definitions...</guibutton>
+            </term>
+            <listitem>
+               <para>By default, the custom tag and flagged text definitions are
+                  		  global and shared by all projects. They are found in the
+                  <link linkend="dialogs.preferences.tag.processing" endterm="dialogs.preferences.tag.processing.title"></link>
+                  		  preferences.</para>
+               <para>Use this button to create local tag definitions specific to
+                  		  your project. Check the
+                  <option>Use local tag definitions</option> box, and adjust the two
+                  		  regular expressions as desired. An empty field means the project
+                  		  does not use that expression at all.</para>
+               <para>The project will store the expressions in the
+                  <link linkend="project.folder.omegat.tag.patterns" endterm="project.folder.omegat.tag.patterns.title"></link> file located in its
+                  <link linkend="project.folder.omegat.folder" endterm="project.folder.omegat.folder.title"></link> folder. These
+                  		  settings will supersede the global tag definitions.</para>
+               <note>
+                  <para>To disable local tag definitions, uncheck the box, or remove
+                     			that file while the project is closed.</para>
+               </note>
+            </listitem>
+         </varlistentry>
       </variablelist>
    </section>
    <section xml:id="dialogs.project.properties.file.locations">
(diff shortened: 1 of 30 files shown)


brandelune:
The setting can work like the "local external searches" button right above: click on it and open a dialog where the data is entered.

No need to clutter the project properties dialog more than what it already is.

src/docs/manual/en/Dialogs_ProjectProperties.xml
```diff
@@ -196,6 +196,25 @@
                </note>
             </listitem>
          </varlistentry>
+         <varlistentry xml:id="dialogs.project.properties.tag.definitions">
+            <term xml:id="dialogs.project.properties.tag.definitions.title">
+               <option>Tag definitions</option>
+            </term>
+            <listitem>
+               <para>Custom tags and flagged text are normally defined for all
```
brandelune: Please, use the same wording as for other similar "local" options. That way users learn from the redundancy and OmegaT localizers can reuse previous translations.

brandelune (changes requested):

src/docs/manual/en/Dialogs_ProjectProperties.xml
zs-stpa: will do

zs-stpa: Done - the section is gone; there is now a 'Local Tag Definitions...' button below 'Local External Searches...' that opens a small dialog, and the manual entry follows the wording of the other local options (tag_patterns.xml is now also documented in the project-folder chapter).

zs-stpa:
Added team distribution for the project tag definitions: opening a team project whose local `omegat/tag_patterns.xml` is unknown to the repository now asks once whether to share it with the team, remove it (match the team), or keep it local. The tag definitions editor gains a "Share with the Team" button. Publishing refreshes the checkout, commits the file and reports rejected pushes; older OmegaT versions simply ignore the file.

Bildschirmfoto 2026-08-20 um 16 49 41

Bildschirmfoto 2026-08-20 um 16 50 40

zs-stpa:
added another additional safety net more, if the local protected tag settings is changed and not shared with team and next time the project is opened and the teamproject does not _not contain a protected tag setting_ but contains a different protected tag setting, then this dialog:

Bildschirmfoto 2026-08-21 um 00 45 36

Force-pushed an update to the team distribution commit. Sharing definitions the team repository already carries is a no-op now instead of an error (git reports an empty commit the same way as a rejected push).
Round-trip and detection tests extended accordingly.

zs-stpa:
positive user feedback on original ticket https://sourceforge.net/p/omegat/feature-requests/824/#b832/b861

miurahr:
This PR includes a OmegaT core basic API exntentions. It is not only single method but several method additions.
You need to share a core API change design in dev list and made consensus and record as ADR is mandatory.
Please post a core API change proposal in the dev list.

zs-stpa:
okay. i am refactoring again, because a similar (better) pattern for registering project settings exists in #2335 (stacked on #2319) - so this PR here gets converted to 'draft' + refactored + restacked on top of #2335 and once that is merged, it gets reopened.

PR #2252, opened 2026-08-15
refactor(docs): remove version properties and simplify manual lang handling

miurahr:
## Pull request type

- Other (describe below)
refactor a behavior of the help -> manual link

## Which ticket is resolved?
None
## What does this PR change?

- Now we always bundle supported manual so version check of the manual is no-sense.

### DIFF
--- a/build-logic/src/main/groovy/org.omegat.document-conventions.gradle
+++ b/build-logic/src/main/groovy/org.omegat.document-conventions.gradle
@@ -1,5 +1,4 @@
 import org.omegat.documentation.*
-import org.omegat.gradle.PropertiesUtil
 
 plugins {
     id 'org.omegat.documentation'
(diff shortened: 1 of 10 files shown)


brandelune:
We need a fallback when the internet connexion is not available, not just an empty page.

Also, if the web page does not offer the selected language, what is the point sending the user to the site?

miurahr:
When internet connection is not available, OmegaT does not show anythings, and browser will show error.
OmegaT point manual selection page.

miurahr:
> We need a fallback when the internet connexion is not available, not just an empty page.
> Also, if the web page does not offer the selected language, what is the point sending the user to the site?

This PR does not change behavior. Web browser will show error when internet connection does not exist even without the PR.
The comment is out of scope of the PR change.

brandelune (approved):

PR #2253, opened 2026-08-16
[6.1] chore(ci): fix dest URL and hardening connection

miurahr:
Release upload failed in previous 6.1.0 and 6.1.1. It is because the $dest has missed the "sftp://"

## Pull request type

## What does this PR change?

- Fix the error that missing "sftp://" in the $dest
- Hardening timeout/retry/reconnect to failed early in several minutes when somethings become wrong


miurahr:
This is obvious change and fix for the release process. I will merge now.

PR #2249, opened 2026-08-15
feat(editor): show optional metadata column with translation state and live length counter (SF-RFE-420)

zs-stpa:
## Pull request type

- Feature enhancement -> [enhancement]

## Which ticket is resolved?

- Optional extra columns in edit pane for various things
  * https://sourceforge.net/p/omegat/feature-requests/420/

## What does this PR change?

- Adds an optional segment metadata gutter to the editor: segment number, translation state (T/U, M/M+ non-unique, I/C/A/E provenance, * recorded origin), author, change date, bound marker colours with underline miniatures, segment ID, text lengths, alternative translations.
- Adds a settings menu to the editor pane with a small live-apply dialog: column choice and order per drag'n'drop or buttons or keyboard; spelled-out tooltips; localized right-aligned numbers; live lengths for the active segment.
- The gutter is off by default, keeps no own state (rows painted at document positions) and resolves colours at paint time.

## Other information

Stacked on #2239 because this needs the instantanious viewport refreshing; its commits leave the diff once that PR is merged.

### screenshots

Bildschirmfoto 2026-08-15 um 09 10 32
Bildschirmfoto 2026-08-15 um 09 13 17

### Demo videos

https://github.com/zs-stpa/omegat/raw/media/demo-videos/videos/2026-08-15-metadata-gutter-demo-0216.mp4
https://github.com/zs-stpa/omegat/raw/media/demo-videos/videos/2026-08-15-metadata-gutter-demo-0222.mp4

PR #2257, opened 2026-08-16
feat(editor): enhanced customizability of the editor view (SF-RFE-1028)

zs-stpa:
## Pull request type

## Which ticket is resolved?

- Create two-column edit pane, Pootle style* (*: i have never seen pootle)
  * URL https://sourceforge.net/p/omegat/feature-requests/1028/

## What does this PR change?

- Source and translation of the inactive segments can be shown side by side or stacked in either order; the active segment keeps the classic editing layout.
- The text pair joins the layout dialog of the metadata columns: order, coupled width shares (25–75 %), alignment presets for the pair, and a draggable boundary in the editor.
- The layout can be reset to the classic default, exported to and imported from a properties file.

## Other information

Stacked on top of #2249

Demo video: https://github.com/zs-stpa/omegat/blob/media/demo-videos/videos/2026-08-16-editor-layout-demo.mp4 (not part of commits in this PR, only linked in this description, and again as an attachment so it plays inline. demo video will probably not updated if details of the implementation change)

https://github.com/user-attachments/assets/42b82dc8-459a-4f93-8378-3170f5c85d79


brandelune:
This ideas is very nice.

There are visual issues with the current segment in "alternative background".

The grey surrounds the white/green backgrounds that "stick" to the text. It is not elegant.

zs-stpa:
next step on the accessibility ladder would be to stronger emphasize the current segment with frame / background / announcments (bells+whistle+earcons when entering/leaving a segment), etc.

zs-stpa:
> This ideas is very nice.
> 
> There are visual issues with the current segment in "alternative background".
> 
> The grey surrounds the white/green backgrounds that "stick" to the text. It is not elegant.

is why i initially left out the alternating background for source/target columns. thousands of possibilities: more adjustment options for alternating background: which columns to apply or not apply it, color strength, padding, etc., and it also depends on the users' marker color palette settings

brandelune:
> > This ideas is very nice.
> > There are visual issues with the current segment in "alternative background".
> > The grey surrounds the white/green backgrounds that "stick" to the text. It is not elegant.
> 
> is why i initially left out the alternating background for source/target columns. thousands of possibilities: more adjustment options for alternating background: which columns to apply or not apply it, color strength, padding, etc., and it also depends on the users' marker color palette settings

One option is to disable the "alternate" background for the current segment, or eventually leave a "trace" of that alternate color as cell borders.

Also, that display option is a major request for users (and sometimes a deal breacker for some users). Its availability should not be buried in the list of possible segment meta info display options.

zs-stpa:
i am unhappy too with the many new options living in a separate window, i would love for all options affecting editor appearance to be in one single place where we would then have a unified simple intuitive automagic way to set the preferences just the way the user can use them most efficiently without the user having to understand every last detail of every possible sub-option… the usual dilemma, make it able to do everything but do the right thing at the right time for everyone - virtually impossible to solve, but open to suggestions ^.^

edit: could offer certain presets in the editor-preferences panel with sample screenshots (or even renderings based on current document) of what it would look like - e.g. one default legacy appearance as has always been, and then four or five wildly different new editor-layout-configurations to chose from for the users who do not want to go into the depths of all possible settings, and selecting one of the presets then does the same as would happen if you click 'Import' and then we just deliver a couple of useful configurations to chose from

edited: scrap the preview idea, is not necessary in wysiwyg. i'll just make radio buttons instead of the checkbox in the new metadata/layout-options window over the table with some presets and you/we/someone can then later change or add to those presets

... edited again … but then, there are still lots more options to come, because full accessibility setup would include .. well, a lot more of possible settings. maybe wait with the whole presets-offerings-idea until all those new settings will have arrived as well maybe?

zs-stpa:
realizing just now that there is no public discussion, i will post something on the mail list.

PR #2261, opened 2026-08-18
feat(editor): show spelling suggestions in tooltip of misspelled words

zs-stpa:
## Pull request type

## Which ticket is resolved?

- (no ticket)

## What does this PR change?

- Hovering a misspelled word in the editor shows its spelling suggestions in the tooltip (bold, at most five). Until now suggestions were only reachable through the context menu, and only in the active segment.
- The lookup runs lazily on hover and once per word: the spell checker caches the suggestions and serializes the provider access, because the providers are not thread safe.


brandelune:
An accessibility improvement would be to put the spellchecking suggestions as an "auto-completion" dialog where the whole word is the target for replacement.

PR #2263, opened 2026-08-18
feat(stat): add match-category filter buttons in Match Statistics window

zs-stpa:
## Pull request type

## Which ticket is resolved?

- No ticket. Related: Separate Statistics for Context-Bound (ICE/101%) and Exact (100%) Matches
  * https://sourceforge.net/p/omegat/feature-requests/1782/

## What does this PR change?

- Makes the Match Statistics window non-modal, so the editor stays usable while it is open
- Adds a Filter button per category row that restricts the editor to the segments of that category, asking before replacing an active filter
- Keeps the last result for the session with a timestamp; a Recalculate button refreshes it

### screenshot
(not part of this PR, only as an attachment, will not be updated if implementation details should change)
Bildschirmfoto 2026-08-18 um 14 46 01

### DIFF
--- a/src/docs/manual/en/Menus_Tools.xml
+++ b/src/docs/manual/en/Menus_Tools.xml
@@ -182,6 +182,11 @@
             <para>Displays the match statistics for the project, which consist of 
                the number of repetitions, exact matches, fuzzy matches and no-matches 
                for segments, words and characters.</para>
+            <para>Every category row has a
+               <guibutton>Filter</guibutton> button that restricts the editor to
+               the segments of that category. The window is not modal, and the
+               last result is kept for the session, with its date and time,
+               until you use <guibutton>Recalculate</guibutton>.</para>
             <para>The data is saved in the 
                <link linkend="project.folder.omegat.project.stats.match" endterm="project.folder.omegat.project.stats.match.title"></link> file located in the 
                <link linkend="project.folder.omegat.folder" endterm="project.folder.omegat.folder.title"></link> folder of the
--- a/src/main/java/org/omegat/core/statistics/CalcMatchStatistics.java
+++ b/src/main/java/org/omegat/core/statistics/CalcMatchStatistics.java
@@ -8,6 +8,7 @@
                2013 Alex Buloichik
                2015 Aaron Madlon-Kay
                2024 Hiroshi Miura
+               2026 Stephan Pakebusch
                Home page: https://www.omegat.org/
                Support center: https://omegat.org/support
 
@@ -41,10 +42,14 @@ import java.time.ZoneId;
 import java.time.format.DateTimeFormatter;
 import java.time.format.FormatStyle;
 import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.IntPredicate;
 import java.util.logging.Logger;
@@ -95,6 +100,11 @@ public class CalcMatchStatistics extends CalcStandardStatistics implements ICalc
             OStrings.getString("CT_STATS_Words"), OStrings.getString("CT_STATS_Characters_NOSP"),
             OStrings.getString("CT_STATS_Characters") };
 
+    /**
+     * Row labels of the total table. Its final shape (all rows except index 1,
+     * plus the total row) is mirrored by the filter buttons of
+     * org.omegat.gui.stat.MatchStatisticsPanel.
+     */
     protected final String[] rowsTotal = new String[] { OStrings.getString("CT_STATSMATCH_RowRepetitions"),
             OStrings.getString("CT_STATSMATCH_RowExactMatch"), OStrings.getString("CT_STATSMATCH_RowMatch95"),
             OStrings.getString("CT_STATSMATCH_RowMatch85"), OStrings.getString("CT_STATSMATCH_RowMatch75"),
@@ -115,6 +125,17 @@ public class CalcMatchStatistics extends CalcStandardStatistics implements ICalc
 
     /** Already processed segments. Used for repetitions detect. */
     private final Set<String> alreadyProcessedInProject = new HashSet<>();
+
+    /**
+     * Category row index per entry number, following the row semantics of
+     * {@link MatchStatCounts}. Concurrent because the similarity pass may run
+     * on a parallel stream. Only recorded for total calculations that output
+     * data; the per-file mode reuses calcSimilarity but never delivers the map.
+     */
+    private final Map<Integer, Integer> entryRowIndexes = new ConcurrentHashMap<>();
+
+    /** Volatile because calcSimilarity may record from parallel stream workers. */
+    private volatile boolean recordEntryRows;
 
     private final ThreadLocal<ISimilarityCalculator> distanceCalculator = ThreadLocal
             .withInitial(LevenshteinDistance::new);
@@ -210,6 +231,8 @@ public class CalcMatchStatistics extends CalcStandardStatistics implements ICalc
     MatchStatCounts calcTotal(boolean outData) {
         MatchStatCounts result = new MatchStatCounts();
         alreadyProcessedInProject.clear();
+        entryRowIndexes.clear();
+        recordEntryRows = outData;
 
         final List<SourceTextEntry> untranslatedEntries = new ArrayList<>();
 
@@ -222,10 +245,17 @@ public class CalcMatchStatistics extends CalcStandardStatistics implements ICalc
                 // segment has translation - should be calculated as "Exact
                 // matched"
                 result.addExact(count);
+                if (recordEntryRows) {
+                    entryRowIndexes.put(ste.entryNum(),
+                            MatchStatCounts.getRowByPercent(Statistics.PERCENT_EXACT_MATCH));
+                }
                 entryProcessed();
             } else if (!isFirst) {
                 // already processed - repetition
                 result.addRepetition(count);
+                if (recordEntryRows) {
+                    entryRowIndexes.put(ste.entryNum(), MatchStatCounts.ROW_REPETITIONS);
+                }
                 entryProcessed();
             } else {
                 // first time
@@ -241,11 +271,16 @@ public class CalcMatchStatistics extends CalcStandardStatistics implements ICalc
         }
 
         calcSimilarity(untranslatedEntries).ifPresent(result::addCounts);
+        // calcSimilarity swallows cancellation to deliver partial results to
+        // the closing window, but a cancelled run must not be published as a
+        // complete result (entryRowIndexes would have gaps).
+        cancellationToken.throwIfCancelled();
 
         if (outData) {
             String[][] table = result.calcTable(rowsTotal, i -> i != 1);
             String outText = TextUtil.showTextTable(header, table, align);
             showText(outText);
+            callback.setEntryRowIndexes(Collections.unmodifiableMap(new HashMap<>(entryRowIndexes)));
             callback.setTable(header, table);
             String fn = project.getProjectProperties().getProjectInternal() + OConsts.STATS_MATCH_FILENAME;
             writeStat(fn, outText);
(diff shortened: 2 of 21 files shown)


miurahr (changes requested):
## Design review: MatchStatisticsCache / MatchStatisticsPanel state handling

This is a design-level review rather than line comments, since it touches how
state flows across three files (`CalcMatchStatistics`, `MatchStatisticsCache`,
`MatchStatisticsPanel`). I'm leaving the concrete refactor approach to you as
the author — the points below explain *why* each concern exists and what
constraints a fix needs to satisfy, but the implementation choice is yours.

### 1. `FINAL_TABLE_ROWS = 8` in `MatchStatisticsPanel` duplicates a fact that belongs to `MatchStatCounts`

```java
static final int FINAL_TABLE_ROWS = 8;
```

This constant exists to distinguish a completed *total* scan (whose table has
a fixed shape: all categories + the total row) from an intermediate or
per-file table, inside `onComplete`:

```java
if (completion.isSuccess() && rows != null && headers != null && data != null
        && data.length == FINAL_TABLE_ROWS && projectRoot != null) {
    MatchStatisticsCache.store(headers, data, rows, lastTextData, projectRoot);
}
```

The actual shape it mirrors is defined in `CalcMatchStatistics`:

```java
/**
 * Row labels of the total table. Its final shape (all rows except index 1,
 * plus the total row) is mirrored by the filter buttons of
 * org.omegat.gui.stat.MatchStatisticsPanel.
 */
protected final String[] rowsTotal = new String[] { ... };
```

That javadoc already flags this as a manual sync point — but right now the
sync point is a *comment*, and the number `8` is duplicated as a bare literal
in a different package. If `rowsTotal` ever gains or loses a row, nothing
will fail loudly; caching will just silently stop firing (or start firing on
the wrong table shape) because the `data.length == FINAL_TABLE_ROWS` check
quietly stops matching.

This PR already establishes the right pattern for this exact kind of
constant: `MatchStatCounts.ROW_REPETITIONS` (a public static final int
describing row semantics) and `MatchStatCounts.getRowByPercent(int)` (made
`public static` in this PR, previously private). Both live in
`MatchStatCounts`, which is becoming the authoritative place for "what a row
index means" in the match statistics table.

**Ask:** move `FINAL_TABLE_ROWS` into `MatchStatCounts` as a public static
final constant, following the `ROW_REPETITIONS` pattern, with a javadoc note
that ties it explicitly to `CalcMatchStatistics.rowsTotal`'s length — e.g.:

```java
/**
 * Number of rows in the final total match statistics table, including the
 * total row. Must be kept in sync with {@link CalcMatchStatistics#rowsTotal}
 * (all categories plus the total row). Used by
 * {@link org.omegat.gui.stat.MatchStatisticsPanel} to distinguish a
 * completed total scan from an intermediate or per-file table.
 */
public static final int FINAL_TABLE_ROWS = 8;
```

`MatchStatisticsPanel` then references `MatchStatCounts.FINAL_TABLE_ROWS`
instead of defining its own copy. This doesn't remove the manual-sync
obligation entirely (that would mean deriving the row count from
`rowsTotal.length` dynamically, which is a bigger change and out of scope
here), but it collapses the duplication down to one number in one place with
one javadoc, referenced bidirectionally from both classes — which is the
minimum needed to make future drift visible instead of silent. This is
deliberately scoped as the minimal fix, not a request to redesign how row
counts are derived.

### 2. `MatchStatisticsCache.Snapshot` has asymmetric defensive copying

```java
Snapshot(String[] headers, String[][] data, Map<Integer, Integer> entryRowIndexes,
        @Nullable String textData, String projectRoot, Instant lastScan) {
    this.headers = headers.clone();
    this.data = new String[data.length][];
    for (int i = 0; i < data.length; i++) {
        this.data[i] = data[i].clone();
    }
    this.entryRowIndexes = entryRowIndexes;   // <-- no copy
    this.textData = textData;
    this.projectRoot = projectRoot;
    this.lastScan = lastScan;
}
```

The class javadoc says "Immutable result of one finished match statistics
scan," but `entryRowIndexes` is stored by reference while `headers`/`data`
are cloned. In practice this is safe today only because the caller
(`CalcMatchStatistics.calcTotal`) happens to pass
`Collections.unmodifiableMap(new HashMap<>(entryRowIndexes))` — i.e. the
safety is an accident of the caller's behavior, not a guarantee the class
itself enforces. Anyone constructing a `Snapshot` from a different call site
in the future could pass a live, mutable map without the compiler or the
class contract stopping them.

### 3. Related: `MatchStatisticsPanel` holds parallel state as four independent volatile fields, which is where I'd like the fix for #2 to also land

```java
private volatile @Nullable Map<Integer, Integer> entryRowIndexes;
private volatile String @Nullable [] lastHeaders;
private volatile String @Nullable [][] lastData;
private volatile @Nullable String lastTextData;
```

These are written independently by `setEntryRowIndexes`, `setTable`, and
`setTextData` (three different `IStatsConsumer` callback methods), then all
read together in `onComplete` to build the cache entry. I confirmed that
`setTable` itself is fine — it assigns `lastHeaders`/`lastData` synchronously
before deferring the Swing repaint via `invokeLater`, and rendering uses the
method's own parameters rather than the fields, so the paint path has no
torn-read risk. The risk is narrower than I first thought: it's specific to
`onComplete` reading four separately-written fields as if they were one
atomic unit.

Per the `IStatsConsumer` contract, `setEntryRowIndexes` → `setTable` →
`onComplete` are called synchronously in that order *within one scan*, so
under normal sequential use this is fine. The open question is whether two
scans can ever be in flight concurrently against the same
`MatchStatisticsPanel` instance — e.g. the menu item invoked again before a
prior scan's `onComplete` fires. I'd also ask whether the CLI stats command
or the team-project auto-stat-file feature can ever target the same
`IStatsConsumer` callback as an open GUI window, or whether they always use
a separate consumer — if the latter, this concern is scoped entirely to the
GUI menu path; if the former, it's broader. If overlapping invocations
against the same panel are possible by any of these paths, `onComplete`
could read a "torn" combination — `entryRowIndexes` from one scan,
`lastHeaders`/`lastData` from another — and persist that inconsistent
combination into the cache.


### Suggested direction for #2 and #3 together

Rather than treating "fix the Snapshot copy asymmetry" and "the panel's four
parallel fields" as separate patches, I think they're the same underlying
issue: match statistics results are being modeled as loose, independently
mutable pieces (arrays + a map + a string) in two different places
(`MatchStatisticsPanel`'s fields and `MatchStatisticsCache.Snapshot`),
instead of as one cohesive value type.

Concretely, I'd suggest:

- Define a single record type for a match-statistics result — something like
  `MatchStatisticsResult(String[] headers, String[][] data,
  Map<Integer, Integer> entryRowIndexes, @Nullable String textData)` — and
  use **the same type** in both `MatchStatisticsPanel` and
  `MatchStatisticsCache`, rather than `Snapshot` being a cache-specific
  reshaping of what the panel already assembled piecemeal.
- Give it a builder that `MatchStatisticsPanel` accumulates into across the
  three `IStatsConsumer` callbacks, replacing the four independent volatile
  fields with a single builder reference. `onComplete` then calls one
  `build()` instead of reading four fields separately — which also directly
  addresses #3, since there's no longer a way to read a partially-updated
  combination.
- **Important:** a Java `record` only makes the *reference* to
  `headers`/`data`/`entryRowIndexes` final — it does not by itself protect
  against a caller mutating the array/map contents after construction, and it
  does not replace the need for defensive copying. To actually close the gap
  described in #2, the copying needs to happen in exactly one place that
  cannot be bypassed — e.g. a compact constructor on the record:

```java
  public record MatchStatisticsResult(
          String[] headers,
          String[][] data,
          Map<Integer, Integer> entryRowIndexes,
          @Nullable String textData) {

      public MatchStatisticsResult {
          headers = headers.clone();
          data = Arrays.stream(data).map(String[]::clone).toArray(String[][]::new);
          entryRowIndexes = Map.copyOf(entryRowIndexes);
      }
  }
```

  If the builder's `build()` always routes through this constructor, the
  headers/data/map copying becomes symmetric and enforced by the type
  itself, regardless of what any future caller passes in — closing the gap
  permanently rather than relying on today's caller behaving correctly.

- If a builder-per-panel-instance is used, please double check its
  lifecycle: if the same builder instance is reused across multiple scans
  (rather than a fresh builder being created per scan), the torn-read
  concern from #3 isn't actually closed, just moved — two overlapping scans
  would still race to fill the same builder. Whether this matters depends on
  the answer to the "can two scans overlap" question above.

### Two ways to place the cache-specific metadata (projectRoot, lastScan) — author's choice

`Snapshot` currently bundles the pure statistics data together with
cache-specific metadata (`projectRoot`, `lastScan`). Once there's a shared
`MatchStatisticsResult` type, there are two reasonable ways to fold that
metadata in, and I don't have a strong preference between them — this is a
style choice, not a correctness issue:

- **(a) One record for everything.** `MatchStatisticsResult` itself carries
  `projectRoot` and `lastScan` alongside the statistics data, and
  `MatchStatisticsPanel`'s builder fills them in (`Core.getProject()...`,
  `Instant.now()`) when it builds. Simpler — one type total — but the panel
  ends up carrying fields (project root, timestamp) that it doesn't
  otherwise care about for rendering.
- **(b) Two types: a pure result, wrapped for caching.** Keep
  `MatchStatisticsResult` limited to `headers`/`data`/`entryRowIndexes`/
  `textData`, and have `MatchStatisticsCache` wrap it in something like
  `record Entry(MatchStatisticsResult result, String projectRoot, Instant
  lastScan)`. Slightly more types, but a cleaner separation: the panel deals
  only in "what the scan produced," and the cache is the only place that
  knows about project-scoping and staleness.

Either is fine with me — please pick whichever reads better to you, as long
as the defensive-copy guarantee in the constructor described above is
preserved in whichever type ends up holding the arrays/map.

zs-stpa:
ok you can make walls of text, too, i see ^.^ but i read it later and act accordingly, have a nice weekend

zs-stpa:
9f5dfb7: intermediate table shown during scan was misaligned - now uses same rows as final table, recalculating forgets previous entry mapping, so filter buttons never act on stale categories (mapping of running scan arrives before its final table, tested). Side effect: MatchStatCounts.calcTableWithoutPercentage no longer used in production code.

Requested changes were implemented in two commits:

fa68c56: FINAL_TABLE_ROWS moved into MatchStatCounts next to ROW_REPETITIONS, javadoc ties it to CalcMatchStatistics row labels, test pins constant to actually produced table so drift fails loudly.

8038d1c: new record MatchStatisticsResult(headers, data, entryRowIndexes, textData) with compact constructor cloning arrays and Map.copyOf - defensive copy enforced in one place no caller can bypass. Panel accumulates callbacks into nested Builder; builder replaced wholesale when calculation starts, so parts of two scans cannot combine into one result and build() in onComplete reads one accumulator instead of four fields. Cache stores record Entry(result, projectRoot, lastScan) - option (b), pure result wrapped with cache metadata.

On overlap question: CalcMatchStatistics is instantiated only via StatisticsWindow (STAT_TYPE.MATCHES); CLI stats command and team auto-stat files go through separate StatisticsTextWriter consumer, never through open panel. On GUI path overlap is prevented by disabled recalculate button until onComplete and by one-window-per-type map, so concern was scoped to GUI only; builder-per-scan closes it structurally anyway. Builder lifecycle: fresh instance per startCalculation and per cache restore, never reused across scans.

brandelune:
Here again, you modify a GUI function to fit your purpose without discussing it first.

You make the window not an information tool, but a filtering interface. Which it was never meant to be.

zs-stpa:
yes correct. still getting used to how things are suppoed to work here. sorry for that. i will post something on the mail list.

the intent of the match statistics window seemed to be* to give information to the translator user. but that information
- takes long to calculate in a big project
- was not actionable
- not usable
just informative.

i turned it into something i could really usefully use: filter for good matches to rapidly turn them into good translations, did not know how to do that efficiently before in released versions.

*: yeah i probably really should read all the documentation from start to finish myself instead of relying on search functions and filtering there)

i did NOT do this to personally offend anyone, i did this because i thought this could help translators improve efficiency of their work.

ciao.

zs-stpa:
just found the exact matching ticket: https://sourceforge.net/p/omegat/feature-requests/863/ other users need this too.

zs-stpa:
i keep this part out of the mailing list mail:

the accusation was, that "You [i.e. me] make the window not an information tool, but a filtering interface. Which it was never meant to be." which confirms a pattern with me: i try to improve existing things. which is not to say they were not good before, on the contrary. omegaT is great and loveable and all previous work deserves the greatest respect. so i am trying to pay my respects by making it this tiny bit better and more usable. and get the impression this is not welcome at all, unless there is at least ten times more prose than code written. is my impression only, is not a counter accusation … is just personal feeling,.

brandelune:
> i keep this part out of the mailing list mail:
> 
> the accusation was, that "You [i.e. me] make the window not an information tool, but a filtering interface. Which it was never meant to be." which confirms a pattern with me: i try to improve existing things. which is not to say they were not good before, on the contrary. omegaT is great and loveable and all previous work deserves the greatest respect. so i am trying to pay my respects by making it this tiny bit better and more usable. and get the impression this is not welcome at all, unless there is at least ten times more prose than code written. is my impression only, is not a counter accusation … is just personal feeling,.

https://omegat.readthedocs.io/en/latest/04.ContributingOmegaT.html#start-a-conversation

Start a conversation with relevant RFEs, or your ideas and then we move on.

PR #2264, opened 2026-08-18
fix(languagetools+spellcheck): never show matches inside protected placeholders where translator is not supposed to change anything anyway

zs-stpa:
## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?

- (no ticket)

## What does this PR change?

- LanguageTool matches lying entirely inside a protected part in the translation are suppressed, in the editor marker and in the Issues window. Style rules routinely misread placeholder syntax, for example the currency rule firing on the "3$" inside "%3$@". Matches reaching beyond a placeholder are kept.

### DIFF
--- a/src/main/java/org/omegat/core/data/ProtectedPart.java
+++ b/src/main/java/org/omegat/core/data/ProtectedPart.java
@@ -4,6 +4,7 @@
           glossaries, and translation leveraging into updated projects.
 
  Copyright (C) 2013 Alex Buloichik
+               2026 Stephan Pakebusch
                Home page: https://www.omegat.org/
                Support center: https://omegat.org/support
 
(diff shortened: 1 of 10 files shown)


zs-stpa:
Added follow-up commit 15394734b: same suppression for spell checker. Spelling marks and spelling issues no longer flag words inside protected placeholders (example: "ld" inside "%1$ld"). Occurrence computation moved to shared ProtectedPart.occurrencesIn, used by LanguageTool filter and spell checker filter. Context menu suggestions inside placeholders remain; separate change.

zs-stpa:
The right-click popup now follows the same filter: no suggestions and no learn/ignore inside placeholders.

miurahr (approved):
LGTM.

PR #2274, opened 2026-08-22
fix(core): tolerate unknown elements in the project file parser

zs-stpa:
## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?

- No ticket; preventive robustness fix observed while extending the project schema.

## What does this PR change?

- `ProjectFileStorage` no longer rejects unknown elements in `omegat.project`: one team member saving with a newer OmegaT (extended schema) locked every member on older versions out of the project with "Unable to read project file".
- Unknown elements are now ignored on read, matching how the other configuration files behave; known fields are unaffected (covered by new test).


miurahr:
Good catch! LGTM.
This worth backport to 6.1.x

miurahr (approved):

PR #2276, opened 2026-08-23
fix(team): repositories mapping cancel

zs-stpa:
## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?

- (none)

## What does this PR change?

- Repositories mapping dialog returns result only after confirmed OK. Since #1752, every exit path returned current table snapshot written straight into live project properties; one cancelled dialog plus next auto-save silently dropped `repositories` element from `omegat.project`.
- Properties dialog buffers mapping edits like SRX and filters, applies on OK; window close counts as Cancel.
- Validation rejects two mappings onto same local folder (previously saved fine, project then failed to load with "Multiple mappings for file").
- Project load drops exact duplicate mapping entries with logged warning instead of failing.
- Tests cover dialog exit paths, validation, duplicate cleanup.

Regression ships in 6.1.0/6.1.1; backport PR for `releases/6.1` follows.


miurahr (approved):

miurahr:
There is no objection from other reviewers and code owners of Bundle.properties in 2 weeks. I would like to merge it.

brandelune (approved):

PR #2286, opened 2026-08-26
chore: tweak Gradle warnings

miurahr:
Implicit reference to root project preference in subproject will become an error in Gradle 10. This PR to tweak several `build.gradle` to avoid warnings of Gradle 10 compatibility.

## Pull request type

## Which ticket is resolved?
none
## What does this PR change?

- Replace implicit property reference as explicit `rootProject.property()`
- avoid implicit file test and use explicit file()
- change dependency alias as `assertj.core`
- use a root project dependency `project(':')` instead of implicit `project.rootProject`
- migrate `ext.providedModuleLib()` to `org.omegat.omegatModule` function.


miurahr:
@zs-stpa  I modified your contribution #2227 to support a migration to Gradle 10.
Could you review the changes?

build-logic/src/main/groovy/org/omegat/module/OmegatModulePlugin.groovy
```diff
@@ -210,6 +212,48 @@ class OmegatModulePlugin implements Plugin<Project> {
...
+        private def providedLib(File libsDir, String... artifacts) {
+            def patterns = artifacts.collect { "**/${it}-*.jar".toString() }
+            def tree = project.fileTree(dir: libsDir, includes: patterns)
+            if (tree.isEmpty()) {
+                throw new GradleException("No jar matches ${patterns} in ${libsDir}; "
+                        + "the provided libraries are incomplete.")
+            }
+            return tree
+        }
+    }
```
miurahr: I moves @zs-stpa contribution to define `providedModuleLib` into `OmegatModulePlugin` with genelarization for both module and core.
Now we can remove `project.extention` definition and access the function through plugin API.

zs-stpa: agreed, good direction. possible follow-ups to keep the #2227 guarantees intact:

- Per-artifact check in `providedLib`: current `tree.isEmpty()` now succeeds when any of several artifacts matches - the aligner call has 9 -> one missing jar goes unnoticed. Loop over artifacts, require ≥1 match each
- `providedCoreLib` turns ~20 formerly silent-empty core-lib lookups into failures - ok, but maybe worth a line in the commit message, also maybe restore "module"/"core" in the error text + explanation comment
- Path literals now duplicated (extension class + root build.gradle sourceDist block), maybe `static final` constants on plugin/extension, root script imports them, single source again

Rest looks good; all call sites converted cleanly, no dangling `ext` refs found.

miurahr: I will fix the detection of the `providedLib` and improve error message.

miurahr: I believe set all changes. Now ready.

PR #2293, opened 2026-08-27
refactor: remove OpenXmlFilterTest and simplify OpenXmlFilter logic

miurahr:
## Pull request type
- Other (describe below)
refactor
## Which ticket is resolved?
None; a work to reduce doddy codes.

## What does this PR change?
- Refactored OpenXmlFilter to improve null safety and readability.
- Replaced usage of LinkedList with ArrayList for performance optimization.
- Because OpenXmlFilter.doCompactTags never become true in the ordinal execution path, so streamline unreached codes and obsolete OpenXmlFilterTest class.


miurahr:
There is no feedback in 2 weeks. I would like to merge it soon.

t-cordonnier:
Hi Hiroshi

Even if StaX filter generates less tags than the old Filters3 native filter, compating tags can sometimes be a good option. If you are right and the way I activate it is wrong, shouldn't you try to find another way to activate it, rather than immediately delete it?

By the way, isFileSupported is a part of IFilter interface and even if OpenXmlFilter is an internal class, it implements it. I initially wanted to use OpenXmlFilter.isFileSupported inside AbstractZipFilter.isFileSupported but browsing a zip generates ZipEntry which is not compatible with java.io.File, reason why I created other methods. But now I realize that my methods do not take care of configuration,  while isFileSupported does, so the contents of OpenXmlFilter.isFileSupported becomes useful again... Let me have a look.
In any case, when I see a bug I try to correct the bug, not to remove the feature.

In any case for the moment I cannot approve this pull request.

Regards
Thomas

t-cordonnier:
> There is no feedback in 2 weeks. I would like to merge it soon.

 Really? Please see my message from 4/9/2026, sent as reply to one of your mails
Just pushed a copy in the PR, just in case
RegardsThomas

    Le vendredi 11 septembre 2026 à 16:08:31 UTC+2, Hiroshi Miura ***@***.***> a écrit :  
 
 miurahr left a comment (omegat-org/omegat#2293)
There is no feedback in 2 weeks. I would like to merge it soon.

t-cordonnier:
Except of that, did you really see differences when using ArrayList? A test?

From what I know, ArrayList is good when
- you frequently make use of list.get(i), based on integer index
- you have an idea of the size of the list before filling it (else, list.add(...) has an extra cost)
On the contrary I selected LinkedList here because:
- this filter always works with iterators with prev and next, never list.get(i)
- it is difficult to know in advance the size of the list

For these reasons LinkedList sounds better in theory here. But if you have practical results about performance, I remain open

Thomas

miurahr:
When force enabling the feature I found as a dead code, many acceptance and functional test failed. It is reason I consider the code is really dead.
It is good chance for you to fix the code enabling a feature,  and write test in patterns. 

About LinkedList.  
You're right to push back, and I should have read the method more carefully. Looking at `buildTags()` again, the attribute-collapsing block in `LOOP2` does use `ListIterator.remove()/add()` in the middle of the run list, and there is `currentPara.remove(j)` and `nList.addFirst(...)` as well. So my premise that the lists are append-then-iterate only was wrong.

For what it's worth, the method also uses `get(0)`, `get(1)`, `currentPara.get(i)` and `subList(...)`, so neither structure is a clean fit in theory. In practice the lists are per-paragraph and per-run, i.e. small enough that the difference is not measurable either way, and I don't have numbers to justify overriding your choice. I'll revert this part of the change and keep the PR to the other cleanups.

One unrelated thing I noticed while reading: in `LOOP2`, `currentPara.remove(j); continue;` skips the element that shifts into position j. I'll open a separate issue for that so it doesn't get lost here.

src/main/java/org/omegat/filters4/xml/openxml/OpenXmlFilter.java
```diff
@@ -118,15 +110,18 @@ protected boolean processStartElement(StartElement startElement, XMLStreamWriter
             if ("p".equals(name.getLocalPart()) // word
                     || "si".equals(name.getLocalPart()) || "comment".equals(name.getLocalPart()) // excel
             ) {
-                currentBuffer = new LinkedList<>();
-                currentPara = new LinkedList<>();
+                currentBuffer = new ArrayList<>();
+                currentPara = new ArrayList<>();
```
miurahr:
```suggestion
                currentBuffer = new LinkedList<>();
                currentPara = new LinkedList<>();
```

miurahr:
I reverted many of changes to focus a dead code only.

t-cordonnier:
The code is not dead, it is unreachable, which is a bug. This PR should not be merged.

PR #2301, opened 2026-08-27
fix(gui): stop menu hover from parsing the current source file

zs-stpa:
## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?

- (no SourceForge ticket; found during development)

## What does this PR change?

- Run enablement checks of the access-project-contents submenu in a SwingWorker instead of the EDT; the target-path lookup there can parse the whole current source file (XHTML for example) and froze the UI on every hover.
- Cache source-to-target path mapping in RealProject, invalidated when another FilterMaster instance is installed.
- Add RealProjectTest coverage for cache hit and invalidation.

## Other information

First hover after project load shows previous enablement until the background check finishes; clicking a stale entry degrades to the existing status-bar message.

### DIFF
--- a/src/main/java/org/omegat/core/data/RealProject.java
+++ b/src/main/java/org/omegat/core/data/RealProject.java
@@ -55,13 +55,21 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.OptionalLong;
 import java.util.Set;
 import java.util.TreeMap;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
+import javax.cache.Cache;
+import javax.cache.CacheManager;
+import javax.cache.Caching;
+import javax.cache.expiry.CreatedExpiryPolicy;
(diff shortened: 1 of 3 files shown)


miurahr:
@zs-stpa We already use a JCache Java Standard cache API in the OmegaT especially for MT and dictionary lookup results. We use a Caffeine library as the jcache implementation. The target path cache you propose is better to replace with the jcache. JCache supports LRU expiration and better memory pressure handling.

miurahr (changes requested):
Please consider jcache. ref:  `o.o.core.machinetranslators.BaseCachedTransate#getCacheLayer` 

And why invalidate with FIlterMaster instance for the cache here? ProjectChange Event is better abstraction of the change, is not it?

zs-stpa:
Agreed for project lifecycle (now done via CLOSE event), the instance check covers filter-preference editing, which install a new FilterMaster without a project event (which is optional, user can decline)

miurahr (approved):

PR #2303, opened 2026-08-27
feat(matches): character equivalence classes in fuzzy matching and search (SF-RFE-1681)

zs-stpa:
## Pull request type

## Which ticket is resolved?

- Straight quotes match curly quotes
  * https://sourceforge.net/p/omegat/feature-requests/1681/

## What does this PR change?

- Fuzzy matching folds character variant classes together: quotation marks and apostrophes, dashes, spaces, invisible formatting characters; NFC applies before comparison.
- Classes active by default, configurable per project via dialog with character inventories and live test area; opt-out persists as optional `match_equivalence_disabled` element in `omegat.project` ~(written only when disabling; reading requires tolerant parser from #2274).~ (written as sidecar outside project file so that older released versions simply ignore the new setting).
- Exact and keyword search match same variants pattern-side; offsets, highlighting, replace stay on raw text. New checkbox row per search/replace panel, prefilled from project setting, disabled in regexp mode.
- Replaces "Space matches nbsp" option; preference keys deprecated, `SearchExpression.spaceMatchNbsp` replaced by `equivalences`, regexp-mode nbsp rewriting removed.
- Match statistics windows and files name active classes.

## Other information

Folded match still differs literally from source; replacing span with tolerated invisible characters removes them.

### screenshots

(only attached to this comment, not part of the PR, will not be updated if details change)

with new default project setting active, all 3 fuzzy match passes (stem, nostem, verbatim) show a 100% match now, even though ‘ is not exactly equal '
Bildschirmfoto 2026-08-27 um 20 39 05

in the project settings is a new button "Character equivalence…"
Bildschirmfoto 2026-08-27 um 20 39 25

that opens a window with scrollable character info list per equivalence class and a test-text field with preset example text where user can experiment with text and equivalence classes for their own use case. characte lists of classes are not user editable, only opt-outable.
Bildschirmfoto 2026-08-27 um 20 39 36

in the search window, the NBSP checkbox is obsolete and thus removed, instead a new line with four checkboxes for the new equivalence classes, preset with project setting, disabled when regex, search for ‘ with the equivalence checkbox unchecked does not find the segment
Bildschirmfoto 2026-08-27 um 20 43 36

search for ‘ with the checkbox checked does find the segment
Bildschirmfoto 2026-08-27 um 20 43 26

the replace window also has the new checkboxes and only applies them in the search, not in the replace.
Bildschirmfoto 2026-08-27 um 20 44 29

### DIFF
--- a/src/docs/manual/en/Dialogs_ProjectProperties.xml
+++ b/src/docs/manual/en/Dialogs_ProjectProperties.xml
@@ -196,23 +196,43 @@
                </note>
             </listitem>
          </varlistentry>
-         <varlistentry xml:id="dialogs.project.properties.match.numbers">
-            <term xml:id="dialogs.project.properties.match.numbers.title">
-               <option>Consider numbers in fuzzy matching</option>
+         <varlistentry xml:id="dialogs.project.properties.match.equivalence">
+            <term xml:id="dialogs.project.properties.match.equivalence.title">
+               <option>Character equivalence...</option>
             </term>
             <listitem>
-               <para>By default, the word-based match percentages ignore numbers, and
-                  		  a segment that consists only of numbers (a date, a bare quantity)
-                  		  scores too low against a segment with different numbers to be
-                  		  offered as a match.</para>
-               <para>Activating this option makes fuzzy matching compare number
-                  		  tokens by their numeric value: numbers written in different
-                  		  numbering systems (Roman numerals, ideographic numbers,
-                  		  Arabic-Indic or full-width digits, and so on) count as equal to
-                  		  their Western counterpart of the same value. Two segments that
-                  		  consist only of numbers are scored against each other as number
-                  		  sequences, with a small fixed penalty, so differing dates or bare
-                  		  numbers surface as fuzzy matches at all.</para>
+               <para>Opens a dialog with the character equivalence classes used by
+                  		  fuzzy matching and search: quote, dash, space and
+                  		  invisible-format variants fold onto each other. Each class
+                  		  lists its complete character inventory, a test area compares
+                  		  two sample texts with the current selection, and the dialog
+                  		  also holds the number matching option described below.</para>
+            </listitem>
+         </varlistentry>
+         <varlistentry xml:id="dialogs.project.properties.match.numbers">
+            <term xml:id="dialogs.project.properties.match.numbers.title">
+               <option>Number value equality</option>
+            </term>
+            <listitem>
+               <para>This option is part of the equivalence dialog opened with the
+                  		  <option>Character equivalence...</option> button. It is active
+                  		  by default.</para>
+               <para>While active, fuzzy matching compares number tokens by their
+                  		  numeric value: numbers written in different numbering systems
+                  		  (ideographic numbers, Arabic-Indic or full-width digits, and so
+                  		  on) count as equal to their Western counterpart of the same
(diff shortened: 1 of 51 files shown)


brandelune:
The search window does not need "find next / find previous" buttons. We already have documented shortcuts.

Remove them from master please.

zs-stpa:
> The search window does not need "find next / find previous" buttons. We already have documented shortcuts.
> 
> Remove them from master please.

out of scope, sorry, that was another PR and seems to be merged already in master.

clarification: the buttons came from #2129 in response to https://sourceforge.net/p/omegat/feature-requests/1380/ which is titled literally 'Add a "Find Previous" button to the Search window' and in response to https://sourceforge.net/p/omegat/feature-requests/1125/ with title 'Find Next button on Search window', and also their functionality is also different to the preexisting functionality in three points:
- before there was double click on search results, which lost search focus to the editor
- arrow key navigation in search results (up/down) only focus editor with auto-sync checked did not cycle through ends
- the documentation of the shortcuts cmd+n/p is also slightly different from what the new (shift+)f3 does and mainly from where it does it (old only when result list focussed, new from anywhere in search window)
and plus: the new does not replace the old behavior, is just added to it, so no old muscle memory forced to retraining

brandelune:
> * before there was double click on search results, which lost search focus to the editor

??? There is no need to double-click a search result to open the segment in the editor.

> * arrow key navigation in search results (up/down) only focus editor with auto-sync checked did not cycle through ends

I am not talking about arrow navigation, but about documented shortcuts : C-n / C-p

If you need to add a cycling through ends, that’s a different (and unrequested) PR. And you can add that to the current behavior.

> * the documentation of the shortcuts cmd+n/p is also slightly different from what the new (shift+)f3 does and mainly from where it does it (old only when result list focussed, new from anywhere in search window)

That’s not a requested feature.

>   and plus: the new does not replace the old behavior, is just added to it, so no old muscle memory forced to retraining

You’re adding an unrequested behavior, buttons and a shortcut.

zs-stpa:
new text:

## Stacking

Stacked on #2319 (team settings registry) and #2228 (number value matching, SF-465): base #2319 → #2228 → this PR. Diff shows both until they merge; review only the top commit.

## What does this PR change?

- Fuzzy matching and exact/keyword search fold curated character equivalence classes: quote, dash, space and invisible-format variants compare as equal; NFC always applies; replaces search-window nbsp option
- Per-project configuration behind `Character equivalence...` button: per-group summaries with per-character tooltips, copy buttons with full inventory, live test areas with multilingual example pairs; match statistics label active classes
- Number matching (option from #2155) moves into equivalence dialog, renamed "Number equality by value across writing systems"; now ON by default (opt-out), new opt-in sub-option reads Latin-letter Roman numerals (dedicated code points always count)
- Number-only search term matches by value across numeral systems and source/target locale separator formats ("1.234" finds 1234 and vice versa in de-en project); literal writings ranked first; number checkboxes in search and replace panels, prefilled from project
- omegat.project: `match_numbers` written only on opt-out, new optional `match_numbers_roman`; note: OmegaT without this change cannot open project file carrying unknown elements, relevant for mixed-version teams once someone opts out
- Manual and tip page updated; regression tests across parser, storage, matching, substitution, search and dialog

zs-stpa:
latest screenshot, still not final.

Bildschirmfoto 2026-09-08 um 21 06 27

PR #2273, opened 2026-08-22
feat(editor): additional editing shortcuts for more efficient editing

zs-stpa:
## Pull request type

- Feature enhancement -> [enhancement]

## Which ticket is resolved?

- Additional editing functions based on Readline shortcuts
  * https://sourceforge.net/p/omegat/feature-requests/536/

- Placeables auto-completer view (partially)
  * https://sourceforge.net/p/omegat/feature-requests/1266/

- Shortcut for inserting "tag pairs" (partially)
  * https://sourceforge.net/p/omegat/feature-requests/1302/

## What does this PR change?

- New editor shortcuts, each user-adjustable via `EditorShortcuts.properties` in config dir: delete to segment start/end, forward word delete without Forward-Delete key, caret/selection to segment bounds, cycle through source placeables (numbers, tags, URLs, e-mails) missing from target, no-break space, first missing tag pair (around selection if any), move word at caret over neighbours.
- Range deletes and word move clamp to editable range, respect tag protection and form single undo steps via new `TranslationUndoManager.runAtomic` — undo restores text and caret position.
- Pure text computations live in new, unit-tested `SegmentEditingOps`; tag-pair offerings reuse `TagUtil.getGroupedMissingTagsFromTarget`.
- Tip of the day (en+de) documents shortcuts, separate commit.

Notes: `ctrl+shift+SPACE` on non-mac platforms yields to open auto-completer popup; on Linux `ctrl+PERIOD` may be taken by IBus emoji hotkey and `ctrl+alt+arrows` by desktop workspace switching — remappable per user config.

## Other information

Tip of the day text:

> **Faster editing**
> Extra editing shortcuts (macOS in parentheses):
> - Delete everything left or right of cursor: Ctrl+Shift+Backspace / Ctrl+Shift+Delete (Cmd+Backspace / Cmd+Shift+Backspace)
> - Delete next word, no Forward-Delete key needed: Alt+Shift+Backspace
> - Jump to segment start or end, Shift selects: Ctrl+Home / Ctrl+End (Cmd+Up / Cmd+Down)
> - Insert next number, tag or URL still missing from translation: Ctrl+Comma; last one: Ctrl+Shift+Comma
> - Insert non-breaking space: Ctrl+Shift+Space
> - Insert first missing tag pair, around selection if any: Ctrl+Period
> - Move word at cursor left or right: Ctrl+Alt+Left / Ctrl+Alt+Right (Cmd+Alt+Left / Cmd+Alt+Right)

PR #2305, opened 2026-08-29
[BUGS#1334]fix(editor): identical translation is broken

miurahr:
PR#1803 to fix [BUGS1226](https://sourceforge.net/p/omegat/bugs/1226/) enbugs the identical translation.

## Pull request type
- Bug fix -> [bug]

## Which ticket is resolved?

- Register Identical Translation is broken
- https://sourceforge.net/p/omegat/bugs/1334/

## What does this PR change?

- Boolean values `isNewDefaultTrans` and `isNewAltTrans` detect wrongly because the `forceTranslation` block is located after the status detection.
- This PR moves boolean creation after `forceTranslation` operation.

## Other information

The bug is affected in 6.1.0 and after, so this should be backported to `releases/6.0` branch.


miurahr:
I have added an acceptance test which reproduces the bug and keeps behavior from a regression.

zs-stpa:
i can verify that cmd+shift+s now inserts identical translation as well as right click context menu action. but the undo history is still broken, could you please try and fix that as well while you are at it? maybe apply it as a single atomic undoable edit in the active field before committing, so undo can revert it (and the new acceptance tests could assert that)?

miurahr:
>  the undo history is still broken, could you please try and fix that as well while you are at it? 

`EditorController.commitAndDeactivate` method calls `undoManager.reset()` from OmegaT 3.0.x era.
The insert identical translation menu item calls `commitAndDeactivate` in that days. This is why we could not undo the identical translation action.

It might be in design that the undo is not for the identical translation. When you think we need to consider a design issue, please raise new RFE or BUGS.

(Edit: Add more explanation and clarify what I mean.) call for comment @zs-stpa @brandelune

brandelune:
> `EditorController.commitAndDeactivate` method calls `undoManager.reset()` from OmegaT 3.0.x era. The insert identical translation menu item calls `commitAndDeactivate` in that days. This is why we could not undo the identical translation action.
> 
> It might be in design that the undo is not for the identical translation. When you think we need to consider a design issue, please raise new RFE or BUGS.

I don’t remember any design discussion about not accepting undo for that feature.

As far as I’m concerned, any modification to the edit field should be undoable.

miurahr:
> > `EditorController.commitAndDeactivate` method calls `undoManager.reset()` from OmegaT 3.0.x era. The insert identical translation menu item calls `commitAndDeactivate` in that days. This is why we could not undo the identical translation action.
> > It might be in design that the undo is not for the identical translation. When you think we need to consider a design issue, please raise new RFE or BUGS.
> 
> I don’t remember any design discussion about not accepting undo for that feature.
> 
> As far as I’m concerned, any modification to the edit field should be undoable.

Ok, Let's  postpone the merge and keep identical translation broken. is it ok?

zs-stpa:
> > > `EditorController.commitAndDeactivate` method calls `undoManager.reset()` from OmegaT 3.0.x era. The insert identical translation menu item calls `commitAndDeactivate` in that days. This is why we could not undo the identical translation action.
> > > It might be in design that the undo is not for the identical translation. When you think we need to consider a design issue, please raise new RFE or BUGS.
> > 
> > 
> > I don’t remember any design discussion about not accepting undo for that feature.
> > As far as I’m concerned, any modification to the edit field should be undoable.
> 
> Ok, Let's postpone the merge and keep identical translation broken. is it ok?

why not?! please do fix it. just forget i asked about undo history or let's worry about that later and do not make the fix dependent on that. not necessary to mix fix with enhancement.

i just remember in earlier versions there was effects where you could insert long text with a single shortcut but then undoing cmd+z only removed the last inserted letter so you had to type a hundred times cmd+z to reverse one single cmd+shift+i for example (not checked if this was the accurate example) - and now we have the atomic undo operations on bord .. might as well use them? or not? but no matter if yes or no just do the fix first

brandelune:
> > > `EditorController.commitAndDeactivate` method calls `undoManager.reset()` from OmegaT 3.0.x era. The insert identical translation menu item calls `commitAndDeactivate` in that days. This is why we could not undo the identical translation action.
> > > It might be in design that the undo is not for the identical translation. When you think we need to consider a design issue, please raise new RFE or BUGS.
> > 
> > 
> > I don’t remember any design discussion about not accepting undo for that feature.
> > As far as I’m concerned, any modification to the edit field should be undoable.
> 
> Ok, Let's postpone the merge and keep identical translation broken. is it ok?

You asked be a history question, I gave you an answer. As Stephan wrote, we can put aside the undo tree for the time being, but keep it in mind for later fixes.

zs-stpa:
sorry my mistake - the atomic undo is so new you didnt knew – https://github.com/omegat-org/omegat/pull/2273/changes#r3943313378 - i had a draft PR with more editing shortcuts where this method takes care that multiple edit steps can be combined into single undo steps … to be worried about later.

miurahr:
> > > I don’t remember any design discussion about not accepting undo for that feature.
> > > As far as I’m concerned, any modification to the edit field should be undoable.
> > 
> > 
> > Ok, Let's postpone the merge and keep identical translation broken. is it ok?
> 
> You asked be a history question, I gave you an answer. As Stephan wrote, we can put aside the undo tree for the time being, but keep it in mind for later fixes.

There is similar but different feature "Edit/Replace with Source" that make translation as identital as source and undoable.

This PR is about "Edit/Register Identical Translation" that COMMIT source as translation, check issues, execute user script(optional), and synchorize with team.

You (and I) might confuse two different feature. I don't think we need to change the behavior of identical translation.

PR #2314, opened 2026-08-29
fix(mt/mymemory): restore no-arg constructor of MyMemoryHumanTranslate

zs-stpa:
## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?

- (none — master regression from #2290)

## What does this PR change?

- `MachineTranslatorsManager` instantiates registered connectors reflectively through no-arg constructor. Cleanup in #2290 removed seemingly unused constructor of `MyMemoryHumanTranslate`; result: NoSuchMethodException on every GUI start, connector missing from machine translation menu.
- Restore constructor, mark both MyMemory constructors as reflectively used, add regression test mirroring manager's instantiation path (fails with NoSuchMethodException without fix).

## Other information

Sibling `MyMemoryMachineTranslate` kept its pair all along; sweep over all machinetranslators modules found no further connector with parameterized constructor lacking no-arg one.


miurahr:
Oh, sorry. This is my mistake.

PR #2315, opened 2026-08-29
chore(ci): revive source distribution duplicate dependency check

zs-stpa:
## Pull request type

## Which ticket is resolved?

- (none — CI hygiene)

## What does this PR change?

- Old bash checker looked for `lib/provided/modules`, real path is `lib/provided/module`: both directory guards failed silently, CI step never checked anything. Version comparison also existed only within one directory, missing cross-directory skews (e.g. icu4j 78.1 beside 78.3); name-identical jars in both directories would have raised 19 false errors (by design, module packaging skips application jars by file name).
- Replace with python checker: errors on multiple versions within directory, on new cross-directory skew, on missing directories (guards against layout moves). 18 existing skews tolerated via `tool/known_dependency_skews.txt`; resolved skew turns baseline entry into error, so list only shrinks.
- Workflow step calls python script directly.

## Other information

Verified against locally built source distribution: green run, all three failure modes exit 1.


miurahr (approved):
LGTM.

PR #2313, opened 2026-08-29
chore(build): drop bundled icu4j copy from Japanese language module

zs-stpa:
## Pull request type

- Build and release changes -> [build/release]

## Which ticket is resolved?

- (none — dependency hygiene)

## What does this PR change?

- Exclude icu4j runtime dependency of `language-ja`: leftover of gosen tokenizer; omt build tokenizes through kuromoji, no class on module classpath references `com.ibm.icu`.
- Module jar filter matches application jars by file name; 78.1/78.3 skew baked unreachable icu4j copy into module jar. `omegat-language-ja.jar` shrinks 15.1 MB → 82 KB; single icu4j left in `lib/`.
- Add tests: NFKC surface cross-checked against JDK reference (golden sentences, kana voicing sequences, stable blocks, invariants; identical on icu4j 72.1/78.1/78.3), plus LanguageTool Japanese smoke test (tokenizer, tagger, rule check).

## Other information

Full test suite green; Mac distribution rebuilt and inspected, distribution size shrank by 15MB without loss of functionality.

PR #2316, opened 2026-08-29
chore(build): align module dependency resolution with application-provided versions

zs-stpa:
## Pull request type

## Which ticket is resolved?

- (none — distribution dependency hygiene, follow-up to #2315)

## What does this PR change?

- Module fat jars and source distribution's `lib/provided/module` carried 18 version skews against application libs (icu4j, guava, jackson, jna, morfologik, slf4j, …). Parent-first class loading serves shared libraries from application copy regardless of module-resolved version: skewed bundled copies are unreachable dead weight, and modules compile against versions they never run with.
- New `DependencyAlignment` helper (build-logic) rewrites module classpaths and root `moduleRuntimeLibs` resolution to application-resolved versions; propagates automatically on future core bumps.
- Aligner drops `tokyo.northside:segment`: fork keeps `net.loomchild` package names, fully shadowed by application's loomchild copy; only maligna class touching segment is SrxSplitAlgorithm, whose members all exist in application copy — and OmegaT's aligner never invokes that split path.
- Duplicate dependency checker baseline (#2315) shrinks to empty; any future skew fails CI.
- Distribution `modules/` folder shrinks 235 MB → 191 MB (−43 MB, −19%); ~15 MB of that overlaps with #2313's icu4j removal, ~28 MB is net new.

## Other information

Stacked on #2315. Verified: checker green (92 module jars, was 95), full test suite, distributionSmokeTest, configuration-cache run green, source distribution self-build resolution replayed locally.

### DIFF
--- a/aligner/build.gradle
+++ b/aligner/build.gradle
@@ -13,6 +13,15 @@ dependencies {
         // Aligner
         implementation(libs.tokyo.northside.maligna) {
             exclude module: 'commons-logging'
+            // Fork of net.loomchild:segment keeping original package names;
+            // parent-first loading shadows it completely behind application's
+            // loomchild copy (LanguageTool dependency), so bundling it only
+            // adds unreachable dead weight. OmegaT's aligner never invokes
+            // maligna's SRX split path - only maligna class touching segment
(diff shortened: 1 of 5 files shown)


miurahr (approved):
LGTM.

PR #2318, opened 2026-08-29
fix(glossary): search performance

zs-stpa:
## Pull request type

Bug fix

## Which ticket is resolved?

- 5.2.0 very slow
  * https://sourceforge.net/p/omegat/bugs/981/

## What does this PR change?

- Glossary pane renders matches into offline document swapped in once; fixes quadratic EDT rendering introduced by 45b9c44ca (800 matches: 4.5 s → 20 ms). Linkifier styles swapped-in documents once, font fallback listener follows document swaps.
- TransTipsMarker matches displayed entries in one batched call; segment tokenized once per marker pass instead of once per entry.
- searchSourceMatches prefilters entries via source token hashes and source characters; scan of 300k entries per segment switch drops from 1.6 s to 62 ms.
- Load tests pin interactive budgets for reported setup (ja→en, glossary of tens of megabytes, segments of 150+ characters).

## Other information

Search code was byte-identical between 4.3.2 and 5.2.0 and runs on background thread; freezes came from EDT rendering, matching reported threshold behaviour (3+ glossary files, long segments worst).

### DIFF
--- a/src/main/java/org/omegat/gui/glossary/GlossaryTextArea.java
+++ b/src/main/java/org/omegat/gui/glossary/GlossaryTextArea.java
@@ -58,6 +58,8 @@ import javax.swing.KeyStroke;
 import javax.swing.ToolTipManager;
 import javax.swing.text.AttributeSet;
 import javax.swing.text.Caret;
+import javax.swing.text.DefaultEditorKit;
+import javax.swing.text.DefaultStyledDocument;
 import javax.swing.text.Element;
 import javax.swing.text.JTextComponent;
 import javax.swing.text.StyledDocument;
@@ -252,9 +254,18 @@ public class GlossaryTextArea extends EntryInfoThreadPane<List<GlossaryEntry>>
         }
 
         IGlossaryRenderer entryRenderer = GlossaryRenderers.getPreferredGlossaryRenderer();
+        // Render into an offline document and swap it in whole: every insert
+        // into the displayed document costs time proportional to the document
+        // length, which stalls the EDT for seconds once a large glossary
+        // produces hundreds of matches (SF bug #981).
+        StyledDocument newDoc = new DefaultStyledDocument();
+        // getText() falls back to the platform line separator without this
+        // property; setText-created documents carry it, so match them.
+        newDoc.putProperty(DefaultEditorKit.EndOfLineStringProperty, "\n");
         for (GlossaryEntry entry : entries) {
-            entryRenderer.render(entry, getStyledDocument());
+            entryRenderer.render(entry, newDoc);
         }
+        setStyledDocument(newDoc);
 
         firePropertyChange("entries", oldEntries, entries);
     }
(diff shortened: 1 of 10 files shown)


src/main/java/org/omegat/gui/glossary/GlossaryTextArea.java
```diff
@@ -252,9 +254,18 @@ protected void setFoundResult(SourceTextEntry en, List<GlossaryEntry> entries) {
...
+        // produces hundreds of matches (SF bug #981).
+        StyledDocument newDoc = new DefaultStyledDocument();
+        // getText() falls back to the platform line separator without this
+        // property; setText-created documents carry it, so match them.
+        newDoc.putProperty(DefaultEditorKit.EndOfLineStringProperty, "\n");
         for (GlossaryEntry entry : entries) {
-            entryRenderer.render(entry, getStyledDocument());
+            entryRenderer.render(entry, newDoc);
         }
+        setStyledDocument(newDoc);
```
miurahr, line 268: Good catch for this!

src/main/java/org/omegat/gui/glossary/GlossaryTextArea.java
miurahr:
```suggestion
    protected void setFoundResult(@Nullable SourceTextEntry en, List<GlossaryEntry> entries) {
```

Here is nullable and pass null in the test case.

src/test/java/org/omegat/gui/glossary/GlossarySearchLoadTest.java
```diff
@@ -0,0 +1,431 @@
...
+ * the measured magnitudes.
+ *
+ * The paths mirror production wiring: one long-lived source tokenizer whose
+ * per-instance token cache stays warm across segment switches (as in
+ * {@code GlossaryManager#buildSearcher}), while every switch presents a new
+ * segment text.
+ *
+ * @author stephan.pakebusch at zollsoft.de
+ */
+public class GlossarySearchLoadTest extends TestCore {
```
miurahr, line 74: This test class mixes algorithmic load tests with UI responsiveness tests. The searcher/tokenization checks are useful as unit-level regression guards, but the rendering test is trying to validate an EDT/Swing freeze in a non-real-GUI fixture. 

The unit/function test run on `ConsoleWindow` stub. 

Since we have an AssertJ-Swing acceptance harness `src/testAcceptance`, I think the UI regression should be moved there and exercised through the real GUI/EDT path. 

The unit/function tests should be narrowed to GlossarySearcher/tokenization behavior and avoid “interactive” wording in method name and javadoc, while the acceptance test should cover segment switching/glossary pane rendering end-to-end.

miurahr (changes requested):
Thank you for your analysis and the fix! there is a space to improve in the test code.

miurahr:
Note: This fix should be backported into `releases/6.1` maintainance branch

zs-stpa:
i tried to improve as suggested.

miurahr (approved):

PR #2319, opened 2026-08-29
feat(team): negotiate per-project settings with team through sidecar file

zs-stpa:
## Pull request type

## Which ticket is resolved?

- (none — infrastructure base for SF-RFE-465 et al., some 'older' open PR's will be restacked onto this one to reuse the infrastructure)

## What does this PR change?

- Per-project settings a team should share had no home: project file parser of released OmegaT versions rejects unknown elements in omegat.project, locking team members on older versions out. Registered settings now live in `omegat/project_settings.properties`; older versions ignore extra file, file only materialises once setting is used, default projects stay byte-identical.
- Generic, feature-neutral machinery: `TeamSetting` descriptor + `TeamSettingsRegistry` (empty registry = every code path no-op), string-keyed `ProjectSettingsStorage` (deterministic write, byte-identity no-op detection on team commit), `IProject` default API (publish/useLocal/superseded), load flow activates team values and asks per superseded local value (share / take team / keep this time), properties dialog offers distribution on change; dialogs use generic {0} texts, en+de.
- 13 tests against real git repositories: publish/no-op guard, team round trip, resolution truth table, descriptor and persistence contracts.
- Follow-up PRs register concrete settings (fuzzy number matching, numeral value check, character equivalence) as thin one-line registrations.

## Other information

Extracted feature-neutrally from the number-substitution stack. Behavior unchanged until first setting registers.

### DIFF
--- a/src/main/java/org/omegat/core/data/IProject.java
+++ b/src/main/java/org/omegat/core/data/IProject.java
@@ -29,6 +29,7 @@
 package org.omegat.core.data;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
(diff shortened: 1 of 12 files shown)


brandelune:
We are starting to have a number of settings that need proper management:

- project specific settings (mostly with configurations files located in `omegat`)
- application settings through the main configuration folder
- project specific settings through a specific configuration folder (only available through CLI)
- team settings

All this is very messy in terms of UI and UX. We need to fix that.

zs-stpa:
totally agreed and am trying to do so, bit by bit, but is not easy in small atomic reviewable steps to get from messy to clean without breaking some old patterns here or there …

i was just thinking (after waking up, i probebly even dream about omegaT …) that it is a nuisance to have to change some settings everytime after opening a different project, just one example: the checkbox for automatically fetching MT proposal when entering untranslated segments, needs a project specific override.…

open to suggestions – the direction* i was aiming for, is motivated by exactly what you have described, but i also do not know in detail how to solve it. *: that direction being: **every** possible user settable setting needs a project-override and needs to be team-compatible i.e. negotiating differences between local and team settings for every single setting. moving in such a direction would require this kind of infrastructural change here.

zs-stpa:
also, i would suggest, that we maybe try and distinguish between "ordinary user" who uses only GUI and never CLI or config folder, and "super user / admin" who are open to advanced below-surface features… and that we then aim to make all settings accessible to the ordinary user as well. manually changing the default for a regex or display pattern with placeholders, that is something i would like to make easier for ordinary users, many of whom struggle with such features or simply never use them.

brandelune:
> also, i would suggest, that we maybe try and distinguish between "ordinary user" who uses only GUI and never CLI or config folder, and "super user / admin" who are open to advanced below-surface features…

No, we don’t. We teach users how to use complex features instead. And we generally make complex features easily accessible when possible.

"expert" features were never designed as so because of a need to limit them to expert use, but rather because developers did not have the time to make them easier to access, and for other historical reasons.

In our case, an "easy" way to solve the difference between "project specific settings" and "CLI based project specific configuration folder" is to have OmegaT create a project specific configuration folder for each project and to put project specific settings there.

But that is a discussion we need to have on the user list.

There definitely is some sort of "cascading" effect that we need to take into account. And also we need to have clear visual hints for the status of settings (local/general/team) etc.

zs-stpa:
i am not active on the user list. there, i can probably spend much more time discussing than producing. not saying that discussing can not be productive as well. but the dis- prefix makes it sounds so negative … is there a concus... never mind. maybe syncussion? not much better .. cheears and good night 🍻

PR #2321, opened 2026-08-30
chore(filters): refactor XMLFilter constructors and SAX feature configuration

miurahr:
fix [this-escape] java compiler warning.

## Pull request type

- Other (describe below)
refactor
## Which ticket is resolved?
 none; there is no behavior change but reduce Java compiler warning.

## What does this PR change?

- Introduced parameterized constructors with default and overrideable SAX features.
- Removed redundant `setSAXFeature` calls in subclasses.
- Simplified XML processing and test setup.


miurahr (commented):
We can remove exception catch from filter classes and the all the exceptions are catched in abstract XMLFilter class. This is good direction.
We can also reduce code from test code for setup a filter class. It clarify the logic.

PR #2322, opened 2026-08-30
feat(gui): matches deleted parts are red by default

zs-stpa:
## Pull request type

## Which ticket is resolved?

- "Match display template" including ${diff}
  * https://sourceforge.net/p/omegat/feature-requests/1793/ (colour part)

## What does this PR change?

- Deleted match text defaults to red tones (#ff3399 active, #ff0000 inactive) in both bundled schemes instead of plain foreground grey/black, so removed words stand out beyond strike-through alone.

## Other information

Stacked on #2208 (last commit only)


brandelune:
I’m seeing things here that have nothing to do with "matches deleted parts are red by default - #2322".

zs-stpa:
> I’m seeing things here that have nothing to do with "matches deleted parts are red by default - #2322".

this PR relies on #2208 so is stacked on top of it, that should be reviewed first, only the very last commit actually belongs to this PR #2322 here.

brandelune:
Do we need all this stack to change a default color? If yes, that’s a problem. Just change the default color please.

PR #2329, opened 2026-08-31
fix(gui): change table selection configuration and add test for row selection behavior

miurahr:
This PR fixes the regression caused by PR #2073 that file list table does not allow the user to select a file by cursol key and add the regression test to keep a configuration.

## Pull request type
- Bug fix -> [bug]

## Which ticket is resolved?
None; report in the mail list but issued the ticket.

## What does this PR change?

- A file list table properties to set properly seletion limitation.

### DIFF
--- a/src/main/java/org/omegat/gui/filelist/ProjectFilesListController.java
+++ b/src/main/java/org/omegat/gui/filelist/ProjectFilesListController.java
@@ -779,10 +779,20 @@ public class ProjectFilesListController implements IProjectFilesList {
 
     private void createTableFiles() {
         applyColors(list.tableFiles);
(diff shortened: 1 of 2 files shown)


brandelune (approved):
Thank you. This should be ported to 6.1 too.

miurahr:
> This should be ported to 6.1 too.

No, I don't. The change #2073 is not in `releases/6.1` and the bug does not exist in the 6.1.1.

miurahr:
The CI failure is caused by another place and #2333 fix one. So I want to ignore it.

brandelune:
> > This should be ported to 6.1 too.
> 
> No, I don't. The change #2073 is not in `releases/6.1` and the bug does not exist in the 6.1.1.

Sorry I got it wrong.

PR #2332, opened 2026-08-31
[6.0] docs: Warning in the manual for macOS users regarding notarization

brandelune:
The macOS notarized bundle does not allow modifications of files inside `OmegaT.app`.

Such files were used as launch parameters for OmegaT. macOS users can now only use command line options.

This manual modifications warns users about that.

https://sourceforge.net/p/omegat/documentation/453/


miurahr (approved):
It is great if `releases/6.1` have a same change.

PR #2334, opened 2026-09-01
Add warning in the manual for macOS users re. notarization (#2332)

brandelune:
The macOS notarized bundle does not allow modifications of files inside OmegaT.app.

Such files were used as launch parameters for OmegaT. macOS users can now only use command line options.

This manual modifications warns users about that.

https://sourceforge.net/p/omegat/documentation/453/

### DIFF
--- a/src_docs/manual/en/HowTo_Run.xml
+++ b/src_docs/manual/en/HowTo_Run.xml
@@ -61,65 +61,17 @@
          <filename>OmegaT.app/Contents/MacOS/Java/</filename> that
          	you can use to launch OmegaT from the command line. See the 
          <link endterm="launch.with.command.line.title" linkend="launch.with.command.line"></link> section for details.</para>
-      <para>You can modify the behaviour of OmegaT.app by editing the
-         	
-         <filename>Configuration.properties</filename> (OmegaT configuration) as well
-         	as the 
-         <filename>Info.plist</filename> (Java configuration) files located
-         	inside the OmegaT.app package.</para>
-      <para>The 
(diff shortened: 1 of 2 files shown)


src_docs/manual/en/HowTo_Run.xml
```diff
@@ -61,65 +61,17 @@
...
-            </listitem>
-         </varlistentry>
-      </variablelist>
+
+	  <warning>
+	     <para>OmegaT is now a macOS notarized application. Any
+	     modification to files inside <filename>OmegaT.app</filename>
+	     will keep OmegaT from running.</para>
+
+	     <para>To change OmegaT’s launch parameters (alloted memory,
```
miurahr, line 70: Users can override JVMOptions with a command line like 
```shell
defaults write org.omegat.OmegaT "JVMOptions/" -dict-add Xmx "-XmxNNNNm"
```

It may be a similar way for `user.country`, `user.language`, `http.proxyHost`, `http.proxyPort`.

miurahr (changes requested):

src_docs/manual/en/HowTo_Run.xml
brandelune: Yes, that’s what is written after that paragraph. Do you want me to add anything?

miurahr: Because we don't reach concensus for adding a command line in the dev-ml discussion, I think it is ok now for the current proposal.

miurahr (approved):

PR #2343, opened 2026-09-05
chore(ci): update integration test server base image to debian:bookworm

miurahr:
Recent weekly release build failed with Integration test failure.
The integration test use docker for server and client.
A server image uses old debian base version so now it has been arachived then download become error.

## Pull request type

## Which ticket is resolved?
None; but add note in developer/changes/6.2.0.md
## What does this PR change?

- bump to bookworm from bullseye debian version


miurahr:
This also cause the error on releases/6.1 branch. The fix #2360 is also mandatory to fix in 6.1.2 weekly.

PR #2350, opened 2026-09-08
fix(gui): distinguish dictionary sources in spell checker language list

zs-stpa:
## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?

- No dedicated ticket; duplicate entries stem from bundled dictionaries introduced with "Support Morfologik spell checker and use LT bundled dictionary"
  * https://sourceforge.net/p/omegat/feature-requests/1748/

## What does this PR change?

- Preferences > Spell checker listed same language up to three times with identical labels (bundled Hunspell, bundled Morfologik, installed file) - list selection has no consumers, so rows were pure duplicates.
- Available-languages list now aggregates all dictionary sources into one line per language, naming engine and origin: `de_CH - German (Switzerland): Hunspell (installed), Hunspell + Morfologik (bundled)`.
- `SpellingDictionaryEntry` fields encapsulated (`custom` -> `isLocalFile()`); source-incompatible for plugins reading fields directly - no in-repo usages outside `DictionaryManager`.
- Flat `getLocalDictionaryNameList` deprecated in favor of aggregated display list.

### screenshots

before (in arabic)

Bildschirmfoto 2026-09-08 um 09 43 48

after (in english)

Bildschirmfoto 2026-09-07 um 21 36 43


brandelune:
> ### screenshots
> before (in arabic)

What’s the point sending a screenshot in Arabic? It does not make the comparison easier.

> after (in english)

Can you resent the set so that we can make an easier comparison?

zs-stpa:
> What’s the point sending a screenshot in Arabic? It does not make the comparison easier.

several points.
- you can see identical repitions without having to understand the language's script
- you can verify this in the released version or when you build from current master locally in your Ωt instance
- you can see i am trying to get a holistic feel of parts of omegaT i usually do not use, like arabic UI
- and you can see the arabic UI is mostly english

> Can you resent the set so that we can make an easier comparison?

yes. soon.

brandelune:
> > What’s the point sending a screenshot in Arabic? It does not make the comparison easier.
> 
> several points.

If you want us to compare something, make that "all other things being equal". That decreases the time spent on pointless exchanges.

brandelune:
> after (in english)
> 
> Bildschirmfoto 2026-09-07 um 21 36 43

Ok. Keep the original layout and just add the dictionary name after each language please:

de_DE German (Germany) Hunspell (installed)
de_DE German (Germany) Morfologik (bundled)

I guess Hunspell is the package that’s been installed by the user (from LO’s list)?
Also "bundled" won’t be bundled anymore when we implement @miurahr’s l10n packages idea. So what will we have then?

zs-stpa:
the label above the table says 'available languages' i see no reason for a language to be available more than one time (it is available or it is not available) so it should be one line per distinct language code (…or someone allowed an inaccurate label to be committed at some point in the past… not pointing fingers, i did not prevent it either, but hey we're trying to fix it now, right? ^.^) may i maybe also change the label to 'available language resources' if one line per resource is what it has to be?

and i was not really that aware about the upcoming refactorings so i did not think of that. but i imagine that can be dealt with during the upcoming refactoring instead of having to already present a post-refactoring state before the refactoring has even begun.., right? i imagine i'll be able to help with that when the time comes.

brandelune:
> the label above the table says 'available languages' i see no reason for a language to be available more than one time 

Then we change the label to "available dictionaries".

*This* is more elegant than wasting time on restructuring the whole table and making it less readable.

> and i was not really that aware about the upcoming refactorings

But I guess your "bundled" comes from actually checking a resource availability, right? So there is nothing wrong with *that*, only with the fact that the other resource does not have a qualifier.

miurahr (approved):
LGTM. of the code quality.

miurahr:
Legacy hunspell spellchecker that uses native library bridge of Dumont Hunspell library, which only supports intel/arm macOS, windows and linux on amd64, no support for windows/linux on aarch64 architecture.
The dumont hunspell native library only supports files on file systems, but no support for resources in the jar files.

OmegaT copies hunspell `.aff` and `.dic` file under the `$HOME/.omegat/spelling/` or `~/Library/Preferences/OmegaT/spelling`

When user manually install spell dictionary, they copy files on the same place. It is a reason OmegaT does not recognize the files are custom dictionary which user installed or copied from bundled dictionary.


We have Lucene Hunspell spellchecker engine as a bundled OmegaT plugin, which supprts both files on filesystem and java resources.
When using the Lucene Hunspell spellchecker,  when user works on Linux/Windows on aarch64 like me, there is no dictionary-copy to user config folder when using bundled dictionary.

PR #2354, opened 2026-09-08
fix(gui): give Forward in History its own shortcut

zs-stpa:
## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?

- No ticket; found by conflict analysis of the bundled shortcut files.

## What does this PR change?

- Forward in History ships with the same shortcut as New Project (`ctrl/cmd shift N`) in both platform shortcut files - one of the two is silently shadowed; the manual documents the duplication
- Forward in History moves to `ctrl/cmd alt shift N` (`ctrl/cmd alt N/P` are taken by unique-segment navigation, `ctrl/cmd shift F` by find-in-project). Back in History keeps `ctrl/cmd shift P`.
- Manual shortcut updated (Menus_GoTo.xml).

### DIFF
--- a/src/docs/manual/en/Menus_GoTo.xml
+++ b/src/docs/manual/en/Menus_GoTo.xml
@@ -305,6 +305,7 @@
             
             <keycombo>
                <keycap>C</keycap>
+               <keycap>A</keycap>
                <keycap>S</keycap>
                <keycap>N</keycap>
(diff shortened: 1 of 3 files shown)


zs-stpa:
as pointed out on the dev list - this is useless, since 'new project' is not available anyways in an already open project.

PR #2360, opened 2026-09-08
chore(CI/test-integ): fix ssh login error in the integration test

miurahr:
We recently change docker image base to bookworm from bullseye for server side.
The server image configuration lacks home directly and depends on an implicit behavior of useradd.
Also logging configuration can be broken in client side.

## Pull request type

## Which ticket is resolved?
none; recent weekly release failed with integration test failure
## What does this PR change?

- server/Dockerfile: add '--home /home/git' in useradd shell
- TestLogFileHanlder.java: check logging folder exists
- client/entrypoint.sh simplify command line


miurahr:
With this fix, integration test with docker run well in local environment.

PR #2364, opened 2026-09-10
feat(editor): add glossary insertion shortcuts

zs-stpa:
## Pull request type

## Which ticket is resolved?

- Replace glossary hits when inserting fuzzy match
  * https://sourceforge.net/p/omegat/feature-requests/1566/
- Repair fuzzy matches with glossary items
  * https://sourceforge.net/p/omegat/feature-requests/369/

## What does this PR change?

- New Edit action "Insert Match with Glossary Hits Replaced" (ctrl/cmd alt I): inserts active fuzzy match with glossary repairs - where match source differs from current segment and both sides of the difference are glossary terms, old target term is swapped for the new one (whole-word, capitalization-matched); number substitution and MT-provenance marking behave like plain insert
- New Edit action "Insert Source with Glossary Hits Replaced" (ctrl/cmd alt shift I): source text with glossary source terms replaced, regardless of replace-on-insert preference
- New Edit submenu "Insert Glossary Hit" # 1-5 (ctrl/cmd alt 1-5): inserts first target term of the nth Glossary pane entry
- Actions registered in shortcut catalog and editable in the shortcuts preferences (stacked on shortcut-management branches)
- Manual: Edit menu entries and shortcut appendix; tests for the repair engine (LCS diff, word boundaries, capitalization, multiple differences)

Design notes: glossary lookup for repairs matches source spans exactly (case-insensitive, no stemming); one glossary query per insertion since external glossaries answer over the network. Follow-up planned on this stack: configurable actions toolbar per SF-642/SF-1170, feeding from the shortcut catalog - conceptual feedback welcome there.

## Stacking

Stacked on the shortcut-management branches (catalog and API -> live rebind -> preferences GUI, SF-443/SF-1301): best reviewable only after the groundwork PRs are merged. hence created as draft.

### screenshot

the seven circled shortcuts are new in this PR here and work as advertised in my real life in-production projects running for close to 10 years now.

Bildschirmfoto 2026-09-10 um 15 49 48


brandelune:
> * New Edit action "Insert Match with Glossary Hits Replaced" (ctrl/cmd alt I): inserts active fuzzy match with glossary repairs - where match source differs from current segment and both sides of the difference are glossary terms, old target term is swapped for the new one (whole-word, capitalization-matched); number substitution and MT-provenance marking behave like plain insert
> * New Edit action "Insert Source with Glossary Hits Replaced" (ctrl/cmd alt shift I): source text with glossary source terms replaced, regardless of replace-on-insert preference

So, basically what you propose is do a slightly improved manual "replace glossary terms on match insertion" (that includes numbers conversion). Is that correct?

If that is the case, there are 2 patterns:
- the preference is enabled and your function is not very useful (we could improve the preference setting so that it includes numbers)
- the preference is not enabled and the user activates the "preference" function case-by-case

That makes me wonder whether the preference is still useful if your function is implemented.


> * New Edit submenu "Insert Glossary Hit" # 1-5 (ctrl/cmd alt 1-5): inserts first target term of the nth Glossary pane entry

We already have a UI to insert glossary terms. That’s the autocompletion window. It seems to me that we don’t need redundancy here and we’d rather find a way to implement glossary insertion either within the autocompletion mechanism, or like the Search layer does: hit a shortcut, OmegaT puts an indexed layer on the source glossary terms, user hits the relevant key and the associated term is inserted at point.

That layered UI that @yu-tang devellopped is already an often used feature of OmegaT’s GUI and extending it to other parts of the editor is a smarter move than adding new menu shortcuts.

> * Actions registered in shortcut catalog and editable in the shortcuts preferences (stacked on shortcut-management branches)

We certainly can add those items as disabled items in the menu UI, without associated shortcuts, for people who really want them. After your shortcut setting GUI is implemented, this will only be a matter of personal preference. But we need a better default UI for that first.

brandelune:
Also, I’m a bit overwhelmed by the amount of stuff you’re working on, but it seems to me that this feature has not formally been introduced to the dev list.

PR #2366, opened 2026-09-11
[6.1] fix(glossary): search performance (backport PR#2318)

miurahr:
* fix(glossary): render the glossary pane as one document swap (SF-981)

Since 5.x every entry was inserted separately into the displayed StyledDocument on the EDT; cost per insert grows with document length, so panes with hundreds of matches froze the UI for seconds. Render into an offline document and swap it in once; the linkifier styles links of a swapped-in document one time, and the font fallback listener moves along with the document.

* fix(glossary): tokenize the segment once per marker pass and prefilter the entry scan (SF-981)

TransTipsMarker re-tokenized the segment for every displayed entry, and searchSourceMatches ran the full token scan for every glossary entry. New batched searchSourceMatchTokens tokenizes the segment once; the per-search token-hash and character prefilters skip entries that cannot match (necessary condition for both searchAll modes, hash collisions fall through to the full check).

* test(glossary): pin time budgets for large glossaries (SF-981)

Guards the SF-981 fixes against regression: unit-level load tests pin the searcher and tokenization budgets, and an AssertJ-Swing acceptance test exercises the glossary pane end to end through the real GUI and EDT, switching segments against a glossary that yields hundreds of rendered matches.

(cherry picked from commit c9f7e459931cc7989ed086d442aad363b173d0b6)

## Pull request type

- Bug fix -> [bug]

## Which ticket is resolved?

- https://sourceforge.net/p/omegat/bugs/981

## What does this PR change?

-
-
-

### DIFF
--- a/src/org/omegat/gui/common/EntryInfoPane.java
+++ b/src/org/omegat/gui/common/EntryInfoPane.java
@@ -30,6 +30,8 @@ import java.awt.Rectangle;
 
 import javax.swing.JTextPane;
 
+import javax.swing.text.Document;
+
 import org.omegat.core.Core;
 import org.omegat.core.CoreEvents;
 import org.omegat.core.events.IProjectEventListener;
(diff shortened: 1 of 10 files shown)


brandelune (approved):

zs-stpa:
possible duplicate: https://github.com/omegat-org/omegat/pull/2324 - i only just now saw that it was set to draft, sorry

miurahr:
Oh...

PR #2370, opened 2026-09-12
chore(doc-devel): git workflow to add guide of a smart assistant

miurahr:
Require an editable source next to SVG diagrams in the developer
documentation, and tell contributors how to write the PR explanation.

the Diagrams section already expects SVGs to be prepared with drawing
tools, but nothing asks for the tool's file. Without it the SVG cannot be
edited later. Requiring the source keeps SVG available for drawings Mermaid
cannot express, while keeping every diagram maintainable. Only SVG files
added under `src/docs/developer/` are checked; existing files, PNG/JPEG
screenshots, and the user manual are unaffected.

## Pull request type

## Which ticket is resolved?

none;
## What does this PR change?

- `48.DeveloperDocumentation.md`, section "Diagrams": states the order
  of preference (plain text or list, then Mermaid, then SVG), and
  requires that an SVG be committed together with the editable source
  it was exported from, under the same base name.
- `06.GitWorkFlows.md`, "Open a Pull Request": a short paragraph on the
  PR explanation (a few plain sentences, written by the author even
  when an AI assistant was used) with a pointer to the diagram rule.
- `.github/workflows/developer-docs-diagram-source.yml`: new check on pull
  requests touching `src/docs/developer/`. Fails if an SVG is added
  without a matching `.drawio`, `.excalidraw`, `.puml`, `.dot`, or
  `.mmd` file.
- `96.ContinuousIntegration.md`: the new workflow added to the table
  and file listing.

### DIFF
--- a/src/docs/developer/48.DeveloperDocumentation.md
+++ b/src/docs/developer/48.DeveloperDocumentation.md
@@ -212,12 +212,17 @@ locally before opening a pull request.
 Diagrams help explain architecture, workflows, and data flow. The developer
 documentation supports both static SVG images and Mermaid diagrams.
 
-Use the format that best fits the contribution:
+Start from the cheapest form that carries the information:
 
-- Use SVG for polished infographics, screenshots, and diagrams prepared with
-  drawing tools.
-- Use Mermaid for diagrams that are easier to maintain as text in version
-  control.
+- Use plain Markdown text or a list when the content is a sequence or a set
+  of items. Do not turn a list into an image.
+- Use Mermaid for structural diagrams (flows, dependencies, state, sequence).
+  Mermaid is text, so it is reviewed as a diff and can be edited later by
+  anyone.
+- Use SVG for drawings prepared with a drawing tool, and commit the tool's
+  editable source next to it with the same base name (for example
+  `build-pipeline-overview.svg` with `build-pipeline-overview.drawio`).
+  An SVG without its source cannot be maintained and is rejected by CI.
 
 ### Embedding SVG images
 
--- a/.github/workflows/developer-docs-diagram-source.yml
+++ b/.github/workflows/developer-docs-diagram-source.yml
@@ -1,0 +1,34 @@
+name: Developer docs diagram source check
+
+on:
+  pull_request:
+    paths:
+      - 'src/docs/developer/**'
+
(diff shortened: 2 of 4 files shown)


src/docs/developer/48.DeveloperDocumentation.md
```diff
@@ -212,12 +212,17 @@ locally before opening a pull request.
...
+
+- Use plain Markdown text or a list when the content is a sequence or a set
+  of items. Do not turn a list into an image.
+- Use Mermaid for structural diagrams (flows, dependencies, state, sequence).
+  Mermaid is text, so it is reviewed as a diff and can be edited later by
+  anyone.
+- Use SVG for drawings prepared with a drawing tool, and commit the tool's
+  editable source next to it with the same base name (for example
+  `build-pipeline-overview.svg` with `build-pipeline-overview.drawio`).
+  An SVG without its source cannot be maintained and is rejected by CI.
```
zs-stpa, line 225: question mark? an svg is human readable and can be manually edited in any text editor, see also https://en.wikipedia.org/wiki/SVG#Authoring "SVG images can be hand-coded […]" and it is common practice to generate SVG programmatically either by XSL transformations of other XML data or with simple python scripts or as output of any other language … i do not agree that you have to use a graphical editor

brandelune: > question mark? an svg is human readable and can be manually edited in any text editor, see also https://en.wikipedia.org/wiki/SVG#Authoring "SVG images can be hand-coded […]" and it is common practice to generate SVG programmatically either by XSL transformations of other XML data or with simple python scripts or as output of any other language … i do not agree that you have to use a graphical editor

Any XML file is "human readable", because it is not a binary, and _can_ be hand-coded. SVG is no exception. But that is not addressing Hiroshi’s objection. Your SVGs are not hand-coded and come from a data set.

miurahr: I remember the Inkscape -- the graphic SVG editor -- which has `.svg` as default saved filetype.
I tried to use inkscape to evaluate editing capability, and recognize that the explanation is wrong. Yes, I can edit easily the SVG diagram.

I would like to remove the SVG clause here.

PR #2373, opened 2026-09-13
chore(UI/QuickAccess): enable QuickAccess UI

miurahr:
QuickAccess UI was discussed in April 2023, 3 years ago, and postponed the discussion to 6.2 development cycle.  
This is a enablement of the in-source plugin and refactring for null safety, typo correction and so on.

## Pull request type

## Which ticket is resolved?

- Adding a toolbar in OmegaT 
- https://sourceforge.net/p/omegat/feature-requests/642/

## What does this PR change?

- enable in-source plugin QuickAccess
- fix typo of the interface name Preferece -> Preference
- Null safety annotations

## Other information

https://github.com/omegat-org/omegat/pull/622


zs-stpa:
i have no problem with this nullability refactoring and typo correction (prefere(n)ce), but i suspect this could be more clearly divided into separate narrow-scoped PRs.. as i sayd, not really a prbolem with me.

but just to make sure i understand the 'real' change is only the very last uncommenting line in the properties?

zs-stpa:
the 'needs restart' notice is unreliable when setting the Quick Access UI style in preferences. With both these fixes, the restart notice appears exactly when the menu style actually changed:

- `isModified()` compares selection against raw preference, which is empty until Appearance panel is persisted once — so the panel reports "modified" right after opening, before any change. Comparing against same default as used by `initFromPrefs()` fixes it:

```java
Object selected = panel.cbMenustyleSelect.getSelectedItem();
if (selected != null && !selected.toString().equals(Preferences
        .getPreferenceDefault(Preferences.MENUUI_CLASS_NAME, MainMenuUI.class.getName()))) {
    return true;
}
```

- combo listener registered in `initGui()`, so the programmatic `setSelectedItem()` in `initFromPrefs()` already fires it during panel construction; theme combos register their listeners in `initListeners()`, which runs after `initFromPrefs()`; moving the line there keeps initialization silent:

```java
panel.cbMenustyleSelect.addActionListener(e -> setRestartRequired(isModified()));
```


and on macOS the toolbar never appears when the setting is active: `MainWindow.initMainMenu` instantiates the class via `getDeclaredConstructor(MainWindow.class, MainWindowMenuHandler.class)`, but `AccessTools` only has `AccessTools(BaseMainWindowMenuHandler)` — `NoSuchMethodException`, silently swallowed by the empty catch. Since the menu stays `MainWindowMenu` on macOS by design, enabling Quick Access UI there changes nothing visible.

Minimal fix: matching overload delegating to the existing constructor, plus `Log.log(e)` instead of the silent catch.

```java
public AccessTools(MainWindow mainWindow, MainWindowMenuHandler mainWindowMenuHandler) {
    this(mainWindowMenuHandler);
}
```

now i can finally see the toolbar

Bildschirmfoto 2026-09-14 um 14 40 10

src/main/resources/org/omegat/Bundle.properties
brandelune: Why call the keys "new UI" ? Shouldn’t it be "Quick access" ? (whatever that means, since shortcuts are also quick access UI).

brandelune:
> now i can finally see the toolbar
> 
> Bildschirmfoto 2026-09-14 um 14 40 10

Ok, so this toolbar comes from Hiroshi’s proposal and yours is in the mail that you sent to the list, is that correct?

My question is: which of the two uses already available possibilities from our docking library? Because it seems to me that it’s the one we want to use, instead of adding yet some new code to our GUI.

Also, I was never fond of that "burger" thing that is fashionable on web sites and touch UIs, but not appropriate for standalone applications.

zs-stpa:
> > now i can finally see the toolbar
> Ok, so this toolbar comes from Hiroshi’s proposal and yours is in the mail that you sent to the list, is that correct?

yes.


> My question is: which of the two uses already available possibilities from our docking library? Because it seems to me that it’s the one we want to use, instead of adding yet some new code to our GUI.

this here is not docking but a permanent north fixture in the app window, i did something different and have yet to open PR's (soon).

brandelune:
> > My question is: which of the two uses already available possibilities from our docking library? Because it seems to me that it’s the one we want to use, instead of adding yet some new code to our GUI.
> 
> this here is not docking but a permanent north fixture in the app window, i did something different and have yet to open PR's (soon).

What I am asking is: do your widgets come from the windowing library that we use, or not? Because if they don’t, that means that we can’t have a "toolbar" in OmegaT, but only a "tool pane". Which is not really what users expect.

It is OK to limit oneself to what (very few) users want: just buttons at the top of the editor. If we can’t deliver that, that’s OK and we don’t need to hack OmegaT’s GUI to make it look like we can.

miurahr:
> but just to make sure i understand the 'real' change is only the very last uncommenting line in the properties?

Yes, it is.

miurahr:
> Also, I was never fond of that "burger" thing that is fashionable on web sites and touch UIs, but not appropriate for standalone applications.

Firefox has had a "burger" menu at the top-right of its window since 2014, so it is not only a web/touch pattern. On macOS the menu is always in the system menu bar, so another menu style is unnecessary there and the Quick Access UI keeps the standard menu. On Windows and Linux, where the menu bar occupies space inside the window, a compact menu has real value. Either way it is opt-in; the default remains the classic menu.

That said, the menu style itself is not what this PR decides. The Quick Access code has been in the tree since 2023; the actual change here is a single line that enables the in-source plugin, as Stephan confirmed above.

To keep the review focused, I will split this into three PRs, as Stephan suggested:

1. Enable the QuickAccess in-source plugin (this PR, reduced to that one change)
2. Typo fix (Preferece → Preference) and nullability annotations / API documentation. 
3. Rename the toolbar button label "GO" to "Open project"

PR 2 will be the right place to discuss how an alternative UI implementation, such as the one Stephan is preparing, can build on the same API instead of adding separate GUI code.

Design discussion about the toolbar and menu style is welcome, but let's take it there or to the dev list, so this PR can be judged on whether enabling the plugin is acceptable for 6.2.

brandelune:
> > Also, I was never fond of that "burger" thing that is fashionable on web sites and touch UIs, but not appropriate for standalone applications.
> 
> Firefox has had a "burger" menu at the top-right of its window since 2014, so it is not only a web/touch pattern.

And Firefox also has a tool bar where this buttons fits, as a button.

So I stick to my opinion. The way you designed that thing does not fit OmegaT’s UI. I opposed it when you put it there, without discussion, and I still oppose it now.

> Design discussion about the toolbar and menu style is welcome,

??? It is not welcome, it is compulsory.

> but let's take it there or to the dev list, so this PR can be judged on whether enabling the plugin is acceptable for 6.2.

Right now it is not. And Stephan’s button pane is not either. The way you two envision OmegaT’s GUI is really not going in the right direction. If you want an external plugin for your code, I’m fine with that though.

We’ll need to discuss a lot more before we change OmegaT’s external aspect, default or not.

miurahr:
I rebuild commits; enable in-source plugin and fix error on macOS.

miurahr:
I have split the PR to focused PRs; 
- #2378 -- fix API name typo, now working on writing API document as ADR
- #2379 -- null safety in `o.o.g.accesstools` package
- #2381 -- handles document load event to enable/disable tool.

miurahr:
> And Firefox also has a tool bar where this buttons fits, as a button.

Yes, so we can provide tool buttons, combobox, and burger menu only on Windows and Linux on a menu bar.

brandelune:
> > And Firefox also has a tool bar where this buttons fits, as a button.
> 
> Yes, so we can provide tool buttons, combobox, and burger menu only on Windows and Linux on a menu bar.

I don’t think so. We provide a unified experience for all users of OmegaT. And profound changes in the UX/GUI must first discussed.

The way You and Stephan think of integrating a toolbar in OmegaT is not elegant. We need to think more.

